PR #11, opened 2026-05-22
chore: add CODEOWNERS (cli-maintainers)

lml2468:
Add CODEOWNERS file pointing to `@Mininglamp-OSS/cli-maintainers`.

This is part of the sub-maintainer team setup. All code review assignments will be managed via team membership going forward — no need to edit this file when members change.

/cc @lml2468

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,5 @@
+# Code owners for octo-cli
+# Managed via GitHub Team membership — do not edit directly.
+# To change reviewers, update the team: @Mininglamp-OSS/cli-maintainers
+
+* @Mininglamp-OSS/cli-maintainers


yujiawei (changes requested):
# Code Review — PR #11 (octo-cli)

## Summary

This PR adds a new top-level `CODEOWNERS` file pointing all paths to `@Mininglamp-OSS/cli-maintainers`. Stated intent (per PR body): hand off review routing to the `cli-maintainers` sub-team via GitHub team membership.

Verdict: **CHANGES_REQUESTED**. The change as written has **no functional effect** — it is silently shadowed by an existing `.github/CODEOWNERS` and points to a team that does not have access to this repository. Both issues need to be resolved before merging, otherwise the PR ships dead code and the sub-maintainer routing the description promises will not actually happen.

## Findings

### P0-1 — New `CODEOWNERS` is shadowed by existing `.github/CODEOWNERS`

`CODEOWNERS` (lines 1–5, new file) is created at the repository root, but `.github/CODEOWNERS` already exists on `main` and on this PR's head (`9b3b831`):

```
# .github/CODEOWNERS (existing, unchanged by this PR)
* @Mininglamp-OSS/maintainers
```

Per the GitHub docs, when a `CODEOWNERS` file exists in more than one of the supported locations, GitHub searches `.github/`, then the repo root, then `docs/`, **and uses the first one it finds — the others are ignored**. Because `.github/CODEOWNERS` is present, the new root file will never be read.

Direct evidence on this PR itself: PR #11's auto-requested reviewers are the `maintainers` team, not `cli-maintainers`:

```
$ gh api repos/Mininglamp-OSS/octo-cli/pulls/11/requested_reviewers --jq '.teams[].slug'
maintainers
```

So after merging, behavior is unchanged: `@maintainers` continues to be the sole code owner; the new file is dead.

**Required fix (pick one):**

- **Option A (recommended):** Modify `.github/CODEOWNERS` in place — replace `* @Mininglamp-OSS/maintainers` with `* @Mininglamp-OSS/cli-maintainers` (or add a more specific rule above it, since later rules override earlier rules in CODEOWNERS). Do not create a second root file.
- **Option B:** Delete `.github/CODEOWNERS` in this same PR and keep the root one. Less preferable because `.github/` is the conventional location and tooling/audits tend to look there first.

### P0-2 — `@Mininglamp-OSS/cli-maintainers` has no access to `octo-cli`

For a team to be a valid code owner, the team must be visible to the repository (added with at least Triage permission; Write is the norm so reviews can actually be merged). Verification:

```
$ gh api repos/Mininglamp-OSS/octo-cli/teams --jq '.[] | {slug, permission}'
{"permission":"admin","slug":"maintainers"}
{"permission":"push","slug":"dev-team"}
```

`cli-maintainers` is not in the list. Looking at the team itself confirms it has no repos attached at all:

```
$ gh api orgs/Mininglamp-OSS/teams/cli-maintainers --jq '{slug, members_count, repos_count}'
{"slug":"cli-maintainers","members_count":3,"repos_count":0}
```

Result: even after fixing P0-1, the new owner line would silently produce no review requests, because the team is not attached to the repo. The `repos/.../codeowners/errors` endpoint currently returns `[]`, but that endpoint only evaluates the first-found CODEOWNERS file — once the root file becomes the active one (or `.github/CODEOWNERS` is rewritten to use this team), the missing-access error will surface there.

**Required fix:** Add the `cli-maintainers` team to the `octo-cli` repository with at least Write (or Triage at minimum) permission **before** the CODEOWNERS rule starts referencing it, otherwise the rule will be flagged as invalid and reviews will not be auto-requested. This is an org/admin action, not a code change, but it must happen as part of landing this PR.

## Nits / suggestions (non-blocking)

- The header comment says "Managed via GitHub Team membership — do not edit directly." That guidance is good, but it would be clearer to also link to the docs page (`https://docs.github.com/en/repositories/.../about-code-owners`) for consistency with the existing `.github/CODEOWNERS` style.
- Consider adding a follow-up rule for higher-risk paths (e.g. release tooling, CI workflows) so those still escalate to `@maintainers`, rather than the entire repo being owned by a single sub-team. Not required for this PR.

## What would unblock approval

1. Drop the new root `CODEOWNERS` and instead update `.github/CODEOWNERS` to point to `@Mininglamp-OSS/cli-maintainers` (or whatever ownership split is desired). Verify no other `CODEOWNERS` location remains.
2. Ensure `@Mininglamp-OSS/cli-maintainers` is added to `Mininglamp-OSS/octo-cli` with Write permission **before** the CODEOWNERS change is merged.
3. After re-push, sanity-check with:
   ```
   gh api repos/Mininglamp-OSS/octo-cli/codeowners/errors?ref=<head> --jq '.errors'
   gh api repos/Mininglamp-OSS/octo-cli/pulls/11/requested_reviewers --jq '.teams[].slug'
   ```
   The first should be `[]`; the second should list `cli-maintainers`.

PR #15, opened 2026-05-27
fix: move permissions to workflow top level for pr-contributor-welcome

lml2468:
## 问题

`.github/workflows/pr-contributor-welcome.yml` 存在权限配置 bug：

- 顶层 `permissions: {}` 清空了所有权限
- job 级别的 `permissions: issues: write` 对 reusable workflow 调用（`uses:`）无效
- 导致 `GITHUB_TOKEN` 没有写权限，发欢迎评论时出现 403 错误

## 修复

将权限配置移至 workflow 顶层，并移除 job 级别的 `permissions` 块：

```yaml
# 修复前（错误写法）
permissions: {}

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
    permissions:
      issues: write

# 修复后（正确写法）
permissions:
  issues: write

jobs:
  welcome:
    uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
```

## 参考

GitHub 文档：[reusable workflow 中的权限继承](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow) — job 级别的 `permissions` 对 `uses:` 类型的 job 不适用，必须在 workflow 顶层设置。

### DIFF
--- a/.github/workflows/pr-contributor-welcome.yml
+++ b/.github/workflows/pr-contributor-welcome.yml
@@ -4,10 +4,9 @@ on:
   pull_request_target: # zizmor: ignore[dangerous-triggers] metadata-only automation; no PR code executed
     types: [opened]
 
-permissions: {}
+permissions:
+  issues: write
 
 jobs:
   welcome:
     uses: Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml@main
-    permissions:
-      issues: write


yujiawei (approved):
# Code Review — PR #15 (octo-cli)

## Summary

Minimal, surgical fix for a real GitHub Actions permission bug. The diff moves `permissions: issues: write` from the calling job to the workflow top level and drops the now-redundant empty `permissions: {}` ceiling. Verified against the GitHub reusable-workflow permission model and the reusable callee's actual API needs.

## Verification

### 1. Root cause is real and correctly diagnosed ✅

Before this PR, the caller had:

```yaml
permissions: {}            # workflow-level ceiling = nothing
jobs:
  welcome:
    uses: ...reusable-pr-contributor-welcome.yml@main
    permissions:           # job-level on a `uses:` job
      issues: write
```

Two compounding issues:

- **Workflow-level `permissions: {}` is the maximum** the `GITHUB_TOKEN` can ever be granted in this run. Job-level `permissions` on any job — including a reusable-workflow caller — can only **narrow** that ceiling, never elevate it. With the ceiling at empty, `issues: write` at the job level had nothing to grant.
- Net effect: `GITHUB_TOKEN` reaches the reusable workflow with no `issues` scope, and `POST /repos/{owner}/{repo}/issues/{number}/comments` returns 403 — which matches the live failure on this PR run: `Unhandled error: HttpError: Resource not accessible by integration` ([run 26491739152](https://github.com/Mininglamp-OSS/octo-cli/actions/runs/26491739152)).

### 2. Fix is correct ✅

After this PR:

```yaml
permissions:
  issues: write            # workflow-level ceiling now grants issues:write
jobs:
  welcome:
    uses: ...reusable-pr-contributor-welcome.yml@main
```

The caller workflow ceiling now permits `issues: write`. The reusable callee defines `permissions: {}` at its own workflow level and `permissions: issues: write` at its job level, which is the documented pattern for reusable workflows — the called workflow's job-level permission is honored as long as the caller granted at least that scope. After merge, future `pull_request_target: opened` runs will run the fixed file from `main` (since `pull_request_target` always uses the base ref), so the welcome comment will succeed.

Note: the currently-failing check on this PR is **expected and not a regression** — `pull_request_target` ran the still-broken workflow from `main`, not the PR's version. Merging this PR is what flips main green for the next first-time contributor.

### 3. Scope of permission is minimal and appropriate ✅

I checked what API calls the reusable workflow actually makes (in `Mininglamp-OSS/.github/.github/workflows/reusable-pr-contributor-welcome.yml`):

| API call | Required scope |
| --- | --- |
| `GET /orgs/{org}/members/{username}` | none (public) |
| `search.issuesAndPullRequests` | none |
| `issues.listComments` | `issues: read` (covered by write) |
| `issues.createComment` | `issues: write` ✅ |

PR comments are issue comments, so `issues: write` alone is sufficient — `pull-requests: write` is **not** needed for posting the welcome comment. The minimum-privilege principle is respected.

### 4. Security review of `pull_request_target` ✅

`pull_request_target` runs in the context of the base repo with access to the base-repo `GITHUB_TOKEN`, which is normally dangerous when paired with `actions/checkout`. The reusable callee explicitly **does not check out PR code** and the file already carries the appropriate `# zizmor: ignore[dangerous-triggers] metadata-only automation; no PR code executed` annotation. The new permissions are narrow (`issues: write` only — no `contents`, no `pull-requests: write`), so this PR does not widen the attack surface beyond what's needed.

## Findings (P0 / P1 / P2)

None.

## Suggestions (non-blocking)

- **(nit)** Consider adding a one-line comment above the workflow-level `permissions:` block explaining the constraint, e.g. `# Must be at workflow top level: reusable-workflow caller jobs cannot elevate above the workflow ceiling.` This preserves the lesson learned so a future contributor doesn't re-introduce the bug by "tidying" permissions back to job level. Strictly optional — the PR description already documents it.

## Verdict

This is a textbook small, well-scoped fix backed by a correct root-cause analysis and a documented reference. No correctness, security, or scope concerns.

Jerry-Xin (approved):
This PR is in scope for `Mininglamp-OSS/octo-cli` and the workflow remains narrowly scoped to PR welcome automation.

💬 Non-blocking

🟡 Warning — `.github/workflows/pr-contributor-welcome.yml:7`: Moving `issues: write` to workflow level is safe today because this workflow has only one job, but it broadens the default permission for any future jobs added to this file. The existing repository pattern in workflows like `labeler.yml` and `issue-welcome.yml` keeps permissions at the reusable-workflow call job. Consider keeping that pattern if future jobs are added.

🔵 Suggestion — The PR description says job-level `permissions` do not apply to reusable workflow calls, but GitHub Actions supports `permissions` on jobs that call reusable workflows. The change is still acceptable, but the rationale may be misleading for future maintainers.

✅ Highlights

🔵 The resulting workflow grants only `issues: write`, which matches the called reusable workflow’s need to create an issue comment on the PR.

🔵 The `pull_request_target` usage remains constrained to metadata-only automation and does not check out or execute PR code.

Validation note: I inspected the local workflow context and the called reusable workflow. `actionlint` and `zizmor` were not installed in this environment, so I could not run those linters locally.

PR #16, opened 2026-05-27
feat(cmd): add octo skills command for embedded skill discovery

liuooo:
## Summary

Add an `octo skills` command and embed the agent `SKILL.md` docs into the
binary, so a released `octo` ships its skills and can list / print / export
them with no network and no token.

## Changes

- **skills/skills.go** — `//go:embed */SKILL.md` exposes every bundled skill as
  an `embed.FS`.
- **cmd/skills.go** — three modes: `octo skills` (list name+description),
  `octo skills <name>` (print one skill's full content), and
  `octo skills --install <dir>` (write all to `<dir>/<name>/SKILL.md`).
  Combining a name with `--install` is rejected with a validation error rather
  than silently ignoring the name.
- **cmd/root.go** — register the command and add `skills` to the
  skipValidation list (the command is offline, like `schema` / `config`).
- **cmd/skills_test.go** — unit tests for the frontmatter parser, the loader
  (sorted, non-empty, contains `octo-shared`), and all three command modes,
  including the unknown-skill and name-plus-`--install` error paths.
- **README.md** — document the command under Agent Skills.

## Motivation

Released binaries didn't ship the agent `SKILL.md` docs. `//go:embed` makes the
binary self-contained, and `octo skills --install` lets an agent runtime
extract them on demand — no separate files to distribute.

## Testing

- [x] `go vet ./...` passes
- [x] `go test ./...` passes (new `cmd/skills_test.go`)
- [x] `go build ./...` passes
- Smoke-tested the built binary: list / show / install exit 0; unknown skill
  and name+`--install` exit 2 with a `VALIDATION_ERROR` envelope; list runs
  without a token (skipValidation).

## Checklist

- [x] New commands added to README Command Reference table (Agent Skills section)
- [ ] CLAUDE.md command list updated — `skills` could be added to the
  hand-written-leaves line; deferred to keep this PR scoped strictly to the command
- [x] New features include tests (`cmd/skills_test.go`)
- [x] No unrelated changes included — `.goreleaser.yaml` archive bundling and the
  `.gitignore dist/` change were deliberately excluded as separate concerns

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,14 @@ Machine-readable usage docs for AI Agents live under [`skills/`](./skills/):
   threads, event polling.
 - [`octo-files`](./skills/octo-files/SKILL.md) — files and bot housekeeping.
 
+These docs are also **embedded in the binary**, so a released `octo` ships them:
+
+```bash
+octo skills                       # list embedded skills
+octo skills octo-messaging        # print one skill (name, description, content)
+octo skills --install ~/.config/octo/skills   # write all SKILL.md to a dir
+```
+
 ## Shell Completion
 
 ```bash
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -49,6 +49,7 @@ func NewRootCmd(f *cmdutil.Factory) *cobra.Command {
 	root.AddCommand(newVersionCmd(f))
 	root.AddCommand(newAPICmd(f))
 	root.AddCommand(newConfigCmd(f))
+	root.AddCommand(newSkillsCmd(f))
 	service.RegisterServiceCommands(root, f)
 
 	return root
@@ -61,7 +62,7 @@ func NewRootCmd(f *cmdutil.Factory) *cobra.Command {
 func skipValidation(cmd *cobra.Command) bool {
 	for c := cmd; c != nil; c = c.Parent() {
 		switch c.Name() {
-		case "version", "help", "schema", "config", "completion", "":
+		case "version", "help", "schema", "config", "completion", "skills", "":
 			return true
 		}
 	}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,8 +1,8 @@
 // Package cmd is the command tree. It wires cobra commands with the Factory DI
 // container and holds root-level persistent flags. Service-domain commands are
 // auto-registered from the embedded OpenAPI registry via cmd/service — the
-// only hand-written leaves are `schema`, `version`, and `api` (generic
-// passthrough).
+// hand-written leaves are `schema`, `version`, `api` (generic passthrough),
+// `config`, and `skills`.
 package cmd
 
 import (
--- a/cmd/skills.go
+++ b/cmd/skills.go
@@ -86,6 +86,14 @@ func newSkillsCmd(f *cmdutil.Factory) *cobra.Command {
   octo skills --install <dir>    write every skill to <dir>/<name>/SKILL.md`,
 		Args: cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if cmd.Flags().Changed("install") && install == "" {
+				ee := output.ErrValidation(
+					"--install requires a directory",
+					"pass a target like --install ~/.config/octo/skills",
+				)
+				_ = f.EmitError(ee) //nolint:errcheck // best-effort emit before returning err
+				return ee
+			}
 			entries, err := loadSkills()
 			if err != nil {
 				_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
--- a/cmd/skills.go
+++ b/cmd/skills.go
@@ -1,0 +1,178 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Mininglamp-OSS/octo-cli/internal/cmdutil"
+	"github.com/Mininglamp-OSS/octo-cli/internal/output"
+	"github.com/Mininglamp-OSS/octo-cli/skills"
+)
+
+type skillEntry struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	path        string // embed path, e.g. "octo-messaging/SKILL.md"
+	content     []byte
+}
+
+// loadSkills reads every embedded SKILL.md and parses its frontmatter
+// description. Sorted by name for deterministic output.
+func loadSkills() ([]skillEntry, error) {
+	paths, err := fs.Glob(skills.FS, "*/SKILL.md")
+	if err != nil {
+		return nil, err
+	}
+	out := make([]skillEntry, 0, len(paths))
+	for _, p := range paths {
+		b, err := skills.FS.ReadFile(p)
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, skillEntry{
+			Name:        strings.SplitN(p, "/", 2)[0],
+			Description: parseSkillDescription(b),
+			path:        p,
+			content:     b,
+		})
+	}
+	sort.Slice(out, func(i, j int) bool { return out[i].Name < out[j].Name })
+	return out, nil
+}
+
+// parseSkillDescription pulls `description:` out of the leading `---` YAML
+// frontmatter block with a line scan (no YAML dependency). It assumes a
+// single-line `description:` value — the shape every bundled SKILL.md uses;
+// multi-line or block-scalar descriptions are not supported.
+func parseSkillDescription(b []byte) string {
+	inFrontmatter := false
+	for _, ln := range strings.Split(string(b), "\n") {
+		t := strings.TrimSpace(ln)
+		if t == "---" {
+			if inFrontmatter {
+				break
+			}
+			inFrontmatter = true
+			continue
+		}
+		if inFrontmatter && strings.HasPrefix(t, "description:") {
+			return strings.TrimSpace(strings.TrimPrefix(t, "description:"))
+		}
+	}
+	return ""
+}
+
+// newSkillsCmd returns `octo skills`. Three mutually exclusive modes:
+//   - `octo skills`: list embedded skills (name + description).
+//   - `octo skills <name>`: print one skill's name, description, and content.
+//   - `octo skills --install <dir>`: write every skill to <dir>/<name>/SKILL.md.
+func newSkillsCmd(f *cmdutil.Factory) *cobra.Command {
+	var install string
+
+	cmd := &cobra.Command{
+		Use:   "skills [name]",
+		Short: "List, print, or install the embedded agent skill docs",
+		Long: `Agent-facing SKILL.md docs are embedded in this binary.
+
+  octo skills                    list available skills
+  octo skills <name>             print one skill (name, description, content)
+  octo skills --install <dir>    write every skill to <dir>/<name>/SKILL.md`,
+		Args: cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			entries, err := loadSkills()
+			if err != nil {
+				_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
+				return err
+			}
+			switch {
+			case install != "":
+				if len(args) == 1 {
+					ee := output.ErrValidation(
+						"cannot combine a skill name with --install",
+						"drop the name to install all skills, or drop --install to print one",
+					)
+					_ = f.EmitError(ee) //nolint:errcheck // best-effort emit before returning err
+					return ee
+				}
+				return runSkillsInstall(f, entries, install)
+			case len(args) == 1:
+				return runSkillsShow(f, entries, args[0])
+			default:
+				return runSkillsList(f, entries)
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&install, "install", "", "write all skills to this directory (<dir>/<name>/SKILL.md)")
+	return cmd
+}
+
+// runSkillsList emits every skill as {name, description}.
+func runSkillsList(f *cmdutil.Factory, entries []skillEntry) error {
+	buf, err := json.Marshal(map[string]any{"skills": entries})
+	if err != nil {
+		_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
+		return err
+	}
+	return f.EmitSuccess(buf)
+}
+
+// runSkillsShow emits one skill's full content, or a validation error naming
+// the available skills when name doesn't match.
+func runSkillsShow(f *cmdutil.Factory, entries []skillEntry, name string) error {
+	for _, s := range entries {
+		if s.Name == name {
+			buf, err := json.Marshal(map[string]any{
+				"name":        s.Name,
+				"description": s.Description,
+				"content":     string(s.content),
+			})
+			if err != nil {
+				_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
+				return err
+			}
+			return f.EmitSuccess(buf)
+		}
+	}
+	names := make([]string, len(entries))
+	for i, s := range entries {
+		names[i] = s.Name
+	}
+	ee := output.ErrValidation(
+		fmt.Sprintf("unknown skill %q", name),
+		fmt.Sprintf("available: %s", strings.Join(names, ", ")),
+	)
+	_ = f.EmitError(ee) //nolint:errcheck // best-effort emit before returning err
+	return ee
+}
+
+// runSkillsInstall writes every skill to <dir>/<name>/SKILL.md and reports the
+// files written.
+func runSkillsInstall(f *cmdutil.Factory, entries []skillEntry, dir string) error {
+	written := make([]string, 0, len(entries))
+	for _, s := range entries {
+		dst := filepath.Join(dir, s.path)
+		if err := os.MkdirAll(filepath.Dir(dst), 0o755); err != nil {
+			_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
+			return err
+		}
+		if err := os.WriteFile(dst, s.content, 0o644); err != nil {
+			_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
+			return err
+		}
+		written = append(written, dst)
+	}
+	buf, err := json.Marshal(map[string]any{"dir": dir, "installed": written})
+	if err != nil {
+		_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
+		return err
+	}
+	return f.EmitSuccess(buf)
+}
--- a/cmd/skills_test.go
+++ b/cmd/skills_test.go
@@ -1,0 +1,173 @@
+package cmd
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-cli/internal/config"
+)
+
+func TestParseSkillDescription(t *testing.T) {
+	cases := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{"basic", "---\nname: x\ndescription: hello world\n---\nbody", "hello world"},
+		{"trims surrounding space", "---\ndescription:   spaced   \n---", "spaced"},
+		{"no frontmatter", "# title\ndescription: ignored", ""},
+		{"description only in body is ignored", "---\nname: x\n---\ndescription: in body", ""},
+		{"empty input", "", ""},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			if got := parseSkillDescription([]byte(c.in)); got != c.want {
+				t.Errorf("parseSkillDescription(%q) = %q, want %q", c.in, got, c.want)
+			}
+		})
+	}
+}
+
+func TestLoadSkills(t *testing.T) {
+	entries, err := loadSkills()
+	if err != nil {
+		t.Fatalf("loadSkills: %v", err)
+	}
+	if len(entries) == 0 {
+		t.Fatal("expected at least one embedded skill")
+	}
+	for i, s := range entries {
+		if s.Name == "" || s.Description == "" || len(s.content) == 0 {
+			t.Errorf("entry %d incomplete: name=%q desc=%q contentLen=%d", i, s.Name, s.Description, len(s.content))
+		}
+		if i > 0 && entries[i-1].Name >= s.Name {
+			t.Errorf("entries not sorted ascending: %q before %q", entries[i-1].Name, s.Name)
+		}
+	}
+	// octo-shared is the foundational skill; it must always be embedded.
+	if !hasSkill(entries, "octo-shared") {
+		names := make([]string, len(entries))
+		for i, s := range entries {
+			names[i] = s.Name
+		}
+		t.Errorf("octo-shared not found among %v", names)
+	}
+}
+
+func hasSkill(entries []skillEntry, name string) bool {
+	for _, s := range entries {
+		if s.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+// A config without a token must still let `octo skills` run — the command is
+// offline and on the skipValidation list.
+func TestCmd_SkillsList(t *testing.T) {
+	f := newTestFactoryWithReg()
+	f.SetConfig(&config.Config{Format: "json"})
+
+	out, _, err := execRoot(t, f, "skills")
+	if err != nil {
+		t.Fatalf("skills: %v", err)
+	}
+	var env map[string]any
+	if jerr := json.Unmarshal([]byte(out), &env); jerr != nil {
+		t.Fatalf("unmarshal: %v\n%s", jerr, out)
+	}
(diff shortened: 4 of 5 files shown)


lml2468 (dismissed):
## Summary

Adds `octo skills` command for embedded skill discovery — list, show one, or install all to disk. Clean use of Go's `embed.FS`, proper error handling, and comprehensive tests (173 lines covering unit + command-level).

## Findings

### 🔴 Blocking
None.

### 🟡 Non-blocking

1. **`parseSkillDescription` quoted strings** — if any future SKILL.md uses `description: "quoted: value"`, the quotes would be included verbatim in the output. Current bundled skills don't hit this, and the comment documents the limitation, so low risk.

### ✅ Highlights

- Unexported fields (`path`, `content`) correctly excluded from JSON serialization — no info leak.
- `skipValidation` addition means the command works offline without auth. Correct for a local-only discovery tool.
- Tests cover all modes including conflict (`--install` + name) and unknown skill error paths.
- `fs.Glob` over embedded FS guarantees no path traversal risk on `--install`.
- Error handling is consistent: `f.EmitError` + return err throughout.

## Verdict
APPROVED — well-structured feature with no security or correctness concerns.

Jerry-Xin (approved):
Summary: The PR is in scope for `octo-cli`, adds a relevant offline `octo skills` command, and the implementation is consistent with the project’s command/envelope patterns.

💬 Non-blocking
- 🔵 Suggestion — [cmd/root.go](/Users/mlamp/.openclaw/tmp/pr-review-gclz09uo/repo/cmd/root.go:4): The package comment still says the only hand-written leaves are `schema`, `version`, and `api`; it was already stale for `config`, and this PR adds `skills`. Consider updating it in this PR or the follow-up noted in the checklist.
- 🔵 Suggestion — [cmd/skills.go](/Users/mlamp/.openclaw/tmp/pr-review-gclz09uo/repo/cmd/skills.go:50): The frontmatter parser is intentionally narrow and matches current bundled skills. If future skill docs may use quoted YAML strings or block scalars, this should either gain parser support or have a test documenting the unsupported cases.
- 🔵 Suggestion — [cmd/skills.go](/Users/mlamp/.openclaw/tmp/pr-review-gclz09uo/repo/cmd/skills.go:166): `--install` overwrites existing `SKILL.md` files silently. That is reasonable for an export-style command, but if users may edit installed skills locally, a future `--force` or “skipped existing” mode could prevent surprises.

✅ Highlights
- Relevant project change: the PR extends the CLI with embedded repo-local skill docs and documents the command.
- `skills` is correctly added to validation skipping for offline/no-token use.
- Validation errors for unknown skills and `name + --install` follow the existing structured error envelope.
- Tests cover parsing, loading, list/show/install modes, and error paths.

Verification: `go test ./...`, `go vet ./...`, and `go build ./...` all pass locally.

yujiawei (approved):
# Code Review — PR #16 (octo-cli)

## Summary

This PR adds an `octo skills` command backed by `//go:embed */SKILL.md`, exposing three modes (list / show / install) so a released binary can ship and emit its agent-facing skill docs without any network or token. The change is well-scoped (≈370 LOC across 5 files, including 173 LOC of unit tests), local in surface area, and consistent with existing offline commands (`schema`, `config`, `version`).

Verified locally at head `03a8a36`: `go build ./...`, `go vet ./...`, and `go test ./cmd/...` all pass; the smoke flow (list / show / install / unknown skill / name+`--install`) produces the documented envelopes.

**Verdict: APPROVED.** No P0/P1 findings. A handful of P2 polish suggestions and items worth a human eye for the `security_sensitive` classification — none block merge.

## 1. Verification

| Item | Status | Evidence |
|---|---|---|
| Command compiles, vets, and tests pass | ✅ | `go build/vet/test ./cmd/...` clean at `03a8a36` |
| Embedded FS layout matches glob | ✅ | `skills/skills.go:10` uses `//go:embed */SKILL.md`; `skills/{octo-files,octo-matter,octo-messaging,octo-shared}/SKILL.md` all match |
| Three documented modes are mutually exclusive | ✅ | `cmd/skills.go:140-156` switch handles `install != ""` (with name-rejection at `:142-149`), then single-name show, then list |
| Offline command added to `skipValidation` | ✅ | `cmd/root.go:65` |
| Frontmatter description parser handles edge cases (no FM, body-only `description:`, empty input, surrounding spaces) | ✅ | `cmd/skills_test.go:44-69` |
| Loader is deterministic / sorted | ✅ | `cmd/skills.go:92` + assertion at `cmd/skills_test.go:84-87` |
| `--install` writes expected on-disk layout | ✅ | `cmd/skills.go:206-217` + `TestCmd_SkillsInstall` (`cmd/skills_test.go:147-162`) |
| Name + `--install` returns `VALIDATION_ERROR` | ✅ | `cmd/skills.go:142-149` + `TestCmd_SkillsInstallRejectsName` (`cmd/skills_test.go:165-176`) |
| Unknown skill returns `VALIDATION_ERROR` and lists available skills | ✅ | `cmd/skills.go:194-198` + `TestCmd_SkillsUnknown` |
| `octo skills` runs without `OCTO_BOT_TOKEN` | ✅ | `TestCmd_SkillsList` uses an empty `config.Config{Format: "json"}`; passes |
| Path-traversal safety in `--install` | ✅ | `s.path` is sourced exclusively from the embedded FS via `fs.Glob(skills.FS, "*/SKILL.md")` and cannot contain `..` or absolute components; only `dir` is user-controlled and each write lands at `<dir>/<embedded-skill>/SKILL.md` |

## 2. Findings

No P0 or P1 issues.

### P2 — `--install` follows existing symlinks at the destination (`cmd/skills.go:212`)

`os.WriteFile` opens with `O_WRONLY|O_CREATE|O_TRUNC`, which **follows** symlinks at the destination path. If a malicious or stale `<dir>/<skill>/SKILL.md` symlink points at, e.g., `~/.bashrc` or `/etc/hosts`, the install would clobber the link target rather than refusing or replacing the link.

This is a low-severity, classic "extract into a directory" risk shared by tools like `tar -x` and `unzip`, and the user picks `<dir>` themselves — but since the PR is flagged `security_sensitive`, it is worth either:

- using `os.OpenFile(dst, O_WRONLY|O_CREATE|O_TRUNC|O_NOFOLLOW, 0o644)` to refuse symlink targets, **or**
- preceding each write with an `os.Lstat(dst)` and removing it if it is a symlink.

Either is a 5-line change. Not a merge blocker.

### P2 — Partial-failure state on `--install` (`cmd/skills.go:204-217`)

If the loop fails on skill *N+1* after writing skills 1..N, the JSON envelope is never emitted (early return), but the partially-written files remain on disk and the user has no machine-readable summary of what landed. Two reasonable options:

- Stage to a temp directory and `os.Rename` into place at the end (atomic install).
- On error, include the partial `installed` list inside the `ExitError`'s `Detail` field so agent consumers can clean up.

Optional polish — the simpler "abort and leak partial files" behavior is acceptable for a CLI install command, but a one-line note in the help text ("partial writes may remain on failure") would close the surprise gap.

### P2 — `--install ""` silently degrades to list mode (`cmd/skills.go:141`)

`install != ""` means an explicit empty string passed via `--install ""` is treated as "no flag set" and the command silently lists skills instead. Cheap to harden:

```go
if cmd.Flags().Changed("install") && install == "" {
    return output.ErrValidation("--install requires a directory", "pass a target like --install ~/.config/octo/skills")
}
```

### P3 / nit — Description parser ignores unquoted multi-line YAML

`parseSkillDescription` (`cmd/skills.go:100-115`) stops at the first newline — block scalars (`description: |`) or folded values would be silently truncated to "" or to the marker line. The bundled SKILL.md files all use single-line descriptions today (verified), and the in-code comment at `cmd/skills.go:96-99` is honest about this assumption, so this is documentation-only. Worth a one-line CI guard or a parser-level "if value starts with `|`/`>`, return everything until next top-level YAML key" if multi-line descriptions ever land — but no change needed for this PR.

### P3 / nit — `runSkillsList` description in `octo skills --help`

The help text for the `--install` flag (`cmd/skills.go:159`) is good. The command `Long` description (`cmd/skills.go:128-132`) does not mention that the list output is JSON-only by default — minor since it matches the rest of the CLI, but a one-line "output is the standard JSON envelope; pipe to `jq` for filtering" would help first-time users.

### P3 / nit — `runSkillsList` builds an envelope with `skills` as a list of structs

`json.Marshal(map[string]any{"skills": entries})` (`cmd/skills.go:165`) marshals `skillEntry` directly, exposing only the JSON-tagged `Name`/`Description` fields and dropping the unexported `path` and `content` (Go's encoder ignores unexported fields). This is correct and intentional, but a quick reader might suspect the unexported fields would leak — adding a one-line code comment confirming the JSON shape ("only `name`+`description` are serialized; `path`/`content` are intentionally unexported") would save a future maintainer five minutes.

## 3. Suggestions

- (Optional) tighten `os.WriteFile` to `os.OpenFile` with `O_NOFOLLOW` for the install path, or `Lstat`+remove symlinks first. Smallest, highest-signal hardening for the `security_sensitive` flag.
- (Optional) reject `--install ""` with `VALIDATION_ERROR`.
- (Optional) document partial-failure semantics in the command `Long` text.
- The PR description notes that `.goreleaser.yaml` archive bundling and the `CLAUDE.md` command-list update were deliberately deferred. Both are reasonable scope boundaries; please open a follow-up to add `skills` to the hand-written-leaves comment in `cmd/root.go:1-5` so the package doc stays accurate.

## 4. Items Recommended for Human Review (security_sensitive)

The `security_sensitive` classification appears to have triggered on the `token` keyword (the command joins `skipValidation` and bypasses `OCTO_BOT_TOKEN` validation). Items a human owner should sanity-check:

1. **Token-validation bypass is appropriate.** The command is purely offline (reads only the embedded `embed.FS`, makes no HTTP calls, touches no credentials). Adding `skills` to `skipValidation` (`cmd/root.go:65`) is the correct mirror of `schema`/`config`/`version`. ✅ Confirmed by reading the diff; no auth surface introduced.
2. **No new attack surface in the embed FS itself.** `//go:embed */SKILL.md` only matches files committed to `skills/<name>/SKILL.md`. Anyone who can land a `SKILL.md` already has merge access; the embed does not pull from any external source at build time.
3. **Symlink-overwrite risk on `--install`** (see P2 above). Worth a human eye since the command writes to a user-specified directory; current behavior matches `tar -x`-style tools but could be hardened with one extra `O_NOFOLLOW` flag.
4. **No path-traversal vector via the skill name.** The user-supplied `<name>` argument is only used for an exact-match lookup against `skillEntry.Name` (`cmd/skills.go:177`). It is **not** joined into any filesystem path, so `octo skills ../../etc/passwd` is just an unknown-skill error.
5. **README install example writes under `~/.config/octo/skills`** (`README.md:217`) — a sensible per-user location. Confirm there is no documented flow that runs `octo skills --install <dir>` as root into a system directory; if there is, the symlink-hardening suggestion becomes more important.

Overall this is a tightly-scoped, well-tested, low-risk change. Recommend merge after a quick human read of items (3) and (5) above.

Jerry-Xin (dismissed):
Project relevance gate passed; this PR extends the existing `octo-cli` skill documentation workflow and is in scope. No blocking issues found.

💬 Non-blocking

- 🔵 Suggestion: [cmd/root.go](/Users/mlamp/.openclaw/tmp/pr-review-jcjp7t08/repo/cmd/root.go:3) has a stale package comment saying the only hand-written leaves are `schema`, `version`, and `api`. This was already incomplete for `config`, and this PR adds `skills`; updating it would avoid misleading future maintainers.
- 🔵 Suggestion: [CLAUDE.md](/Users/mlamp/.openclaw/tmp/pr-review-jcjp7t08/repo/CLAUDE.md:24) still omits `skills` from the hand-written command list. The PR notes this as deferred, so not blocking, but it should be kept in sync soon.
- 🔵 Suggestion: [cmd/skills.go](/Users/mlamp/.openclaw/tmp/pr-review-jcjp7t08/repo/cmd/skills.go:95) treats an explicitly empty `--install ""` the same as no install flag. That is a minor edge case, but validating `cmd.Flags().Changed("install") && install == ""` would make the mode selection stricter.

✅ Highlights

- The command is correctly registered and added to skip-validation for offline use.
- Embedded skill loading is deterministic and covered by tests.
- Unknown skill and name-plus-install validation paths use the existing structured error envelope.
- `--install` writes the expected `<dir>/<skill>/SKILL.md` layout from embedded content.

Verification run: `go test ./...`, `go vet ./...`, and `go build ./...` all passed.

yujiawei (dismissed):
# Code Review — PR #16 (octo-cli)

## Summary

Adds an offline `octo skills` command that exposes the bundled `SKILL.md` docs from a `//go:embed` FS via three modes (list / show / install). The change is small (~370 LOC, mostly tests + a new file), self-contained, and follows the project's existing factory + error-envelope patterns. `go vet ./...` and `go test ./cmd/...` pass on the head SHA.

**Verdict: APPROVED.** No correctness, security, or build-break blockers. A few non-blocking suggestions below.

## What I checked

- `cmd/skills.go` — three-mode dispatch, frontmatter parser, install side effects.
- `cmd/root.go` — registration and the `skipValidation` change.
- `skills/skills.go` — `embed.FS` declaration.
- `cmd/skills_test.go` — coverage of the three modes and error paths.
- Compatibility with the existing `EmitSuccess` / `EmitError` / `ErrValidation` envelope contract in `internal/output` and `internal/cmdutil`.
- Local `go vet ./...` and `go test ./cmd/...` runs.

## Verification — claims in the PR description

| Claim | Status | Evidence |
| --- | --- | --- |
| `//go:embed */SKILL.md` exposes every bundled skill | ✅ | `skills/skills.go` declares `var FS embed.FS` with the glob; `fs.Glob(skills.FS, "*/SKILL.md")` returns the four bundled directories under `skills/`. |
| Three modes, mutually exclusive | ✅ | `cmd/skills.go:88-110` switch-case; `--install` + name combination explicitly rejected with `ErrValidation`. |
| `skills` added to `skipValidation` so it runs token-less | ✅ | `cmd/root.go:65`. The command never touches the API client, so this is correct — same shape as `schema` / `config`. |
| `go vet ./...` and `go test ./...` pass | ✅ | Reproduced locally; no warnings. |
| Sorted, deterministic output | ✅ | `sort.Slice` on name; `TestLoadSkills` asserts ascending order. |

## Security review (PR is `security_sensitive`-flagged)

I gave the install path-handling and the `skipValidation` change extra scrutiny. Both are clean:

1. **`--install <dir>` is not vulnerable to path traversal.** `dst := filepath.Join(dir, s.path)` joins a user-supplied destination with `s.path`, but `s.path` comes from `fs.Glob(skills.FS, "*/SKILL.md")` over a compile-time `embed.FS`. The set of paths is fixed at build time and cannot contain `..`. The user is choosing where to write their own files, which is the expected CLI behavior.
2. **`skipValidation` change is correct.** The command performs no network I/O, reads only embedded data, and writes only to a user-supplied directory. Adding it to the skip list mirrors the precedent of `schema` / `config` and avoids forcing a `OCTO_BOT_TOKEN` for an offline doc-export command.
3. **No secrets, tokens, or credentials are read or written.** The embedded SKILL.md contents are static documentation.
4. **File permissions** (`0o755` for dirs, `0o644` for files) match standard CLI behavior; no setuid/setgid concerns.

Nothing here needs human-only manual verification beyond a sanity check that no future SKILL.md will contain content the team doesn't want shipped in the binary.

## Findings

### P2 / Nits (non-blocking — fix-on-merge or follow-up at the author's discretion)

1. **`--install` silently overwrites existing files.** `runSkillsInstall` always calls `os.WriteFile` without checking for an existing destination. If a user has hand-edited `<dir>/octo-shared/SKILL.md`, it will be clobbered with no warning. Consider one of: (a) document this in the flag help, (b) skip files that already exist with matching contents (no-op), or (c) add an opt-in `--force` flag. Low priority since the command's purpose is to install the canonical bundled copy, but worth a one-line help-string note (`"… (overwrites existing files)"`).

2. **`parseSkillDescription` has implicit assumptions worth documenting or tightening.**
   - It splits the entire file with `strings.Split(string(b), "\n")`, which means a `\r\n` SKILL.md would yield trailing `\r` on every line, and `t == "---"` would never match. All four bundled skills use LF today, so this is not a current bug — but if a contributor ever saves a SKILL.md from a Windows editor it will silently degrade the description to `""`. A `strings.TrimRight(t, "\r")` or switching to `bufio.Scanner` would harden it.
   - A leading UTF-8 BOM on the first line would similarly defeat the `t == "---"` check. Same fix.
   - The doc-comment at `cmd/skills.go:50-53` already calls out the "single-line `description:`" assumption, which is good — extending it to mention "LF line endings, no BOM" would prevent future surprise.

3. **`--install ""` (empty string) silently degrades to list mode.** The switch at `cmd/skills.go:94-95` uses `install != ""`. If a script does `octo skills --install "$DIR"` and `$DIR` is unset, the user gets a list of skills back instead of an error. Considering treating empty string as a validation error, since it's almost certainly user error rather than intent.

4. **`runSkillsInstall` partial-failure leaves a half-written tree.** If the 3rd of 4 skills fails to write (e.g. disk full, permission flip mid-loop), the first two are already on disk and there's no rollback. This matches typical CLI tool behavior, but a one-line note in the command's `Long` help — "writes are not transactional; partial failures leave installed files in place" — would set expectations. Optional.

5. **`fs.Glob` returns paths in lexical order already**, so the explicit `sort.Slice` after the loop is defensive but not strictly necessary. Harmless and the test pins the contract — leaving as-is is fine.

### Test coverage gaps (non-blocking)

- No test for the partial-failure / clobber scenarios above.
- No test for `--install ""` (which currently silently switches to list mode).
- `TestParseSkillDescription` does not cover the CRLF/BOM cases. If you adopt the trim suggestion above, add a test row for `"---\r\ndescription: x\r\n---\r\n"`.

These are coverage-completeness suggestions, not regressions — current tests adequately exercise the documented behavior.

## Build / CI

- `go vet ./...` — clean.
- `go test ./cmd/...` — `ok` on `cmd`, `cmd/octo`, `cmd/service`. New `TestParseSkillDescription`, `TestLoadSkills`, `TestCmd_Skills*` all pass.
- `go build ./...` is implicit in `go test` and not separately exercised, but the test binary builds successfully.

## Recommendation

Merge as-is. The suggestions above are quality-of-life improvements that can ship in a follow-up; none of them block this PR's stated goal of letting a released `octo` binary export its bundled skill docs offline.

lml2468 (approved):
## Re-review (1f0902e1df87)

Previous non-blocking feedback addressed:
- ✅ Package comment in cmd/root.go updated (lists config + skills)
- ✅ `--install ""` now returns validation error instead of silently no-op
- ✅ New test `TestCmd_SkillsInstallEmptyDir` covers the edge case

APPROVED — clean polish commit.

yujiawei (approved):
# Code Review — PR #16 (octo-cli)

## Verdict
**APPROVED.** The change is small, self-contained, well-tested, and operationally benign. No blockers. A few P2 nits below worth addressing in this PR or a follow-up, plus one item the dispatcher labelled `needs-human-review` that I want a human to glance at.

## 1. Validation summary

| Item | Result | Evidence |
|---|---|---|
| Embed wires every `<skill>/SKILL.md` | ✅ | `skills/skills.go:10` (`//go:embed */SKILL.md`) |
| `octo skills` lists name + description | ✅ | `cmd/skills.go:147` (`runSkillsList`) |
| `octo skills <name>` prints one skill | ✅ | `cmd/skills.go:155` (`runSkillsShow`) |
| `octo skills --install <dir>` writes all | ✅ | `cmd/skills.go:174` (`runSkillsInstall`) |
| Name + `--install` rejected | ✅ | `cmd/skills.go:122-128` |
| `--install` with empty value rejected | ✅ | `cmd/skills.go:108-114` |
| Unknown skill returns validation error | ✅ | `cmd/skills.go:165-171` |
| `skills` added to `skipValidation` (offline command) | ✅ | `cmd/root.go:64` |
| Test coverage for parser, loader, all 3 modes, error paths | ✅ | `cmd/skills_test.go` (parser, `loadSkills`, list, show, unknown, install, empty-install, name+install) |
| README updated under Agent Skills | ✅ | `README.md:213-220` |
| Path traversal risk in `--install` | ✅ safe | `s.path` is from `fs.Glob` over an embedded FS; values are compile-time-controlled (e.g. `octo-messaging/SKILL.md`). User-supplied `dir` is the install target by design. |

## 2. Findings

### P0 / P1
None.

### P2 — frontmatter parser matches indented `description:` keys (`cmd/skills.go:99-107`)
```go
t := strings.TrimSpace(ln)
...
if inFrontmatter && strings.HasPrefix(t, "description:") {
    return strings.TrimSpace(strings.TrimPrefix(t, "description:"))
}
```
Because the line is `TrimSpace`'d before the prefix check, a nested `description:` key (e.g. under a `metadata:` parent) would be accepted as the top-level description. None of today's bundled `SKILL.md` files exercise this, but a future schema change could silently flip the wrong value into the listing. Two cheap fixes: either require the line to be unindented (`if !strings.HasPrefix(ln, " ") && !strings.HasPrefix(ln, "\t")` before the `HasPrefix(t, "description:")` check) or add a `nested description ignored` test case to lock in the intended behavior either way.

### P2 — `--install` silently overwrites existing files (`cmd/skills.go:184-191`)
```go
if err := os.WriteFile(dst, s.content, 0o644); err != nil {
```
If a user has hand-edited a previously installed `<dir>/<name>/SKILL.md`, a re-run of `--install` clobbers it with no warning. Not a security issue, but a destructive default. Options: document the overwrite behavior in the command's `Long` help, or add a `--force` flag (default no-overwrite). Small UX win for not much code.

### P2 — `description` field is not validated to be non-empty
`TestLoadSkills` already asserts every entry has a non-empty description, which is good. But `parseSkillDescription` returns `""` silently when the frontmatter is missing/malformed. In production loading, an empty description simply shows as a blank in `octo skills` output. A defensive option would be to fail fast in `loadSkills` if any embedded skill is missing a description — that turns "I forgot the frontmatter" into a build-time-test failure rather than a quiet UX regression. Optional.

### P2 — `f.EmitError(err) … return err` double-emit shape
Pattern repeats at `cmd/skills.go:111`, `127`, `136`, `150`, `160`, `170`, `181`, `186`, `192`. If cobra is wired to print the returned error to stderr while the Factory also emits the structured envelope, callers see the same error twice. This is consistent with how other commands in the tree appear to be written, so I'm flagging it as "follow existing convention" rather than asking for change here — but worth a sanity check that other leaf commands (`schema`, `config`) use the same idiom and the resulting stderr is acceptable in `--output text` mode.

### Nit — `cmd/root.go` package comment
Comment is updated to mention `skills`, but the README "Command Reference" tables (if there is one beyond the Agent Skills section) and the `CLAUDE.md` hand-written-leaves line aren't synced. The PR description acknowledges the `CLAUDE.md` deferral; that's fine. No action needed here.

## 3. Security note (`security_sensitive` routing)

The dispatcher labelled this `security_sensitive` and tagged `needs-human-review`. The keyword hit is almost certainly `install` (or `credential`/`token` from the README). The actual surface area is:

- **Read-only embed of static markdown.** Content is shipped at compile time; no network, no external input.
- **No token/credential handling.** `skills` is on `skipValidation`, so it explicitly does not call any auth path.
- **`--install` writes to a user-supplied directory.** `filepath.Join(dir, s.path)` with `s.path` controlled (`*/SKILL.md` glob over `embed.FS`), so the output path tree is fixed at build time. No traversal vector.
- **File modes:** `0o755` for dirs, `0o644` for files — standard.

I see no concrete security concern. The `needs-human-review` label is precautionary; a human reviewer can confirm and remove it after merge if desired.

## 4. Suggestions (non-blocking)

1. Address the indented-`description:` parser edge case in this PR (one-line fix + test) — cheapest of the three to land.
2. Decide on `--install` overwrite semantics: keep silent overwrite + document, or add `--force`. Either is fine, just be explicit.
3. Consider failing `loadSkills` fast on empty `Description` so a missing frontmatter is caught at test time rather than as a blank in user-facing output.

## 5. Approve rationale

Functionality matches the PR description, tests are thorough across happy + error paths (including the error envelopes), the embed wiring is the canonical Go idiom, and the only items I flagged are P2 polish. Nothing here blocks merge.

Jerry-Xin (approved):
Summary: This PR is relevant to `octo-cli`, fits the existing command/envelope architecture, and the implementation is ready to merge.

💬 Non-blocking

🔵 Suggestion: `cmd/skills.go:126` wraps list output as `{"skills": [...]}`. That is fine for JSON and `--jq`, but `--format table/csv/ndjson` renders as one `skills` column instead of one row per skill. If tabular list output matters, consider emitting the skill slice directly or teaching the formatter to unwrap this shape.

🔵 Suggestion: `cmd/skills.go:54-66` uses a deliberately narrow frontmatter parser. It works for the current bundled `SKILL.md` files, but future quoted YAML values or block-scalar descriptions would be returned literally or missed. Not a blocker given the documented constraint.

✅ Highlights

- `cmd/root.go:52` and `cmd/root.go:65` correctly register `skills` and skip credential validation for offline use.
- `skills/skills.go:10-11` embeds only `*/SKILL.md`, keeping the binary payload scoped.
- `cmd/skills.go:89-112` handles invalid `--install` combinations with structured validation errors.
- `cmd/skills_test.go:13-186` covers parser, loader, list/show/install flows, and error paths.

Verification: I reviewed the full PR diff and related command/output code, then ran `go test ./...`, `go vet ./...`, and `go build ./...`; all passed.

PR #17, opened 2026-05-27
feat(auth): encrypted credential profiles and octo auth commands

liuooo:
## Summary

Add encrypted credential storage and an `octo auth` command family so an AI
agent can use `octo` without ever putting a bot token on the command line.
Today a token can only come from `OCTO_BOT_TOKEN`, which leaks into the agent's
command string, shell history, and transcripts. Now a token is stored encrypted
on disk and referenced at runtime by `--bot-id` / `--profile`; the raw token
never appears in argv.

Zero server changes — this is purely the CLI credential layer.

## Changes

**Storage — `internal/authstore` (new)**
- `credentials.enc` (AES-256-GCM, tokens) + plaintext `config.json` (non-secret
  metadata) under `~/.octo-cli` (override `OCTO_CONFIG_DIR`); 0600 / 0700.
- Key = `SHA256(machineID‖salt)`; per-OS machine id (darwin/linux/windows) with
  a stable salt-only fallback on other platforms. A transient machine-id read
  failure surfaces rather than silently changing the key.
- `ActiveProfile` resolves a single profile from the selectors with a four-state
  outcome (found / none / ambiguous / missing). Saves write the token before the
  metadata (and deletes metadata before the token) so a crash never leaves a
  listed profile without a usable token.

**Credential chain — `internal/credential`**
- `FileProvider` implements the existing `Source` interface; the chain is now
  FileProvider → EnvProvider. `BotCredential` gains Profile / RobotID / BotKind.
- `MaskToken` / `TokenKind` centralise token display (kind prefix + head +
  `***` + tail, e.g. `app_ab***5678`) and classification.

**Commands — `cmd/auth.go` (new) + wiring**
- `octo auth login | status | logout | list`. `login` reads the token from a
  hidden, asterisk-masked prompt (and prompts for the bot id interactively on a
  terminal), or from stdin (`--with-token`) / `--token-file` — never argv. The
  masked reader accepts only printable ASCII and drops escape sequences so arrow
  keys / bracketed-paste markers can't pollute the token.
- `status` is whoami (one active identity with masked token + source, or
  `active:null` + a pointer); `list` is the catalog (never prints tokens).
  Neither errors on ambiguity.
- Persistent `--bot-id` (env `OCTO_BOT_ID`) and `--profile` selectors. The
  envelope `identity` is now always an object
  `{type, profile?, robot_id?, bot_kind?, source}` so the active bot is visible
  on every response. `config show` reflects the resolved credential and the
  dry-run `Authorization` header reuses the same masking.

**Docs** — `CLAUDE.md` (identity model, auth command group, `--bot-id` /
`--profile` flags, `OCTO_CONFIG_DIR` / `OCTO_BOT_ID` env, `x/term` dependency
exception) and the `octo-shared` skill (stored-profile login as the primary auth
path, selection, identity echo).

## Precedence and anti-misuse

`--bot-id`/`--profile` selector > sole stored profile > `OCTO_BOT_TOKEN`
(fallback only when no profile is stored). With two or more profiles and no
selector, runtime commands fail rather than silently guessing an identity; a
selector that matches nothing is a hard auth error. `OCTO_BOT_ID` is a selector,
not a secret — if it matches no profile the command fails rather than using a
raw `OCTO_BOT_TOKEN` under that id.

## Security / threat model (honest)

The trust boundary is the OS user account. The machine-derived key protects the
encrypted file against off-machine exposure (accidental commit, backup, cloud
sync), but any process running as the same user can decrypt it. Isolate
mutually-distrusting bots with separate OS users or `OCTO_CONFIG_DIR` values.
The CLI's guards (ambiguity = hard error, `identity` echo, `--bot-id` assertion)
prevent *accidental* wrong-identity use, not a malicious co-resident process.

## Dependencies

Adds `golang.org/x/term` (used only for the hidden token prompt), pinned to
v0.39.0 so the module keeps `go 1.24.0` to match CI.

## Testing

- [x] `gofmt`, `go vet ./...`, `go test ./...` all green; `internal/authstore`
  cross-compiled for linux / windows / other.
- [x] New tests: authstore round-trip, ciphertext-hides-token, file perms,
  four-state `ActiveProfile`, orphaned-token safety; `FileProvider` selection
  matrix (name / bot-id / both / mismatch / ambiguous / none); `MaskToken`;
  `cmd` auth login / status / list / logout; envelope `identity` object shape;
  `OCTO_CONFIG_DIR` test isolation.
- [x] Manually smoke-tested the built binary against a live backend: interactive
  login (hidden masked prompt; bracketed-paste markers and a mid-paste arrow key
  stripped, exact token stored), `auth status` / `list`, `group list` resolving
  the sole stored profile, and the precedence / ambiguity / mismatch exit codes
  (0 / 2 / 3).

## Checklist

- [x] English (code, comments, commit messages)
- [x] Conventional Commits
- [x] New behavior covered by tests
- [x] Docs updated (CLAUDE.md, `octo-shared` skill)
- [ ] No linked issue — this implements the design in
  `docs/credential-profile-design.md`; happy to convert to a tracking issue if
  maintainers prefer.


Jerry-Xin (changes requested):
## Summary

Well-architected credential management system with solid cryptographic design. The encrypted profile store, profile-switching logic, and security boundaries are all thoughtfully implemented. However, **CI is red** — golangci-lint reports 16 issues that must be fixed before merge.

## 🔴 Blocking

### 1. CI failure: golangci-lint (16 issues)

The `build & test` check fails on lint. All must be addressed:

**errcheck (8 instances)** — unchecked error returns:
- `cmd/auth.go:273` — `fmt.Fprint(f.IOStreams.ErrOut, "Paste bot token: ")`
- `cmd/auth.go:328` — `fmt.Fprint(out, "\r\n")`
- Plus 6 more `fmt.Fprint` calls in `readPasswordMasked` and `os.Remove`/`tmp.Close` in `atomicWrite`

Fix: Add `//nolint:errcheck` with justification for the terminal I/O cases (best-effort writes to stderr/tty), or use `_ = fmt.Fprint(...)`. For `atomicWrite` defer, the existing comment says "no-op after rename" — just suppress it:
```go
defer func() { _ = os.Remove(tmpName) }()
```

**gocyclo (2)** — high cyclomatic complexity:
- `cmd/auth.go:300` — `readPasswordMasked` (21 > 15)
- `internal/cmdutil/factory.go:71` — `NewDefaultFactory` (17 > 15)

Fix for `readPasswordMasked`: extract the escape-sequence state machine into a helper (e.g. `processEscByte(state, c) (newState, emit bool)`). For `NewDefaultFactory`: extract the credential-overlay logic into a named method.

**revive (4)** — exported const comments must start with the const name:
```go
// StatusFound: exactly one profile...    ← wrong
// StatusFound indicates exactly one...   ← correct
```

**gocritic (1)** — `ProfileMeta` passed by value is 80 bytes:
```go
func (s *Store) SaveProfile(name string, meta ProfileMeta, ...) // pass *ProfileMeta instead
```

**staticcheck (1)** — redundant embedded field selector in test:
```go
tf.Factory.cred = ...  →  tf.cred = ...
```

## 💬 Non-blocking

### 2. Key derivation — SHA-256 is functional but not ideal

`deriveKey()` uses `SHA256(machineID || salt)`. Since `machineID` is not a password (it's a high-entropy machine identifier), the computational-cost argument for PBKDF2/Argon2 doesn't apply. However, **HKDF** (RFC 5869) would be more formally correct for extracting a key from non-uniform keying material:

```go
hkdf.New(sha256.New, []byte(id), salt, []byte("octo-cli/credentials/v1"))
```

This adds domain separation (the `info` parameter) and is the standard construction for "derive a key from a shared secret + salt". Not blocking since the current approach is cryptographically sound for the stated threat model.

### 3. No migration path when machine ID changes

If the machine ID changes (hardware swap, VM migration, OS reinstall on Linux regenerating `/etc/machine-id`), all stored tokens become undecryptable. The error message says "corrupt, or key changed" — consider detecting this case specifically and suggesting `octo auth login` to re-store the tokens. Low priority since this is documented behavior.

### 4. Crash ordering correctness (positive note)

The comment-documented ordering invariant for `SaveProfile` (token first → metadata second) and `RemoveProfile` (metadata first → token second) is excellent — orphaned tokens are invisible and harmless, while the alternative (listed profile with no token) would cause runtime errors.

## ✅ Highlights

- **Cryptography**: AES-256-GCM with random nonce, proper nonce-prefix format, `crypto/rand` throughout — textbook correct.
- **Token hygiene**: Never on argv, never in shell history, masked in all output with fixed-width `***` so length isn't leaked.
- **File security**: 0700 dir, 0600 for cred/salt, atomic rename-into-place writes.
- **Profile resolution**: Clean 4-state machine with deterministic (sorted) tie-breaking and comprehensive error messages guiding the user.
- **Test coverage**: Roundtrip, permissions, corruption, orphan tokens, all resolution paths — thorough.
- **Identity echo in envelope**: Smart defensive pattern making "acting as wrong bot" immediately visible.

---

**Verdict: Fix the CI lint issues (mechanical fixes, ~30 min), then this is ready to merge.** The architecture, security model, and test coverage are all solid.

yujiawei (dismissed):
# Code Review — PR #17 (octo-cli)

Independent code review of the encrypted credential profile feature. I read every changed file plus the surrounding code (factory wiring, root command, client transport, envelope) and ran `go build`, `go vet`, `go test ./...` against the head commit — all green.

## TL;DR

**Verdict: APPROVED.** No P0/P1 blockers. The PR delivers what its summary promises: tokens never touch argv, are stored AES-256-GCM at rest, are machine-bound (resists off-machine leakage), and are masked everywhere they surface (login output, `auth status`, `config show`, `--dry-run` headers). Threat model is honestly documented in the package comment and `CLAUDE.md` (`trust boundary = OS user`), and the implementation matches it. Tests are thorough — round-trip, ciphertext-hides-secret, file perms, four-state `ActiveProfile`, orphaned-token safety, masking invariants, ambiguity / mismatch / fallthrough provider matrix, hidden-prompt escape-sequence stripping is exercised in spirit by the auth-login table tests, and the package isolates `OCTO_CONFIG_DIR` cleanly via `t.Setenv` + `TestMain`.

A few P2 maintainability / UX nits below — none block merge.

## 1. Verification against the PR description

| Claim | Evidence | Status |
|---|---|---|
| Token never on argv | `cmd/auth.go:251-285` reads from `--token-file`, stdin (`--with-token`), or hidden prompt only | ✅ |
| `credentials.enc` AES-256-GCM | `internal/authstore/crypto.go:120-156` — `cipher.NewGCM` with random nonce prefix | ✅ |
| `0600 / 0700` perms | `internal/authstore/authstore.go:29-32`; verified by `TestStore_FilePermissions` (authstore_test.go:71-85) | ✅ |
| Key = SHA256(machineID‖salt) | `crypto.go:77-96` | ✅ |
| Per-OS machineID + sentinel fallback | `machineid_{linux,darwin,windows,other}.go` + `errMachineIDUnsupported` (machineid.go:9) | ✅ |
| Transient machineID failure must NOT silently re-key | `crypto.go:86-89` returns the error rather than folding empty id into the hash; comment is explicit | ✅ |
| Token-first save / metadata-first delete | `authstore.go:143-158` (save) and `authstore.go:166-181` (remove); `TestStore_OrphanedTokenIsHarmless` (authstore_test.go:140-164) covers the post-crash residue | ✅ |
| Four-state `ActiveProfile` | `authstore.go:204-242` + `TestActiveProfile` table (authstore_test.go:166-253) | ✅ |
| Selector precedence: selector > sole > env | `cmdutil/factory.go:108-133` chains `FileProvider → EnvProvider`; `FileProvider.Resolve` returns hard error on `Missing`/`Ambiguous`, nil on `None`, so env is only reached when zero profiles exist (file_provider.go:36-67). Matches `OCTO_BOT_ID` "selector, not secret" docs. | ✅ |
| Bracketed-paste / arrow-key stripping | `cmd/auth.go:287-353` — explicit ESC / CSI / SS3 state machine; only printable ASCII (0x20-0x7E) appended to token | ✅ |
| Identity echoed on every response | `cmdutil/factory.go:247-265` builds `{type, profile, robot_id, bot_kind, source}`; `TestFactory_IdentityEcho` and `TestFactory_IdentityDefaultsToBot` (factory_test.go:424-462) | ✅ |
| `OCTO_BOT_ID` matches no profile → hard auth error (no silent fallback to `OCTO_BOT_TOKEN`) | `file_provider.go:58-59` + `missingErr` (file_provider.go:69-84). `TestFileProvider_BotIDNoMatchErrors` confirms exit type. | ✅ |

## 2. Findings

### P0 / P1: none

### P2 (non-blocking suggestions, ordered by usefulness)

**P2-1 — `MaskToken` reveals body chars for unknown-prefix tokens.**
`internal/credential/token.go:40-56`: the masking parameters are `head=2, tail=4`. For a recognized `app_*` / `bf_*` token this is fine — a meaningful prefix already classifies the kind. But for an unknown prefix (`TestMaskToken` asserts `"unknown_format_token" → "un***oken"` at token_test.go:33), six characters of body text leak with no prefix identifying *what* was leaked. Treat unknown-prefix tokens like the "too short" branch and return `"***"` unconditionally. The asserted test case is the wrong invariant — it makes a sharp edge into the spec.

**P2-2 — `SaveProfile` requires the existing store to be decryptable to add a *new* profile.**
`authstore.go:143-158` calls `loadTokens()` to merge into the existing map, then calls `saveTokens()`. If the existing `credentials.enc` is undecryptable (corrupt, key changed because the user restored from backup, machine-id changed because the host was re-imaged) `octo auth login` fails with a decrypt error — the user can't recover without `rm ~/.octo-cli/credentials.enc` (and they have to know that). Worth either (a) detecting "decrypt failure" specifically and surfacing a hint that names the file and the consequence, or (b) offering a `--reset` flag that drops the existing store and starts fresh.

**P2-3 — Plain errors from the auth store are silently swallowed in `ConfigFunc`.**
`internal/cmdutil/factory.go:91-94`: `if output.AsExitError(err) != nil { return nil, err }`. Only structured `*ExitError`s propagate. The intent (per the comment) is to keep the "no credential" fallthrough working, but this also masks legitimate plain errors from earlier in the chain — e.g. `authstore.New()` failing because `os.UserHomeDir()` errored, or a permission error reading `~/.octo-cli/cred.salt` (`crypto.go:99-117`). Those would be reported to the user as "OCTO_BOT_TOKEN required", which sends them down the wrong investigation path. Either wrap home-dir / IO errors as `output.ErrConfig` / `output.ErrAuth` so they survive the filter, or only swallow the specific `errors.New("no credential found ...")` from `provider.go:68`.

**P2-4 — No file lock around `auth login` / `logout`.**
Two concurrent `octo auth login` invocations both load → modify in-memory → atomic-write. Last writer wins; the earlier writer's profile is lost. Atomic *file* writes don't compose into atomic *operations* across two files. CLI users rarely do this so it's a real but low-probability hazard, but worth a flock or `.lock` sentinel in a future pass. (Not introduced by this PR — flagging because the multi-profile model makes it newly relevant.)

**P2-5 — `--with-token` and `--token-file` aren't mutually exclusive.**
`cmd/auth.go:253-285` — the `switch` checks `tokenFile` first, so passing both silently uses the file and ignores stdin. `cobra.Command.MarkFlagsMutuallyExclusive("with-token", "token-file")` would catch the misuse explicitly and is a one-liner.

**P2-6 — `os.MkdirAll(dirPerm)` doesn't tighten permissions on a pre-existing directory.**
`authstore.go:126`, `crypto.go:48,106`: if a user previously created `~/.octo-cli` with wider permissions (or restored it from a backup that didn't preserve mode), `MkdirAll` is a no-op and the directory stays e.g. `0755`. The secret files inside are still `0600`, but the directory listing leaks profile names. A one-shot `os.Chmod(s.dir, dirPerm)` after `MkdirAll`, or a startup check that warns on too-loose dir perms, would close the gap.

**P2-7 — `readPasswordMasked` translates Ctrl-C to a returned error.**
`cmd/auth.go:330-332`: returns `errors.New("interrupted")` rather than re-raising SIGINT or exiting non-zero directly. The user-visible effect is that a `validation` error envelope is emitted on stderr, which exits 2 — close enough to abort, but slightly surprising for an interactive prompt where Ctrl-C usually exits 130. Minor — flagging for future polish.

**P2-8 — `SECURITY.md` not updated.**
The PR adds a new on-disk credential store with non-trivial threat-model claims (machine-bound key, salt-only fallback on unsupported platforms, OS-user trust boundary). The package doc-comment and `CLAUDE.md` § *Identity Model* say it well, but `SECURITY.md` is the canonical place a security-minded user / auditor will look first. A short pointer there would close the documentation loop.

### P2 — observations not requiring action

- **`identityValue()` invokes `credential.TokenKind(f.cred.Token)` on every emit** (factory.go:258). Cheap (prefix compare), but if you ever have to call this many times per second it's an obvious cache point. Not relevant today.
- **`overlayProfileBaseURL` swallows `LoadProfiles` errors silently** (factory.go:185-200). Consistent with the rest of the file's "config show should still work" stance, just worth a one-line comment saying why.
- **`f.IOStreams.In.(*os.File)` cast in `readToken`** (auth.go:274) is gated by `isTerminal` (auth.go:356-359) which itself does the type-check, so the cast is safe. Mentioning because it's the kind of thing that looks like a panic waiting to happen.
- **`json.Marshal(map[string]string)` on `tokens`** before encryption (crypto.go:57): Go's JSON encoder sorts map keys deterministically, so the ciphertext is reproducible across runs apart from the random nonce — fine, no leak via ordering.

## 3. Suggestions

1. Convert P2-1 (`MaskToken` unknown-prefix leak) into a P2 fix: change the test expectation and return `"***"` when no recognized prefix matches. Tightens an asserted invariant for negligible cost.
2. Add a friendlier diagnostic for "existing `credentials.enc` cannot be decrypted" inside `SaveProfile` (P2-2) — suggest the file path and either a `--reset` escape hatch or a `octo auth reset` subcommand.
3. Tighten the swallow filter in `ConfigFunc` (P2-3) to only swallow the specific "no credential found" error, so home-dir / IO failures don't masquerade as "set OCTO_BOT_TOKEN".
4. `MarkFlagsMutuallyExclusive("with-token", "token-file")` (P2-5).
5. Add a brief `Credential storage` section to `SECURITY.md` (P2-8) that links to the package comment.

## 4. Additional observations (out of scope for this PR)

- `parseRetryAfter` (`internal/client/client.go:429-443`) does not cap large `Retry-After` values from the server. Pre-existing — not changed by this PR — but the new path through `attempt → retryableErr` keeps it relevant. A malicious or misconfigured upstream could pin retries open for arbitrary durations.
- The dropped `redactToken` test in `internal/client/client_test.go` is correct (the helper was removed in favor of the centralised `credential.MaskToken`), and the new behaviour is covered by `TestMaskToken` in the credential package.
- `cmd/cmd_test.go` and `cmd/octo/main_test.go` adjustments correctly add `OCTO_CONFIG_DIR` isolation so a developer's real `~/.octo-cli` doesn't bleed into tests now that `NewDefaultFactory` reads the on-disk store.

## Notes for the maintainer

- The PR is labelled `needs-human-review`. From this read I'd flag the following items as worth a human eye even if they're not blockers:
  1. The threat-model trade-off ("trust boundary = OS user") is correct but is also the *largest* security claim of the PR. It is honestly disclosed, and the code does not over-promise. Confirm the team is comfortable shipping with that boundary explicit.
  2. The Linux machineID source (`/etc/machine-id`) is world-readable, so the *secret* in the key derivation is the per-user `cred.salt` (0600). On a fresh machine without `/etc/machine-id` (some hardened containers, minimal LXC images) the read fails and `octo` errors out rather than silently re-keying — confirm this is acceptable for your CI / agent host shapes.
  3. The "darwin" backend shells out to `ioreg`. If the binary is sandboxed (App Sandbox) the call may fail; that's a deployment-time concern, not a code issue.

Jerry-Xin (dismissed):
The PR is in scope for `octo-cli` and the implementation is broadly coherent: credential storage, command wiring, identity echo, and tests all align with the repository’s CLI architecture.

💬 Non-blocking

🟡 Warning — `skills/octo-shared/SKILL.md:80` says env-token output uses `identity: "bot"`, but the code now always emits an object via `internal/output/envelope.go:22` and env credentials will include at least `{type, bot_kind, source}` once resolved. This should be corrected so agent consumers do not implement the wrong response-shape handling.

🟡 Warning — existing credential directories are not permission-hardened. `internal/authstore/authstore.go:125`, `internal/authstore/crypto.go:48`, and `internal/authstore/crypto.go:106` use `os.MkdirAll(..., 0700)`, but `MkdirAll` does not fix permissions when the directory already exists. Given the PR’s security framing around `0700`, consider `Chmod` or a fail-fast permission check for existing `OCTO_CONFIG_DIR` paths.

🔵 Suggestion — `internal/authstore/authstore.go:244` resolves duplicate `robot_id` values by sorted profile name and silently picks the first. That is deterministic, but for an anti-misuse feature, duplicate robot IDs may be better treated as ambiguous.

✅ Highlights

The file provider fallback behavior is well tested, and the “selector mismatch must not fall back to env token” rule is correctly enforced.

Token display is centralized in `internal/credential/token.go`, and dry-run authorization masking reuses the same helper.

`go test ./...` and `go vet ./...` both pass.

yujiawei (dismissed):
# Code Review — PR #17 (octo-cli)

## Summary

A clean, security-conscious implementation of an encrypted on-disk credential store and the `octo auth` command family. Tokens never reach `argv`; AES-256-GCM with a random nonce protects them at rest; the trust model is documented honestly. `go vet`, `go test ./...`, and the new test files all pass against the head SHA. The chain wiring (file → env), four-state `ActiveProfile` resolution, write-ordering invariant (token first on save, metadata first on remove), envelope `identity` echo, and `--bot-id`/`--profile` selectors all match the PR description and behave as advertised.

No correctness, security, or build blockers were found. The remaining items are P2/maintainability and a couple of explicit notes for the human security reviewer.

## 1. Verification

| Item | Status | Evidence |
| --- | --- | --- |
| AES-256-GCM with random nonce, prefixed | ✅ | `internal/authstore/crypto.go:120-130` |
| File perms 0600 / dir 0700 | ✅ | `internal/authstore/authstore.go:29-32`, test at `authstore_test.go:71-85` |
| Token never in argv | ✅ | `cmd/auth.go:122-124` flags only toggle source; reads from prompt / stdin / file at `cmd/auth.go:253-285` |
| Token never in `config.json` | ✅ | `authstore_test.go:46-69` (ciphertext + plaintext both checked) |
| Atomic write (temp + rename) | ✅ | `internal/authstore/authstore.go:273-297` |
| Save ordering (token then metadata) | ✅ | `internal/authstore/authstore.go:143-158`; orphan-token resilience covered by `authstore_test.go:140-164` |
| Remove ordering (metadata then token) | ✅ | `internal/authstore/authstore.go:166-181` |
| Four-state `ActiveProfile` | ✅ | table-driven test in `authstore_test.go:166-253` |
| Selector precedence and ambiguity errors | ✅ | `internal/credential/file_provider.go:36-67` + `file_provider_test.go:40-128` |
| Mid-paste arrow keys / bracketed-paste markers stripped | ✅ | escape-state machine in `cmd/auth.go:288-318` (only printable ASCII written to the buffer at `cmd/auth.go:323-339`) |
| Token masked everywhere it's displayed | ✅ | `internal/credential/token.go:40-56`, also reused by `--dry-run` Authorization header in `internal/client/client.go:288` |
| Envelope `identity` is always an object | ✅ | `internal/output/envelope.go:25-31`, `factory_test.go:424-462` |
| `auth` subtree skips `cfg.Validate` | ✅ | `cmd/root.go:65-73` (`auth` listed) |
| `OCTO_BOT_ID` is a selector, not a fallback | ✅ | When `--bot-id` / `OCTO_BOT_ID` is set with no matching profile, `FileProvider` returns `auth_error` (`file_provider.go:58-67` → `missingErr`) and the chain stops before `EnvProvider`. Verified via `TestFileProvider_BotIDNoMatchErrors`. |
| Backwards compatibility for env-only users | ✅ | Zero profiles → `StatusNone` → `(nil, nil)` → falls through to `EnvProvider`. `TestFileProvider_ZeroProfilesFallsThrough` confirms. |
| `go vet ./...` and `go test ./...` | ✅ | All packages pass at 1f58794. |

## 2. Findings

### P2 — Salt regeneration on corruption silently invalidates all stored tokens

`internal/authstore/crypto.go:98-117`. `getOrCreateSalt` accepts the salt only when `err == nil && len(salt) == saltLen`. If the salt file exists but is the wrong length (e.g. truncated by a partial write or a half-applied recovery), the function falls through to the regeneration branch, writes a *new* random salt, and silently changes the derived key. Every existing entry in `credentials.enc` then fails to decrypt. The user's first signal is "decrypt … (corrupt, or key changed)" on the next command, but the salt has already been overwritten so they can't recover the old key. Recommend treating "salt file present but wrong length" as a hard error and asking the user to remove the directory explicitly.

### P2 — Concurrent `SaveProfile` / `RemoveProfile` is last-writer-wins

`internal/authstore/authstore.go:143-181` reads the full `profiles` and `tokens` maps, mutates them in memory, and writes them back. There is no file lock. Two `octo auth login` invocations racing in the same `OCTO_CONFIG_DIR` (e.g. an agent retrying, or a script paralleling `auth login` over a list of bots) can lose a profile. The same applies to `auth logout`. For a single-user CLI this is usually tolerable, but at minimum it's worth documenting; a `flock`-based critical section would close the gap cheaply on POSIX.

### P2 — `auth login` silently overwrites an existing profile

`cmd/auth.go:100-119` calls `SaveProfile(name, …)` unconditionally; `SaveProfile` does `profiles[name] = *meta` (`authstore.go:152`). Reusing `--profile <name>` with a different `--bot-id` rewrites the entry without warning. The stored profile count in the response hint hides that an existing entry was replaced (count stays the same). Recommend either prompting on overwrite when interactive, or returning a `replaced: true` field plus a stderr line in the response.

### P2 — Duplicate `robot_id` across profiles is silently accepted

`authstore.go:244-253` resolves duplicates deterministically (alphabetical), which is a reasonable tiebreaker, but the PR never warns the user that two different profile names point at the same `robot_id`. That's a foot-gun: `auth list` shows two entries that both think they're the same bot, and `--bot-id` will always pick the alphabetically first one. Either reject duplicates at `SaveProfile` time, or surface a `duplicate: true` flag in `auth list`.

### Nit — macOS/Windows machine-id commands rely on `PATH`

`internal/authstore/machineid_darwin.go:15` runs `ioreg` and `machineid_windows.go:14` runs `reg` via `exec.Command` with bare names. A shell with a hostile or shadowed `PATH` (a malicious binary called `reg` in the user's `PATH` ahead of `C:\Windows\System32`) becomes the source of the encryption key. Locking these to absolute paths (`/usr/sbin/ioreg`, `%SystemRoot%\System32\reg.exe`) closes that hole. Realistic risk is low — same-user code execution already breaks the trust boundary — but on Windows in particular it costs nothing to fix.

### Nit — Orphan tokens never garbage-collected

The "token written, metadata not yet" crash window is documented and handled correctly on read, but `loadTokens()` never reconciles the two maps. Orphans sit in `credentials.enc` indefinitely. A cheap opportunistic cleanup (drop tokens whose profile is missing, on `SaveProfile`) would tidy up over time.

### Nit — `atomicWrite` has no `fsync` before rename

`authstore.go:273-297` writes, chmods, closes, renames. There is no `tmp.Sync()` before close, so a kernel/host crash between rename and writeback can yield a renamed-but-empty file. Standard Go practice, and the impact is bounded to needing a fresh login, but worth a sentence in the doc-comment.

### Nit — No syntactic validation of bot id / "unknown" token kind

`auth login` accepts any non-empty `--bot-id` value and any non-empty token. A typo like `app_demo` vs `app demo` (trailing space already trimmed, but a leading space inside the token would survive) is only diagnosed on the first API call. Warning when `credential.TokenKind(token) == "unknown"` (i.e. the token has neither `app_` nor `bf_` prefix) at login time would catch most paste accidents.

## 3. Notes for the human security reviewer (per `needs-human-review`)

- **Trust-boundary scope is correctly described.** The PR body and `internal/authstore/authstore.go:1-10` are explicit that the boundary is the OS user account; the encryption protects the file off-machine, not against a same-user process. This is the right framing.
- **Off-machine claim is conditional.** The encryption key is `SHA256(machineID || salt)`. An attacker who exfiltrates `credentials.enc` *and* the host's machine id (`/etc/machine-id`, `IOPlatformUUID`, or `MachineGuid`) *and* `cred.salt` can decrypt offline. The salt lives next to the ciphertext, so any backup/sync that captures `~/.octo-cli/` exports both. The remaining secret is the machine id — which is per-host but not designed to be confidential. Worth being explicit in the doc-comment that "off-machine binding" means the *machine id has to be re-acquired* — not that it's a true secret.
- **No KDF.** SHA-256 of high-entropy material is fine here (unlike password-derived keys, there's no brute-force surface), so PBKDF2/Argon2 would be ceremony, not security. Calling that out so a future reader does not "fix" it by adding one.
- **`x/term` raw-mode handling restores state via `defer`.** `cmd/auth.go:346-378`. Looks correct; Ctrl-C exits the loop returning `errors.New("interrupted")` and the deferred `term.Restore` still runs. Worth confirming on a real Windows console — that's the platform most likely to surface escape-sequence edge cases — but the test matrix (`TestAuth_*` in `cmd/auth_test.go`) doesn't and can't cover the raw terminal.
- **No constant-time comparison needed in this PR.** `ActiveProfile` matches by string equality, but neither `name` nor `robot_id` is secret, so timing leaks are not relevant.

## 4. Additional observations (out of scope, not blocking)

- `factory.go:138-202` resolves config and credentials lazily; calling `Config()` triggers `Credential()` which triggers `AuthStore()`. The error swallowing at `factory.go:147-153` (`output.AsExitError(err) != nil` → propagate, anything else → ignore) is the right shape, but the test coverage for the swallow path is implicit. A direct `TestBuildConfig_PlainErrorSwallowed` would document the contract.
- `cmd/config.go:62-72` reads `f.Credential()` to enrich `config show`, which means running `octo config show` *now* triggers credential resolution, while previously it didn't. With ambiguous profiles this surfaces a structured error rather than masking it (which is the right call), but it changes the failure shape for env-only users with malformed `OCTO_BOT_TOKEN` — they will now get the new auth-error text path. Worth a release-note line.
- The `file_provider.go:69-84` `missingErr` hint suggests `octo auth login --bot-id <id>` to fix a missing match; if the actual failure was "you typed the wrong bot id", that hint pushes the user to *create another profile* under the wrong id rather than fixing the typo. Minor wording polish.

Jerry-Xin (approved):
The PR is in scope for `octo-cli` and the credential profile/auth command implementation is generally well-structured, tested, and aligned with the CLI’s existing factory/provider patterns.

💬 Non-blocking

🟡 Warning — [cmd/config.go](/Users/mlamp/.openclaw/tmp/pr-review-2ap8o6gg/repo/cmd/config.go:40): `config show` only surfaces structured `ExitError`s. Plain auth-store failures such as corrupt `credentials.enc`, corrupt salt, or machine-id read errors fall back to `config.Load()`, and the later `f.Credential()` error is ignored at [cmd/config.go](/Users/mlamp/.openclaw/tmp/pr-review-2ap8o6gg/repo/cmd/config.go:65). That can make a broken stored credential look like an unconfigured/env-only state. Consider surfacing non-`ErrNoCredential` credential-store errors here too.

✅ Highlights

- Storage ordering in `SaveProfile` / `RemoveProfile` preserves the advertised “listed profile has token” invariant.
- Selector behavior is explicit and avoids silently falling back to env when `--bot-id` / `--profile` misses.
- Token masking is centralized and reused in status/config/dry-run paths.
- Tests cover the main resolution matrix, auth commands, encryption round-trip, permissions, and identity envelope shape.

Verification run: `go test ./...` and `go vet ./...` both pass.

yujiawei (approved):
# Code Review — PR #17 (octo-cli)

## Verdict
Approve. The PR delivers what the description claims: tokens never appear on argv, are encrypted at rest, and the runtime credential chain is honest about ambiguity rather than guessing. Threat model is documented accurately. Code quality is high, tests are comprehensive, cross-compiles cleanly for linux/darwin/windows, and `go test ./internal/authstore ./internal/credential ./cmd` passes locally.

No P0/P1 blockers found. A handful of P2 nits and one usability concern noted below; none gate merge.

## What I verified

- ✅ **No token on argv.** `cmd/auth.go` accepts the token only via hidden prompt, `--with-token` (stdin), or `--token-file`. There is no `--token <value>` flag.
- ✅ **AES-256-GCM with random nonce per write.** `internal/authstore/crypto.go:128-139`. Nonce is generated per `seal()` call, prefixed to ciphertext; the entire blob is rewritten on each save, so there is no nonce-reuse risk.
- ✅ **Key derivation surfaces real failures.** `deriveKey` only degrades to salt-only material when `errors.Is(err, errMachineIDUnsupported)` (`crypto.go:88-91`); a transient read failure on a supported platform surfaces, which is the correct behavior — silently changing the key would render every stored token undecryptable.
- ✅ **Crash-safe write ordering.** `SaveProfile` writes the encrypted token before the metadata, and `RemoveProfile` deletes the metadata before the token (`authstore.go:163-201`). The orphaned-token case is exercised by `TestStore_OrphanedTokenIsHarmless`.
- ✅ **File permissions enforced.** Dir `0700`, `credentials.enc` and `cred.salt` `0600`; `ensureDir` tightens a pre-existing dir if its mode allows group/other access (`authstore.go:129-143`). Verified by `TestStore_FilePermissions`.
- ✅ **Ambiguity is a hard error.** `FileProvider.Resolve` returns a `validation` ExitError on `StatusAmbiguous` and an `auth_error` on `StatusMissing` (`file_provider.go:36-83`); a selector that matches nothing never silently falls through to `OCTO_BOT_TOKEN` (`TestFileProvider_AmbiguousErrors`, `TestFileProvider_BotIDNoMatchErrors`).
- ✅ **Duplicate-bot-id rejection.** `assertBotIDFree` blocks a second profile claiming an existing robot id (`auth.go:253-267`), preventing `--bot-id` selection from becoming non-deterministic. Re-login under the same name is allowed (`TestAuth_DuplicateRobotIDRejected`).
- ✅ **Salt corruption is loud.** A wrong-length `cred.salt` is not silently regenerated — `getOrCreateSalt` returns a clear error pointing at re-running `octo auth login` (`crypto.go:106-111`).
- ✅ **Token never echoed in plain.** `auth list` never returns a token; `auth status` and `config show` both go through `credential.MaskToken`. `--dry-run` `Authorization` header is masked too (`internal/client/client.go:288`). Verbose logs print URL/body but not the `Authorization` header.
- ✅ **Hidden prompt is robust.** `readPasswordMasked` puts the terminal in raw mode with a deferred restore, accepts only printable ASCII, and uses an explicit state machine to swallow ESC/CSI sequences and bracketed-paste markers so arrow keys / `ESC[200~` cannot land inside the token (`auth.go:314-405`). Backspace handling and Ctrl-C abort are present.
- ✅ **Identity echo in envelope.** `Factory.identityValue()` populates `{type, profile, robot_id, bot_kind, source}` only when a credential was actually resolved, never forcing resolution for diagnostic commands. Defaults to `{type:"bot"}` otherwise (`factory.go:270-288`, `envelope.go:29-46`). `TestFactory_IdentityEcho` and `TestFactory_IdentityDefaultsToBot` cover both cases.
- ✅ **`go test ./...`** for the touched packages passes locally; build succeeds for `GOOS=linux|darwin|windows`.

## Findings

No P0/P1. The items below are P2 / suggestions and do not block merge.

### P2 — Linux without `/etc/machine-id` fails hard rather than degrading

`internal/authstore/machineid_linux.go` returns a plain error when neither `/etc/machine-id` nor `/var/lib/dbus/machine-id` is present, so `deriveKey` surfaces `read machine id: machine-id not found` and every `octo auth …` (and any service command resolving from a stored profile) fails. This is the documented intent ("a missing id on a supported platform must surface"), but minimal containers — including some agent runtimes — ship without those files, and the failure mode is opaque to the operator who has not yet stored a profile. Two alternatives worth considering:

- Detect the "no profiles + no machine-id" early-bootstrap case and let it slide (it cannot have invalidated a key that does not exist), or
- Return `errMachineIDUnsupported` from the Linux path when both files are missing **and** the running environment has no other plausible source, with a one-line stderr note that off-machine binding is not in effect.

Either approach keeps the existing safety property (a host that *had* an id and lost it must not silently change the key) while reducing footgun friction in agent containers. Not a merge blocker — current behavior is defensible and documented in the decrypt error.

### P2 — `--token-file` does not warn on permissive mode bits

`readToken` reads the file unconditionally (`auth.go:282-288`). If an operator passes a `--token-file` that is `0644` or world-readable, the token has already leaked at rest before login. A best-effort `Stat` + stderr warning when `mode & 0o077 != 0` would catch the common mistake without forcing a hard failure (an operator with a defensible reason can still proceed). Not blocking; nice-to-have.

### P2 — `config.json` mode is `0644`

`metaPerm = 0o644` (`authstore.go:31`) means profile names, robot ids, and api base URLs are world-readable on the filesystem. The parent dir is `0700` so this is moot in practice, but if a backup tool or bind-mount ever exposes the file, the catalogue leaks. Tightening to `0600` costs nothing — there is no inter-user reader for this file. Pure defense in depth.

### P2 — Hidden prompt silently drops non-printable bytes

`applyMaskByte` ignores any byte outside `0x20..0x7e` (`auth.go:350-366`). For a paste that smuggled a non-breaking space, a stray UTF-8 byte, or a tab, the operator has no signal that the stored token differs from what they pasted (the tail-mask in `auth status` reveals only the last four chars). Two small mitigations:

- Beep / print a stderr note ("ignored 1 non-ASCII byte") if any input byte was dropped, and/or
- After Enter, length-check the trimmed token against a sane minimum and reject empty/short tokens with a clearer hint than the current `"empty token"`.

This is purely a UX safety net; tokens that match `app_*` / `bf_*` are ASCII, so practical harm is low.

### nit — `cmd/octo/main_test.go` env handling

The `OCTO_BOT_TOKEN=`, `OCTO_BOT_ID=`, `OCTO_CONFIG_DIR=…` triple is correct, but it relies on `os.Environ()` first then the empty-string overrides last. That works on every platform Go currently targets, but a future refactor that uses an env-merging helper could miss the ordering invariant. A short comment naming the contract ("later entries win") would prevent that. Optional.

### nit — `extractRetryAfterFromErr` comment block in `client.go`

Pre-existing, unchanged by this PR, but flagged because it confused me on first read: the dangling "No — we pass it in via lastRetryAfter below." mid-sentence at `client.go:155-159` is a left-over thought. Worth a follow-up pass to clean up.

## Notes for the human reviewer (`needs-human-review`)

- **Threat-model framing in `SECURITY.md`** is honest and explicit: same-user processes can decrypt the store; the encryption defends against off-machine leakage. Recommend the human reviewer confirm this matches the project's published security stance for what `octo` claims to protect against.
- **Machine id sources** read from system files / spawn `ioreg` and `reg query`. PATH for `ioreg` and `reg` is implicit; on a hardened agent image with a minimal PATH, these could fail. The macOS / windows paths are well-known absolute locations, so an explicit absolute path (`/usr/sbin/ioreg`, `%SystemRoot%\System32\reg.exe`) would harden against PATH-injection by a malicious co-resident process — not in-scope for the stated threat model, but a cheap belt-and-braces step.
- **`x/term` dependency.** Pinned to `v0.39.0` in `go.mod`; CI's `go 1.24.0` keeps the existing toolchain. Confirmed that the PR description's exception ("used only for the hidden token prompt") matches the actual import sites — only `cmd/auth.go` references `golang.org/x/term`.

## Summary

This is the right shape of fix for the underlying problem (token in argv → leak in transcript / shell history). The crypto layer is conservative and correct, the four-state `ActiveProfile` outcome is the right primitive for the precedence rules in the PR description, and the test matrix covers the awkward cases (orphaned token, ambiguous selection, duplicate robot id, ciphertext-hides-token, file perms). Approving as-is; the P2s above are good follow-ups but do not need to land in this PR.

PR #19, opened 2026-05-27
feat(cli): withhold the matter domain behind an x-octo-disabled spec flag

liuooo:
## Summary

Temporarily withhold the `matter` command domain from the CLI surface while its
backend API stabilizes, driven by a `x-octo-disabled` flag on the service spec.
The spec stays embedded — introspection and the metadata-driven engine are
unaffected — and re-enabling is a one-line flag flip.

## Changes

**Mechanism — `x-octo-disabled` (spec-driven, not hardcoded to matter)**
- `internal/registry/specs/matter.json`: top-level `"x-octo-disabled": true`.
- `internal/registry/loader.go`: `ServiceDisabled` (truthy = bool `true` or
  string `"true"`, shared with the skill convention) plus `EnabledServices` /
  `EnabledOperations` as the single caller-facing view. `ListServices` /
  `ListAllOperations` are deliberately left unfiltered so the engine fixture and
  explicit introspection still see matter.

**Presentation layers (honour the flag)**
- `cmd/root.go`: `NewRootCmd` drops disabled service subtrees *after*
  registration, so `octo matter ...` disappears from help/completion. Done after
  `RegisterServiceCommands` so the engine tests that drive it directly keep their
  full matter fixture.
- `cmd/schema.go`: global `schema --list` uses the enabled views; explicit
  `schema --list matter` / `schema matter.list` still resolve (read-only
  introspection preserved).
- `cmd/skills.go` + `skills/octo-matter/SKILL.md`: `disabled: true` frontmatter
  removes octo-matter from list/print/install; the file stays embedded.

**Tests**
- New regression tests lock all three surfaces (command tree absent, global list
  hidden, explicit introspection preserved, skill unlisted) plus the registry
  enabled-view invariants and `truthy`. Mutation-checked: flipping the flag off
  makes the guards fail.
- Three existing auth-gate tests swap their incidental `matter list` fixture to
  `event list`.

**Docs**
- README / CLAUDE.md annotate matter as withheld; octo-shared / octo-files swap
  runnable `octo matter` examples to live domains.

## Motivation

The matter backend API is not yet stable; offering `octo matter ...` to agents
now risks them building on a moving surface. This withholds it cleanly and
reversibly without removing the spec — which is also the engine's primary test
fixture, so a hard removal would gut loader/service coverage and the matter-only
transition aliases.

## Testing

- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] `go build ./...` passes
[...]

PR #20, opened 2026-05-28
feat(thread): remove the destructive thread.delete command

liuooo:
## What

`thread.delete` was the only `x-octo-risk: high-risk-write` operation left
in any enabled spec, and it sat at odds with the project's "bots get no
destructive operations" convention. The classic destructive endpoints —
message revoke, group disband, conversation delete — are user-only on the
server and were never exposed to bots; `thread.delete` was the lone
exception. Threads also have no bot-accessible archive or soft-close path,
so a hard delete was the only way for a bot to "finish" a thread, which
made the inconsistency particularly sharp.

This PR removes the spec entry. Shape mirrors #19 (matter-domain withhold),
except the operation is deleted outright rather than marked `x-octo-disabled`,
because — unlike `matter` — there is no plan to re-enable `thread.delete`.
The backend route is unchanged; callers that genuinely need to delete a
thread can still construct the call via
`octo api DELETE /v1/bot/groups/{group_no}/threads/{short_id}`.

## How

- **`internal/registry/specs/thread.json`** — drop the `delete` block under
  `/v1/bot/groups/{group_no}/threads/{short_id}`; the path keeps its `get`
  sibling untouched. The metadata-driven engine then removes the command
  from the cobra tree, the `schema --list` output, and shell completions
  with zero code change.
- **`internal/registry/loader_test.go`** — `thread: 9 → 8` in
  `TestAllDomainOperationCounts`. The total is summed from the map, so no
  other assertion change is needed.
- **`CLAUDE.md` / `README.md`** — `48 operations / 51 commands → 47/50`;
  the command-tree snippet in `CLAUDE.md` drops `delete` from the `thread`
  row.
- **`CHANGELOG.md`** — new `### Removed` entry under `[Unreleased]`;
  `[0.4.0]` history is left untouched (Keep a Changelog).
- **`docs/octo-cli-design.md`** — four independent count copies updated:
  domain heading (`Domain 4: thread (... 9 → 8 commands)`), the numbered
  command table (drop `#34 thread delete` and renumber `#35..#51 → #34..#50`
  for contiguity), the summary table (`thread 9→8`, `Total 51→50`,
  `App-blocked 13→12`), and the App/User availability lines
  (`38/51 → 38/50 (76%)`, `51/51 → 50/50`).
- **`skills/octo-messaging/SKILL.md`** — `9 commands → 8 commands` and the
  `octo thread delete` line removed.

Untracked draft docs in the working tree are intentionally excluded from
this PR.

## Tests

No code logic changed; only spec data and one assertion. CI commands
locally:

- `go build ./cmd/octo` — green
- `go vet ./...` — green
- `go test ./... -count=1` — all packages green, including the updated
  `TestAllDomainOperationCounts` (thread: 8, total: 47).

Manually verified after rebuild:

- `octo thread --help` lists 8 subcommands, no `delete`.
- `octo schema --list thread` no longer returns `thread.delete`.

## Breaking change

`octo thread delete` is no longer registered. Callers will receive
`unknown command`. Use the `octo api DELETE ...` passthrough as a fallback
if a thread genuinely needs deletion. This is intentional and aligned with
the "no destructive bot operations" convention.

## Notes

The count synchronisation across 6+ files in this PR highlights an existing
tech-debt pattern: the same number is duplicated by hand in many places,
guarded only on the code side (`loader_test.go`). A separate chore issue
is being filed to track de-duplication; this PR does not address it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+- `octo thread delete` — withdrew the thread-deletion command. Threads expose no
+  bot-accessible archive or soft-close path, so a hard delete was inconsistent
+  with the convention that bots get no destructive operations. The backend route
+  is untouched; only the CLI command and its `thread.delete` spec entry are
+  removed. Command/operation totals drop 51/48 → 50/47.
+
 ## [0.4.0] — 2026-05
 
 ### Architectural overhaul — metadata-driven command tree
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   is untouched; only the CLI command and its `thread.delete` spec entry are
   removed. Command/operation totals drop 51/48 → 50/47.
 
+### Changed
+- Service-domain parent commands (`octo thread`, `octo group`, `octo file`, …)
+  now reject unknown subcommands with `unknown subcommand %q for %q` and exit 2
+  instead of silently printing help and exiting 0. Required by the `thread
+  delete` removal above so automation can detect a missing operation; also
+  catches typos like `octo group lisst`. Parent help (`octo thread`,
+  `octo thread --help`) is unchanged and still works without a token; only
+  leaf operations require authentication.
+
 ## [0.4.0] — 2026-05
 
 ### Architectural overhaul — metadata-driven command tree
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -22,7 +22,7 @@
 - Each Bot has an **owner**; operations are attributed to the Bot identity. For LLM-backed paths (`matter extract`) the bot acts on behalf of its owner — pass `owner_uid` as `creator_uid`.
 - `OCTO_SPACE_ID` (or `--space`) supplies space context for platform-scoped bots. Space-scoped bots resolve their space server-side.
 
-## Command Structure (7 domains, 48 operations / 51 commands incl. 3 matter transition aliases)
+## Command Structure (7 domains, 47 operations / 50 commands incl. 3 matter transition aliases)
 
 Service commands are auto-registered. The hand-written leaves are `schema`, `version`, `api` (generic passthrough), `config`, `auth`, and the cobra-generated `completion`.
 
@@ -41,7 +41,7 @@ octo matter    create | list | get | update | delete        (withheld)
 octo message   send | edit | sync | read-receipt
 octo group     list | get | members | md-get | md-update
                create | update | member-add | member-remove       (User Bot only)
-octo thread    create | list | get | delete | members
+octo thread    create | list | get | members
                join | leave | md-get | md-update                  (User Bot only)
 octo file      upload | download | credentials | presigned
 octo bot       register | set-commands | user-info | space-members | typing | heartbeat
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Key properties:
 |-----------|-----|----------------------------------------------------------------|
 | `matter`  | 14  | Todos/tasks — **temporarily withheld** while the backend API stabilizes |
 | `group`   | 9   | Groups — list, get, members, metadata; create/update (User Bot)|
-| `thread`  | 9   | Threads — create, list, get, members, join/leave, metadata     |
+| `thread`  | 8   | Threads — create, list, get, members, join/leave, metadata     |
 | `bot`     | 6   | Bot lifecycle — register, user-info, space-members, heartbeat  |
 | `message` | 4   | Messaging — send, edit, sync, read-receipt                     |
 | `file`    | 4   | Files — upload, download, credentials, presigned URLs          |
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ deterministic taxonomy. There is no interactive I/O.
 
 ## Architecture
 
-octo-cli is **metadata-driven**. The entire command tree — 48 operations
+octo-cli is **metadata-driven**. The entire command tree — 47 operations
 across 7 domains — is auto-registered at startup from OpenAPI 3.x specs
 embedded into the binary. Adding or changing an endpoint means editing a
 spec, not the code.
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -88,6 +88,13 @@ func withholdDisabledServices(root *cobra.Command, f *cmdutil.Factory) {
 // `octo auth`, and cobra's generated `completion`). Walks the parent chain so
 // leaves like `config show` match through their parent.
 func skipValidation(cmd *cobra.Command) bool {
+	// Per-command annotation, deliberately NOT inherited through the parent
+	// chain: service parents (`octo thread`, `octo group`, …) opt out of the
+	// auth gate so their help / unknown-subcommand path works without a token,
+	// but their leaf operations (`octo thread create …`) must still authenticate.
+	if cmd.Annotations["skipValidation"] == "true" {
+		return true
+	}
 	for c := cmd; c != nil; c = c.Parent() {
 		switch c.Name() {
 		case "version", "help", "schema", "config", "completion", "skills", "auth", "":
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -1,0 +1,114 @@
+package cmd
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-cli/internal/cmdutil"
+	"github.com/Mininglamp-OSS/octo-cli/internal/config"
+	"github.com/Mininglamp-OSS/octo-cli/internal/output"
+)
+
+// TestNewRootCmd_ParentCommandSkipAuthAndRejectUnknown drives the real
+// NewRootCmd so PersistentPreRunE and WrapCLIError are in the loop. Service
+// parents must:
+//  1. skip the auth gate so help works zero-config, and
+//  2. reject unknown subcommands with a validation envelope (not auth_error).
+//
+// Regression guard for PR #20 review: the parent-`RunE` change made parents
+// Runnable() and routed them through the auth gate, turning
+// `octo thread bogus` into UNAUTHORIZED instead of "unknown subcommand".
+func TestNewRootCmd_ParentCommandSkipAuthAndRejectUnknown(t *testing.T) {
+	cases := []struct {
+		name          string
+		args          []string
+		wantErr       bool
+		wantErrType   string
+		wantErrSubstr string
+	}{
+		{
+			name: "thread no args prints help without a token",
+			args: []string{"thread"},
+		},
+		{
+			name:          "thread unknown subcommand -> validation, not auth_error",
+			args:          []string{"thread", "bogus"},
+			wantErr:       true,
+			wantErrType:   "validation",
+			wantErrSubstr: "unknown subcommand",
+		},
+		{
+			name:          "thread delete (removed) -> validation, not auth_error",
+			args:          []string{"thread", "delete", "g", "s"},
+			wantErr:       true,
+			wantErrType:   "validation",
+			wantErrSubstr: "unknown subcommand",
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			f := newTestFactoryWithReg()
+			// Empty token: the auth gate would fail loudly if it ran. The
+			// whole point of the annotation is that it does NOT run for
+			// service parents.
+			f.SetConfig(&config.Config{APIBaseURL: "http://localhost", BotToken: ""})
+			root := NewRootCmd(f.Factory)
+			root.SetArgs(tc.args)
+			var buf bytes.Buffer
+			root.SetOut(&buf)
+			root.SetErr(&buf)
+			err := root.Execute()
+
+			if !tc.wantErr {
+				if err != nil {
+					t.Fatalf("expected nil error, got %v; output=%s", err, buf.String())
+				}
+				return
+			}
+			if err == nil {
+				t.Fatalf("expected error, got nil; output=%s", buf.String())
+			}
+			// Same path main.go uses to derive exit code and envelope shape.
+			wrapped := cmdutil.WrapCLIError(err)
+			ee := output.AsExitError(wrapped)
+			if ee == nil {
+				t.Fatalf("expected ExitError after WrapCLIError, got %T: %v", wrapped, wrapped)
+			}
+			if ee.Type != tc.wantErrType {
+				t.Errorf("error type: got %q, want %q (msg=%s)", ee.Type, tc.wantErrType, ee.Message)
+			}
+			if !strings.Contains(strings.ToLower(ee.Message), tc.wantErrSubstr) {
+				t.Errorf("error msg %q should contain %q", ee.Message, tc.wantErrSubstr)
+			}
+		})
+	}
+}
+
+// TestNewRootCmd_LeafStillRequiresAuth proves the per-command skipValidation
+// annotation does NOT inherit through the parent chain: a leaf operation
+// under a service parent (which has the annotation) must still authenticate.
+// Without this guard the annotation fix would silently exempt every real
+// command from auth.
+func TestNewRootCmd_LeafStillRequiresAuth(t *testing.T) {
+	f := newTestFactoryWithReg()
+	f.SetConfig(&config.Config{APIBaseURL: "http://localhost", BotToken: ""})
+	root := NewRootCmd(f.Factory)
+	// `octo group list` is a leaf under `group` (which has the annotation).
+	root.SetArgs([]string{"group", "list"})
+	var buf bytes.Buffer
+	root.SetOut(&buf)
+	root.SetErr(&buf)
+	err := root.Execute()
+	if err == nil {
+		t.Fatalf("expected leaf to require auth, got nil; output=%s", buf.String())
+	}
+	wrapped := cmdutil.WrapCLIError(err)
+	ee := output.AsExitError(wrapped)
+	if ee == nil {
+		t.Fatalf("expected ExitError, got %T: %v", wrapped, wrapped)
+	}
+	if ee.Type != "auth_error" {
+		t.Errorf("leaf should hit auth gate (type=auth_error), got type=%q msg=%s", ee.Type, ee.Message)
+	}
+}
--- a/cmd/service/service.go
+++ b/cmd/service/service.go
@@ -51,9 +51,10 @@ func ensureServiceCmd(parent *cobra.Command, svc string) *cobra.Command {
 		return existing
 	}
 	c := &cobra.Command{
-		Use:   svc,
-		Short: fmt.Sprintf("Operations on the %s domain", svc),
-		RunE:  rejectUnknownSubcommand,
+		Use:         svc,
+		Short:       fmt.Sprintf("Operations on the %s domain", svc),
+		RunE:        rejectUnknownSubcommand,
+		Annotations: map[string]string{"skipValidation": "true"},
 	}
 	parent.AddCommand(c)
 	return c
@@ -74,9 +75,10 @@ func attachOperation(svcCmd *cobra.Command, f *cmdutil.Factory, d *registry.Oper
 		sub := findChild(cur, name)
 		if sub == nil {
 			sub = &cobra.Command{
-				Use:   name,
-				Short: fmt.Sprintf("%s %s operations", cur.Use, name),
-				RunE:  rejectUnknownSubcommand,
+				Use:         name,
+				Short:       fmt.Sprintf("%s %s operations", cur.Use, name),
+				RunE:        rejectUnknownSubcommand,
+				Annotations: map[string]string{"skipValidation": "true"},
 			}
 			cur.AddCommand(sub)
 		}
--- a/cmd/service/service.go
+++ b/cmd/service/service.go
@@ -53,6 +53,7 @@ func ensureServiceCmd(parent *cobra.Command, svc string) *cobra.Command {
 	c := &cobra.Command{
 		Use:   svc,
 		Short: fmt.Sprintf("Operations on the %s domain", svc),
+		RunE:  rejectUnknownSubcommand,
 	}
 	parent.AddCommand(c)
 	return c
@@ -75,6 +76,7 @@ func attachOperation(svcCmd *cobra.Command, f *cmdutil.Factory, d *registry.Oper
 			sub = &cobra.Command{
 				Use:   name,
 				Short: fmt.Sprintf("%s %s operations", cur.Use, name),
+				RunE:  rejectUnknownSubcommand,
 			}
 			cur.AddCommand(sub)
 		}
@@ -91,6 +93,18 @@ func findChild(parent *cobra.Command, name string) *cobra.Command {
 		}
 	}
 	return nil
+}
+
+// rejectUnknownSubcommand makes parent commands fail loudly on an
+// unrecognised subcommand. Without it, cobra treats the unknown token
+// as args, finds no RunE on the parent (Runnable()==false), prints
+// help, and exits 0 — which can let automation treat a removed
+// command as a success.
+func rejectUnknownSubcommand(cmd *cobra.Command, args []string) error {
+	if len(args) > 0 {
+		return fmt.Errorf("unknown subcommand %q for %q", args[0], cmd.CommandPath())
+	}
+	return cmd.Help()
 }
 
 // buildOperationCmd builds the leaf cobra.Command for one operation.
--- a/cmd/service/service_test.go
+++ b/cmd/service/service_test.go
@@ -521,3 +521,37 @@ func TestBuildMultipartBody_MissingFile(t *testing.T) {
 		t.Errorf("expected validation ExitError, got %T: %v", err, err)
 	}
 }
+
+// TestParentCommand_RejectsUnknownSubcommand guards against cobra's default
+// fallback: a parent command with no RunE silently prints help and exits 0
+// when given an unknown token, which can let automation treat a removed or
+// mistyped command as a success. Regression test for the removal of
+// thread.delete (where `octo thread delete ...` previously printed help).
+func TestParentCommand_RejectsUnknownSubcommand(t *testing.T) {
+	cases := []struct {
+		name string
+		args []string
+	}{
+		{"thread + removed leaf (delete)", []string{"thread", "delete", "g", "s"}},
+		{"thread + bogus", []string{"thread", "bogus"}},
+		{"group + bogus", []string{"group", "nope"}},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+				t.Fatalf("HTTP handler must not be called when subcommand is unknown")
+			})
+			root.SetArgs(tc.args)
+			var buf bytes.Buffer
+			root.SetOut(&buf)
+			root.SetErr(&buf)
+			err := root.Execute()
+			if err == nil {
+				t.Fatalf("expected error for %v, got nil; output=%s", tc.args, buf.String())
+			}
+			if !strings.Contains(err.Error(), "unknown subcommand") {
+				t.Errorf("error should mention 'unknown subcommand', got %q", err.Error())
+			}
+		})
+	}
+}
--- a/docs/octo-cli-design.md
+++ b/docs/octo-cli-design.md
@@ -109,19 +109,18 @@ Notes:
 
 ---
 
-## Domain 4: thread (dmworkim bot_api, 9 commands)
+## Domain 4: thread (dmworkim bot_api, 8 commands)
 
 | # | Command | Method | Path | App Bot | User Bot |
 |---|---------|--------|------|---------|----------|
 | 31 | `octo thread create <group_no>` | POST | /v1/bot/groups/:group_no/threads | **N** | Y |
 | 32 | `octo thread list <group_no>` | GET | /v1/bot/groups/:group_no/threads | **N** | Y |
 | 33 | `octo thread get <group_no> <short_id>` | GET | .../:short_id | **N** | Y |
-| 34 | `octo thread delete <group_no> <short_id>` | DELETE | .../:short_id | **N** | Y |
-| 35 | `octo thread members <group_no> <short_id>` | GET | .../:short_id/members | **N** | Y |
-| 36 | `octo thread join <group_no> <short_id>` | POST | .../:short_id/join | **N** | Y |
-| 37 | `octo thread leave <group_no> <short_id>` | POST | .../:short_id/leave | **N** | Y |
-| 38 | `octo thread md-get <group_no> <short_id>` | GET | .../:short_id/md | **N** | Y |
-| 39 | `octo thread md-update <group_no> <short_id>` | PUT | .../:short_id/md | **N** | Y |
+| 34 | `octo thread members <group_no> <short_id>` | GET | .../:short_id/members | **N** | Y |
+| 35 | `octo thread join <group_no> <short_id>` | POST | .../:short_id/join | **N** | Y |
+| 36 | `octo thread leave <group_no> <short_id>` | POST | .../:short_id/leave | **N** | Y |
+| 37 | `octo thread md-get <group_no> <short_id>` | GET | .../:short_id/md | **N** | Y |
+| 38 | `octo thread md-update <group_no> <short_id>` | PUT | .../:short_id/md | **N** | Y |
 
 Flags:
 - create: --name* (thread name), --data (additional fields)
@@ -137,10 +136,10 @@ Notes:
 
 | # | Command | Method | Path | App Bot | User Bot |
 |---|---------|--------|------|---------|----------|
-| 40 | `octo file upload` | POST | /v1/bot/file/upload | Y | Y |
-| 41 | `octo file download <path>` | GET | /v1/bot/file/download/*path | Y | Y |
-| 42 | `octo file credentials` | GET | /v1/bot/upload/credentials | Y | Y |
-| 43 | `octo file presigned` | GET | /v1/bot/upload/presigned | Y | Y |
+| 39 | `octo file upload` | POST | /v1/bot/file/upload | Y | Y |
+| 40 | `octo file download <path>` | GET | /v1/bot/file/download/*path | Y | Y |
+| 41 | `octo file credentials` | GET | /v1/bot/upload/credentials | Y | Y |
+| 42 | `octo file presigned` | GET | /v1/bot/upload/presigned | Y | Y |
 
 Flags:
 - upload: --file* (multipart), --type(default:"chat"), --path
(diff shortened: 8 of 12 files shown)


Jerry-Xin (changes requested):
Summary: The PR is relevant to `octo-cli`, but the removed command currently fails open as a successful no-op/help response instead of an unknown command.

**🔴 Blocking**
- 🔴 Critical: `octo thread delete ...` now exits successfully and prints `octo thread` help instead of returning an unknown-command error. Since [internal/registry/specs/thread.json](/Users/mlamp/.openclaw/tmp/pr-review-1fdq_jb_/repo/internal/registry/specs/thread.json:49) removes the `delete` operation, Cobra falls back to the parent service command created at [cmd/service/service.go](/Users/mlamp/.openclaw/tmp/pr-review-1fdq_jb_/repo/cmd/service/service.go:53), which has no `RunE`/argument validation to reject extra args. I verified locally that `go run ./cmd/octo thread delete g s --dry-run` exits `0` and prints help. This contradicts the PR’s breaking-change note and can make automation treat a removed destructive command as successfully completed. Add a nested unknown-command regression test and make service parent commands reject unexpected args or otherwise return an error.

**💬 Non-blocking**
- 🟡 Warning: [README.md](/Users/mlamp/.openclaw/tmp/pr-review-1fdq_jb_/repo/README.md:43) still lists `thread` as `9` ops. The architecture count above it was updated to `47`, so this table should be changed to `8` to avoid count drift.

**✅ Highlights**
- The spec removal itself is clean: `thread.delete` no longer appears in `octo thread --help` or `octo schema --list thread`.
- `go test ./... -count=1` passes locally.

lml2468 (dismissed):
Clean, well-scoped removal. No blockers.

**Verified:**
- `thread.json` remains valid JSON after the `delete` block excision; the `/v1/bot/groups/{group_no}/threads/{short_id}` path retains its `get` sibling intact.
- Count cascade is consistent across all 6 touch-points: spec (47 ops via test), CLAUDE.md (47 ops / 50 cmds), README.md (47), design doc (renumbered #34–#50, summary table 50/38/12), SKILL.md (8 commands), CHANGELOG (51/48 → 50/47).
- No dangling `thread.delete` or `thread delete` references remain outside the main-branch search index (which reflects pre-merge state).
- Renumbering in `docs/octo-cli-design.md` is contiguous from #34 onward through Domain 7.
- `build & test (1.24.x)` CI green; `welcome / welcome` failure is a non-blocking contributor-greeting bot.

**Non-blocking note:**
The PR description correctly identifies the count-duplication tech debt (same number hand-copied in 6+ places, only guarded by `loader_test.go`). Agreed a de-dup chore would be worthwhile — but out of scope here.

LGTM. 👍

yujiawei (dismissed):
# Code Review — PR #20 (octo-cli)

**Verdict: APPROVED** — single P2 doc nit, no blocker.

## Summary

This PR removes the `thread.delete` operation from the metadata-driven CLI by
deleting the spec entry in `internal/registry/specs/thread.json` and
synchronising the operation/command counts across docs and the loader test.
The change is small (35 / -40), well-scoped, and the rationale (no other
high-risk-write op is exposed to bots; no soft-close path makes hard delete
the lone destructive exception) is sound and consistent with the existing
"no destructive bot operations" convention. The chosen shape — delete the
spec entry outright rather than `x-octo-disabled` — is appropriate given
there is no plan to re-enable it; this differs from the matter domain
(PR #19) where re-enable is on the roadmap.

## 1. Verification (✅ / ⚠️ / ❌ + evidence)

- ✅ Spec entry removed cleanly. `internal/registry/specs/thread.json:1-110`
  now contains 8 `operationId` entries (`thread.create`, `list`, `get`,
  `members`, `join`, `leave`, `md-get`, `md-update`); the `delete`
  operation under `/v1/bot/groups/{group_no}/threads/{short_id}` is gone
  and the path keeps its `get` sibling intact.
- ✅ `internal/registry/loader_test.go:40` updated to `"thread": 8` and
  the total in `TestAllDomainOperationCounts` is summed from the map, so
  no other assertion needed adjustment.
- ✅ `CLAUDE.md:25` headline updated to `47 operations / 50 commands incl.
  3 matter transition aliases`. Math checks: matter 14 + group 9 + thread 8
  + bot 6 + message 4 + file 4 + event 2 = **47** operations ✓.
- ✅ `CLAUDE.md:44` command tree drops `delete` from the thread row.
- ✅ `docs/octo-cli-design.md` — all four count copies updated coherently:
  domain heading (`9 → 8`), numbered command table (rows `#34..#50`
  contiguous after dropping the old `#34 thread delete`), summary table
  (thread `9→8`, total `51→50`, app-blocked `13→12`), availability lines
  (`38/51 → 38/50 (76%)`, `51/51 → 50/50 (100%)`).
- ✅ `skills/octo-messaging/SKILL.md:108` updated to `8 commands` and the
  `octo thread delete` example is removed.
- ✅ `CHANGELOG.md` adds a `### Removed` entry under `[Unreleased]`;
  `[0.4.0]` history is left untouched (correct per Keep a Changelog).
- ✅ Build / vet / test all green at HEAD `d516d45`:
  - `go build ./cmd/octo` — green
  - `go vet ./...` — green
  - `go test ./... -count=1` — all packages pass, including the updated
    `TestAllDomainOperationCounts`.
- ✅ Runtime surface verified after rebuild:
  - `octo thread --help` lists 8 subcommands (`create`, `get`, `join`,
    `leave`, `list`, `md-get`, `md-update`, `members`); no `delete`.
  - `octo schema --list` returns 8 thread operations; no `thread.delete`.
- ✅ No orphan references to `thread.delete` / `thread delete` in the
  tree except the intentional CHANGELOG mentions.
- ⚠️ One missed count in `README.md` — see Finding 1 below.

## 2. Findings

### P2 — `README.md` Domains table still shows `thread | 9` (stale count)

`README.md:43` was not updated:

```
| `matter`  | 14  | Todos/tasks — **temporarily withheld** while the backend API stabilizes |
| `group`   | 9   | Groups — list, get, members, metadata; create/update (User Bot)|
| `thread`  | 9   | Threads — create, list, get, members, join/leave, metadata     |
```

The PR correctly updated the `README.md:15` headline to `48 operations →
47 operations`, but the per-domain table immediately below still has
`thread | 9`. With `thread = 9` the table sums to 48 ops, contradicting
the new headline (47). This is the same per-file count-duplication tech
debt the PR description already calls out as a separate chore — the same
mistake the engine-side `TestAllDomainOperationCounts` is designed to
catch on the code side. Not a blocker (no behaviour or build impact),
but worth a one-line fix in this PR while the count change is fresh:

```diff
-| `thread`  | 9   | Threads — create, list, get, members, join/leave, metadata     |
+| `thread`  | 8   | Threads — create, list, get, members, join/leave, metadata     |
```

After the fix the table will sum to 47, matching the headline, `CLAUDE.md`,
the design doc, the SKILL doc, and the loader test.

## 3. Suggestions

- **Sweep with the chore issue.** The PR description already plans a
  separate chore to de-duplicate the operation/command counts. Worth
  considering as part of that chore: (a) a single `cmd/internal` constant
  or generated build tag exposing `OperationCount`, and (b) a `go test`
  that reads the markdown files and asserts the counts match the
  registry. That would have caught Finding 1 automatically.
- **Document the recovery path more discoverably.** The PR description
  notes that callers who genuinely need to delete a thread can fall back
  to `octo api DELETE /v1/bot/groups/{group_no}/threads/{short_id}`.
  Consider mirroring that one-line note into the `### Removed`
  CHANGELOG entry (it's already in the PR body, but the CHANGELOG is
  what end-users will read after upgrading and getting `unknown command`).
  Optional, not required.

## 4. Additional observations (out of scope, no action required)

- The PR’s asymmetric-treatment rationale (delete-outright vs.
  matter’s `x-octo-disabled`) is correct, but it relies on the spec
  loader not breaking when an entire HTTP method is dropped from a path
  while sibling methods on the same path remain. The current loader
  handles this — `internal/registry/specs/thread.json` keeps `get` under
  `/v1/bot/groups/{group_no}/threads/{short_id}` after `delete` is
  removed, and tests pass — but it would be worth a small unit test
  guarding "method removal under a multi-method path" so the loader
  can never silently regress this shape. Out of scope for this PR.
- The `thread` row in `docs/octo-cli-design.md` summary table reads
  `App-blocked 13 → 12 = group write (4) + thread all (8)` ✓ (math
  checks). Good catch keeping that arithmetic in sync.

## 5. Risk assessment

- **Behaviour:** breaking change for any caller of `octo thread delete`
  (now `unknown command`). PR description flags this clearly and offers
  the `octo api DELETE …` passthrough as a fallback. Backend route is
  untouched, so server-side semantics are unchanged. Acceptable.
- **Security / safety:** moves the project further toward the
  "no destructive bot operations" invariant; net positive. No new
  surface introduced.
- **Build / CI:** green locally; no code logic changed; only spec data
  and one test assertion.

Recommend merging once Finding 1 (single-line README fix) is addressed,
either in this PR or a same-day follow-up — it does not block correctness
or CI.

lml2468 (dismissed):
Re-review after e628574. Both prior findings resolved cleanly.

**Previous blocking — unknown subcommand exits 0:** Fixed via `rejectUnknownSubcommand` RunE on all parent commands. Cobra's resolution order (match subcommand first, fall back to parent RunE) means registered children route normally while unknown tokens get a loud `fmt.Errorf("unknown subcommand %q …")`. Regression test covers the removed `thread delete`, a bogus `thread` leaf, and a bogus `group` leaf — solid.

**Previous non-blocking — README thread count:** Fixed, now shows 8.

**New code review (e628574 delta):**
- `rejectUnknownSubcommand` is minimal and well-placed — applied at both `ensureServiceCmd` (domain roots) and `attachOperation` intermediate nodes.
- The `if len(args) > 0` guard is correct: bare `octo thread` with no args falls through to `cmd.Help()` (expected UX), while `octo thread <unknown>` errors.
- Test uses `rootWithService` with an HTTP handler that `t.Fatalf`s on call — confirms no network request is made for invalid subcommands.

No blockers. LGTM.

liuooo:
Related: #21 — tracks the hand-maintained-count tech debt this PR's diff surfaced (the missed `Domain 4: thread (... 9 commands)` heading and the README row that triggered the review fix). Not in scope for this PR.

Jerry-Xin (changes requested):
Summary: The spec removal is in scope and mostly correct, but the new parent-command `RunE` introduces an auth-gated help/unknown-command regression in the real root command.

🔴 Blocking

- 🔴 Critical — Parent service commands now require auth before showing help or reporting removed/unknown subcommands.  
  In [cmd/service/service.go](/Users/mlamp/.openclaw/tmp/pr-review-lbeh1kbc/repo/cmd/service/service.go:53), adding `RunE: rejectUnknownSubcommand` makes service parents runnable. With the real root, [cmd/root.go](/Users/mlamp/.openclaw/tmp/pr-review-lbeh1kbc/repo/cmd/root.go:25) runs `PersistentPreRunE` before that `RunE`, and [cmd/root.go](/Users/mlamp/.openclaw/tmp/pr-review-lbeh1kbc/repo/cmd/root.go:90) does not skip validation for service parents. As a result, unauthenticated `octo thread`, `octo group`, and even `octo thread delete g s` return `UNAUTHORIZED` instead of help or the intended unknown-subcommand error. This contradicts [cmd/service/service.go](/Users/mlamp/.openclaw/tmp/pr-review-lbeh1kbc/repo/cmd/service/service.go:103) and the PR’s stated breaking-change behavior. The new test at [cmd/service/service_test.go](/Users/mlamp/.openclaw/tmp/pr-review-lbeh1kbc/repo/cmd/service/service_test.go:530) misses this because it builds a minimal root without the real `PersistentPreRunE`. Please cover this through `cmd.NewRootCmd` or adjust validation so parent help/unknown-subcommand paths do not require credentials.

💬 Non-blocking

- 🔵 Suggestion — The PR description says this is “zero code change” from metadata removal, but the diff includes a real Cobra behavior change in `cmd/service/service.go`. It would be worth updating the PR text so reviewers and release notes reflect the CLI behavior change explicitly.

✅ Highlights

- The `thread.delete` operation is removed cleanly from [internal/registry/specs/thread.json](/Users/mlamp/.openclaw/tmp/pr-review-lbeh1kbc/repo/internal/registry/specs/thread.json:49), while `thread.get` remains intact.
- Operation counts and docs were updated consistently in the main touched files.
- `go test ./... -count=1` passes locally.

yujiawei (approved):
# Code Review — PR #20 (octo-cli)

## Verdict

Approve. The PR does exactly what the title says and nothing more: it removes the lone destructive `thread.delete` spec entry, syncs every count it touches, and adds a defensive guard so cobra parents no longer swallow removed/mistyped subcommands as exit-0. Net diff is +84/−41 across 9 files; logic surface is essentially one new helper.

## What I verified

### 1. Spec removal is the right shape

- `internal/registry/specs/thread.json` drops only the `delete` operation under `/v1/bot/groups/{group_no}/threads/{short_id}`; the `get` sibling and every other thread op stay intact (verified by reading the JSON; no orphaned refs).
- The metadata-driven engine picks up the removal automatically — no command-tree code touches `thread`. The PR description's claim that this also drops the command from `schema --list` and shell completions matches how `RegisterServiceCommands` walks `reg.ListOperations(svc)` (`cmd/service/service.go:33`).
- `thread.delete` was indeed the only `x-octo-risk: high-risk-write` op left in any *enabled* spec. `matter.delete` is also `high-risk-write` (`internal/registry/specs/matter.json:132`) but the matter spec carries `x-octo-disabled: true` (line 9), so it does not surface in `EnabledOperations` (`internal/registry/loader_test.go:201-212`).

### 2. Count sync across docs is consistent

Spot-checked every count touched by the diff. Math: 14 (matter) + 9 (group) + 8 (thread) + 6 (bot) + 4 (message) + 4 (file) + 2 (event) = **47 ops**, plus 3 matter status aliases = **50 commands**. Every reference matches:

| File | Reference | Value |
|---|---|---|
| `internal/registry/loader_test.go:40` | thread map entry | `8` ✓ |
| `CLAUDE.md:25` | header total | `47 operations / 50 commands` ✓ |
| `CLAUDE.md:44` | thread row | `create | list | get | members` ✓ (no `delete`) |
| `README.md:15` | architecture line | `47 operations` ✓ |
| `README.md:43` | domain table | `thread | 8` ✓ |
| `docs/octo-cli-design.md:112` | domain heading | `8 commands` ✓ |
| `docs/octo-cli-design.md:113-122` | numbered table | `#34..#38` (renumbered contiguously) ✓ |
| `docs/octo-cli-design.md:200` | summary row | `thread | 8 | 0 | 8 | 8 | 0` ✓ |
| `docs/octo-cli-design.md:204` | totals | `50 | 38 | 12 | 50 | 0` ✓ |
| `docs/octo-cli-design.md:206-208` | App Bot lines | `38/50 (76%)`, `50/50`, `4 + 8 = 12` ✓ |
| `skills/octo-messaging/SKILL.md:108` | thread heading | `8 commands` ✓ |
| `CHANGELOG.md:11-15` | Removed entry, `51/48 → 50/47` | ✓ |

I `grep`ed for any lingering `thread.{0,5}delete`, `9 commands` / `9 operations`, and `DELETE.*threads` outside of the expected places (CHANGELOG narrative, the regression test, the `/v1/bot/groups/{group_no}/threads/{short_id}` path in the spec for the surviving `get`). Nothing stale found.

### 3. Parent-command guard is a real win

The follow-up commit `e628574` is the part of this PR worth flagging as load-bearing: before it, removing a leaf was incomplete. Cobra's behavior on a parent with no `RunE` is to silently `cmd.Help()` with exit 0 when handed extra args, so `octo thread delete g s` would have *successfully exited* after the spec removal — which is the exact failure mode the "no destructive bot operations" convention is trying to prevent (automation can't tell removal from success).

The fix:

- `cmd/service/service.go:56,79` — adds `RunE: rejectUnknownSubcommand` to both `ensureServiceCmd` (service-level parent) and the intermediate-resource creation path in `attachOperation`. This covers all generated parents, not just `thread`.
- `cmd/service/service.go:103-108` — helper returns `unknown subcommand %q for %q` on extra args; falls back to `cmd.Help()` (no error) when called bare. So `octo thread` with no args still prints help and exits 0, which is the intended UX.
- Cobra still routes `octo thread create ...` to the leaf `RunE` because `findChild` matches first — the parent `RunE` only fires when no child matches.
- `cmd/service/service_test.go:530-557` — `TestParentCommand_RejectsUnknownSubcommand` covers three angles: a real-removed leaf (`thread delete`), an arbitrary unknown leaf on the same domain (`thread bogus`), and another domain (`group nope`). The HTTP handler is wired with `t.Fatalf` on entry, which gives a strong assertion that the parent rejects *before* any backend round-trip — important for "removed destructive op" semantics.

I ran `go build ./...`, `go vet ./...`, and `go test ./... -count=1` against the head SHA `e628574d410c46d24a9d134b3d98ecd5687fb1c6`. All green, including `TestRegisterServiceCommands_TreeShape` which would have caught a stray `thread.delete` re-introduction.

### 4. Pre-existing tests still hold

- `TestAllDomainOperationCounts` — updated `thread: 8`, total derived; passes.
- `TestRegisterServiceCommands_TreeShape` — does not enumerate `thread` children, so no incidental update needed; the test still passes against the new tree.
- The `--data` / `octo api DELETE` fallback path is unchanged, matching the PR description's "callers that genuinely need to delete a thread can still construct the call via `octo api DELETE /v1/bot/groups/{group_no}/threads/{short_id}`."

## Findings

### P2 — `WrapCLIError` heuristic does not match the new error string

**Location:** `internal/cmdutil/factory.go:351-365` vs. `cmd/service/service.go:103-108`.

The heuristic in `WrapCLIError` classifies cobra-framework error messages into envelope types by substring match: `"unknown command"`, `"unknown flag"`, `"required flag"`, `"invalid argument"`, `"accepts "`, `"requires at "`, `"arg(s)"` → `validation`. The new helper returns `unknown subcommand %q for %q`, which contains *none* of these substrings, so it falls through to the default branch → `output.ErrWithHint("config", "CLI_ERROR", ...)`.

Effects:
- Exit code is unaffected — `validation` and `config` both map to exit 2 (`internal/output/errors.go:38-39`), so the PR description's "exits 2 with a validation envelope" claim is half-right: code is correct, type is `config`/`CLI_ERROR`, not `validation`/`VALIDATION_ERROR`.
- Agent-facing JSON envelope `type` field is therefore inconsistent with the analogous "unknown command" path that cobra's own framework would produce.

Two equally cheap fixes:
1. Word the helper as `unknown command %q under %q` so the existing heuristic catches it.
2. Add `strings.Contains(lower, "unknown subcommand")` to the validation case in `WrapCLIError`.

I'd lean toward (2) — it's one line and keeps the more precise "subcommand" wording in user-visible output.

Not a merge blocker; the core safety property (non-zero exit) holds.

### P2 — CHANGELOG entry under-sells the second commit

The CHANGELOG `[Unreleased] ### Removed` block only mentions `octo thread delete`. The second commit (`e628574`) is also user-visible: every parent command now rejects unknown subcommands with exit ≠ 0, where it previously exited 0 with help text. That's a behaviour change for *any* mistyped subcommand across the whole CLI, not just `thread delete`. Worth a one-line `### Changed` entry under the same `[Unreleased]` so users grepping for "why does my pipeline now error on a typo" can find it.

### P2 (nit) — Test could assert exit-code path end-to-end

`TestParentCommand_RejectsUnknownSubcommand` confirms `err != nil` and the message contains `unknown subcommand`. It does *not* exercise `WrapCLIError` / `output.AsExitError`, which is the path `cmd/octo/main.go:34` actually uses to derive the exit code. A small follow-up assertion (`AsExitError(WrapCLIError(err)).ExitCode() == 2`) would lock down the property that matters at the process boundary and would have surfaced finding 1 above. Not blocking for this PR.

## Things I considered and ruled out

- **Tree-shape test on disabled `matter`**: `TestRegisterServiceCommands_TreeShape` (line 60) still lists `delete` under `matter`. This is unrelated to PR #20 — `matter` is `x-octo-disabled` but still loaded into the raw tree (`TestEnabledServicesExcludesDisabledButKeepsLoaded`), and `RegisterServiceCommands` registers all loaded services. Pre-existing behaviour, out of scope.
- **`octo schema --list thread` no longer surfacing `thread.delete`**: confirmed this flows from the registry change, not a separate code path. No additional code update needed.
- **`octo thread` with zero args**: helper returns `cmd.Help()` (nil error), so the historical UX of "show help on bare parent" is preserved. Verified by reading `rejectUnknownSubcommand`.
- **Backend route still reachable via `octo api DELETE ...`**: passthrough is unchanged by this PR; PR description's fallback advice is accurate.

## Summary

Single-purpose, well-scoped, count-perfect, with a real teeth-fix (parent `RunE` guard) that closes the silent-exit-0 hole that would otherwise have undermined the spec removal. The two P2 items are small consistency/UX polish and do not block merge. Approve.

yujiawei (changes requested):
# Code Review — PR #20 (octo-cli) — Revised

Revising my earlier APPROVED review on this same head SHA. Another reviewer flagged a regression in the parent-`RunE` guard that I missed; I reproduced it locally and the finding is correct. Switching my verdict to CHANGES_REQUESTED.

## Blocker (P1) — Parent service commands now demand auth before they can show help or report unknown subcommands

**Location:** `cmd/service/service.go:53-57,76-80,103-108` interacting with `cmd/root.go:25-34,90-99`.

### What broke

Adding `RunE: rejectUnknownSubcommand` to every service parent and every intermediate-resource parent makes those nodes `Runnable()`. In cobra, when a parent is not runnable and gets unknown args, `c.execute` returns `flag.ErrHelp` *before* the `PersistentPreRunE` chain fires — that is exactly why pre-PR `octo thread delete g s` printed help with exit 0. Once the parent has `RunE`, cobra now flips into the runnable branch, the `PersistentPreRunE` on root runs first, and that hook calls `cfg.Validate()` for any command that doesn't pass `skipValidation` (`cmd/root.go:90-99`). `skipValidation` allow-lists only `version | help | schema | config | completion | skills | auth | ""`, so service parents (`thread`, `group`, `bot`, `file`, `event`, `message`) are *not* exempt.

Net effect: auth is now required to either bring up a parent's help text or to surface the new "unknown subcommand" error.

### Reproduced locally on `e628574d410c46d24a9d134b3d98ecd5687fb1c6`

```
$ unset OCTO_BOT_TOKEN OCTO_API_BASE_URL
$ go build -o /tmp/octo ./cmd/octo

$ /tmp/octo thread             # used to print help, exit 0
{"error":{"code":"UNAUTHORIZED","type":"auth_error",
          "message":"OCTO_BOT_TOKEN is required (App Bot token, app_*)"}, ...}
exit=3

$ /tmp/octo thread bogus       # should report "unknown subcommand"
{"error":{"code":"UNAUTHORIZED","type":"auth_error", ...}}
exit=3

$ /tmp/octo thread delete g s  # should report "unknown subcommand"
{"error":{"code":"UNAUTHORIZED","type":"auth_error", ...}}
exit=3
```

For comparison, the explicit `--help` flag still works (cobra short-circuits on the help flag before `PersistentPreRunE`):

```
$ /tmp/octo thread --help
Operations on the thread domain
...
exit=0
```

### Why this matters

1. **Contradicts the PR's stated breaking-change behavior.** The fix's whole point is to surface a `unknown subcommand %q for %q` error so automation cannot mistake a removed destructive op for a no-op. Right now an unauthenticated caller — including most CI agents that haven't yet picked up a token — sees `auth_error / UNAUTHORIZED`, not the precise removal signal. Authenticated callers still get the intended "unknown subcommand" envelope, so the *safety property* (non-zero exit) is preserved end-to-end, but the *diagnostic* is wrong for the unauthenticated half of the matrix.
2. **UX regression on command discovery.** `octo thread`, `octo group`, etc. used to be a zero-config way to see what subcommands exist. After this PR, you must be authenticated to do that — even though `octo thread --help` still works. Easy to mistype while exploring; less easy to remember to add `--help`.
3. **The new test does not cover this path.** `TestParentCommand_RejectsUnknownSubcommand` (`cmd/service/service_test.go:530-557`) builds a minimal `cobra.Command{Use:"octo"}` root with no `PersistentPreRunE`, so the `cfg.Validate()` step never runs in the test. That is why the regression slipped through. My earlier review missed this for the same reason — I read only the service-level test and did not run the binary under the real root.

### Suggested fixes (any one is fine)

- **Annotate the parents and let `skipValidation` see it.** When `ensureServiceCmd` / `attachOperation` set `RunE: rejectUnknownSubcommand`, also set `cmd.Annotations = map[string]string{"skipValidation": "true"}` (or similar). Extend `skipValidation` to honor that annotation. Keeps all the policy in `cmd/root.go`.
- **Use a sentinel `RunE`-backed allowlist.** In `skipValidation`, treat any cmd whose `RunE` is `rejectUnknownSubcommand` as auth-skipped. Slightly more coupled but no annotations to thread through.
- **Re-validate Cobra-level help inside `rejectUnknownSubcommand`.** Move the auth gate into a `PreRunE` that the leaf operation commands set explicitly, instead of a global root `PersistentPreRunE`. Bigger refactor, lower priority for this PR.

In all variants, please add a regression test that drives `cmd.NewRootCmd(f)` (the real root) with no token set and asserts:

- `octo thread`               → exit ≠ 0, error type `validation`/`config` (not `auth_error`), or exit 0 with help (matching pre-PR UX).
- `octo thread bogus`         → exit ≠ 0, error type `validation`, message contains `unknown subcommand`.
- `octo thread delete g s`    → same as above (this is the actual safety property).

## Other findings (still P2, restated from prior review)

These are unchanged and remain non-blocking on their own — but if you're respinning the PR for the P1 above, they are cheap to roll in.

### P2 — `WrapCLIError` does not classify "unknown subcommand"

`internal/cmdutil/factory.go:351-365` wraps cobra-framework messages by substring. The phrases it looks for are `"unknown command"`, `"unknown flag"`, `"required flag"`, `"invalid argument"`, `"accepts "`, `"requires at "`, `"arg(s)"`. The new helper emits `unknown subcommand %q for %q`, which matches none of them, so the envelope `type` becomes `config` / `CLI_ERROR` instead of `validation` / `VALIDATION_ERROR`. Exit code is unaffected (both map to 2 in `internal/output/errors.go:38-39`), but the agent-facing JSON `type` field is inconsistent with the analogous "unknown command" path. Two equally cheap fixes:
1. Word the message as `unknown command %q under %q` so the existing heuristic catches it.
2. Add `strings.Contains(lower, "unknown subcommand")` to the validation case in `WrapCLIError`.

### P2 — CHANGELOG underspecifies the second commit

The `[Unreleased] ### Removed` block only mentions `octo thread delete`. Commit `e628574` is also user-visible: parent commands now reject unknown subcommands (and, per the P1 above, also reject unauthenticated callers). Worth a one-line `### Changed` entry so users grepping for "why does my pipeline now error on a typo / on a missing token" can find it.

### P2 (nit) — Test exit-code path is not exercised end-to-end

`TestParentCommand_RejectsUnknownSubcommand` confirms `err != nil` and the error string contains `unknown subcommand`. It does not exercise `WrapCLIError` / `output.AsExitError`, which is the path `cmd/octo/main.go:34` actually uses to derive the exit code, and it does not exercise `PersistentPreRunE` at all. Driving the assertion from `cmd.NewRootCmd(f)` would have caught both the P2 envelope-type miss above and the P1 auth-gate regression in this review.

## What still verifies cleanly

The spec / count work in the first commit (`d516d45`) is unaffected by the P1 regression and remains a clean removal:

- `internal/registry/specs/thread.json` — drops only the `delete` operation; `get` sibling intact.
- All 6+ count touch-points (CLAUDE.md, README.md, design doc, SKILL.md, CHANGELOG, loader_test) sum to 47 operations / 50 commands consistently.
- `thread.delete` was the only `x-octo-risk: high-risk-write` op in any *enabled* spec — `matter.delete` is also `high-risk-write` but the matter spec is `x-octo-disabled: true`.
- `go build ./...`, `go vet ./...`, `go test ./... -count=1` all green at HEAD.

The intent of the parent-`RunE` guard is correct and worth keeping; it just needs to interact with the auth gate cleanly before this can ship.

Jerry-Xin (approved):
Project relevance gate passed; this PR is in scope for octo-cli and is ready to merge.

🔴 Blocking: None.

💬 Non-blocking: None.

✅ Highlights:
- 🔵 The `thread.delete` operation is removed cleanly while preserving the `thread.get` sibling in `internal/registry/specs/thread.json:49`.
- 🔵 Service parent commands now reject removed or mistyped subcommands instead of silently printing help and exiting success in `cmd/service/service.go:53` and `cmd/service/service.go:105`.
- 🔵 The auth-skip annotation is scoped to the exact parent command, so service leaves still require credentials as intended in `cmd/root.go:90`.
- 🔵 Regression coverage is solid: parent unknown-subcommand behavior and leaf auth requirements are tested in `cmd/root_test.go:22` and `cmd/root_test.go:93`.

Verification run:
- `go test ./... -count=1`
- `go vet ./...`
- Manual checks confirmed `octo thread --help` lists 8 subcommands, `schema --list thread` omits `thread.delete`, and `octo thread delete ...` returns a validation error.

yujiawei (approved):
# Code Review — PR #20 (octo-cli)

## Verdict

Approve at `afc015f`. Every P1 and P2 from the prior `CHANGES_REQUESTED` review is addressed cleanly, and no new regressions were introduced. The PR now does what its title promises (remove `thread.delete`), adds a real safety net (parent commands reject unknown subcommands instead of silently exiting 0), and keeps the auth-gate semantics correct for both the parent help path and the leaf operation path.

## What changed since the last review (`e628574` → `afc015f`)

The previous reviewers flagged one P1 (auth gate now runs in front of parent help / unknown-subcommand) and three P2 items. `afc015f` lands all four fixes in a single, well-scoped commit:

1. `cmd/service/service.go:53-58, 75-82` — both `ensureServiceCmd` and the intermediate-parent creation path in `attachOperation` now also set `Annotations: map[string]string{"skipValidation": "true"}` alongside `RunE: rejectUnknownSubcommand`.
2. `cmd/root.go:90-105` — `skipValidation` now consults `cmd.Annotations["skipValidation"]` first and **deliberately does not walk the parent chain for that branch**. This is the correct shape: the exemption applies to `octo thread` itself, but not to `octo thread create …`, because leaves built in `buildOperationCmd` never set the annotation. The pre-existing name-based parent-chain allowlist (`version`, `help`, `schema`, `config`, `completion`, `skills`, `auth`, `""`) is kept untouched.
3. `internal/cmdutil/factory.go:355-360` — `WrapCLIError` recognises `"unknown subcommand"` alongside `"unknown command"`. The new helper now produces the same `validation` / `VALIDATION_ERROR` envelope the analogous cobra-framework "unknown command" path produces, instead of falling through to `config` / `CLI_ERROR`. Exit code was already 2 in both cases; the agent-facing envelope `type` is now also consistent.
4. `CHANGELOG.md:17-23` — adds a `### Changed` entry under `[Unreleased]` documenting the new "service parent rejects unknown subcommand" behaviour. This is the user-visible change that wasn't called out before.

## Verification at `afc015f`

### Behavioural matrix (binary built locally, `OCTO_BOT_TOKEN` unset)

| Invocation | Pre-PR | After `e628574` (broken) | After `afc015f` |
|---|---|---|---|
| `octo thread`              | help, exit 0 | UNAUTHORIZED, exit 3 | help, exit 0 ✅ |
| `octo thread --help`       | help, exit 0 | help, exit 0 | help, exit 0 ✅ |
| `octo thread bogus`        | help, exit 0 (silent fail-open) | UNAUTHORIZED, exit 3 | `unknown subcommand`, validation, exit 2 ✅ |
| `octo thread delete g s`   | help, exit 0 (silent fail-open) | UNAUTHORIZED, exit 3 | `unknown subcommand`, validation, exit 2 ✅ |
| `octo group list` (leaf)   | UNAUTHORIZED, exit 3 | UNAUTHORIZED, exit 3 | UNAUTHORIZED, exit 3 ✅ |

The last row is the load-bearing one for the annotation fix: leaves under a parent that carries `skipValidation: "true"` must **not** inherit the exemption. They don't, because the annotation lookup is on `cmd` itself, not the parent chain. Verified both at the binary level and via the new `TestNewRootCmd_LeafStillRequiresAuth` test.

### Test coverage

- `TestNewRootCmd_ParentCommandSkipAuthAndRejectUnknown` (`cmd/root_test.go:22-86`) — drives the **real** `NewRootCmd(f.Factory)`, so `PersistentPreRunE` and `WrapCLIError` are both in the loop. Asserts both the error type (`validation`) and the substring (`unknown subcommand`) for `octo thread bogus` and `octo thread delete g s`, and asserts nil-error / help for bare `octo thread`. This is exactly the path the prior review asked for.
- `TestNewRootCmd_LeafStillRequiresAuth` (`cmd/root_test.go:93-114`) — pins down the non-inheritance property of the annotation. Without this guard the annotation fix could silently exempt every operation; with it, a leaf-level regression would fail loudly.
- Existing `TestParentCommand_RejectsUnknownSubcommand` (`cmd/service/service_test.go:530`) still passes; the HTTP fatal handler proves no backend round-trip happens for an unknown subcommand.
- `TestAllDomainOperationCounts` (`internal/registry/loader_test.go:40`) still asserts `"thread": 8`, total derived; passes.

### Build / vet / test

```
go build ./...        — green
go vet ./...          — green
go test ./... -count=1 — all packages green
```

### Count consistency (re-verified at `afc015f`)

The first commit's count cascade is unaffected by the second-and-third commit. Spot-checked again:

| File | Reference | Value |
|---|---|---|
| `internal/registry/loader_test.go:40` | thread map entry | `8` ✓ |
| `CLAUDE.md:25` | header total | `47 operations / 50 commands` ✓ |
| `CLAUDE.md:44` | thread row | no `delete` ✓ |
| `README.md:15` | architecture line | `47 operations` ✓ |
| `README.md:43` | domain table thread row | `8` ✓ |
| `docs/octo-cli-design.md:112,113-122,200,204,206-208` | heading / numbered table / summary / availability | all 47/50/38/12 consistent ✓ |
| `skills/octo-messaging/SKILL.md:108` | thread heading | `8 commands` ✓ |
| `CHANGELOG.md` | `### Removed` (51/48 → 50/47) and new `### Changed` | both present ✓ |

`grep`s for `thread.{0,5}delete`, `9 commands` / `9 operations`, and `DELETE.*threads` outside the expected places (CHANGELOG narrative, regression test, surviving `get` path) returned nothing stale.

### Runtime surface

- `octo thread --help` lists 8 subcommands (`create`, `get`, `join`, `leave`, `list`, `md-get`, `md-update`, `members`); no `delete`.
- `octo schema --list` returns 8 thread operations; no `thread.delete`.

## Things I considered and ruled out

- **Annotation accidentally exempting too much.** `Annotations["skipValidation"]` is set only by `ensureServiceCmd` (service-level parents) and the intermediate-parent branch of `attachOperation` (the `for i := 1; i < len(segs)-1` loop). The leaf `buildOperationCmd` path (line 114) never sets it. So `octo matter assignee list` (a leaf under an intermediate `assignee` parent) still authenticates; the exemption stops at the parent.
- **Nil-map safety.** `cmd.Annotations["skipValidation"]` is safe even when `cmd.Annotations` is nil — Go indexes into a nil map and returns the zero value (`""`), no panic. No defensive `if cmd.Annotations == nil` needed.
- **`octo thread --invalid-flag`.** Cobra rejects unknown flags during flag parsing, before `PersistentPreRunE` runs. The error string contains `"unknown flag"`, which `WrapCLIError` already classifies as `validation`. Unchanged path; unaffected by this PR.
- **`matter` regression risk.** `matter` carries `x-octo-disabled: true`, so `withholdDisabledServices` strips its subtree from the root after registration. The `skipValidation` annotation on the (now-removed) `matter` parent is therefore unreachable from the CLI; only seen by engine tests that drive `RegisterServiceCommands` directly, which already account for `matter`. No issue.
- **`octo schema --list thread`.** Flows from the registry change, not a separate code path. Confirmed not surfacing `thread.delete`.

## Non-blocking suggestions (P3, defer or roll into the chore in #21)

- The `### Changed` CHANGELOG entry mentions service parents reject unknown subcommands, which is good. It might also be worth a one-liner that the **error envelope type** for those rejections is `validation` (exit 2), so automation that branches on `error.type` doesn't need to discover this empirically. Pure documentation polish.
- The new test at `cmd/root_test.go` is excellent. If you wanted to fully nail down the safety property for downstream automation, one extra assertion could check `ee.ExitCode == 2` for `validation` and `== 3` for `auth_error` — that locks the contract the dispatcher observes at `os.Exit`. Optional.
- The hand-maintained-count tech debt (already tracked in #21) remains the structural fix for what this PR had to chase across 6+ files. Not in scope here, but a `go test` that reads the markdown files and asserts each count matches `len(reg.ListOperations(svc))` would have caught both the original `README.md` miss and any future drift in one stroke.

## Risk assessment

- **Behaviour:** breaking change for `octo thread delete` callers (now `unknown subcommand`, exit 2). Also a behaviour change for **any** mistyped subcommand across the whole CLI: `octo group lisst` now fails loudly with exit 2 instead of silently printing help with exit 0. Both are user-visible, both are documented in CHANGELOG, both are aligned with the "no destructive bot operations" invariant. Acceptable.
- **Security / safety:** net positive. Removes the lone `high-risk-write` enabled op and closes the silent-exit-0 hole that would otherwise let automation mistake removal for success. The auth-gate semantics for leaf operations are unchanged — the annotation is non-inherited by design.
- **Build / CI:** green at HEAD. No code logic regressions.

Recommend merge.

PR #24, opened 2026-05-28
ci: add build-artifacts job to release-publish workflow

lml2468:
## 背景

`release.yml` 靠 `push tag v*` 触发 goreleaser，但 tag protection ruleset 让 `GITHUB_TOKEN` 永远无法推 tag，该 workflow 事实上是死代码。

## 本次改动

在 `release-publish.yml` 的 `publish` job 之后追加 `build-artifacts` job：

| 项目 | 说明 |
|------|------|
| 触发条件 | `needs: publish` + `inputs.draft == false` |
| 构建 | goreleaser `--skip=publish,announce`（只构建打包，不操作 GitHub Release） |
| 产物 | linux/darwin/windows × amd64/arm64 的 .tar.gz / .zip + checksums.txt |
| 上传 | `gh release upload --clobber` 附加到已发布的 release |

## 使用前提

**tag 必须由授权用户（`cli-maintainers` team 成员）提前推送**，`GITHUB_TOKEN` 没有 tag protection bypass 权限。

## 发布流程（v0.5.0 起）

1. `lml2468` 或 `liuooo` 推 tag：`git tag v0.5.0 b5a4d6a3 && git push origin v0.5.0`
2. Actions → Release Publish → Run workflow，填入：
   - tag: `v0.5.0`
   - validate_run_id: `26559942509`
   - draft: `false`
3. `publish` job 发布 release，`build-artifacts` job 构建并上传二进制

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -29,3 +29,38 @@ jobs:
     permissions:
       contents: write
       actions: read
+
+  build-artifacts:
+    needs: publish
+    if: ${{ inputs.draft == false }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.tag }}
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.24.x"
+          cache: true
+
+      - name: Build with GoReleaser
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          version: "~> v2"
+          args: release --clean --skip=publish,announce
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: List dist artifacts
+        run: ls -lh dist/
+
+      - name: Upload artifacts to release
+        run: gh release upload "${{ inputs.tag }}" dist/*.tar.gz dist/*.zip dist/checksums.txt --clobber
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -60,7 +60,14 @@ jobs:
       - name: List dist artifacts
         run: ls -lh dist/
 
-      - name: Upload artifacts to release
-        run: gh release upload "${{ inputs.tag }}" dist/*.tar.gz dist/*.zip dist/checksums.txt --clobber
+      - name: Validate tag before upload
+        run: |
+          [[ "$TAG" =~ ^v[0-9]+\.[0-9]+(\.[0-9]+)?([-.][0-9A-Za-z.-]+)?$ ]] || { echo "Invalid tag: $TAG"; exit 1; }
         env:
+          TAG: ${{ inputs.tag }}
+
+      - name: Upload artifacts to release
+        run: gh release upload "$TAG" dist/*.tar.gz dist/*.zip dist/checksums.txt --clobber
+        env:
+          TAG: ${{ inputs.tag }}
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}


lml2468 (commented):
[APPROVE]

Clean addition — the `build-artifacts` job correctly chains after `publish`, builds with GoReleaser `--skip=publish,announce`, and uploads binaries to the existing release. Logic, permissions, and conditions are all sound.

💬 **Non-blocking notes**

1. **Defense-in-depth: use `env` instead of inline `${{ inputs.tag }}` in `run:` steps** — The `publish` job upstream validates the tag as strict semver before `build-artifacts` runs, so there is no actual injection vector today. However, the reusable workflow itself consistently uses the `env: TAG: ${{ inputs.tag }}` → `"$TAG"` pattern in all its `run:` blocks. Matching that convention here would be more defensive:
   ```yaml
   - name: Upload artifacts to release
     run: gh release upload "$TAG" dist/*.tar.gz dist/*.zip dist/checksums.txt --clobber
     env:
       TAG: ${{ inputs.tag }}
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   ```
   Low priority since the semver gate in `publish` already sanitises the value.

✅ **Highlights**

- `needs: publish` + `if: inputs.draft == false` correctly gates artifact builds on non-draft publishes
- `--skip=publish,announce` avoids double-publishing or overwriting the release created by the reusable workflow
- `--clobber` on upload makes the job idempotent on re-runs
- `fetch-depth: 0` ensures GoReleaser can resolve version from git tags
- Permissions are minimal (`contents: write` only)

Jerry-Xin (changes requested):
This PR is relevant to `Mininglamp-OSS/octo-cli`: it updates the project’s release workflow to build and attach CLI artifacts.

🔴 Blocking

- 🔴 Critical: `.github/workflows/release-publish.yml:64` interpolates the manual `inputs.tag` value directly into a shell command:
  ```yaml
  run: gh release upload "${{ inputs.tag }}" ...
  ```
  Because this workflow runs with `contents: write`, a malicious or malformed tag name containing shell metacharacters can break out of the quoted string and execute commands with the workflow token. Git refs can contain characters such as quotes and semicolons, so this should not rely on tag naming discipline or validation inside the reusable workflow. Fix by validating the tag before use and passing it through an environment variable:
  ```yaml
  - name: Validate tag
    run: |
      [[ "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+([-.][0-9A-Za-z.-]+)?$ ]]
    env:
      TAG: ${{ inputs.tag }}

  - name: Upload artifacts to release
    run: gh release upload "$TAG" dist/*.tar.gz dist/*.zip dist/checksums.txt --clobber
    env:
      TAG: ${{ inputs.tag }}
      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  ```

💬 Non-blocking

- 🟡 Warning: `.github/workflows/release-publish.yml:33-64` publishes the release before building/uploading artifacts. If GoReleaser or upload fails, the release remains published without binaries. That may be acceptable for the chosen flow, but a draft-first flow or an explicit recovery note would reduce release risk.

✅ Highlights

- 🔵 Suggestion: The new job matches the repository’s Go version (`1.24.x`) and existing GoReleaser v2 configuration.
- 🔵 Suggestion: The artifact upload glob matches `.goreleaser.yaml`, which produces Linux/macOS tarballs, Windows zip archives, and `checksums.txt`.

yujiawei (dismissed):
# Code Review — PR #24 (octo-cli)

## Summary

Adds a `build-artifacts` job to `release-publish.yml` that runs after the reusable `publish` job, builds cross-platform binaries via GoReleaser with `--skip=publish,announce`, and uploads the archives + checksums to the existing release with `gh release upload --clobber`. The change is small (35/+0/−0, single workflow file) and well scoped.

## Verdict: APPROVED

No P0/P1 blockers. A couple of minor suggestions below — none gate merge.

## What works well

- **Ordering is correct.** `needs: publish` ensures the release exists before the upload step runs, and `if: ${{ inputs.draft == false }}` skips the build for draft releases (binaries aren't attached to drafts in this design — matches the PR description).
- **No double-publish.** `goreleaser release --clean --skip=publish,announce` builds + archives + computes checksums but does not touch the GitHub release. The release lifecycle stays owned by the reusable `publish` job, and `build-artifacts` only attaches files. This avoids the most common GoReleaser-in-CI footgun of two jobs racing on the same release.
- **Idempotent upload.** `gh release upload --clobber` makes re-runs safe (e.g. if the upload step fails halfway and is retried).
- **Permissions are minimum-viable.** Only `contents: write` on the new job; `actions: read` is correctly limited to `publish` (which needs it for the `validate_run_id` lookup).
- **Tag-protection workaround is documented.** The PR body and commit message both call out that the tag must be pre-pushed by a `cli-maintainers` member because `GITHUB_TOKEN` cannot bypass tag protection. This matches the dead-`release.yml` analysis.
- **GoReleaser config alignment.** `.goreleaser.yaml` is `version: 2` and the action pin `~> v2` matches; `goos`/`goarch` matrix covers linux/darwin/windows × amd64/arm64, archives produce `tar.gz` (and `zip` for Windows) plus `checksums.txt`, which is exactly what the upload glob targets.
- **`fetch-depth: 0`** is necessary for GoReleaser to see the tag and compute `{{.Version}}` / `{{.Commit}}` / `{{.Date}}` in the ldflags — correctly set.

## Suggestions (non-blocking, P2 / nit)

### 1. Add a `concurrency` group keyed on the tag

Two manual dispatches against the same tag could race the upload step. `--clobber` mostly saves you, but you'd still pay double the build cost and have a small window where partial uploads land. A single line on the job (or workflow) prevents it:

```yaml
concurrency:
  group: release-publish-${{ inputs.tag }}
  cancel-in-progress: false
```

`cancel-in-progress: false` is intentional here — you want the second dispatch to queue, not cancel, in case the first run is mid-upload.

### 2. The `List dist artifacts` step is debug-only

```yaml
- name: List dist artifacts
  run: ls -lh dist/
```

Useful while bringing the workflow up; once green, it's noise in the run log. Optional to remove — harmless either way.

### 3. Consider pinning actions to commit SHAs

`actions/checkout@v4`, `actions/setup-go@v5`, `goreleaser/goreleaser-action@v6` follow the rest of the repo's convention, so this is purely a future hardening note rather than a request: pinning to immutable SHAs (with Renovate/Dependabot to update) closes a tag-mutation supply-chain vector. Not a change I'd block this PR on.

### 4. Minor: `setup-go` cache hit

`cache: true` relies on `go.sum` at the checked-out ref. Since GoReleaser doesn't touch `go.sum` between runs, the cache will be reused across release builds — fine, just noting it works correctly.

## What I checked but found no issue with

- **Failure modes of `publish`.** If the reusable `publish` job fails, `needs: publish` short-circuits `build-artifacts` — no orphan binaries get attached to a non-existent release.
- **`inputs.draft == false` semantics.** GitHub expression evaluates the boolean correctly; default is `false`, so omitting the input also runs `build-artifacts` (consistent with the documented intent).
- **Glob expansion.** `dist/*.tar.gz dist/*.zip dist/checksums.txt` runs in bash on `ubuntu-latest`; globs expand before `gh release upload` sees them. With the GoReleaser config's archive matrix, all six archives + `checksums.txt` will exist, so no "no matches" error.
- **Token scope.** `GITHUB_TOKEN` with `contents: write` is sufficient for `gh release upload` against a release in the same repo.
- **Old `release.yml` (push-tag trigger).** Out of scope for this PR; the PR body correctly identifies it as effectively dead under the current tag protection ruleset. A follow-up to delete it would be reasonable but is not required here.

lml2468:
Fixed in ff0533e.

**Jerry-Xin 🔴 blocking → addressed:**
- Added `Validate tag before upload` step: regex check via `$TAG` env var, fails fast if format invalid
- Changed `gh release upload` to use `$TAG` env var instead of direct `${{ inputs.tag }}` interpolation

**Jerry-Xin 🟡 non-blocking (publish-before-artifacts):**
- Current flow is intentional — publish job validates CI run and publishes the draft, then artifacts are uploaded. If goreleaser fails, release exists without binaries. This is an acceptable tradeoff for the manual workflow_dispatch flow; a recovery path is to re-run the `build-artifacts` job independently once goreleaser is fixed.

Jerry-Xin (approved):
The PR is in scope and the added job fits this repository’s release workflow; I found no blocking correctness or security issues.

💬 Non-blocking

🟡 Warning — Move tag validation before checkout/build for defense in depth. In [.github/workflows/release-publish.yml](/Users/mlamp/.openclaw/tmp/pr-review-i23kkd2l/repo/.github/workflows/release-publish.yml:63), validation happens after the tag input is already used by `actions/checkout` at line 43 and GoReleaser at line 56. The upstream reusable `publish` job validates the tag first, so this is not blocking, but this job would be safer and easier to reason about if it validated before checkout and reused the same strict SemVer pattern.

🟡 Warning — A public release can briefly or permanently exist without binaries if `build-artifacts` fails after `publish`. The dependency order at [.github/workflows/release-publish.yml](/Users/mlamp/.openclaw/tmp/pr-review-i23kkd2l/repo/.github/workflows/release-publish.yml:34) means `publish` completes before artifacts are built and uploaded. This matches the PR design, but operationally a failed GoReleaser build or upload leaves an incomplete published release until rerun or manually fixed.

🔵 Suggestion — Consider making the upload target explicit in logs. The upload command at [.github/workflows/release-publish.yml](/Users/mlamp/.openclaw/tmp/pr-review-i23kkd2l/repo/.github/workflows/release-publish.yml:70) is safe with `"$TAG"` and `GH_TOKEN`; adding a short `gh release view "$TAG"` or echo before upload could make release troubleshooting easier.

✅ Highlights

The job uses scoped `contents: write` permissions, checks out the requested release tag, builds through the existing GoReleaser config, skips publish/announce correctly for artifact-only generation, and uses `--clobber` so reruns can repair assets.

Validation performed: `go test ./...` passed locally. `actionlint` was not available in the environment, so I did not run a workflow linter.

yujiawei (approved):
# Code Review — PR #24 (octo-cli)

## Summary

Re-review of `ff0533e` after the previous round flagged `${{ inputs.tag }}` interpolated directly into a `gh release upload` shell command. The new commit:

1. Adds a `Validate tag before upload` step that regex-checks `$TAG` (env-var indirection) and fails fast on a bad tag.
2. Switches the `gh release upload` step from inline `${{ inputs.tag }}` to `"$TAG"` sourced via `env:`.

Both edits address the prior 🔴 blocking finding correctly. The 🟡 publish-before-artifacts concern was addressed by the author with an explicit tradeoff rationale (manual `workflow_dispatch` flow with `--clobber` re-run path) — acceptable for this design.

## Verdict: APPROVED

No P0/P1 blockers remain. A few minor observations below, none gating merge.

## What the fix does correctly

- **Env-var indirection cuts the injection vector.** `${{ inputs.tag }}` is now expanded into the YAML `env:` mapping, not into the shell command string. Bash sees `"$TAG"` as a literal variable reference; shell metacharacters in the value cannot escape the quoted argument. This matches the convention already used by the reusable `publish` workflow upstream.
- **Validation runs immediately before the only shell-executed use of the tag.** The upload step is the sole place `$TAG` reaches a shell command line. Placing validation right before it (rather than at job start) is sufficient for the injection threat model — the other tag consumers (`actions/checkout`'s `ref:`, `goreleaser-action`'s args, GoReleaser reading the git tag) are not shell-evaluated.
- **Regex is restrictive enough.** `^v[0-9]+\.[0-9]+(\.[0-9]+)?([-.][0-9A-Za-z.-]+)?$` admits only digits, dots, hyphens, and ASCII letters — no shell metacharacters, no spaces, no quotes. Even a near-miss (a value that bypassed upstream validation but still matched this regex) cannot break out of the quoted `"$TAG"`.
- **Failure message is safe.** `echo "Invalid tag: $TAG"; exit 1` prints the value as a literal argument to `echo`; bash does not re-evaluate substitutions performed at variable expansion. No re-injection risk in the error path.
- **Idempotency / ordering / permissions are unchanged and still sound.** `needs: publish`, `if: ${{ inputs.draft == false }}`, `--skip=publish,announce`, `--clobber`, and the minimal `contents: write` permission all hold up.

## Suggestions (non-blocking — P2 / nit)

### 1. Patch version is now optional in the regex

Jerry-Xin's suggested pattern required `\.[0-9]+` for the patch component; this PR makes it optional via `(\.[0-9]+)?`. That admits two-segment tags like `v1.4`. The PR description and the existing release history all use three-segment tags (`v0.5.0`, `v1.4.0`), so two-segment matches are unlikely in practice and are still injection-safe. If the intent is strict semver, tightening to `^v[0-9]+\.[0-9]+\.[0-9]+([-.][0-9A-Za-z.-]+)?$` would match the project's actual tagging convention more precisely. Either form is acceptable from a security standpoint.

### 2. Validation could move earlier as defense-in-depth

The current order is Checkout → Setup Go → Build → List → Validate → Upload. Moving the validate step to the very top of the job (before Checkout) would also short-circuit GoReleaser cost on a bad input. Not necessary for the security property — Checkout's `ref:` parameter is not a shell vector — but a small efficiency / fail-fast win.

### 3. `concurrency` group keyed on the tag

Two concurrent manual dispatches on the same tag could race the upload step. `--clobber` mostly saves you, but you'd double the build cost and have a small window where partial uploads land. A single line prevents it:

```yaml
concurrency:
  group: release-publish-${{ inputs.tag }}
  cancel-in-progress: false
```

`cancel-in-progress: false` is intentional so a second dispatch queues instead of cancelling an in-flight upload.

### 4. The `List dist artifacts` step is debug-only

Useful while bringing the workflow up; once green, it's noise. Optional to drop.

### 5. Action pinning to commit SHAs

`actions/checkout@v4`, `actions/setup-go@v5`, `goreleaser/goreleaser-action@v6` follow the rest of the repo's convention, so this is purely a future hardening note rather than a request: pinning to immutable SHAs (with Renovate/Dependabot) closes a tag-mutation supply-chain vector. Not a change I'd block this PR on.

## What I verified but found no issue with

- **`needs: publish` short-circuits on `publish` failure** → no orphan binaries against a non-existent release.
- **`inputs.draft` default** — boolean default `false` means an omitted input still runs `build-artifacts`, matching documented intent.
- **Glob expansion** — `dist/*.tar.gz dist/*.zip dist/checksums.txt` expands in bash before `gh release upload` sees argv. The `.goreleaser.yaml` archive matrix (`tar.gz` for linux/darwin, `zip` for windows) plus `checksum.name_template: checksums.txt` guarantees all three globs match at least one file. No "no matches" failure.
- **GoReleaser config alignment** — `version: 2` config + `~> v2` action pin, `goos`/`goarch` linux/darwin/windows × amd64/arm64, `fetch-depth: 0` lets GoReleaser resolve `{{.Version}}` / `{{.Commit}}` / `{{.Date}}` from the tag. All correct.
- **Token scope** — `GITHUB_TOKEN` with `contents: write` is sufficient for `gh release upload` against a release in the same repo.
- **Old `release.yml` (push-tag trigger)** — out of scope here; the PR body correctly notes it is effectively dead under the current tag-protection ruleset. A follow-up deletion would be reasonable but is not required for this PR.

PR #37, opened 2026-06-04
ci: add concurrency group to CI workflow

lml2468:
## Summary

Add concurrency group to CI workflow to cancel redundant runs when multiple pushes happen on the same PR.

## Changes

Added the org-standard concurrency block between `permissions` (or `on`) and `jobs`:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

## Why

Without concurrency control, rapid successive pushes to the same PR trigger parallel CI runs that all run to completion, wasting CI minutes. This change cancels in-progress runs when a newer push arrives on the same PR, while preserving independent runs for different PRs and branches.

## Validation

- [x] YAML syntax verified (`yaml.safe_load`)
- [x] actionlint passed with no errors
- [x] Follows the same pattern used by octo-server, octo-matter, octo-web, octo-speech, and other org repos

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     name: build & test


Jerry-Xin (approved):
The PR is in scope for `Mininglamp-OSS/octo-cli` and safely adds CI concurrency control to the project’s main validation workflow.

💬 Non-blocking
No non-blocking issues found.

✅ Highlights
🔵 Suggestion / Positive: The concurrency group in `.github/workflows/ci.yml:12`-`.github/workflows/ci.yml:14` correctly keys PR runs by `github.event.pull_request.number` and push runs by `github.ref`, so separate PRs and branches do not cancel each other.

🔵 Suggestion / Positive: `cancel-in-progress: true` is appropriate here because this workflow only runs formatting, vetting, linting, tests, coverage summary, and build steps; it does not publish artifacts or perform deployment side effects.

Validation note: I verified the workflow parses as YAML locally. `actionlint` was not installed in this environment, so I could not independently rerun that check.

yujiawei (approved):
# Code Review — PR #37 (octo-cli)

## Verdict: APPROVED

A clean, minimal, single-file change that adds a standard concurrency block to the CI workflow. No correctness, security, or build-breaking issues found.

## What the change does

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

Placed between `permissions` and `jobs` in `.github/workflows/ci.yml`.

## Correctness review

✅ **Group key is well-formed.** The workflow triggers on `push` (branches: `main`) and `pull_request` (branches: `main`).
- On a `pull_request` event, `github.event.pull_request.number` is populated → group resolves to `CI-<PR#>`, so successive pushes to the same PR cancel the older run.
- On a `push` to `main`, `pull_request.number` is empty → it falls back to `github.ref` (`refs/heads/main`) → group `CI-refs/heads/main`.

✅ **No cross-PR / cross-branch interference.** Distinct PRs get distinct group keys (different numbers), and main pushes are isolated under `github.ref`. In-progress runs are only cancelled within the same logical stream, which is the intended behavior.

✅ **No double-trigger overlap.** `push` is scoped to `main` only, so feature-branch pushes fire just the `pull_request` event — there is no situation where a single commit creates two competing runs in the same group.

✅ **Placement is valid.** `concurrency` is a top-level workflow key; placing it after `permissions` and before `jobs` is syntactically correct and matches the pattern used across the other org repos.

✅ **Indentation / YAML structure** are correct (verified against the file at the head SHA).

## Non-blocking suggestion (nit, optional)

`cancel-in-progress: true` also applies to `push` events on `main`. If two commits merge to `main` in quick succession, the earlier run is cancelled, so you lose a completed CI signal for that intermediate commit. For PR CI this is exactly what you want; for the protected `main` branch some teams prefer to keep every post-merge run. If full per-commit CI history on `main` is ever desired, the group could be conditioned to only cancel on PR events. This is a minor preference and not a reason to block — the current form is the documented org standard and acceptable as-is.

## Conclusion

The change does exactly what it claims, follows the established org pattern, and carries no functional risk. Approving.

PR #42, opened 2026-06-04
fix: skip OSV-Scanner SARIF upload for Dependabot PRs

lml2468:
## What
Add Dependabot exclusion to the OSV-Scanner SARIF upload condition.

## Why
Dependabot-triggered `pull_request` workflows get a read-only `GITHUB_TOKEN`. The existing fork guard (`head.repo.full_name == github.repository`) does not catch Dependabot PRs because they are same-repo branches. This causes `security-events:write` SARIF upload to fail on every Dependabot PR.

## Fix
```yaml
upload-sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.event.pull_request.user.login != 'dependabot[bot]' }}
```

Scan still runs for Dependabot PRs; results visible in workflow output. Only SARIF upload (which would fail anyway) is skipped.

Follow-up to the merged OSV-Scanner PR.

### DIFF
--- a/.github/workflows/osv-scanner.yml
+++ b/.github/workflows/osv-scanner.yml
@@ -30,7 +30,8 @@ jobs:
       scan-args: |-
         -r
         ./
-      # Fork PRs have read-only GITHUB_TOKEN; skip SARIF upload to avoid
-      # security-events:write permission failure. Scan still runs and
-      # results are visible in the workflow output.
-      upload-sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
+      # Skip SARIF upload when GITHUB_TOKEN is read-only:
+      # - Fork PRs: token is inherently read-only
+      # - Dependabot PRs: GitHub enforces read-only token for dependabot[bot]
+      # Scan still runs; results visible in workflow output.
+      upload-sarif: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.event.pull_request.user.login != 'dependabot[bot]' }}


Jerry-Xin (approved):
Summary: The PR is in scope for `Mininglamp-OSS/octo-cli` and correctly narrows SARIF upload for PR scans where the token can write security events.

✅ Highlights

🔵 Suggestion / Positive: `.github/workflows/osv-scanner.yml:37` keeps OSV scanning active for Dependabot PRs while disabling only SARIF upload, which matches the stated failure mode and avoids reducing vulnerability visibility in workflow logs.

🔵 Suggestion / Positive: `.github/workflows/osv-scanner.yml:33` documents both read-only token cases clearly: fork PRs and Dependabot same-repo PRs.

No blocking issues found. The change is small, targeted, and consistent with the existing workflow design. No additional test coverage is needed for this workflow-only condition change.

yujiawei (approved):
# Code Review — PR #42 (octo-cli)

## Verdict: APPROVED

A focused, correct CI fix. Dependabot-triggered `pull_request` workflows run with a read-only `GITHUB_TOKEN`, so the `upload-sarif` step's `security-events: write` requirement fails for every Dependabot PR. The existing fork guard (`head.repo.full_name == github.repository`) does not catch this case because Dependabot PRs are same-repo branches. Adding `&& github.event.pull_request.user.login != 'dependabot[bot]'` is the right, minimal fix.

## Verification

- **Correct condition semantics** — `upload-sarif` is declared `type: boolean` in `osv-scanner-reusable-pr.yml`, and the reusable workflow gates upload on `inputs.upload-sarif == true`. The `${{ ... }}` expression evaluates to a boolean that GitHub coerces correctly. This is the same coercion pattern already used by the original fork guard, so the approach is proven. ✅
- **Correct actor identifier** — `dependabot[bot]` is the canonical `pull_request.user.login` for Dependabot-authored PRs. GitHub expression string comparison is case-insensitive, so no casing pitfall. ✅
- **No security regression** — the change is strictly *more* restrictive: it only ever skips the SARIF upload, never enables anything new. The scan itself still runs and `fail-on-vuln` still gates the PR check, so vulnerability detection/blocking is preserved. ✅
- **No injection surface** — `head.repo.full_name` and `user.login` are used only inside a `${{ }}` comparison expression, not interpolated into a shell `run:` step, so there is no script-injection risk from attacker-controlled fields. ✅
- **Scope is contained** — only the `scan-pr` job (the `pull_request` path) is touched; `scan-scheduled` (push/schedule/dispatch) is unaffected. ✅
- **Comment accuracy** — the updated inline comment correctly documents both read-only-token cases (fork + Dependabot). ✅

## Notes (non-blocking)

- **Visibility trade-off** — when upload is skipped, OSV results for Dependabot PRs no longer surface in the Security → Code Scanning tab; they are only in the workflow logs. This is the intended/unavoidable behavior given the read-only token and matches the existing fork-PR handling, so it is acceptable. Worth a one-time mention to maintainers so nobody is surprised that Dependabot PRs lack code-scanning alerts.
- **Optional hardening** — `github.actor` / `github.triggering_actor` could also identify Dependabot, but `pull_request.user.login` is the most precise signal for "who authored this PR" and is the correct choice here. No change needed.

No P0/P1 issues. The fix is sound and ready to merge.

PR #44, opened 2026-06-04
ci: set fail-on-vuln false on scheduled OSV scan

lml2468:
## What

Add `fail-on-vuln: false` to the `scan-scheduled` OSV job.

## Why

Without this, every push to main and every weekly scheduled run fails
on pre-existing vulnerabilities — creating permanent red CI unrelated
to the change being pushed.

With `fail-on-vuln: false`, results are still uploaded to the Security
tab (SARIF) for tracking, but don't block the workflow. PR scans still
block new vulnerabilities (default `fail-on-vuln: true`).

### DIFF
--- a/.github/workflows/osv-scanner.yml
+++ b/.github/workflows/osv-scanner.yml
@@ -22,6 +22,7 @@ jobs:
       scan-args: |-
         -r
         ./
+      fail-on-vuln: false
 
   scan-pr:
     if: github.event_name == 'pull_request'


Jerry-Xin (approved):
The PR is in scope for `Mininglamp-OSS/octo-cli` and safely updates the repository’s OSV CI behavior without weakening PR-time vulnerability gating.

🔴 Blocking  
None.

💬 Non-blocking  
None.

✅ Highlights  
🔵 Suggestion-free: [.github/workflows/osv-scanner.yml:25](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-a_rbxgcn/repo/.github/workflows/osv-scanner.yml:25) applies `fail-on-vuln: false` only to the `scan-scheduled` job used for `push`, `schedule`, and manual runs. The PR-specific job remains separate at [.github/workflows/osv-scanner.yml:27](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-a_rbxgcn/repo/.github/workflows/osv-scanner.yml:27), preserving blocking behavior for pull requests.

No tests are needed for this workflow-only configuration change.

yujiawei (approved):
# Code Review — PR #44 (octo-cli)

## Summary

This PR adds a single line — `fail-on-vuln: false` — to the `scan-scheduled` job in `.github/workflows/osv-scanner.yml`, so that OSV scans triggered by `push` to `main`, the weekly `schedule`, and `workflow_dispatch` no longer fail CI on pre-existing vulnerabilities. **Approve.**

## Verification

I verified the change against the pinned reusable workflow (`google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@9a49870`, v2.3.8):

- ✅ **Input is valid.** `fail-on-vuln` is a declared `boolean` input on the reusable workflow (default `true`). It is wired straight through to the reporter step as `--fail-on-vuln=${{ inputs.fail-on-vuln }}`, so it does control the failure behavior — no silent no-op from a misspelled key.
- ✅ **PR claim is accurate — SARIF still uploads.** The "Upload to code-scanning" step is gated only on `inputs.upload-sarif == true` (default `true`), which is independent of `fail-on-vuln`. So scheduled/push results continue to flow into the Security tab for tracking, while no longer blocking the run. The PR description matches the code.
- ✅ **New-vuln gate on PRs is preserved.** Only `scan-scheduled` is changed. The `scan-pr` job still uses the PR reusable workflow with its default `fail-on-vuln: true`, so PRs that introduce new vulnerabilities continue to be blocked. The intended protection boundary (block on *new* vulns at PR time; don't block on *pre-existing* vulns on main/scheduled) is correctly placed.

## Security note (for the human reviewer)

This is an intentional, reasonable softening of CI enforcement, not a regression in detection:

- Detection and reporting are unchanged — vulnerabilities are still scanned and surfaced in the Security tab. Only the *gating* on main/scheduled runs is relaxed.
- The trade-off worth being aware of: pre-existing vulnerabilities on `main` will no longer turn CI red, so there is no longer an automatic forcing function to remediate them. Visibility now depends on someone watching the Security tab / code-scanning alerts. Consider pairing this with a Dependabot/alert routing process so known vulns don't sit silently. This is a process consideration, not a blocker for this PR.

## Verdict

No P0/P1 issues. The change is minimal, correctly scoped, and behaves exactly as described.

lml2468 (commented):
**Security review: LGTM** ✅

_(Same GitHub account, comment-based sign-off.)_

1-line change: `fail-on-vuln: false` on scan-scheduled only. Correct — prevents main branch permanent red from pre-existing advisories in transitive deps. PR scan keeps default `fail-on-vuln: true`, so new vulns still block.

— 王大锤 🔨

PR #46, opened 2026-06-04
ci: replace paths-ignore with dorny/paths-filter preflight job

### LINKED ISSUE #50: ci: docs-only PRs deadlock on required status checks (paths-ignore creates no check run)
## Problem

Workflow-level `paths-ignore` on the `pull_request` trigger of `.github/workflows/ci.yml` causes GitHub to create **no** check runs for docs-only PRs. Because `build & test (1.24.x)` is a **required** status check (ruleset 16222449, "main branch protection"), a docs-only PR that produces zero check runs can never satisfy the required gate — it is permanently un-mergeable.

## Fix

Replace workflow-level `paths-ignore` on `pull_request` with a `changes` preflight job using `dorny/paths-filter`:

- Remove `paths-ignore` from the `pull_request` trigger (kept on `push`).
- Add a `changes` preflight job that classifies the PR as code vs docs-only.
- Gate `build` **per-step** (so the required matrix child check `build & test (1.24.x)` is always emitted — success on docs-only PRs, real run on code PRs).
- Gate `npm-test` at the job level (it is not a required check).
- Add draft-PR handling (`converted_to_draft` type + draft `if` guard).

Follows the same pattern as octo-server's CI workflow.

Refs: Mininglamp-OSS workflow optimization T13

## Note for maintainers

This issue tracks the work in #46. To clear the `check-sprint` board-health gate, please add this issue to the Octo Board (https://github.com/orgs/Mininglamp-OSS/projects/2) and assign it to the current Sprint.

lml2468:
## What

Replace workflow-level `paths-ignore` on `pull_request` trigger with a `changes` preflight job using `dorny/paths-filter`.

## Why

When `paths-ignore` is set at the workflow level, GitHub will not create any check run for docs-only PRs. If CI is a required status check, those PRs can never merge. The preflight job produces a "skipped" status instead, which GitHub counts as passing.

## Changes

- Removed `paths-ignore` from `pull_request` trigger (kept on `push`)
- Added `changes` preflight job with `dorny/paths-filter`
- Added `needs: [changes]` condition to build/test jobs
- Added draft PR skip (`converted_to_draft` type + draft condition)

## Reference

Follows the same pattern as octo-server's CI workflow.

Refs: Mininglamp-OSS workflow optimization T13

Closes #50

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,8 +57,11 @@ jobs:
           #   dotfiles:     all negations=true → every=true → CI runs ✅
           #   Dockerfile:   all negations=true → CI runs ✅
           #
-          # Do NOT add '**': picomatch('**') doesn't match dotfiles (e.g. .golangci.yml)
-          # causing them to incorrectly skip CI with the 'every' quantifier.
+          # Do NOT add '**' as an inclusion pattern: with the 'every' quantifier
+          # the bare '**' pattern adds no useful constraint and risks regressing
+          # the negation logic on future picomatch upgrades. Stick to negation-only.
+          # (dorny/paths-filter v3 already sets picomatch dot:true so dotfiles like
+          # .golangci.yml ARE matched by '**' under v3 — earlier versions did not.)
           # Do NOT use 'some' + negation: every changed file matches at least one
           # negation pattern (e.g. main.go matches !**/*.md), so some=true always
           # and the skip logic becomes a no-op.
@@ -77,16 +80,12 @@ jobs:
       needs.changes.outputs.code == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    strategy:
-      fail-fast: false
-      matrix:
-        go-version: ["1.24.x"]
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
 
       - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5.6.0
         with:
-          go-version: ${{ matrix.go-version }}
+          go-version: "1.24.x"
           check-latest: true
 
       - name: gofmt
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,18 @@ name: CI
 on:
   push:
     branches: [main]
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+      - '.github/ISSUE_TEMPLATE/**'
   pull_request:
+    types: [opened, reopened, synchronize, ready_for_review, converted_to_draft]
     branches: [main]
+    # paths-ignore is intentionally NOT set here — workflow-level path filtering
+    # prevents GitHub from creating any check run, which deadlocks required-status
+    # checks on docs-only PRs. Path-based skipping is handled by the `changes`
+    # preflight job below instead (dorny/paths-filter produces "skipped" statuses
+    # that GitHub counts as passing for required checks).
 
 permissions:
   contents: read
@@ -14,8 +24,37 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  changes:
+    name: Detect changed paths
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        if: github.event_name == 'pull_request'
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@6852f92c20ea7fd3b0c25de3b5112db3a98da050 # v3
+        id: filter
+        if: github.event_name == 'pull_request'
+        with:
+          predicate-quantifier: 'every'
+          filters: |
+            code:
+              - '!**/*.md'
+              - '!docs/**'
+              - '!.github/ISSUE_TEMPLATE/**'
+
   build:
     name: build & test
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 30
     strategy:
@@ -66,6 +105,10 @@ jobs:
 
   npm-test:
     name: npm test
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 30
     defaults:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,6 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
-          predicate-quantifier: 'every'
           filters: |
             code:
               - '!**/*.md'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,10 +39,9 @@ jobs:
       # are docs/templates. Push events default to 'true'.
       code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
     steps:
-      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
-        if: github.event_name == 'pull_request'
-        with:
-          persist-credentials: false
+      # No checkout: on pull_request events dorny/paths-filter lists changed
+      # files via the GitHub API (needs only pull-requests: read, granted
+      # above), so a working tree is unnecessary here.
       - uses: dorny/paths-filter@6852f92c20ea7fd3b0c25de3b5112db3a98da050 # v3
         id: filter
         if: github.event_name == 'pull_request'
@@ -154,6 +153,13 @@ jobs:
   npm-test:
     name: npm test
     needs: [changes]
+    # Job-level skip is intentional here and asymmetric with `build` above.
+    # `npm test` is NOT a required status check (ruleset 16222449 pins only
+    # "build & test (1.24.x)" and "code-review"), so skipping the whole job on
+    # docs-only PRs is safe — there is no required check to keep alive. `build`
+    # must use per-step gating instead because skipping it job-level would drop
+    # its required matrix child check and re-create the deadlock. Do not "align"
+    # these two jobs onto the same strategy.
     if: |
       (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
       needs.changes.outputs.code == 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,11 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
+      # code=true when at least one changed file matches the code filter.
+      # predicate-quantifier:every + negation patterns (no bare '**'):
+      # a file matches 'code' iff ALL negation patterns are satisfied
+      # (file is not a doc/template). code=false only when all changed files
+      # are docs/templates. Push events default to 'true'.
       code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
@@ -42,6 +47,22 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
+          # predicate-quantifier: every — a file matches 'code' only when ALL
+          # negation patterns are satisfied (file is NOT a doc/template file).
+          #
+          # Validated by picomatch v4 unit test (same library used internally):
+          #   README.md:    !**/*.md=false → every=false → not match → code=false ✅
+          #   main file:    all negations=true → every=true → match → code=true ✅
+          #   mixed PR:     code file matches → code=true (some over changed files) ✅
+          #   dotfiles:     all negations=true → every=true → CI runs ✅
+          #   Dockerfile:   all negations=true → CI runs ✅
+          #
+          # Do NOT add '**': picomatch('**') doesn't match dotfiles (e.g. .golangci.yml)
+          # causing them to incorrectly skip CI with the 'every' quantifier.
+          # Do NOT use 'some' + negation: every changed file matches at least one
+          # negation pattern (e.g. main.go matches !**/*.md), so some=true always
+          # and the skip logic becomes a no-op.
+          predicate-quantifier: 'every'
           filters: |
             code:
               - '!**/*.md'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,20 +75,43 @@ jobs:
   build:
     name: build & test
     needs: [changes]
-    if: |
-      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
-      needs.changes.outputs.code == 'true'
+    # Draft guard stays at the job level. The code-vs-docs skip is enforced
+    # per-step (below) instead, so the matrix child check
+    # `build & test (1.24.x)` is always emitted (success on docs-only PRs,
+    # real run on code PRs). Job-level `if: code == 'true'` would skip the
+    # whole job BEFORE matrix expansion, so the parenthesised child check
+    # required by the repository ruleset (id 16222449) would never be
+    # created — recreating the very deadlock this workflow is meant to fix,
+    # just on docs-only PRs instead of all PRs.
+    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        # Single-value matrix intentionally retained. The repository ruleset
+        # (id 16222449, "main branch protection") pins the exact check name
+        # "build & test (1.24.x)" — the parenthesised form GitHub emits only
+        # for matrix jobs. Dropping the matrix renames the check to
+        # "build & test" and breaks the required-status gate on every PR.
+        # Add new Go versions here instead of removing the matrix.
+        go-version: ["1.24.x"]
     steps:
+      - name: skip notice (docs-only PR)
+        if: needs.changes.outputs.code != 'true'
+        run: echo "docs-only PR — skipping build & test (matrix=${{ matrix.go-version }})"
+
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        if: needs.changes.outputs.code == 'true'
 
       - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5.6.0
+        if: needs.changes.outputs.code == 'true'
         with:
-          go-version: "1.24.x"
+          go-version: ${{ matrix.go-version }}
           check-latest: true
 
       - name: gofmt
+        if: needs.changes.outputs.code == 'true'
         run: |
           out=$(gofmt -l . | grep -v '^vendor/' || true)
           if [ -n "$out" ]; then
@@ -98,9 +121,11 @@ jobs:
           fi
 
       - name: go vet
+        if: needs.changes.outputs.code == 'true'
         run: go vet ./...
 
       - name: go mod tidy check
+        if: needs.changes.outputs.code == 'true'
         run: |
           go mod tidy
           if ! git diff --exit-code go.mod go.sum; then
@@ -109,17 +134,21 @@ jobs:
           fi
 
       - name: golangci-lint
+        if: needs.changes.outputs.code == 'true'
         uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
         with:
           version: v2.12.2
 
       - name: test (race, coverage)
+        if: needs.changes.outputs.code == 'true'
         run: go test -race -shuffle=on -count=1 -covermode=atomic -coverprofile=coverage.out ./...
 
       - name: coverage summary
+        if: needs.changes.outputs.code == 'true'
         run: go tool cover -func=coverage.out | tail -n 20
 
       - name: build
+        if: needs.changes.outputs.code == 'true'
         run: go build -o octo-cli ./cmd/octo-cli
 
   npm-test:


Jerry-Xin (dismissed):
The PR is in scope for `Mininglamp-OSS/octo-cli` and the CI workflow change is ready to merge.

💬 Non-blocking
- 🟡 Warning: If branch protection currently requires only the old job contexts (`build & test`, `npm test`), consider also requiring `Detect changed paths` or adding an aggregate required CI gate. If the preflight job fails, downstream jobs can be skipped, and skipped job conclusions may not expose the preflight failure through the old required contexts. See `.github/workflows/ci.yml:27` and `.github/workflows/ci.yml:54`.
- 🔵 Suggestion: The checkout in the `changes` job is likely unnecessary for PR events because `dorny/paths-filter` can use the PR API with `pull-requests: read`; keeping it is harmless but adds a small amount of work. See `.github/workflows/ci.yml:37`.

✅ Highlights
- The PR correctly removes workflow-level `paths-ignore` from `pull_request`, avoiding missing check runs for docs-only PRs. See `.github/workflows/ci.yml:10`.
- The `predicate-quantifier: 'every'` setup with negated patterns matches the documented `dorny/paths-filter` approach for excluding file types while still detecting meaningful changes. See `.github/workflows/ci.yml:45`. ([github.com](https://github.com/dorny/paths-filter))
- The added draft handling is consistent and keeps build jobs skipped for draft PRs while rerunning on `ready_for_review`. See `.github/workflows/ci.yml:11` and `.github/workflows/ci.yml:55`.

Validation performed: reviewed the full workflow context and ran `git diff --check` successfully. `actionlint` was not installed in this environment.

yujiawei (dismissed):
# Code Review — PR #46 (octo-cli)

## Verdict: APPROVED

This cleanly fixes a real CI deadlock: workflow-level `paths-ignore` on the `pull_request` trigger causes GitHub to create **no** check runs for docs-only PRs, which permanently blocks them when CI is a required status check. Replacing it with a `changes` preflight job + per-job gating is the correct, well-established pattern. No blocking (P0/P1) issues found.

## Correctness — verified ✅

- **`changes.outputs.code` expression** — `${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}` (`.github/workflows/ci.yml:33`). Logic is sound across all paths:
  - `push` → `event_name == 'pull_request'` is false → output `'true'`, so push always runs code jobs (filter step is correctly `if`-guarded off for push).
  - `pull_request` with code changes → `'true'`; docs-only → `'false'`.
  - Fails **open**: if `steps.filter.outputs.code` were ever empty (step skipped/errored), the `&&` collapses to falsy and the `|| 'true'` kicks in → CI runs rather than silently skipping. Good safety default.
- **Downstream gate** — `needs.changes.outputs.code == 'true'` (`ci.yml:50`, `ci.yml:109`) is a correct string comparison against the string output.
- **dorny negation semantics** — `predicate-quantifier: 'every'` with the three `!`-prefixed filters (`ci.yml:42-47`) is the documented idiom: a changed file counts as `code` only if it is not `.md`, not under `docs/`, and not under `.github/ISSUE_TEMPLATE/`. Docs-only diffs yield `code='false'` → build/npm-test skip → reported as "skipped", which GitHub counts as passing for required checks. Exactly the intended behavior. Note this idiom **requires v3**, which the pin matches.
- **Draft handling** — adding `converted_to_draft`/`ready_for_review` to `types` plus `(github.event_name != 'pull_request' || !github.event.pull_request.draft)` (`ci.yml:49`, `ci.yml:108`) correctly skips the heavy jobs while a PR is a draft and re-runs them on ready-for-review.

## Security — verified ✅

- **Action pins are authentic and immutable (SHA-pinned):**
  - `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5` → tag `v4.3.1` / `v4` ✅
  - `dorny/paths-filter@6852f92c20ea7fd3b0c25de3b5112db3a98da050` → tag `v3` ✅
- `persist-credentials: false` on the preflight checkout (`ci.yml:38`) is good hygiene.
- Permissions are correctly scoped (`contents: read`, `pull-requests: read`) on the `changes` job (`ci.yml:30-31`).

## Non-blocking notes (P2 / nit)

1. **`changes` job has no draft guard**, so it spins up a runner even for draft PRs and for pure-docs drafts. It's cheap (~seconds, `timeout-minutes: 5`), but you could add the same draft `if` to avoid the runner entirely on drafts. Purely optional.
2. **paths-filter does not strictly need the checkout for PR events** — on `pull_request` it can resolve changed files via the API. The explicit `actions/checkout` works fine (git-based diff) and is arguably more robust; just noting the checkout adds a little time. No change required.
3. **Skipped required checks on drafts**: if `build & test` / `npm test` are configured as required status checks, draft PRs will report them as skipped (= passing). That's expected and harmless since drafts aren't merged, but worth keeping in mind when reasoning about branch protection.

Nice work — small, focused, correctly pinned, and it follows the octo-server precedent.

Jerry-Xin (dismissed):
The PR is relevant to `octo-cli`, but the new path filter does not implement the intended docs-only skip behavior.

🔴 Blocking

🔴 Critical — [`.github/workflows/ci.yml`](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-hrh5jffa/repo/.github/workflows/ci.yml:45): `dorny/paths-filter` defaults to `predicate-quantifier: some`, so this negative-only filter will still match ignored paths. For example, `README.md` fails `!**/*.md` but matches `!docs/**` and `!.github/ISSUE_TEMPLATE/**`; `docs/foo.md` similarly matches the issue-template negation. That makes `steps.filter.outputs.code` effectively `true` for docs-only and issue-template-only PRs, so `build` and `npm-test` will run instead of producing the intended skipped required checks. Add `predicate-quantifier: every` to the `dorny/paths-filter` step, or use an equivalent positive include plus exclusions pattern that actually returns false for ignored-only changes.

✅ Highlights

The PR correctly removes workflow-level `paths-ignore` from `pull_request`, which addresses the required-check deadlock class where GitHub creates no CI check run at all.

The job-level permissions for the preflight job are appropriately narrow: `contents: read` and `pull-requests: read`.

yujiawei (dismissed):
# Code Review — PR #46 (octo-cli)

## Summary

This PR removes workflow-level `paths-ignore` from the `pull_request` trigger and introduces a `changes` preflight job (`dorny/paths-filter`) so docs-only PRs produce a passing/skipped status instead of deadlocking required checks. The deadlock fix itself is sound. However, the preflight filter is misconfigured and the path-skipping optimization it adds is currently a **no-op** — it never skips CI for docs-only PRs.

**Verdict: CHANGES_REQUESTED**

## P1 — `code` filter is missing `predicate-quantifier: 'every'`; docs-only skip never triggers

`.github/workflows/ci.yml`, `changes` job:

```yaml
- uses: dorny/paths-filter@6852f92c20ea7fd3b0c25de3b5112db3a98da050 # v3
  id: filter
  if: github.event_name == 'pull_request'
  with:
    filters: |
      code:
        - '!**/*.md'
        - '!docs/**'
        - '!.github/ISSUE_TEMPLATE/**'
```

The filter list contains **only negation patterns** and relies on the action's default `predicate-quantifier: 'some'` (OR). Under `some`, a file matches `code` if it satisfies *at least one* pattern. A negation pattern is "satisfied" when the file is *not* in that path, so any single file almost always satisfies one of the three negations. Concretely, for a docs-only PR:

- changing `docs/foo.md` → `!**/*.md`=false, `!docs/**`=false, `!.github/ISSUE_TEMPLATE/**`=**true** → `some` ⇒ match ⇒ `code=true`
- changing `README.md` → `!docs/**`=**true** → match ⇒ `code=true`
- changing `.github/ISSUE_TEMPLATE/bug.md` → `!docs/**`=**true** → match ⇒ `code=true`

So `steps.filter.outputs.code` evaluates to `true` for **every** docs-only change. The `build`/`npm-test` jobs run full CI regardless, and the path-skipping feature this PR adds never fires. This is the documented dorny/paths-filter gotcha: negation patterns under `some` do not exclude.

This directly contradicts the PR's stated goal and diverges from the octo-server reference the PR says it follows. octo-server's `ci.yml` uses the **identical** negation filters but with `predicate-quantifier: 'every'`, and even carries an inline warning: *"Do NOT use 'some' + negation."* That quantifier line was dropped here.

### Fix (one line, matches the cited reference)

```yaml
  with:
    predicate-quantifier: 'every'
    filters: |
      code:
        - '!**/*.md'
        - '!docs/**'
        - '!.github/ISSUE_TEMPLATE/**'
```

With `every`, a file matches `code` only when it satisfies *all* negations (i.e. it is not a doc/template). `code=false` then occurs exactly when every changed file is a doc/template, correctly skipping `build`/`npm-test`. Note (per the reference): do **not** add a bare `**` inclusion pattern alongside `every` — picomatch's `**` does not match dotfiles (e.g. `.golangci.yml`), which would cause such changes to wrongly skip CI.

## P2 / nits (non-blocking)

- The deadlock fix (removing `paths-ignore` from `pull_request`, so check runs are always created) is correct and works independently of the filter bug above — required checks no longer deadlock on docs-only PRs. The defect is scoped to the skip optimization, not merge safety.
- The `changes` job has no draft guard, so it runs its (cheap, 5-min) detection step even on draft PRs while `build`/`npm-test` correctly skip via the draft `if`. Harmless; leaving it as-is is fine.
- The `outputs.code` ternary (`github.event_name == 'pull_request' && steps.filter.outputs.code || 'true'`) correctly defaults push events to `true`; no issue there.

## Recommendation

Add `predicate-quantifier: 'every'` to the `dorny/paths-filter` step so the optimization actually skips CI for docs-only PRs, matching the octo-server reference. Once that's in, this is good to merge.

yujiawei (dismissed):
# Code Review — PR #46 (octo-cli)

## Verdict: APPROVED

This PR removes workflow-level `paths-ignore` from the `pull_request` trigger and adds a `changes` preflight job (`dorny/paths-filter`) so docs-only PRs report a skipped/passing status instead of deadlocking required checks. At the current head (`538e87e`) the previously-flagged filter bug is fixed and the change is correct, minimal, and well-pinned. No blocking (P0/P1) issues remain.

## Correctness — verified ✅

- **`predicate-quantifier: 'every'` is present** (`.github/workflows/ci.yml`, `changes` job). With only negation patterns (`!**/*.md`, `!docs/**`, `!.github/ISSUE_TEMPLATE/**`), a changed file matches `code` only when it satisfies *all* negations — i.e. it is not a doc/template. `paths-filter` then sets `code=true` iff at least one changed file is a non-doc, so a docs-only PR yields `code=false` and `build`/`npm-test` skip. This is the documented dorny idiom and matches the octo-server reference the PR cites. (Earlier revisions dropped this line, which made the skip a no-op under the default `some` quantifier; that defect is no longer present.)
- **No bare `**` inclusion** — correctly avoided. picomatch's `**` does not match dotfiles (e.g. `.golangci.yml`); combined with `every` it would wrongly skip CI on dotfile-only changes. The negation-only approach lets dotfiles satisfy all negations → `code=true` → CI runs. Correct.
- **`changes.outputs.code` expression** — `${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}`. Push events short-circuit to `'true'` (filter step is `if`-guarded off for push), and the `|| 'true'` fails *open*: if the filter step were ever skipped/empty, CI runs rather than silently skipping. Good safety default.
- **Downstream gates** — `needs.changes.outputs.code == 'true'` is a correct string comparison against the string output.
- **Draft handling** — adding `converted_to_draft`/`ready_for_review` to `types` plus `(github.event_name != 'pull_request' || !github.event.pull_request.draft)` correctly skips the heavy jobs on drafts and re-runs them on ready-for-review.

## Security — verified ✅

- Actions are SHA-pinned to immutable refs:
  - `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5` → `v4` ✅
  - `dorny/paths-filter@6852f92c20ea7fd3b0c25de3b5112db3a98da050` → `v3` ✅ (the `every`+negation idiom requires v3, which the pin matches)
- `persist-credentials: false` on the preflight checkout — good hygiene.
- Permissions scoped to `contents: read` + `pull-requests: read` on the `changes` job.

## Non-blocking notes (P2 / nit)

1. **`changes` job has no draft guard**, so it spins up a runner (cheap, `timeout-minutes: 5`) even on draft PRs while `build`/`npm-test` correctly skip. Optional: add the same draft `if` to skip the runner entirely on drafts.
2. **Skipped required checks on drafts/docs-only PRs** report as "skipped" (= passing). Expected and harmless, but worth keeping in mind if branch protection is later tightened — consider also requiring `Detect changed paths` or an aggregate context so a skipped `build & test` cannot bypass a genuinely-required gate.

Small, focused, correctly pinned, and it follows the octo-server precedent. Good to merge.

Jerry-Xin (dismissed):
Summary: This PR is in scope for `octo-cli`, but the new skip logic can still leave the existing Go CI required check missing on docs-only PRs.

🔴 Blocking
- 🔴 Critical — [.github/workflows/ci.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-rvm_93vc/repo/.github/workflows/ci.yml:74): `build` is a matrix job, and the repo’s current check name is `build & test (1.24.x)`. GitHub evaluates a job-level `if` before matrix expansion, so when `needs.changes.outputs.code == 'false'`, the matrix child check is not created. That can recreate the exact required-check deadlock this PR is meant to fix if branch protection requires `build & test (1.24.x)`. Preserve the required check name by moving the skip condition into the matrix job’s steps with a no-op success path, removing the one-value matrix, or adding a stable always-running wrapper check that branch protection requires.

💬 Non-blocking
- 🟡 Warning — [.github/workflows/ci.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-rvm_93vc/repo/.github/workflows/ci.yml:60): the comment says `picomatch('**')` does not match dotfiles, but `dorny/paths-filter` v3 sets picomatch `dot: true`. The implemented filter is still reasonable, but the comment is misleading and may cause future maintenance mistakes.

✅ Highlights
- Actions are pinned.
- The new `changes` job uses narrow permissions.
- Keeping `paths-ignore` off `pull_request` is the right direction for required-check compatibility.

yujiawei (dismissed):
# Code Review — PR #46 (octo-cli)

## Verdict: CHANGES_REQUESTED

The path-filtering goal is sound and the `dorny/paths-filter` preflight design is correct. However, this PR bundles an **unrelated, unflagged change that re-introduces the exact failure mode the PR claims to fix**: it deadlocks the required-status-check merge gate for every future PR to `main`.

---

## P0 — Removing the `go-version` matrix renames the required status check → merge gate deadlock

The branch ruleset `main branch protection` requires this status check:

```
build & test (1.24.x)
```

That name is **generated by the strategy matrix**: job `name: build & test` + matrix `go-version: ["1.24.x"]` → GitHub appends the matrix value → `build & test (1.24.x)`.

This PR deletes the matrix and hardcodes the version:

```diff
-    strategy:
-      fail-fast: false
-      matrix:
-        go-version: ["1.24.x"]
...
-          go-version: ${{ matrix.go-version }}
+          go-version: "1.24.x"
```

Without the matrix, the job's check run is now named simply **`build & test`** (no `(1.24.x)` suffix). The required check `build & test (1.24.x)` will therefore **never be produced** by this workflow again, so it stays permanently unsatisfied and no PR can merge.

This is observable on this PR itself:

- Required checks (ruleset `main branch protection`): `build & test (1.24.x)`, `code-review`
- Checks actually reported on this PR head: `build & test` (no suffix), `npm test`, `Detect changed paths`, … — `build & test (1.24.x)` is **absent**.

This is the same class of defect — a required check that can never be satisfied — that the PR description sets out to eliminate for docs-only PRs. Here it is reintroduced for *all* PRs via the rename.

### Required fix (pick one)

1. **Keep the `strategy.matrix` block** (recommended, minimal). It preserves the check name `build & test (1.24.x)` that the ruleset depends on, and is orthogonal to the paths-filter change. The matrix removal is unrelated to this PR's stated goal and should be dropped from this change.
2. If removing the matrix is genuinely intended, the ruleset's required-check context must be renamed from `build & test (1.24.x)` to `build & test` **in lockstep** (admin/ruleset change, coordinated so there is no window where merges are blocked). This is not visible in the PR and would need explicit coordination — option 1 is safer.

---

## What's correct (no action needed)

- **`paths-ignore` removal from `pull_request` + preflight job design.** The reasoning is accurate: workflow-level `paths-ignore` suppresses check-run creation entirely, deadlocking required checks on docs-only PRs. A `changes` gate job whose downstream jobs are skipped via `if` yields a "skipped" status that GitHub counts as passing — the right pattern.
- **`predicate-quantifier: every` + negation-only filters.** The logic is correct: a changed file matches `code` only when it satisfies *all* negation patterns (i.e. is not a doc/template); `code=false` only when every changed file is a doc/template; a mixed PR yields `code=true`. The inline rationale against bare `**` and against `some` + negation is accurate.
- **`code` output fallback** `github.event_name == 'pull_request' && steps.filter.outputs.code || 'true'` correctly defaults push events to `true` and avoids an empty output when the filter step is skipped on non-PR events.
- **SHA-pinned actions** with version comments — good supply-chain hygiene, consistent with the rest of the file.
- **Scoped `permissions` on the `changes` job** (`contents: read`, `pull-requests: read`) and `persist-credentials: false` on checkout — appropriate least privilege.
- **Draft handling** (`converted_to_draft` type + `!github.event.pull_request.draft` guard) is internally consistent: draft PRs skip the heavy jobs (reported as skipped→passing), and `ready_for_review` re-triggers a real run.

---

## P2 / nits (non-blocking)

- **Scope creep.** The matrix removal and the `go-version` inlining are unrelated to "replace paths-ignore with paths-filter." Even once the P0 is resolved, unrelated CI-shape changes are best split out so the required-check impact is reviewed deliberately rather than as a side effect.
- **Comment volume.** The `filters`/`outputs` blocks carry ~25 lines of explanatory comments. Useful, but consider trimming the picomatch unit-test walkthrough to a one-line link/summary to keep the workflow readable.

---

## Summary

Approve the paths-filter design; block on the matrix removal. Restore `strategy.matrix` (or coordinate the ruleset rename) so the required check `build & test (1.24.x)` continues to exist — otherwise merging this deadlocks the very gate it aims to unblock.

yujiawei (dismissed):
# Code Review — PR #46 (octo-cli)

## Verdict: APPROVED

This change correctly solves the required-status-check deadlock that workflow-level `paths-ignore` causes on docs-only PRs. The logic is sound, actions are SHA-pinned, and CI is green on the PR itself. No blocking (P0/P1) issues found. A few non-blocking notes below.

## What was reviewed
- `.github/workflows/ci.yml` (only changed file): added a `changes` preflight job using `dorny/paths-filter`, gated `build` and `npm-test` on it via `needs`/`if`, added draft-PR skipping, simplified the Go matrix to a single hardcoded version, and added `paths-ignore` on the `push` trigger.

## Correctness — verified

1. **Output expression is correct (the usual GHA ternary footgun does NOT bite here).**
   ```yaml
   code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
   ```
   The risk with `A && B || C` is that a falsy `B` silently falls through to `C`. Here `B` is the string `'false'`, and GitHub's boolean coercion treats any **non-empty** string as truthy — so `true && 'false'` yields `'false'`, and `'false' || 'true'` short-circuits on the truthy `'false'`, producing `'false'`. Result: docs-only PRs correctly emit `code=false`. Push events emit `'true'` via the left short-circuit. ✅

2. **Skip semantics satisfy required checks.** `changes` has no `needs`, so it always runs; `build`/`npm-test` are skipped only via their own job-level `if`. A job skipped by its own `if` reports a `skipped` conclusion, which GitHub counts as passing for required status checks — this is GitHub's documented pattern and is exactly what avoids the deadlock the PR targets. ✅

3. **Draft handling is consistent.** `converted_to_draft` is added to `types`, and the `if` guard `!github.event.pull_request.draft` skips CI for drafts; `ready_for_review` re-triggers a full run when undrafted. Draft PRs can't be merged regardless, so the skipped-as-passing status is harmless. ✅

4. **Security hygiene preserved.** `dorny/paths-filter` and `actions/checkout` are pinned to commit SHAs with version comments; the preflight checkout uses `persist-credentials: false`. ✅

5. **`predicate-quantifier: every` + negation-only filters** is the right construction for "true unless ALL changed files are docs". A `some` quantifier with negation patterns would always be true (any real file matches at least one negation), so the inline comment's warning against it is accurate. ✅

## Non-blocking notes (nits / P2)

- **Preflight `checkout` is likely unnecessary for PR events.** `dorny/paths-filter` resolves changed files via the GitHub API on `pull_request` events and does not require a working tree. The `actions/checkout` step in the `changes` job is harmless but adds a few seconds; it could be dropped for PR-only filtering. (Keep it only if you later add `push`-event filtering that relies on `base`.)
- **PR description vs. diff.** The description says it "replaces workflow-level `paths-ignore` on the `pull_request` trigger," but the diff (branch vs. `main`) contains no removal of a `pull_request` `paths-ignore` and instead *adds* `paths-ignore` to `push`. Purely cosmetic — the end state matches the stated intent — but worth aligning the wording.
- **Matrix removal** collapses the single-entry Go matrix to a hardcoded `"1.24.x"`. Fine and simpler (YAGNI); just note that re-introducing multi-version testing later means re-adding the matrix.
- **Context:** branch protection on `main` currently has no required status checks enabled, so the deadlock this fixes is forward-looking rather than currently active. The fix is still correct and worth landing ahead of enabling required checks.

Nice touch on the inline documentation explaining the quantifier choice and the picomatch dotfile behavior — it makes the non-obvious filter logic maintainable.

yujiawei (dismissed):
# Code Review — PR #46 (octo-cli)

## Verdict: APPROVED

A well-reasoned, thoroughly documented CI change. It correctly solves the required-status-check deadlock that workflow-level `paths-ignore` causes on `pull_request`. I verified the core claims against the repo's actual configuration and against the check runs on the head commit. No blocking issues; a few optional nits below.

## What this does (and why it's correct)

- Removes `paths-ignore` from the `pull_request` trigger (kept on `push`). This is the crux: with workflow-level `paths-ignore`, GitHub creates **no** check run for docs-only PRs, so a required check stays "expected" forever and the PR can never merge.
- Adds a `changes` preflight job using `dorny/paths-filter` with `predicate-quantifier: every` and negation-only filters, exposing `code` (`true` unless every changed file is a doc/template).
- Gates the `build` job's real steps **per-step** on `code == 'true'`, with a skip-notice step otherwise — so the matrix child check `build & test (1.24.x)` is always emitted (success on docs-only PRs).

### Verified against the live config

- **Required checks** (ruleset `main branch protection`): exactly `build & test (1.24.x)` and `code-review`. `npm test` is **not** required.
  - This validates the two load-bearing decisions: (a) keeping the single-value matrix so the check name stays `build & test (1.24.x)` (the parenthesised form GitHub emits only for matrix jobs), and (b) per-step gating rather than job-level `if`, which would skip before matrix expansion and never create the required child check — re-creating the deadlock on docs-only PRs.
  - Because `npm test` is not a required check, the `npm-test` job using a **job-level** skip (`code == 'true'`) is fine — fully skipping it on docs-only PRs causes no gate deadlock.
- **Filter logic** (`every` + negations): docs-only PR → no file satisfies all negations → `code=false`; any code file (or mixed PR) → `code=true`. The inline note that `some` would make the skip a no-op (every file matches at least one `!` pattern) is accurate.
- **Confirmed on this PR**: it touches `.github/workflows/ci.yml`, which is classified as code (`code=true`), and `build & test (1.24.x)` ran and succeeded on the head commit — the per-step gating path works end-to-end.
- **Scoping/security**: `pull-requests: read` added only on the `changes` job (needed by paths-filter for the PR file list); all actions are SHA-pinned with version comments; `push` retains `paths-ignore`.

## Nits / suggestions (non-blocking)

1. **`actions/checkout` in the `changes` job is unnecessary.** `dorny/paths-filter` on `pull_request` events obtains the changed-file list via the GitHub API and does not need a working tree. The checkout is harmless but adds a few seconds; it can be dropped. (If you keep it for robustness, no objection.)
2. **Asymmetric skip strategy is intentional but undocumented in `npm-test`.** `build` uses per-step gating (to keep the required check alive) while `npm-test` uses a job-level `if`. The reason — `npm test` is not a required check — is sound, but a one-line comment on the `npm-test` job mirroring the rationale already present on `build` would prevent a future maintainer from "fixing" the inconsistency and accidentally hiding `npm test` behind a job-level skip on the build job too.
3. **`**/*.md` root-file matching depends on picomatch globstar semantics.** The skip logic hinges on `**/*.md` matching root-level `README.md` (so `code=false` on a root README-only PR). The inline comment cites a picomatch v4 unit test, which is the right diligence — but this PR's own CI run does not exercise a docs-only path. Consider a follow-up smoke test (or a tracked test PR) so a future picomatch/`dorny` upgrade that changes root-glob behavior is caught rather than silently regressing the skip to a no-op or, worse, blocking real PRs.
4. **Draft guard is job-level**, so for a *draft* PR the `build & test (1.24.x)` child check is not created (job skipped before matrix expansion). This is acceptable — drafts can't merge anyway, and `ready_for_review` re-triggers and creates the check — but worth confirming the team is comfortable with draft PRs showing the required check as "expected" until marked ready.

None of these block merge. The change is correct, secure, and the in-line documentation is unusually good.

lml2468:
Close+reopen to trigger `pull_request.reopened` event so Jerry-Xin review bot re-queues on 190879f. yujiawei already APPROVED on this head. Same fallback pattern as web#288.

lml2468:
Close+reopen to retrigger Jerry-Xin review bot on head 190879f (same symptom as web#288; yujiawei already APPROVED 16:57Z, Jerry-Xin unresponsive). Parallel test of the close/reopen fallback path.

Tuanzi1015:
Branch updated onto `main` (merge `b109033`) and two non-blocking review nits applied in `637931d`:
- Removed the unnecessary `actions/checkout` from the `changes` job (`dorny/paths-filter` uses the PR API on `pull_request` events).
- Documented the intentional job-level vs per-step skip asymmetry between `npm-test` and `build`.

actionlint passes locally; no behavior change to the workflow logic that was already approved.

Tracking issue opened as #50. **`check-sprint` needs `Closes #50` added to the PR description** — I don't have write access to edit this PR's body (it's @lml2468's PR). @lml2468 / maintainer: please add `Closes #50` to the description and add #50 to the Octo Board sprint so `check-sprint` goes green. The two required checks (`build & test (1.24.x)`, `code-review`) plus 2 maintainer approvals (stale-dismissed by this push) are the remaining merge gates.

Jerry-Xin (approved):
Summary: This PR is in scope for `octo-cli` and the CI workflow changes look correct for keeping the required matrix check alive on docs-only PRs while still running full validation for code changes.

💬 Non-blocking
- 🔵 Suggestion `.github/workflows/ci.yml:13`: The comment says `dorny/paths-filter` produces “skipped” statuses. More precisely, it produces outputs that this workflow uses to skip steps/jobs. The behavior is fine, but tightening the wording would reduce future confusion.

✅ Highlights
- ✅ `.github/workflows/ci.yml:40`: The fallback output keeps push events treated as code changes, so main-branch code pushes still run the full CI.
- ✅ `.github/workflows/ci.yml:74`: Keeping `build & test` step-gated instead of job-gated preserves the required `build & test (1.24.x)` matrix check.
- ✅ `.github/workflows/ci.yml:153`: Skipping `npm test` at job level is reasonable given it is documented as non-required.
- ✅ Workflow syntax passed `actionlint v1.7.7` locally.

Jerry-Xin (commented):
Summary: This PR is in scope for `octo-cli` and the CI workflow changes look correct for keeping the required matrix check alive on docs-only PRs while still running full validation for code changes.

💬 Non-blocking
- 🔵 Suggestion `.github/workflows/ci.yml:13`: The comment says `dorny/paths-filter` produces “skipped” statuses. More precisely, it produces outputs that this workflow uses to skip steps/jobs. The behavior is fine, but tightening the wording would reduce future confusion.

✅ Highlights
- ✅ `.github/workflows/ci.yml:40`: The fallback output keeps push events treated as code changes, so main-branch code pushes still run the full CI.
- ✅ `.github/workflows/ci.yml:74`: Keeping `build & test` step-gated instead of job-gated preserves the required `build & test (1.24.x)` matrix check.
- ✅ `.github/workflows/ci.yml:153`: Skipping `npm test` at job level is reasonable given it is documented as non-required.
- ✅ Workflow syntax passed `actionlint v1.7.7` locally.

Jerry-Xin (approved):
Summary: This PR is in scope for `octo-cli` and the CI workflow changes look correct for keeping the required matrix check alive on docs-only PRs while still running full validation for code changes.

💬 Non-blocking
- 🔵 Suggestion `.github/workflows/ci.yml:13`: The comment says `dorny/paths-filter` produces “skipped” statuses. More precisely, it produces outputs that this workflow uses to skip steps/jobs. The behavior is fine, but tightening the wording would reduce future confusion.

✅ Highlights
- ✅ `.github/workflows/ci.yml:40`: The fallback output keeps push events treated as code changes, so main-branch code pushes still run the full CI.
- ✅ `.github/workflows/ci.yml:74`: Keeping `build & test` step-gated instead of job-gated preserves the required `build & test (1.24.x)` matrix check.
- ✅ `.github/workflows/ci.yml:153`: Skipping `npm test` at job level is reasonable given it is documented as non-required.
- ✅ Workflow syntax passed `actionlint v1.7.7` locally.

yujiawei (approved):
# Code Review — PR #46 (octo-cli)

**Verdict: APPROVED** — no blocking issues. The change is correct, well-reasoned, and empirically validated by the PR's own green CI.

## Summary

This PR replaces the workflow-level `paths-ignore` on the `pull_request` trigger with a `changes` preflight job using `dorny/paths-filter`, fixing the well-known deadlock where docs-only PRs never get a required check run created (so a required status check can never be satisfied). The push trigger keeps `paths-ignore` since it has no required-check semantics. The design is sound and the implementation handles two genuinely tricky GitHub Actions semantics correctly.

## Verification

I traced the two non-obvious correctness claims the change depends on; both hold.

### ✅ 1. The `code` output expression coerces correctly

```yaml
code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
```

GHA `&&`/`||` return operands (JS-style), and the non-empty string `'false'` is **truthy** in GHA. Tracing all branches:
- **Docs-only PR**: `true && 'false'` → `'false'`; then `'false' || 'true'` short-circuits on the truthy left operand → **`'false'`** ✅ (the intended "skip" value, not an accidental `'true'`)
- **Push event**: `false && …` → `false`; `false || 'true'` → `'true'` ✅
- **Code PR**: `true && 'true'` → `'true'` ✅

The only failure mode of this ternary idiom — a *falsy* (empty-string) middle operand — cannot occur here because `dorny/paths-filter` outputs the literal strings `'true'`/`'false'`. The downstream gate `needs.changes.outputs.code == 'true'` is a string comparison and behaves as intended.

### ✅ 2. Per-step gating (not job-level) preserves the required matrix child check

The asymmetry between `build` (per-step `if:`) and `npm-test` (job-level `if:`) is deliberate and correct:
- A **job-level** `if:` on a matrix job is evaluated **before** matrix expansion. When false, the parenthesized child check `build & test (1.24.x)` is **never created** (not "skipped" — simply absent), leaving a required gate pinned to that exact name stuck in "Expected — Waiting for status to be reported" forever. This is the documented behavior behind `actions/runner#952`.
- The chosen fix keeps the job always running (job-level `if:` only guards drafts), so the matrix expands and the child check is emitted. A job whose steps are all skipped completes as **success**, satisfying the required gate on docs-only PRs.
- `npm test` is safe to skip job-level because it is **not** a required check (the ruleset pins only `build & test (1.24.x)` and `code-review`).

### ✅ 3. Empirically confirmed by the PR's own CI

`Detect changed paths`, `build & test (1.24.x)`, and `npm test` are all green on this code PR, and `sanity / actionlint` passes — confirming the workflow is syntactically valid and the required matrix child check name is correctly emitted.

## Strengths

- Pinned action SHAs (`dorny/paths-filter@6852f92…`, `checkout`, `setup-go`, etc.) — supply-chain hygiene maintained.
- `changes` job runs with minimal, scoped permissions (`contents: read`, `pull-requests: read`) and no checkout — `dorny/paths-filter` uses the API on `pull_request` events.
- Excellent inline comments documenting *why* each non-obvious choice was made (predicate-quantifier `every`, single-value matrix retention, per-step vs job-level skip). This is exactly the kind of documentation that prevents a future "cleanup" from regressing the gate.

## Suggestions (non-blocking, P2 / maintainability)

1. **Long-term maintainability of the matrix-name coupling.** The whole design is pinned to the exact required-check string `build & test (1.24.x)`. The more robust industry pattern is a small aggregate "gate" job with `if: always()` that inspects `needs.*.result` (accepting `success || skipped`) and is itself the single stable required check — it decouples the gate from matrix dimensions, so adding/removing Go versions no longer risks breaking branch protection. Not required for this PR (the current approach is behaviorally correct), but worth considering as a follow-up to reduce the "do not touch the matrix" fragility.
2. **`code` output `success` vs `skipped` semantics.** On docs-only PRs the required check passes as a green *success* (all steps skipped) rather than a *skipped* conclusion. This is the intended outcome and fully satisfies the gate; just noting it so readers aren't surprised that a docs-only PR shows `build & test (1.24.x)` green rather than skipped.

Neither suggestion blocks merge.

PR #52, opened 2026-06-08
ci: pass TRIAGE_WEBHOOK_URL to octo-issue-notify

lml2468:
The issue notification caller only forwarded OCTO_BOT_TOKEN, so the triage webhook mode was silently skipped and the triage agent was never invoked. Forward the org-level TRIAGE_WEBHOOK_URL secret so triage runs on new issues.

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -18,3 +18,4 @@ jobs:
       event_action: ${{ github.event.action }}
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}


Jerry-Xin (approved):
The PR is relevant to `Mininglamp-OSS/octo-cli`: it updates this repository’s issue notification workflow so the existing reusable issue notifier can invoke triage for new issues.

✅ Highlights
- 🔵 Suggestion: The added secret forwarding in [.github/workflows/octo-issue-feed.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-qqjl0rgd/repo/.github/workflows/octo-issue-feed.yml:21) matches the referenced reusable workflow contract: `TRIAGE_WEBHOOK_URL` is declared optional by `octo-issue-notify.yml@v1` and is passed through as the webhook URL.
- The change is appropriately scoped and does not expose the webhook value in logs or workflow inputs.
- No test changes are needed for this workflow-only secret mapping change.

No blocking issues found.

yujiawei (approved):
# Code Review — PR #52 (octo-cli)

## Summary

Single-line change adding `TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}` to the `secrets:` block of the reusable-workflow call in `.github/workflows/octo-issue-feed.yml`. This forwards the org-level triage webhook secret to `octo-issue-notify.yml@v1`, which is what enables the triage webhook path. The fix is correct and matches the stated intent.

## Verification

I verified the full cross-file call chain against the pinned `@v1` ref of `Mininglamp-OSS/.github`:

- ✅ **Reusable workflow accepts the secret.** `octo-issue-notify.yml@v1` declares `TRIAGE_WEBHOOK_URL` under `workflow_call.secrets` as `required: false`. Without this PR, the caller never passed it, so the downstream `webhook-url` was always empty and triage mode was silently skipped — exactly the bug described. Passing it is the right wiring.
- ✅ **Downstream consumption is correct.** The reusable workflow only builds and forwards a `webhook_payload` to the `octo-notify@v1` composite action via `webhook-url: ${{ secrets.TRIAGE_WEBHOOK_URL }}`. In the action, `webhook_url = os.environ.get('WEBHOOK_URL', '').strip()` and webhook mode is gated on `if webhook_url:`.
- ✅ **Graceful degradation when the secret is unset.** If the org/repo secret `TRIAGE_WEBHOOK_URL` is not configured, the expression resolves to an empty string, `.strip()` yields `''`, and webhook mode is simply skipped. No build break, no hard failure — the IM-feed path still runs. So this change cannot regress repos where the secret isn't provisioned.
- ✅ **Secret hygiene.** The URL is passed as a secret (never hardcoded), consistent with the "capability secret" comment in the reusable workflow. The webhook payload is minimal identifying metadata (`repo`, `number`, `title`, `url`, `author`, `action`) — the issue body is not egressed through Actions.
- ✅ **Explicit secrets mapping** (rather than `secrets: inherit`) keeps the surface tight — only the two intended secrets cross the boundary.

## Notes for human verification (non-blocking)

Because this PR is flagged security-sensitive, two items worth a human confirming — neither blocks merge:

1. **Org secret provisioning.** Confirm `TRIAGE_WEBHOOK_URL` is actually configured at the org (or repo) level. If it is *not* set, this PR is a safe no-op and triage still won't run — the wiring is correct but the secret would also need to exist for the feature to take effect.
2. **No allowlist on the webhook endpoint.** Unlike the IM path (which is constrained to a hardcoded `ALLOWED_API_BASES`), the `webhook-url` transport accepts any URL. That is an acceptable threat model here since the value is a capability secret set only by org admins (not user-influenced input), but it's worth being aware that whoever can set this secret controls the POST destination.

## Verdict

No P0/P1 issues. The change is minimal, correct, cross-file-verified, and fails safe when the secret is absent. Approving.

PR #56, opened 2026-06-25
feat(npm): bundle platform binaries in npm packages

### LINKED ISSUE #57: feat(npm): bundle octo-cli binaries in npm packages
## Problem

`@mininglamp-oss/octo-cli` currently downloads its platform binary from GitHub Releases during npm `postinstall`. This makes npm installation depend on GitHub network availability even when the npm registry is reachable.

## Goal

Ship the platform binaries inside npm packages so `npm install -g @mininglamp-oss/octo-cli` does not download from GitHub at install time.

## Acceptance Criteria

- The main npm package resolves a prebuilt binary from a matching platform optional dependency.
- Platform sub-packages carry the bundled `octo-cli` binary for darwin/linux/win32 on x64/arm64.
- The npm publish workflow publishes platform packages first and the main package last.
- GitHub Release archives remain available for non-npm distribution channels.
- Local packaging tests cover package generation, shim resolution, and executable permissions.

liuooo:
## What

`@mininglamp-oss/octo-cli` used to download the platform binary from GitHub Releases during npm `postinstall`. That makes npm installs depend on GitHub network availability even when the npm registry is reachable.

This PR switches the npm distribution to an esbuild-style package layout: the main package ships a Node shim, and platform-specific optional dependency packages carry the prebuilt Go binary inside npm tarballs.

## How

- **npm/package.json** — removes the `postinstall` downloader, points the `octo-cli` bin at `bin/octo-cli.js`, and leaves `optionalDependencies` to be stamped during packaging.
- **npm/bin/octo-cli.js** — resolves the current platform package and execs its bundled binary, preserving exit codes and signals.
- **npm/scripts/prepare-packages.js** — repacks GoReleaser archives into six platform sub-packages plus the main package with exact-pinned optional dependencies.
- **.goreleaser.yaml** — emits `tar.gz` archives for every platform, including Windows, so npm packaging has one archive format to handle.
- **.github/workflows/release-publish.yml** — builds once after the release evidence gate, uploads the same `goreleaser-dist` artifact to GitHub Releases, and calls npm publishing from that artifact.
- **.github/workflows/npm-publish.yml** — publishes platform packages first and the main package last, supports both `workflow_call` artifact input and manual Release-asset dry-runs, keeps provenance, and uses `--ignore-scripts`.
- **npm/README.md** — documents that npm install no longer downloads from GitHub.

## Tests

- `npm --prefix npm test`
- `actionlint .github/workflows/npm-publish.yml .github/workflows/release-publish.yml`
- `goreleaser check`
- `git diff --check`
- Local packaging smoke test:
  - GoReleaser snapshot produced six platform archives.
  - `prepare-packages.js` generated six platform packages plus the main package.
  - `npm pack` produced npm tarballs.
  - local `npm install` installed the main package and host platform package.
  - installed binary mode was `-rwxr-xr-x`.
  - `node_modules/.bin/octo-cli version` executed through the shim successfully.

Closes #57


Jerry-Xin (dismissed):
Summary: The PR is in scope for `octo-cli` and the npm packaging redesign looks sound; I found no blocking correctness or security issues.

🔴 Blocking

None.

💬 Non-blocking

🟡 Warning: `npm/scripts/prepare-packages.js:104` extracts the full tar archive before checking the expected binary. Since release artifacts are part of the trust root this is not a blocker, but it would be safer to extract only the expected member (`octo-cli` / `octo-cli.exe`) or validate archive member paths before extraction.

🔵 Suggestion: `npm/scripts/prepare-packages.js:119` generates platform packages with only `bin/*` and `package.json`. Consider copying `LICENSE` into each platform sub-package too, so every published npm tarball carries the license text directly.

🔵 Suggestion: `npm/scripts/prepare-packages.test.js:134` covers host-platform shim resolution but does not exercise the Windows package name / `octo-cli.exe` path unless run on Windows. A small injectable resolver test would lock that behavior down without needing a Windows runner.

✅ Highlights

- Good project fit: this directly improves the `@mininglamp-oss/octo-cli` npm distribution.
- The main package now avoids install-time GitHub downloads and pins platform optional dependencies exactly.
- The release workflow builds once and reuses the same artifact for GitHub Release assets and npm packaging.
- Publish ordering puts platform packages before the main package, which avoids a transient broken install window.

Verification performed: `npm --prefix npm test` passed, `git diff --check HEAD~1..HEAD` passed, and a local fake-archive `prepare-packages.js` plus `npm pack` smoke test succeeded. `actionlint` and `goreleaser` were not installed in this environment, so I could not rerun those checks locally.

OctoBoooot (dismissed):
## Review: feat(npm): bundle platform binaries in npm packages (#56)

**Verdict:** Request changes.

CI is red — `npm test` job is failing because this PR deletes `npm/scripts/install.test.js` but leaves the hardcoded reference to it in `.github/workflows/ci.yml:177`. One-line fix, but it has to land in this PR.

The overall design — esbuild-style platform sub-packages + Node shim, build-once / fan-out artifact in `release-publish.yml`, idempotent skip-if-exists on republish — is well-shaped. Tests around `prepare-packages.js` are unusually thorough for packaging glue.

Risk tier: infra (GitHub Actions changes touching the release path). `needs-human-review` auto-applied — 12 files crosses the size threshold and the release workflow change warrants careful merge timing alongside the CI fix.

### Blockers
- `.github/workflows/ci.yml:177` (not in this diff, but stale because of it) — the `npm-test` job runs `node --test scripts/install.test.js`, but this PR deletes `npm/scripts/install.test.js`. CI on this PR has been red on `npm test` since the first push (`Could not find 'scripts/install.test.js'` → exit 1), and the failure will repeat on every subsequent code-touching PR. Fix in this PR: replace with the new test entry point — either `node --test scripts/prepare-packages.test.js` or, cleaner, `npm test` (the `test` script in `npm/package.json` already points at `prepare-packages.test.js`).

### Praise
- `npm/scripts/prepare-packages.test.js:34-49` — cross-validating the npm platform matrix against `.goreleaser.yaml`'s `goos × goarch` is the right shape of guard; an unintended divergence between the two would be caught at PR time rather than discovered during a release.
- `npm/scripts/prepare-packages.js:101-105` — rejecting non-regular archive members (verified by the `/etc/passwd` symlink test) closes the path-traversal-via-symlink class for the unpack step.
- `.github/workflows/release-publish.yml` job graph — build once, fan out the same `goreleaser-dist` artifact to both `release-upload` and `publish-npm` (consumed via `from_artifact: true`). Eliminates the "GitHub Release has a different binary than npm" failure mode that the old re-download path could produce.
- `.github/workflows/npm-publish.yml` `publish_one` — `npm view "$name@$VERSION"` skip-if-exists makes a partial-publish retry safe to rerun without manual cleanup.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-cli PR#56 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-cli/pull/56
**Head SHA:** 9719725554c91dc9b74773ab667dca82b4e48e47
**Title:** feat(npm): bundle platform binaries in npm packages

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Shim binary resolution | ✅ | `npm/bin/octo-cli.js:20-48` — PLATFORM_PACKAGES map → require.resolve → spawnSync |
| Signal/exit-code propagation | ✅ | `npm/bin/octo-cli.js:52-62` — mirrors deleted run.js faithfully |
| Platform matrix parity (npm ↔ goreleaser) | ✅ | `prepare-packages.test.js:47-56` parses .goreleaser.yaml and asserts set-equality |
| Checksum verification in CI | ✅ | `npm-publish.yml:155-162` — sha256sum --check + per-file coverage assertion |
| Security gates preserved | ✅ | Release-existence gate (L108-132), CI-evidence gate (L134-163) — logic identical, only comments trimmed |
| Publish ordering (sub-pkgs → main) | ✅ | `npm-publish.yml:192-198` — loop over `build/octo-cli-*` before `publish_one build/octo-cli` |
| `set -euo pipefail` on failure paths | ✅ | Each shell step declares it independently |
| Symlink/non-regular-file rejection | ✅ | `prepare-packages.js:88-90` + test at `prepare-packages.test.js:113-139` |
| `.gitignore` correctness for shim | ✅ | `!/npm/bin/` re-includes dir; `/npm/bin/octo-cli` and `/npm/bin/octo-cli.exe` ignore only generated binaries |
| workflow_call plumbing | ✅ | release-publish.yml:publish-npm passes `from_artifact: true`, NPM_TOKEN, correct permissions subset |
| Test suite runs pre-publish | ✅ | `npm-publish.yml:151` — `node --test scripts/prepare-packages.test.js` before packaging |
| Artifact pass-through (build → release-upload + publish-npm) | ✅ | `upload-artifact` v4 at release-publish.yml:97-102; both consumers use `download-artifact` v4 |

## 2. Findings

### P2-1: `npm view` dedup check is fragile under registry unavailability

**File:** `.github/workflows/npm-publish.yml:176-179`
**Diff-scope:** new (introduced by this PR)

```bash
if [ -z "$DRY_RUN" ] && npm view "$name@$VERSION" version >/dev/null 2>&1; then
  echo "[npm-publish] $name@$VERSION already on the registry — skipping"
  return 0
fi
```

The `npm view` call swallows all errors via `2>&1 >/dev/null`. If the npm registry is temporarily unreachable (network blip, DNS issue), the command returns non-zero and the script attempts to publish a version that may already exist. `npm publish` then rejects with a 403, `set -e` halts the script, and subsequent platform packages + main package are not published.

On re-run, the already-published packages are skipped (view succeeds this time), but the partial-publish → full-publish cycle requires manual intervention.

**Mitigation:** Distinguish "version exists" (exit 0 from npm view) from "network error" (could retry or at least log a distinct warning). Low risk in practice since GHA→npm registry connectivity is highly reliable.

### P2-2: Windows archive format change (zip → tar.gz) is a breaking change for direct GitHub Release consumers

**File:** `.goreleaser.yaml:21-31`, `release-publish.yml:119`
**Diff-scope:** new (introduced by this PR)

The PR removes the Windows zip format_override so all platforms produce tar.gz. The rationale is sound for the npm path (single-format CI extraction) and notes that Windows 10+ ships bsdtar. However, users/scripts that download `*_windows_*.zip` from GitHub Releases will break.

The upload step (`gh release upload "$TAG" dist/*.tar.gz dist/checksums.txt`) no longer uploads `.zip` files. Any downstream automation keying on `.zip` for Windows will 404.

**Mitigation:** Call out in release notes. Consider keeping zip for Windows GitHub Release consumers while using tar.gz internally for npm packaging (would require format-specific extraction in prepare-packages.js).

### P2-3: Security-rationale comments removed without functional replacement

**File:** `.github/workflows/npm-publish.yml` (entire file), `.github/workflows/release-publish.yml` (entire file)
**Diff-scope:** amplified (pre-existing comments removed; no new comments replace them)

The base branch carries extensive security-rationale comments explaining WHY each gate exists (e.g., "a repo:write actor can create a Release via the API/UI without going through release-publish.yml"). This PR strips ~100 lines of such comments while keeping the gates functionally intact.

The gates work correctly, but the removed comments served as institutional knowledge for future maintainers — explaining the threat model each gate closes. Without them, a future refactor may inadvertently weaken a gate without understanding its purpose.

**Mitigation:** Consider retaining at least one-line security rationales per gate (e.g., "# Gate: prevents publishing without a successful CI run on the tagged commit"). This is a maintainability concern, not a functional defect.

### Nits

- **`npm/bin/octo-cli.js:58`** — `os.constants.signals` lookup fallback to `0` means unknown signals produce exit code 128 (not wrong, but a `console.error` hint could help debug exotic signal scenarios).
- **`prepare-packages.test.js:141-163`** — Shim test only covers the happy path. A negative test (platform package not installed → shim exits 1 with helpful error) would lock in the error message contract.
- **`npm-publish.yml:66`** — Concurrency group `npm-publish-${{ inputs.tag }}` is unchanged but the comment explaining it was removed. One-liner would help.

## 3. Recommendations

1. **For P2-1:** Add a retry or a distinct error path for `npm view` failures — e.g., `npm view "$name@$VERSION" version 2>/dev/null` with an explicit check of the exit code, or a simple retry loop.
2. **For P2-2:** Document the Windows format change prominently in the PR description and release notes.
3. **For P2-3:** Retain terse security-rationale comments at each gate. The functional code is correct; the comments are for future maintainers.

## 4. Additional Observations

- The `os`/`cpu` field removal from the main `package.json` is correct — the main package should install on any platform, with `optionalDependencies` filtered by each sub-package's own `os`/`cpu` constraints. Unsupported platforms get a clear error from the shim.
- `delete manifest.scripts` in `prepare-packages.js:153` correctly strips dev-only test scripts from the published main package.
- The `--ignore-scripts` flag on `npm publish` is defense-in-depth (no lifecycle scripts run during publishing). Combined with `--provenance` and `id-token: write`, the Sigstore attestation chain is intact.
- `from_artifact` plumbing correctly differentiates same-run artifact consumption (workflow_call) from Release-asset download (workflow_dispatch), avoiding double binary builds.

## 5. Data-Flow Trace

| Consumer | Data | Upstream Source | Verified |
|----------|------|-----------------|----------|
| Shim `resolveBinary()` key | `${process.platform}-${process.arch}` | Node.js runtime | ✅ matches PLATFORM_PACKAGES keys |
| Shim `require.resolve(pkg/bin/binName)` | Sub-package binary path | npm install → optionalDependencies → `node_modules/@mininglamp-oss/octo-cli-<platform>/bin/` | ✅ prepare-packages.js creates matching layout |
| prepare-packages archive name | `octo-cli_${version}_${goOs}_${goArch}.tar.gz` | `.goreleaser.yaml:name_template` | ✅ test asserts set-equality |
| Workflow `dist/` path | Archives for packaging | download-artifact (`npm/dist`) or `gh release download` (`npm/dist`) | ✅ all under `working-directory: npm` |
| `NODE_AUTH_TOKEN` for dry-run | Empty string when `DRY_RUN` is set | Expression `steps.version.outputs.DRY_RUN == '' && secrets.NPM_TOKEN \|\| ''` | ✅ npm --dry-run doesn't require auth |
| publish_one loop | `build/octo-cli-*` glob | prepare-packages output dirs | ✅ main package (`build/octo-cli`) excluded from glob |
| CI evidence gate `ref_sha` | Tag → commit SHA resolution | `gh api git/ref/tags/$TAG` + annotated-tag dereference | ✅ logic preserved from base |

## 6. Blind-Spot Checklist (R5)

- **C1 — Dual-path parity:** N/A. No add/remove, subscribe/unsubscribe, or create/delete symmetry in this PR. The two entry points (workflow_call vs workflow_dispatch) are verified to produce equivalent outcomes via different artifact sources.
- **C2 — Control-flow ordering / nested reuse:** Clear. `publish_one` is called sequentially (loop → main), `set -euo pipefail` ensures ordering invariant. No nested reuse of security-critical functions.
- **C3 — Authorization boundary:** N/A. No new endpoints, tools, or capability exposure. Workflow permissions are correctly scoped (build: read-only; release-upload: write; publish-npm: read+id-token).
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth/authz changes.
- **C5 — Build ≠ runtime:** Checked. The npm packages are built on CI (Linux) but consumed on end-user platforms. `require.resolve` uses Node.js module resolution which works across platforms. `process.platform`/`process.arch` are runtime-correct. `chmod 0o755` is set on extraction (prepare-packages.js:92). Windows `.exe` suffix is handled by `binFileName()` and the shim.
- **C6 — Governance / policy / security docs:** N/A. README.md updated to reflect new trust model. No SECURITY.md or disclosure-policy changes.

## 7. Cross-Round Blocker Re-Check (R6)

---

## Verdict

**[Octo-Q] verdict: APPROVE**

All findings are P2 (implementation-nit / maintainability) or lower. No P0/P1 issues. The architecture change from postinstall-download to esbuild-style platform sub-packages is sound: security gates preserved, data-flow verified end-to-end, test coverage adequate, and the CI plumbing is correct for both entry points.

The three P2 findings are advisory — none blocks landing.

Jerry-Xin (changes requested):
🔴 Review REQUEST_CHANGES — correcting my earlier APPROVE.

I previously approved this SHA, but that was wrong: CI is red on this exact commit and the failure is introduced by this PR's own delta, so it is a real blocker, not a PR-unrelated red.

🔴 Blocking
- This PR deletes `npm/scripts/install.test.js`, but `.github/workflows/ci.yml:177` still runs `node --test scripts/install.test.js`. The `npm test` job fails on this SHA (missing file). `ci.yml` is not in this PR's diff, so the test-entry reference was never updated. Fix: update `ci.yml:176-177` to run the new test (e.g. `node --test npm/scripts/prepare-packages.test.js`, or switch the step to `npm test`) so the CI test entry matches the relocated/renamed tests. Once `npm test` is green I will flip back to APPROVE.

The packaging design itself is sound (bundled platform binaries via platform-specific optional deps + Node shim, removing the old install-time GitHub download). My earlier non-blocking notes still stand:
- 🟡 `npm/scripts/prepare-packages.js` extracts the full tar before validating the target binary; prefer extracting only the expected member or validating the archive member path (path-traversal hardening). Release artifacts are inside the trust root, so non-blocking.
- 🔵 Consider copying `LICENSE` into each platform subpackage.
- 🔵 Add an injectable-resolver test covering the Windows `octo-cli.exe` path (no Windows runner needed).

Net: design LGTM, but the CI test reference must be updated to match the deleted test file before merge.

yujiawei (changes requested):
# Code Review — PR #56 (octo-cli)

Reviewed at head `9719725554c91dc9b74773ab667dca82b4e48e47` against `main`. This is a
solid, well-documented migration from postinstall-download to esbuild-style platform
sub-packages. The shim, the repacker, and the build-once/fan-out workflow shape are all
in good order. One blocker, plus a few hardening items.

## Verdict: CHANGES_REQUESTED

The migration is incomplete: the binary still references a test file this PR deletes, so
CI is red on every code PR. Everything else is advisory.

---

## P0 — Build break: CI still runs a deleted test

`.github/workflows/ci.yml:177` (the `npm test` job) runs:

```yaml
- name: node --test
  run: node --test scripts/install.test.js
```

This PR **deletes** `npm/scripts/install.test.js` (and `install.js`, `run.js`) and replaces
the suite with `npm/scripts/prepare-packages.test.js`. `ci.yml` was not updated, so the job
fails immediately:

```
Could not find 'scripts/install.test.js'
##[error]Process completed with exit code 1.
```

This is observable on the PR right now — the `npm test` check is failing. `npm/package.json`
already points its `test` script at the new file, so the fix is to align the workflow:

```yaml
run: node --test scripts/prepare-packages.test.js
# or simply: run: npm test
```

`ci.yml:177` is the only dangling reference to the deleted files in the repo (Makefile,
lefthook, and the other workflows are clean). This single edit is required to land the
refactor — deleting the test without repointing its only caller leaves the migration
half-done.

---

## P1 / P2 — Advisory

### [P2] `publish-npm` and `release-upload` run in parallel (release-publish.yml:102, :123)

Both jobs declare `needs: [publish, build]`, so they race. `npm publish` versions are
**immutable**; `gh release upload` can fail after npm has already published. Result: npm
packages exist with no matching GitHub Release assets/checksums.

Why this is P2 and not a blocker: in the new model npm packages **bundle** the binary and
no longer download from the Release, so a missing Release asset does **not** break `npm
install`. The residual exposure is the `install.sh` / direct-download path (it still pulls
`releases/download/.../*.tar.gz` + `checksums.txt`), which would be temporarily broken until
someone re-runs `release-upload`. `gh release upload --clobber` is idempotent, so recovery
is a re-run, not a stuck state. Still worth tightening: make `publish-npm` depend on
`release-upload` (`needs: [publish, build, release-upload]`) so the public Release is
complete before the immutable npm publish happens.

### [P2] Checksum gate can pass without verifying a tarball (npm-publish.yml:188-194)

```bash
sha256sum --check --ignore-missing checksums.txt
for f in *.tar.gz; do
  grep -qF "  $f" checksums.txt || { echo "::error::$f not covered"; exit 1; }
done
```

`--ignore-missing` skips listed-but-absent files, and the `grep -qF "  $f"` guard is a
substring match: a `checksums.txt` line ending in `octo-cli_x_linux_amd64.tar.gz.old` would
satisfy the guard for `octo-cli_x_linux_amd64.tar.gz`. Not exploitable today (checksums.txt
is goreleaser-generated), but the guard is weaker than it looks. Anchor the match to the end
of line, or build a checksum subset for exactly the present `*.tar.gz` and run a plain
`sha256sum --check` (no `--ignore-missing`) on that subset.

### [P2] Repacker extracts the whole archive before validating (prepare-packages.js:102)

`execFileSync("tar", ["-xzf", archive, "-C", workDir])` unpacks every member, then only the
expected binary is checked. For the manual `workflow_dispatch` path that downloads Release
assets, a malformed/tampered archive could drop extra members into the runner workspace
before any validation. Low risk given the trust chain, but extracting only the expected
member (`tar -xzf archive -C workDir <binFile>`) or listing members first would close it.
The existing symlink/non-regular-file guard is good and worth keeping.

### [P2] `parseArgs` accepts unknown and duplicate flags (prepare-packages.js:63-77)

The loop blindly pairs `argv[i]`/`argv[i+1]`, so unknown keys are silently absorbed and a
duplicate `--out` would override the first — and `--out` feeds the destructive
`fs.rmSync(outDir, { recursive: true, force: true })` at line 90. The current caller is
fixed, so this is hardening for future workflow/manual typos: reject unknown keys and
duplicates.

### [P2] Platform sub-packages ship no LICENSE (prepare-packages.js:120-133)

Each sub-package manifest sets `license: "Apache-2.0"` but `files: ["bin/<binFile>"]`, so the
LICENSE text isn't included in those tarballs. The main package ships LICENSE; the
sub-packages (which carry the actual binary) don't. Minor compliance nit — add `LICENSE` to
each sub-package's `files` and copy it in alongside the binary.

---

## What checks out

- **Shim signal/exit handling** (`npm/bin/octo-cli.js`): re-raises terminating signals and
  maps to `128+signum`, falls back to explicit exit for default-ignored signals, and treats
  `status === null` as exit 1. Consistent with the previous `run.js` behavior. Good error
  messages for the missing-package and unsupported-platform cases.
- **Cross-file matrix consistency**: `PLATFORM_PACKAGES` in the shim and `PLATFORMS` in
  prepare-packages describe the same 6 platforms with the same os/cpu mapping; the guard test
  asserts set-equality against `.goreleaser.yaml`'s `goos × goarch`. No drift.
- **Artifact path wiring**: `download-artifact` → `npm/dist`, then `--dist dist` under
  `working-directory: npm` resolves to `npm/dist`. Consistent across the `from_artifact` and
  `workflow_dispatch` paths.
- **Goreleaser windows tar.gz**: dropping the zip `format_override` to keep the npm Linux
  repack branch-free is a reasonable tradeoff; the comment correctly notes Windows 10+ bsdtar
  reads tar.gz and npm is now the primary Windows path.
- `prepare-packages.test.js` passes locally (5/5).

## Coverage note

Reviewed: the full diff, both workflows end-to-end, the shim, the repacker and its tests,
goreleaser/gitignore/README changes, and all repo references to the deleted files. What I
could **not** verify from the diff alone: actual GoReleaser archive output shape and tar
member layout, npm tarball file-mode preservation on publish, the reusable-workflow
(`reusable-release-publish.yml@v1`) behavior and the OIDC/provenance path end-to-end, and
real Windows `tar.gz` install UX. A second independent (long-context) review leg was
attempted but timed out and did not return findings; this review is the primary reviewer
plus one independent automated pass.

Jerry-Xin (approved):
✅ APPROVE — re-reviewed on the new head `53867152ca85`.

The CI blocker from my earlier REQUEST_CHANGES is fixed. The fix commit changed only `.github/workflows/ci.yml` (+2 -2): the test step is now `npm test` instead of the deleted `node --test scripts/install.test.js`. Verified on the new SHA: `npm test` CI check is now **pass**, and `npm/scripts/install.test.js` is confirmed gone (no dangling reference remains).

🔴 Blocking: none

The packaging design is sound: bundled platform binaries shipped via platform-specific optional-dependency tarballs + a Node shim, removing the old install-time GitHub download (no postinstall network fetch). os/arch matrix, `bin`/`files[]`, and publish ordering (platform packages before the main package) all check out. Binaries come from the GoReleaser-built release artifacts (trust root), so no arbitrary-fetch substitution surface.

💬 Non-blocking (hardening, agree with peers — fine as follow-ups):
- 🟡 `npm/scripts/prepare-packages.js` extracts the full tar before validating the target binary; prefer extracting only the expected member (`octo-cli` / `octo-cli.exe`) or validating the archive member path first (path-traversal hardening).
- 🟡 Publish/Release upload steps could run in parallel; consider serializing or gating so a partial publish can't leave platform packages and the main package out of sync.
- 🟡 `npm view` error handling doesn't distinguish a genuine "not published" from a transient network error — could abort or mis-skip a publish; consider narrowing the error check.
- 🔵 Copy `LICENSE` into each platform subpackage.
- 🔵 Add an injectable-resolver test covering the Windows `octo-cli.exe` path (no Windows runner needed).

Net: CI green, blocker resolved, design LGTM. Approving. Remaining items are non-blocking polish.

yujiawei (approved):
# Code Review — PR #56 (octo-cli)

Re-review at head `53867152ca854a0c5678d2c1aa217075c9927f7e` (was `9719725` at first pass).
The only change since the prior review is commit `5386715`, which resolves the one blocker.

## Verdict: APPROVED

The blocker from the previous pass is fixed and the migration is now complete and green.

### Blocker resolved

The prior CHANGES_REQUESTED was for a single build break: `.github/workflows/ci.yml`'s
`npm test` job still ran the deleted `scripts/install.test.js`. Commit `5386715` repoints it:

```yaml
- name: npm test
  run: npm test        # → node --test scripts/prepare-packages.test.js
```

The `npm test` CI check is now passing, and there are no remaining references to the deleted
`install.js` / `install.test.js` / `run.js` anywhere in the repo. Tests pass locally (5/5).

### Remaining items (all advisory, none block)

- `release-publish.yml` — `publish-npm` and `release-upload` both `needs:[publish,build]`, so
  they race. npm versions are immutable; if upload fails afterward, npm exists without matching
  Release assets. Not a broken `npm install` (the binary is bundled in the sub-package now) —
  only the `install.sh`/direct-download path is briefly affected, and `--clobber` makes recovery
  a re-run. Tightening to `needs:[publish,build,release-upload]` would remove the race.
- `npm-publish.yml` checksum gate — `--ignore-missing` + substring `grep -qF "  $f"` can pass
  without verifying a tarball; anchor the match. Not exploitable today.
- `prepare-packages.js` — extracts the whole archive before validating; `parseArgs` accepts
  unknown/duplicate flags (a duplicate `--out` feeds the destructive `rmSync`). Hardening.
- Platform sub-packages declare `license: Apache-2.0` but ship no LICENSE file. Compliance nit.
- Windows zip → tar.gz: direct GitHub-Release consumers of `*_windows_*.zip` will 404. Worth a
  release-note callout; reasonable tradeoff given npm is now the primary Windows path.

### What checks out

Shim signal/exit handling (faithful to the old run.js), shim↔repacker platform-matrix parity
(guard-tested against `.goreleaser.yaml`), build-once/fan-out artifact wiring, provenance +
`--ignore-scripts` publish chain, and the publish ordering (sub-packages before main).

## Coverage note

Reviewed the full diff at the current head, both publish workflows end-to-end, the ci.yml fix
commit in isolation, the shim, and the repacker + tests. Not verifiable from the diff alone:
actual GoReleaser archive layout, npm tarball file-mode preservation on publish, the
`reusable-release-publish.yml@v1` internals, and real Windows tar.gz install UX. The unrelated
`check-sprint` check is failing (sprint tracking, not this PR's code).

OctoBoooot (approved):
## Re-review on `53867152ca85` (delta)

**Verdict:** Approve.

The fix commit is exactly the minimal change requested: `.github/workflows/ci.yml` (+2 -2) flips the npm-test entry from the deleted `node --test scripts/install.test.js` to `npm test`, which `npm/package.json` already wires to `prepare-packages.test.js`. CI on the new SHA reports `npm test` **pass**, and `build & test (1.24.x)` is green. My prior `REQUEST_CHANGES` blocker is resolved; dismissed.

(Only `check-sprint / check-sprint` is failing — that's a sprint-tracking workflow, not part of this PR's surface, and is not gating merge.)

The packaging design itself was sound from the first SHA; nothing further from me. Non-blocking follow-ups raised by other reviewers (tar extraction hardening, LICENSE in platform sub-packages, narrower argv validation) are worth a separate PR but not required here.

PR #58, opened 2026-06-25
fix(npm): harden bundled package publishing

### LINKED ISSUE #59: fix(npm): harden bundled package publishing
## Description
PR #56 moved `@mininglamp-oss/octo-cli` from install-time GitHub Release
downloads to bundled npm platform sub-packages. Follow-up review identified
several hardening items around the release/npm publishing flow and package
layout.

## Scope
- Make npm publishing wait for GitHub Release asset upload.
- Verify release archive checksums with exact filename matches.
- Distinguish npm registry 404s from transient `npm view` failures.
- Extract only the expected binary member from GoReleaser archives.
- Reject unknown or duplicate `prepare-packages.js` arguments.
- Include `LICENSE` in platform sub-packages.
- Add shim resolver tests for Windows `.exe` paths and missing platform packages.
- Document the npm platform-subpackage install model and Windows `.tar.gz`
  Release archives.

## Expected Outcome
The bundled npm package publishing path is more robust without changing the
runtime behavior introduced by #56.

## Related
- Follow-up to PR #56.

liuooo:
## What
Follow-up to #56. The bundled npm package flow is merged, but reviewers called
out several hardening items around publish ordering, checksum validation,
archive extraction, platform-package metadata, and shim test coverage.

## How
- **release-publish.yml** — makes `publish-npm` wait for `release-upload`, so an
  immutable npm version is not published before the matching GitHub Release
  assets and checksums are attached.
- **npm-publish.yml** — verifies checksums using an exact per-archive subset,
  adds retry/error handling around `npm view`, and restores terse security
  comments for the release/CI gates and publish ordering.
- **prepare-packages.js** — rejects unknown or duplicate arguments before
  touching `--out`, extracts only the expected binary member from each archive,
  and ships `LICENSE` in every platform sub-package.
- **octo-cli.js / prepare-packages.test.js** — makes the shim resolver
  injectable for tests, adds Windows `.exe` path coverage, and locks in the
  missing-platform-package error path.
- **README.md** — documents that npm installs use platform sub-packages and that
  GitHub Release archives are `.tar.gz` for all platforms, Windows included.

## Tests
- `npm --prefix npm test`
- `actionlint .github/workflows/ci.yml .github/workflows/npm-publish.yml .github/workflows/release-publish.yml`
- `goreleaser check`
- `git diff --check`
- Local npm tarball smoke: generated fake release archives, packed the main
  package plus the current platform sub-package, installed them from local
  tarballs with `--offline --omit=optional`, and verified `.bin/octo-cli`
  executes with mode `755`.

Follow-up to #56.

Closes #59.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -54,6 +54,8 @@ on:
         default: false
 
 concurrency:
+  # A package version is immutable on npm; never let two publish attempts for the
+  # same tag race each other.
   group: npm-publish-${{ inputs.tag }}
   cancel-in-progress: false
 
@@ -110,6 +112,8 @@ jobs:
           TAG: ${{ steps.version.outputs.TAG }}
         run: |
           set -euo pipefail
+          # Gate: a repo writer can create a Release outside this workflow; npm
+          # publishing still requires the expected non-draft Release tag.
           if [ -z "$TAG" ]; then
             echo "::error::Empty TAG output from resolver step."
             exit 1
@@ -137,6 +141,7 @@ jobs:
           TAG: ${{ steps.version.outputs.TAG }}
         run: |
           set -euo pipefail
+          # Gate: only publish tags whose target commit has a successful CI run.
           ref_data=$(gh api "repos/$GITHUB_REPOSITORY/git/ref/tags/$TAG")
           ref_type=$(printf '%s' "$ref_data" | jq -r '.object.type')
           ref_sha=$(printf '%s' "$ref_data" | jq -r '.object.sha')
@@ -183,10 +188,26 @@ jobs:
           set -euo pipefail
           shopt -s nullglob
           cd dist
-          sha256sum --check --ignore-missing checksums.txt
+          # Verify exactly the archives we are about to package. Do not use
+          # --ignore-missing: a missing or only substring-matched archive must
+          # fail before npm packages are generated.
+          subset=$(mktemp)
+          trap 'rm -f "$subset"' EXIT
+          archive_count=0
           for f in *.tar.gz; do
-            grep -qF "  $f" checksums.txt || { echo "::error::$f not covered by checksums.txt"; exit 1; }
+            archive_count=$((archive_count + 1))
+            line=$(awk -v file="$f" '$2 == file { print }' checksums.txt)
+            if [ -z "$line" ]; then
+              echo "::error::$f not covered exactly by checksums.txt"
+              exit 1
+            fi
+            printf '%s\n' "$line" >> "$subset"
           done
+          if [ "$archive_count" -eq 0 ]; then
+            echo "::error::No release archives (*.tar.gz) found in dist/."
+            exit 1
+          fi
+          sha256sum --check "$subset"
 
       - name: Generate npm packages
         env:
@@ -203,8 +224,48 @@ jobs:
           set -euo pipefail
           shopt -s nullglob
 
+          npm_view() {
+            local label="$1"
+            shift
+            local output=""
+            for attempt in 1 2 3; do
+              if output=$(npm view "$@" 2>&1); then
+                printf '%s\n' "$output"
+                return 0
+              fi
+              if printf '%s' "$output" | grep -Eiq '(^|[[:space:]])E404([[:space:]]|$)|404 Not Found|is not in this registry'; then
+                return 1
+              fi
+              if [ "$attempt" -eq 3 ]; then
+                echo "::error::npm view failed for $label after $attempt attempts." >&2
+                printf '%s\n' "$output" >&2
+                return 2
+              fi
+              echo "[npm-publish] npm view failed for $label (attempt $attempt/3); retrying" >&2
+              sleep $((attempt * 5))
+            done
+          }
+
+          version_exists() {
+            local name="$1"
+            if npm_view "$name@$VERSION" "$name@$VERSION" version >/dev/null; then
+              return 0
+            fi
+            local status=$?
+            if [ "$status" -eq 1 ]; then
+              return 1
+            fi
+            exit 1
+          }
+
           if [ "$DIST_TAG" = "latest" ]; then
-            reg_latest=$(npm view "@mininglamp-oss/octo-cli" dist-tags.latest 2>/dev/null || true)
+            reg_latest=""
+            if reg_latest=$(npm_view "@mininglamp-oss/octo-cli dist-tags.latest" "@mininglamp-oss/octo-cli" dist-tags.latest); then
+              :
+            else
+              status=$?
+              if [ "$status" -ne 1 ]; then exit 1; fi
+            fi
             if [ -n "$reg_latest" ] && [ "$reg_latest" != "$VERSION" ] && \
                [ "$(printf '%s\n%s\n' "$reg_latest" "$VERSION" | sort -V | tail -n1)" = "$reg_latest" ]; then
               DIST_TAG="release-$VERSION"
@@ -216,9 +277,11 @@ jobs:
             local dir="$1"
             local name
             name=$(node -p "require('./$dir/package.json').name")
-            if [ -z "$DRY_RUN" ] && npm view "$name@$VERSION" version >/dev/null 2>&1; then
-              echo "[npm-publish] $name@$VERSION already on the registry — skipping"
-              return 0
+            if [ -z "$DRY_RUN" ]; then
+              if version_exists "$name"; then
+                echo "[npm-publish] $name@$VERSION already on the registry — skipping"
+                return 0
+              fi
             fi
             echo "[npm-publish] publishing $name@$VERSION"
             if [ -n "$DRY_RUN" ]; then
@@ -228,6 +291,8 @@ jobs:
             fi
           }
 
+          # Publish platform packages first; the main package pins them via
+          # optionalDependencies and must be last.
           for dir in build/octo-cli-*; do
             publish_one "$dir"
           done
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -248,10 +248,12 @@ jobs:
 
           version_exists() {
             local name="$1"
+            local status
             if npm_view "$name@$VERSION" "$name@$VERSION" version >/dev/null; then
               return 0
+            else
+              status=$?
             fi
-            local status=$?
             if [ "$status" -eq 1 ]; then
               return 1
             fi
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -4,15 +4,16 @@
 #
 # Job graph:
 #
-#   publish ── build ──┬─ release-upload: gh release upload
-#                      └─ publish-npm: npm packages
+#   publish ── build ── release-upload: gh release upload ── publish-npm: npm packages
 #
 # Why this shape:
 #   * publish is the CI-evidence gate. Build intentionally waits for it so an
 #     arbitrary tag cannot run build steps before the gate passes.
-#   * the binaries are built EXACTLY ONCE; release-upload and publish-npm both
-#     consume the same `goreleaser-dist` artifact (npm no longer re-downloads
-#     from the Release).
+#   * the binaries are built EXACTLY ONCE; release-upload and publish-npm consume
+#     the same `goreleaser-dist` artifact (npm no longer re-downloads from the
+#     Release).
+#   * npm publish waits for release-upload, so immutable npm versions do not
+#     appear before the matching GitHub Release assets/checksums are attached.
 name: Release Publish
 
 on:
@@ -117,11 +118,12 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: gh release upload "$TAG" dist/*.tar.gz dist/checksums.txt --clobber
 
-  # Repack the SAME build artifact into npm packages and publish. Reuses
-  # npm-publish.yml via workflow_call (from_artifact: true) so there is no
-  # duplicate publish logic and no second download from the Release.
+  # Repack the SAME build artifact into npm packages after the Release assets
+  # are attached. Reuses npm-publish.yml via workflow_call (from_artifact: true)
+  # so there is no duplicate publish logic and no second download from the
+  # Release.
   publish-npm:
-    needs: [publish, build]
+    needs: [publish, build, release-upload]
     if: ${{ inputs.draft == false }}
     uses: ./.github/workflows/npm-publish.yml
     with:
--- a/README.md
+++ b/README.md
@@ -56,7 +56,8 @@ For Node-based agent runtimes (OpenClaw, etc.):
 npm install -g @mininglamp-oss/octo-cli
 ```
 
-A thin wrapper that downloads the matching prebuilt binary on install.
+The npm package resolves the matching platform sub-package, which already
+contains the prebuilt binary. Install does not download binaries from GitHub.
 
 ### Go install
 
@@ -76,12 +77,15 @@ Download the latest binary for your platform from
 [GitHub Releases](https://github.com/Mininglamp-OSS/octo-cli/releases):
 
 ```bash
-# Archives are named octo-cli_<version>_<os>_<arch>.{tar.gz,zip}; pick the one
-# for your platform and substitute <version> (e.g. 0.5.0):
+# Archives are named octo-cli_<version>_<os>_<arch>.tar.gz for every platform,
+# including Windows. Pick the one for your platform and substitute <version>
+# (e.g. 0.5.0):
 curl -LO https://github.com/Mininglamp-OSS/octo-cli/releases/download/v<version>/octo-cli_<version>_linux_amd64.tar.gz
 tar xzf octo-cli_<version>_linux_amd64.tar.gz
 sudo mv octo-cli /usr/local/bin/
 ```
+
+Windows release archives are `.tar.gz` as well; Windows 10+ includes `tar.exe`.
 
 ### install.sh
 
--- a/npm/bin/octo-cli.js
+++ b/npm/bin/octo-cli.js
@@ -16,47 +16,80 @@ const PLATFORM_PACKAGES = {
   "win32-x64": "@mininglamp-oss/octo-cli-win32-x64",
 };
 
-function resolveBinary() {
-  const key = `${process.platform}-${process.arch}`;
+function platformKey(platform, arch) {
+  return `${platform}-${arch}`;
+}
+
+function platformPackageFor(platform, arch) {
+  return PLATFORM_PACKAGES[platformKey(platform, arch)];
+}
+
+function binaryNameFor(platform) {
+  return platform === "win32" ? "octo-cli.exe" : "octo-cli";
+}
+
+function resolveBinary(platform = process.platform, arch = process.arch, resolver = require.resolve) {
+  const key = platformKey(platform, arch);
   const pkg = PLATFORM_PACKAGES[key];
   if (!pkg) {
-    console.error(`[octo-cli] no prebuilt binary for ${key}.`);
-    console.error(
+    const err = new Error(`[octo-cli] no prebuilt binary for ${key}.`);
+    err.detail =
       "[octo-cli] Prebuilt binaries cover darwin/linux/win32 on x64/arm64. " +
-        "Build from source instead: https://github.com/Mininglamp-OSS/octo-cli",
-    );
-    process.exit(1);
+      "Build from source instead: https://github.com/Mininglamp-OSS/octo-cli";
+    throw err;
   }
 
-  const binName = process.platform === "win32" ? "octo-cli.exe" : "octo-cli";
+  const binName = binaryNameFor(platform);
   try {
-    return require.resolve(`${pkg}/bin/${binName}`);
+    return resolver(`${pkg}/bin/${binName}`);
   } catch (_err) {
-    console.error(`[octo-cli] platform package ${pkg} is not installed.`);
-    console.error(
+    const err = new Error(`[octo-cli] platform package ${pkg} is not installed.`);
+    err.detail =
       "[octo-cli] npm skips optionalDependencies when installed with " +
-        "--no-optional / --omit=optional, and some package managers need a " +
-        "lockfile refresh after a platform change.\n" +
-        "[octo-cli] Try reinstalling: npm install -g @mininglamp-oss/octo-cli",
-    );
-    process.exit(1);
+      "--no-optional / --omit=optional, and some package managers need a " +
+      "lockfile refresh after a platform change.\n" +
+      "[octo-cli] Try reinstalling: npm install -g @mininglamp-oss/octo-cli";
+    throw err;
   }
 }
 
-const res = spawnSync(resolveBinary(), process.argv.slice(2), { stdio: "inherit" });
+function main() {
+  let bin;
+  try {
+    bin = resolveBinary();
+  } catch (err) {
+    console.error(err.message);
+    if (err.detail) console.error(err.detail);
+    process.exit(1);
+  }
 
-if (res.error) {
-  console.error(`[octo-cli] ${res.error.message}`);
-  process.exit(1);
+  const res = spawnSync(bin, process.argv.slice(2), { stdio: "inherit" });
+
+  if (res.error) {
+    console.error(`[octo-cli] ${res.error.message}`);
+    process.exit(1);
+  }
+
+  // Re-raise terminating signals so the shell observes the conventional
+  // 128+signum exit code; for default-ignored signals (SIGPIPE, ...) the
+  // explicit exit below sets the code instead.
+  if (res.signal) {
+    process.kill(process.pid, res.signal);
+    const signum = (os.constants && os.constants.signals && os.constants.signals[res.signal]) || 0;
+    process.exit(128 + signum);
+  }
+
+  process.exit(res.status === null ? 1 : res.status);
 }
 
-// Re-raise terminating signals so the shell observes the conventional
-// 128+signum exit code; for default-ignored signals (SIGPIPE, ...) the
-// explicit exit below sets the code instead.
-if (res.signal) {
-  process.kill(process.pid, res.signal);
-  const signum = (os.constants && os.constants.signals && os.constants.signals[res.signal]) || 0;
-  process.exit(128 + signum);
-}
+module.exports = {
+  PLATFORM_PACKAGES,
+  binaryNameFor,
+  platformKey,
+  platformPackageFor,
+  resolveBinary,
+};
 
-process.exit(res.status === null ? 1 : res.status);
+if (require.main === module) {
+  main();
+}
--- a/npm/scripts/prepare-packages.js
+++ b/npm/scripts/prepare-packages.js
@@ -33,6 +33,7 @@ const { execFileSync } = require("child_process");
 
 const SCOPE = "@mininglamp-oss";
 const PROJECT = "octo-cli";
+const ARG_KEYS = new Set(["version", "dist", "out"]);
 
 // goreleaser (os, arch) -> npm (os, cpu). Single source of truth for the
 // platform matrix on the npm side; the guard test asserts set-equality with
@@ -66,7 +67,10 @@ function parseArgs(argv) {
     const key = argv[i];
     const val = argv[i + 1];
     if (!key.startsWith("--") || val === undefined) fail(`bad argument: ${key}`);
-    args[key.slice(2)] = val;
+    const name = key.slice(2);
+    if (!ARG_KEYS.has(name)) fail(`unknown argument: ${key}`);
+    if (Object.prototype.hasOwnProperty.call(args, name)) fail(`duplicate argument: ${key}`);
+    args[name] = val;
   }
   for (const required of ["version", "dist", "out"]) {
     if (!args[required]) fail(`--${required} is required`);
@@ -80,6 +84,12 @@ function parseArgs(argv) {
 
 function writeJson(file, obj) {
   fs.writeFileSync(file, JSON.stringify(obj, null, 2) + "\n");
+}
+
+function copyProjectDoc(npmDir, destDir, doc) {
+  const src = path.join(npmDir, doc);
+  const fallback = path.join(npmDir, "..", doc);
+  fs.copyFileSync(fs.existsSync(src) ? src : fallback, path.join(destDir, doc));
 }
 
 function main() {
@@ -101,7 +111,7 @@ function main() {
     const binFile = binFileName(p.goOs);
     const workDir = path.join(outDir, `.extract-${p.goOs}-${p.goArch}`);
     fs.mkdirSync(workDir, { recursive: true });
-    execFileSync("tar", ["-xzf", archive, "-C", workDir]);
+    execFileSync("tar", ["-xzf", archive, "-C", workDir, binFile]);
     const extracted = path.join(workDir, binFile);
     if (!fs.existsSync(extracted)) fail(`archive ${archive} does not contain ${binFile}`);
     const extractedStat = fs.lstatSync(extracted);
@@ -114,6 +124,7 @@ function main() {
     fs.mkdirSync(path.join(pkgDir, "bin"), { recursive: true });
     fs.copyFileSync(extracted, path.join(pkgDir, "bin", binFile));
     fs.chmodSync(path.join(pkgDir, "bin", binFile), 0o755);
+    copyProjectDoc(npmDir, pkgDir, "LICENSE");
     fs.rmSync(workDir, { recursive: true, force: true });
 
     writeJson(path.join(pkgDir, "package.json"), {
@@ -122,7 +133,7 @@ function main() {
       description: `${PROJECT} prebuilt binary for ${p.npmOs}-${p.npmCpu}.`,
       os: [p.npmOs],
       cpu: [p.npmCpu],
-      files: [`bin/${binFile}`],
+      files: [`bin/${binFile}`, "LICENSE"],
       engines: { node: ">=18" },
       homepage: "https://github.com/Mininglamp-OSS/octo-cli#readme",
       repository: {
@@ -144,11 +155,7 @@ function main() {
     path.join(npmDir, "bin", "octo-cli.js"),
     path.join(mainDir, "bin", "octo-cli.js"),
   );
-  for (const doc of ["README.md", "LICENSE"]) {
-    const src = path.join(npmDir, doc);
-    const fallback = path.join(npmDir, "..", doc);
-    fs.copyFileSync(fs.existsSync(src) ? src : fallback, path.join(mainDir, doc));
-  }
+  for (const doc of ["README.md", "LICENSE"]) copyProjectDoc(npmDir, mainDir, doc);
 
   const manifest = JSON.parse(fs.readFileSync(path.join(npmDir, "package.json"), "utf8"));
   manifest.version = version;
@@ -159,7 +166,7 @@ function main() {
   console.log(`[prepare-packages] ${SCOPE}/${PROJECT}@${version} (main, ${PLATFORMS.length} platform deps)`);
 }
 
-module.exports = { PLATFORMS, binFileName };
+module.exports = { PLATFORMS, binFileName, parseArgs };
 
 if (require.main === module) {
   main();
--- a/npm/scripts/prepare-packages.test.js
+++ b/npm/scripts/prepare-packages.test.js
@@ -10,6 +10,7 @@ const { execFileSync, spawnSync } = require("child_process");
 const SCRIPT = path.join(__dirname, "prepare-packages.js");
 const SHIM = path.join(__dirname, "..", "bin", "octo-cli.js");
 const GORELEASER_YAML = path.join(__dirname, "..", "..", ".goreleaser.yaml");
+const NPM_PUBLISH_YAML = path.join(__dirname, "..", "..", ".github", "workflows", "npm-publish.yml");
 const { PLATFORMS, binFileName } = require(SCRIPT);
 const shim = require(SHIM);
 const VERSION = "9.9.9";
@@ -214,4 +215,78 @@ test("shim reports a missing platform package", (t) => {
   assert.strictEqual(res.status, 1);
   assert.match(res.stderr, new RegExp(`platform package ${scopedPkg} is not installed`));
   assert.match(res.stderr, /Try reinstalling: npm install -g @mininglamp-oss\/octo-cli/);
+});
+
+test("npm publish helper treats registry 404 as not published", () => {
+  if (process.platform === "win32") {
+    return;
+  }
+
+  const workflow = fs.readFileSync(NPM_PUBLISH_YAML, "utf8");
+  const start = workflow.indexOf("          npm_view() {\n");
+  const end = workflow.indexOf('\n          if [ "$DIST_TAG" = "latest" ]; then', start);
+  assert.notStrictEqual(start, -1, "cannot find npm_view helper in npm-publish.yml");
+  assert.notStrictEqual(end, -1, "cannot find helper terminator in npm-publish.yml");
+  const helpers = workflow.slice(start, end).replace(/^ {10}/gm, "");
+
+  const res = spawnSync(
+    "bash",
+    [
+      "-c",
+      `
+set -euo pipefail
+VERSION=9.9.9
+NPM_VIEW_MODE=exists
+
+npm() {
+  if [ "$1" != "view" ]; then
+    echo "unexpected npm command: $*" >&2
+    return 70
+  fi
+  case "$NPM_VIEW_MODE" in
+    exists)
+      printf '9.9.9\\n'
+      return 0
+      ;;
+    missing)
+      printf 'npm ERR! code E404\\n' >&2
+      printf 'npm ERR! 404 Not Found - GET https://registry.npmjs.org/%s - Not found\\n' "$2" >&2
+      return 1
+      ;;
+    hard)
+      printf 'npm ERR! network timeout\\n' >&2
+      return 1
+      ;;
+  esac
+}
+
+sleep() { :; }
+
+${helpers}
+
+NPM_VIEW_MODE=exists
+if ! version_exists "@mininglamp-oss/octo-cli-darwin-arm64"; then
+  echo "expected existing version to return 0" >&2
+  exit 1
+fi
+
+NPM_VIEW_MODE=missing
+if version_exists "@mininglamp-oss/octo-cli-darwin-arm64"; then
+  echo "expected missing version to return 1" >&2
+  exit 1
+fi
+
+NPM_VIEW_MODE=hard
+if (version_exists "@mininglamp-oss/octo-cli-darwin-arm64"); then
+  echo "expected hard npm view failure to exit non-zero" >&2
+  exit 1
+fi
+
+echo helper-ok
+`,
+    ],
+    { encoding: "utf8" },
+  );
+  assert.strictEqual(res.status, 0, res.stderr || res.stdout);
+  assert.match(res.stdout, /helper-ok/);
 });
--- a/npm/scripts/prepare-packages.test.js
+++ b/npm/scripts/prepare-packages.test.js
@@ -11,6 +11,7 @@ const SCRIPT = path.join(__dirname, "prepare-packages.js");
 const SHIM = path.join(__dirname, "..", "bin", "octo-cli.js");
 const GORELEASER_YAML = path.join(__dirname, "..", "..", ".goreleaser.yaml");
 const { PLATFORMS, binFileName } = require(SCRIPT);
+const shim = require(SHIM);
 const VERSION = "9.9.9";
 
 function makeDist(distDir) {
@@ -67,10 +68,11 @@ test("emits one package per platform and a pinned main package", (t) => {
     assert.strictEqual(manifest.version, VERSION);
     assert.deepStrictEqual(manifest.os, [npmOs]);
     assert.deepStrictEqual(manifest.cpu, [npmCpu]);
-    assert.deepStrictEqual(manifest.files, [`bin/${binFileName(goOs)}`]);
+    assert.deepStrictEqual(manifest.files, [`bin/${binFileName(goOs)}`, "LICENSE"]);
     const bin = path.join(pkgDir, "bin", binFileName(goOs));
     assert.ok(fs.existsSync(bin), `missing ${bin}`);
     assert.ok(fs.statSync(bin).mode & 0o100, `${bin} not executable`);
+    assert.ok(fs.existsSync(path.join(pkgDir, "LICENSE")), `missing ${pkgDir}/LICENSE`);
   }
 
   const main = JSON.parse(fs.readFileSync(path.join(outDir, "octo-cli", "package.json"), "utf8"));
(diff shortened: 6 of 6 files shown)


OctoBoooot (dismissed):
## Review: fix(npm): harden bundled package publishing (#58)

**Verdict:** Request changes.

CI is red on `check-sprint` — the project's process gate requires a `Closes #<issue>` reference in the PR description, and `Follow-up to #56` doesn't satisfy it (it points at a PR, not an issue). That's the only blocker. All code-related checks (`npm test`, `actionlint`, `osv-scanner`) are green, and the changes themselves are a clean follow-up to #56 that addresses every hardening note raised in that thread without expanding scope.

Risk tier: infra (`npm-publish.yml`, `release-publish.yml`). `needs-human-review` is already on the PR from earlier labeling.

### Blockers
- `check-sprint` is red because the PR body has no `Closes #<issue>` reference; the gate's message is `No linked issue found. Please add a 'Closes #<issue>' reference to the PR description, then re-run this check`. Either add the link (with a tracking issue on the Octo Board, since `#56` is a PR not an issue) or have a maintainer attach a board issue and re-run. The code itself is fine.

### Praise
- `npm/scripts/prepare-packages.js:114` — narrowing `tar -xzf` to the named binary member is the right shape. Combined with the existing `lstat().isFile()` regular-file gate, the unpack step now refuses both symlinks and stray archive contents being silently written into the workdir. Adversarial check: an archive with `octo-cli/` as a directory entry is rejected by the lstat gate; an archive missing the named member fails `existsSync`; an absolute / `..`-prefixed member won't land at `workDir/octo-cli` and trips the same gate. Defense holds.
- `.github/workflows/npm-publish.yml:188-210` — switching from `--ignore-missing` + substring `grep -qF "  $f"` to `awk '$2 == file'` + a per-archive `subset` file fixes both the "checksums.txt could be empty/missing and pass silently" hole and the `foo.tar.gz`-as-substring-of-`foo.tar.gz.sig` ambiguity. The `archive_count -eq 0` guard closes the empty-dist case too.
- `.github/workflows/npm-publish.yml:225-262` — `npm_view` with E404-vs-transient discrimination is correct: the `E404 / 404 Not Found / is not in this registry` patterns return 1 (no retry, treated as "not on registry"); everything else retries with backoff and finally exits 2 (treated as fatal by `version_exists`). The dist-tag `latest` check uses the same machinery, so a registry blip during release no longer accidentally promotes nothing or skips a publish.
- `.github/workflows/release-publish.yml:127` — making `publish-npm` `needs: [publish, build, release-upload]` closes the only remaining ordering hole from #56: an npm tarball can no longer be published for a tag whose GitHub Release assets haven't been attached yet. Since `release-upload` already needs `[publish, build]`, the explicit listing is belt-and-suspenders rather than strictly required, but it makes the graph self-documenting.
- `npm/bin/octo-cli.js:30-66` — refactor to thread `(platform, arch, resolver)` defaults through `resolveBinary` is exactly the minimal shape needed to make the shim unit-testable without a Windows runner; the new `shim resolver maps windows packages to the .exe binary` test pins the resolution rule with an injected resolver, and `shim reports a missing platform package` covers the error path end-to-end.

### Out of scope (informational)
- The `Verify CI evidence` step in `npm-publish.yml` is now structurally redundant when invoked from `release-publish.yml` (the reusable `publish` job already gates on CI evidence), but it's still needed for the manual `workflow_dispatch` path, so leaving it is fine.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-cli PR#58 Review Report
**PR:** Mininglamp-OSS/octo-cli#58
**Head SHA:** 76a9932452fd694b779d4e8cbed768cf723a6a10
**Title:** fix(npm): harden bundled package publishing
**Scope:** 6 files, +230/-65 — CI workflows, npm shim, packaging script, tests, README

---

## 1. Validation Summary

| Area | Status | Evidence |
|------|--------|----------|
| Workflow ordering (release→npm) | ✅ | `release-publish.yml:121` — `publish-npm` now `needs: [publish, build, release-upload]` |
| Checksum exact-match verification | ✅ | `npm-publish.yml:191-209` — awk `$2 == file` replaces grep substring match; no `--ignore-missing` |
| Tar extraction scoped to binary | ✅ | `prepare-packages.js:114` — `tar -xzf archive -C workDir binFile` (explicit member) |
| Argument allowlist in prepare-packages | ✅ | `prepare-packages.js:36,70-71` — `ARG_KEYS` Set + unknown/duplicate rejection before `--out` is touched |
| LICENSE in platform sub-packages | ✅ | `prepare-packages.js:127,136` — `copyProjectDoc` + `"LICENSE"` in `files` array |
| Shim refactor (throw vs exit, testability) | ✅ | `octo-cli.js:33-93` — `resolveBinary` now throws; `module.exports` + `require.main === module` guard |
| Test coverage additions | ✅ | `prepare-packages.test.js` — unknown/dup args, windows shim resolver, missing-platform error path |

## 2. Findings

### F1 — npm_view `local status=$?` pattern (P2 — nit)

- **File:** `.github/workflows/npm-publish.yml`, `version_exists()` function (~L253)
- **Diff-scope:** new (introduced by this PR)
- **Analysis:** `local status=$?` after `npm_view ...` in the `if` clause. In Bash, `local var=$?` preserves `$?` in the variable value (only `local var=$(cmd)` masks exit status via the `local` builtin's own return code). Verified correct: `status` captures `npm_view`'s return (0/1/2) and the `exit 1` on `status != 1` fires correctly. The pattern is sound, but a future reader could confuse it with the well-known `local var=$(cmd)` masking pitfall. Consider splitting into `local status; status=$?` for explicitness.
- **Severity:** P2 — readability/maintainability nit only, no runtime defect.

### F2 — Checksum subset file not quoted in sha256sum call (P2 — nit)

- **File:** `.github/workflows/npm-publish.yml`, ~L209
- **Diff-scope:** new (introduced by this PR)
- **Analysis:** `sha256sum --check "$subset"` — `$subset` is the output of `mktemp` (always a safe `/tmp/tmp.XXXXXX` path). Double-quoted, so no word-splitting. Safe as-is. Minor: the trap `rm -f "$subset"` also covers cleanup on early exit. No issue.
- **Severity:** P2 — no defect, noting for completeness.

### F3 — npm_view stdout trailing newline in reg_latest comparison (P2 — nit)

- **File:** `.github/workflows/npm-publish.yml`, ~L262
- **Diff-scope:** amplified (old code used `npm view ... 2>/dev/null`; new code wraps in `npm_view` which uses `printf '%s\n'`)
- **Analysis:** `npm_view` prints output via `printf '%s\n' "$output"`, adding a trailing newline. The caller captures with `reg_latest=$(npm_view ...)`, and command substitution strips trailing newlines. Net: `reg_latest` is identical to the old behavior. No defect.
- **Severity:** P2 — no defect.

**No P0 or P1 findings.**

## 3. Suggestions

1. **F1 readability**: Split `local status=$?` → `local status; status=$?` in `version_exists()` for future-proofing against the `local` masking anti-pattern. Low priority.
2. **Checksum subset concurrency**: The `mktemp` + `trap` pattern is clean. If the workflow ever runs matrix builds on the same runner, ensure unique temp names (already handled by `mktemp`). No action needed.

## 4. Extra Observations

- **goreleaser `files` config** (`.goreleaser.yaml:27`): Archives include `README.md` and `LICENSE*` alongside the binary. The new scoped tar extraction (`tar -xzf ... binFile`) correctly ignores these extra members — only the binary enters the npm package. README/LICENSE are separately copied via `copyProjectDoc`. Sound.
- **Platform matrix parity**: The PLATFORMS array in `prepare-packages.js` (6 entries: darwin/linux/win32 × x64/arm64) matches `.goreleaser.yaml` goos × goarch (3 × 2 = 6). The existing test `npm matrix matches .goreleaser.yaml goos x goarch` enforces this at CI time.
- **Publish ordering comment** (`npm-publish.yml:288`): Explicit comment "Publish platform packages first; the main package pins them via optionalDependencies and must be last" — good documentation of the invariant.

## 5. Data-Flow Tracing

| Consumed Data | Upstream Source | Flows Correctly? |
|---------------|----------------|-----------------|
| `checksums.txt` entries | GoReleaser `checksum.algorithm: sha256` → `checksums.txt` with `<hash>  <filename>` format | ✅ awk `$2` matches filename field; sha256sum `--check` verifies hash |
| `*.tar.gz` archives | GoReleaser `archives.formats: [tar.gz]` → `name_template` → downloaded to `dist/` | ✅ nullglob + archive_count gate ensures non-empty set |
| `VERSION` in publish step | `steps.version.outputs.VERSION` — parsed from `inputs.tag` with `v` prefix stripped | ✅ Regex validated: `^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?$` |
| `npm_view` return codes | 0=found, 1=404/not-in-registry, 2=transient-failure-after-3-attempts | ✅ `version_exists` maps: 0→skip, 1→publish, 2→abort. `reg_latest` block: 0→compare, 1→empty (first publish), ≠1→abort |
| `resolveBinary` in octo-cli.js | `PLATFORM_PACKAGES[platformKey(platform, arch)]` → `require.resolve(\`${pkg}/bin/${binName}\`)` | ✅ Unknown platform → throw Error with detail; missing package → throw Error with reinstall hint; both caught by `main()` |
| `binFile` in tar extraction | `binFileName(goOs)` → `"octo-cli"` or `"octo-cli.exe"` | ✅ Scoped extraction prevents extracting unexpected archive members (path-traversal hardening) |
| `ARG_KEYS` in parseArgs | `new Set(["version", "dist", "out"])` — allowlist checked before `--out` directory is used | ✅ Unknown args and duplicates rejected before any filesystem operation |

## 6. Blindspot Checklist (R5)

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths. The workflow changes are linear (release → build → upload → npm).
- **C2 — Control-flow ordering / nested reuse:** Checked. `npm_view` is called from two sites (reg_latest check + version_exists). Both correctly handle the 3-state return (0/1/2). `publish_one` reuses `version_exists` — ordering correct (check-then-publish). No double-action or sequence error.
- **C3 — Authorization boundary ≠ capability boundary:** Checked. `npm-publish.yml` permissions are `contents: read, actions: read, id-token: write`. NPM_TOKEN is a secret, gated by `DRY_RUN == ''`. CI evidence gate and Release verification gate are both in place. No new endpoint or capability exposed.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth/authorization changes to user-facing endpoints.
- **C5 — Build ≠ runtime path:** Checked. The tar extraction change uses GNU tar's explicit member extraction — verified correct syntax (`tar -xzf <archive> -C <dir> <member>`). The scoped extraction runs on the Linux CI runner where GNU tar is available. No browser extension / relative-path / origin concerns.
- **C6 — Governance / policy document consistency:** Checked. README.md changes (npm installs use platform sub-packages; Windows archives are `.tar.gz`) are consistent with the actual distribution model in the workflows and prepare-packages.js. No conflict with SECURITY.md or other governance docs.

## 7. Cross-Round Blocker Re-check (R6)

N/A — automated review review, no prior round.

---

**[Octo-Q] verdict: APPROVE**

All changes are security hardening improvements with solid test coverage. Checksum verification upgraded from substring to exact-match. Tar extraction scoped to prevent unexpected members. Argument parsing hardened with an allowlist. Workflow ordering ensures npm versions never appear before GitHub Release assets. Shim refactored for testability with new test cases covering Windows resolver and missing-package paths. No P0 or P1 findings; only minor readability nits (P2).

Jerry-Xin (dismissed):
Relevant to `Mininglamp-OSS/octo-cli`: this PR hardens the project's npm packaging/release flow and updates matching docs/tests. All five hardening items from #56 are addressed cleanly.

💬 Non-blocking
- 🔵 Suggestion: `npm/scripts/prepare-packages.test.js:108` says unknown/duplicate args are rejected "before touching output paths," but the test only asserts the error messages. Consider also asserting that neither `out` nor `other-out` directories exist after the failure, to lock in the safety property directly.

✅ Highlights
- Checksum verification in `.github/workflows/npm-publish.yml:186` now validates the exact archive subset via `awk '$2 == file'` (anchored, no substring match) and drops `--ignore-missing`, with a zero-archive guard.
- `npm view` handling in `.github/workflows/npm-publish.yml:224` correctly distinguishes E404 (not published) from transient failures (3x retry) from hard errors (abort), with anchored E404 matching — no real errors swallowed, no mis-skips.
- Publish ordering in `.github/workflows/release-publish.yml:125` makes `publish-npm` wait on `release-upload`, so an immutable npm version cannot appear before the matching Release assets/checksums; platform sub-packages still publish before the pinning main package.
- Archive extraction in `npm/scripts/prepare-packages.js:114` is now limited to the expected binary member (`tar -xzf ... binFile`) and retains the post-extraction regular-file check (path-traversal hardening).
- Arg parsing in `npm/scripts/prepare-packages.js:67` adds an allowlist plus duplicate detection before any output path work; covered by tests.
- LICENSE is now shipped in every platform sub-package (`files` array + copy), with a test asserting presence in `npm/scripts/prepare-packages.test.js`.
- Shim resolution in `npm/bin/octo-cli.js:31` is cleaner, injectable, and now test-covered including Windows `.exe` behavior.

Verification:
- `npm --prefix npm test` (clean) and `git diff --check` run during review.
- Independent CI check: `npm test`, `build & test (1.24.x)` pass. `check-sprint` fails on the Project Board "Sprint field" validation gate only — a process/metadata gate unrelated to this PR's code delta, not a code blocker.
- `actionlint`/`goreleaser check` were not run locally (tools not installed in review env); PR author reports both pass.

Jerry-Xin (changes requested):
🔴 Review REQUEST_CHANGES (updated) — two blockers now: one code, one process.

I previously approved this SHA; that was wrong on two counts. Correcting.

🔴 Blocking #1 (code — first-publish failure in `npm-publish.yml`)
`version_exists()` mis-reads the exit status after the `if` and aborts the whole publish job on a first publish. Walk-through under `set -euo pipefail`:
```
if npm_view "$name@$VERSION" ... version >/dev/null; then
  return 0
fi
local status=$?          # <-- $? here is 0, not npm_view's 1
if [ "$status" -eq 1 ]; then
  return 1               # never taken
fi
exit 1                   # taken -> aborts publish
```
When the package is not yet on the registry (E404), `npm_view` returns 1, the `if` condition is false and falls through — but at the fall-through `$?` is **0** (the `if` compound command itself succeeded; the failed-branch status is masked). So `status=0`, the `-eq 1` test is false, and control reaches `exit 1`, killing the `publish-npm` job. I reproduced this with a mock: rc=0 (exists) → skip ✅; rc=1 (E404 / first publish) → **script aborts** instead of publishing; rc=2 (hard error) → exit 1 ✅. Net: the very first publish of any platform package fails. Fix: capture the status before it's masked, e.g.
```
npm_view ... version >/dev/null && return 0
status=$?
[ "$status" -eq 1 ] && return 1
exit 1
```
(assign `status=$?` immediately after the call without the `if` wrapper, or `set +e` around the call). The same `status=$?`-after-`if` shape is used in the `dist-tags.latest` block (`reg_latest` branch) — please audit that one too. (Credit: yujiawei flagged this.)

🔴 Blocking #2 (process — `check-sprint` gate)
`check-sprint` CI fails: `No linked issue found. Please add a 'Closes #<issue>' reference to the PR description, then re-run this check.` The body only says "Follow-up to #56" (a mention, not a closing keyword). Add a `Closes #<issue>` line and re-run.

✅ The rest of the hardening is good and addresses the #56 🟡 notes: targeted tar extraction (only the expected binary member, path-traversal fix), strict checksum verification, the E404-vs-transient *intent* in `npm_view` (retry logic itself is fine — the bug is purely the caller's `$?` handling), sequential publish ordering, and the injectable shim resolver (Windows `.exe` testable). All code-relevant checks pass (`npm test`, `build & test`, `osv-scanner`, `actionlint`, CodeQL).

🟡 Add a regression test for `version_exists` covering the not-published (E404) path so this can't recur.

Net: fix the `version_exists` status capture (Blocking #1), add the `Closes #<issue>` link (Blocking #2), then I'll re-review and approve.

yujiawei (changes requested):
# Code Review — PR #58 (octo-cli)

**Head SHA reviewed:** `76a9932452fd694b779d4e8cbed768cf723a6a10`
**Base:** `main` (merge-base `4da7dc0`, follow-up to #56)
**Classification:** security-sensitive (release/publish pipeline)

## Verdict: Changes Requested

The hardening goals are sound and most of the diff is a clear improvement, but one change in `npm-publish.yml` introduces a release-breaking regression: the new `version_exists()` helper captures the wrong exit status, so the **first real publish of any new version aborts the job**. This must be fixed before merge. Details below.

---

## 1. Scope / intent check

All items described in the PR body are present, nothing out of scope was added:

- ✅ `release-publish.yml` — `publish-npm` now `needs: [publish, build, release-upload]`, so npm versions are not published before the matching Release assets/checksums are attached.
- ✅ `npm-publish.yml` — checksum verification switched to an exact per-archive subset (`awk '$2 == file'`, no `--ignore-missing`), plus an empty-archive guard; retry/error wrapper added around `npm view`.
- ✅ `prepare-packages.js` — unknown/duplicate args rejected before any output path is touched; `tar` now extracts only the expected binary member; `LICENSE` shipped in every platform sub-package.
- ✅ `octo-cli.js` / tests — resolver made injectable, Windows `.exe` path covered, missing-platform-package error path locked in.
- ✅ `README.md` — documents platform sub-packages and `.tar.gz` for all platforms incl. Windows.

The intent of every change is implemented and matches the goreleaser config (`formats: [tar.gz]`, `sha256`, name template). The problem is purely in the *implementation* of the `npm view` error handling.

---

## 2. Findings

### 🔴 P0 (blocking) — `version_exists()` captures the wrong exit status; first real release aborts

`.github/workflows/npm-publish.yml` lines **249–259**:

```bash
version_exists() {
  local name="$1"
  if npm_view "$name@$VERSION" "$name@$VERSION" version >/dev/null; then
    return 0
  fi
  local status=$?            # <-- BUG: this is the status of the *if* compound, not npm_view
  if [ "$status" -eq 1 ]; then
    return 1
  fi
  exit 1
}
```

When `npm_view` returns `1` (the E404 "version not in registry" path — i.e. the **normal** case for a version that has not been published yet), the `if … then` condition is false and there is no `else`, so the completed `if` compound exits `0`. `local status=$?` therefore captures **`0`, not `1`**. The guard `[ "$status" -eq 1 ]` is false, execution falls through to `exit 1`, and the entire **Publish to npm** step dies.

Because `publish_one()` calls `version_exists()` for the very first platform package, the consequence is: **every legitimate new release aborts on the first `octo-cli-*` package**, before publishing anything. The bug is gated behind `[ -z "$DRY_RUN" ]`, so it does not surface in `--dry-run` or in the PR's local tarball smoke test — only in a real production publish.

Reproduction (extracted verbatim from the workflow, `npm_view` returns 1 = E404):
```
npm_view returns 1:  (final step exit: 1)   # publish step dies instead of publishing the new version
npm_view returns 0:  -> EXISTS (skip)       # only the "already published" path works
npm_view returns 2:  (final step exit: 1)   # hard-error path happens to also exit 1
```

Note the sibling `reg_latest=$(npm_view …)` block at lines 261–268 does this **correctly** — it captures `$?` inside an `else` branch. Apply the same shape here:

```bash
version_exists() {
  local name="$1"
  if npm_view "$name@$VERSION" "$name@$VERSION" version >/dev/null; then
    return 0
  else
    local status=$?
    if [ "$status" -eq 1 ]; then
      return 1
    fi
    exit 1
  fi
}
```

(Verified locally: with the `else` branch, `npm_view`→1 yields "not-exists → publish", →0 yields "skip", →2 yields hard `exit 1`.)

### 🟡 P2 (non-blocking) — no regression coverage for the workflow shell helpers

`npm/scripts/prepare-packages.test.js` covers the JS resolver/prepare paths well (8 tests, all pass), but the `npm_view` / `version_exists` / publish-ordering logic lives only in workflow YAML and has no test. This exact class of bug (exit-status capture across an `if` boundary) is invisible to the JS suite and to dry-runs. Consider extracting the publish helper into a small sourced shell script with a `bats`/`node --test` harness, or at minimum a shell unit covering the E404 path, so this can't regress silently again.

### 🟢 Confirmed-good (no action)

- **Selective `tar` extraction** (`prepare-packages.js:114`): `tar -xzf <archive> -C <workDir> <binFile>` correctly extracts only the expected member; a missing member makes `tar` exit non-zero → `execFileSync` throws → fail-loud. Verified against a goreleaser-style archive.
- **Checksum subset verification** (`npm-publish.yml:197–210`): dropping `--ignore-missing` and matching `$2 == file` exactly (rather than `grep -F "  $f"` substring) closes a real gap; the empty-archive guard is a good belt-and-suspenders.
- **Publish ordering** (`release-publish.yml:126`): `release-upload` before `publish-npm` correctly prevents immutable npm versions from preceding the Release assets.
- **Shim refactor** (`octo-cli.js`): injectable resolver + throw-instead-of-exit is clean and the Windows `.exe` mapping is now tested.

---

## 3. Things worth a human's manual eye (security-sensitive)

- The whole point of this PR is publish-ordering and immutability safety; the P0 above means a real release will fail closed (abort), not fail open (mispublish) — so it is a **release-availability** bug, not a supply-chain exposure. Still must be fixed, but it does not weaken the security posture if a release simply can't complete.
- `--provenance --ignore-scripts --access public` on every `npm publish` is correct and preserved.
- No secrets are logged; `npm_view` merges stderr into a captured var and only echoes on the final failed attempt. Good.

---

## 4. Reviewer coverage note

- Independently traced the bash exit-status flow and reproduced it locally; ran the JS test suite (8/8 pass).
- A second independent reviewer pass (line-level) converged on the same `version_exists()` P0 finding — high confidence, two independent derivations.
- A long-context cross-file advisory pass timed out and did not return; it is recorded as **absent**, not as agreement. No silent "pass" is implied by its absence.
- Areas I did **not** execute end-to-end (not reachable from a PR checkout): the live `goreleaser release` build, the real `gh release upload`, and an actual npm registry publish. The P0 is in the publish step that only runs in those real flows.

OctoBoooot (changes requested):
## Review (corrected): fix(npm): harden bundled package publishing (#58)

**Verdict:** Request changes.

Retracting a wrong praise from my prior review and adding a real blocker that yujiawei caught — credit to them; I missed it on the first pass and even praised the broken code path.

### Blockers

- **`.github/workflows/npm-publish.yml:236-247`** — `version_exists` will abort the whole publish job on the very first publish of any platform package. The bash construct

  ```sh
  if npm_view "$name@$VERSION" "$name@$VERSION" version >/dev/null; then
    return 0
  fi
  local status=$?
  if [ "$status" -eq 1 ]; then
    return 1
  fi
  exit 1
  ```

  assumes `$?` after `if cmd; then ...; fi` is the condition's exit code. It isn't. Per the bash spec, an `if/then/fi` compound with a false condition and no `else` returns 0, so `$?` after the `fi` is 0 — not the 1 that `npm_view` returned on E404. `status` is therefore always 0 when the package isn't on the registry, the `-eq 1` branch is never taken, and control falls through to `exit 1`, killing the job.

  Reproduction (no GitHub Actions runner needed):
  ```sh
  set -euo pipefail
  fake_npm_view() { return 1; }
  version_exists() {
    if fake_npm_view; then return 0; fi
    local status=$?
    echo "captured status: $status"
    if [ "$status" -eq 1 ]; then return 1; fi
    exit 1
  }
  version_exists                  # prints "captured status: 0", then exits 1
  ```

  Impact: first publish of any new platform sub-package, and any retry where a sub-package wasn't yet uploaded, both fail. The idempotent skip-if-exists path that this function was supposed to enable is dead on arrival.

  Smallest fix: capture `$?` inside the body of the false branch (`else status=$?`), or inline the check without `local`. Sketch:
  ```sh
  npm_view ... >/dev/null
  status=$?
  case "$status" in
    0) return 0 ;;
    1) return 1 ;;
    *) exit 1 ;;
  esac
  ```
  Or use `||` so the exit code is captured directly:
  ```sh
  npm_view ... >/dev/null || status=$?
  ```

- **`check-sprint`** (carry-over from prior review) — PR description still has no `Closes #<issue>` reference; the process gate is red with `No linked issue found`. Add the link (or have a maintainer attach a board issue) and re-run.

### Retracted praise

- The earlier praise on `.github/workflows/npm-publish.yml:225-262` ("npm_view with E404-vs-transient discrimination is correct") was wrong **for the wrapper**. The retry/backoff inside `npm_view` itself is fine, but `version_exists` mishandles the return values, so the end-to-end behavior is broken. I should have constructed the failing input — a first-publish trace — instead of reading the helper in isolation. Lesson recorded.

### Praise (still stands)

- `npm/scripts/prepare-packages.js:114` — targeted `tar -xzf … binFile` plus the existing `lstat().isFile()` gate keeps the unpack step honest against symlink/path-traversal/missing-member archives.
- `.github/workflows/npm-publish.yml:188-210` — `awk '$2 == file'` + per-archive `subset` + `archive_count -eq 0` guard correctly closes the substring-match and empty-dist holes.
- `.github/workflows/release-publish.yml:127` — `publish-npm needs: [publish, build, release-upload]` correctly serializes npm publish behind GH Release upload.
- `npm/bin/octo-cli.js:30-66` — injectable `(platform, arch, resolver)` defaults make the shim unit-testable on a non-Windows runner; the new Windows `.exe` mapping test and missing-package test are well-scoped.

OctoBoooot (changes requested):
## Review (corrected): fix(npm): harden bundled package publishing (#58)

**Verdict:** Request changes.

Retracting a wrong praise from my prior review and adding a real blocker that yujiawei caught — credit to them; I missed it on the first pass and even praised the broken code path.

### Blockers

- **`.github/workflows/npm-publish.yml:236-247`** — `version_exists` will abort the whole publish job on the very first publish of any platform package. The bash construct

  ```sh
  if npm_view "$name@$VERSION" "$name@$VERSION" version >/dev/null; then
    return 0
  fi
  local status=$?
  if [ "$status" -eq 1 ]; then
    return 1
  fi
  exit 1
  ```

  assumes `$?` after `if cmd; then ...; fi` is the condition's exit code. It isn't. Per the bash spec, an `if/then/fi` compound with a false condition and no `else` returns 0, so `$?` after the `fi` is 0 — not the 1 that `npm_view` returned on E404. `status` is therefore always 0 when the package isn't on the registry, the `-eq 1` branch is never taken, and control falls through to `exit 1`, killing the job.

  Reproduction (no GitHub Actions runner needed):
  ```sh
  set -euo pipefail
  fake_npm_view() { return 1; }
  version_exists() {
    if fake_npm_view; then return 0; fi
    local status=$?
    echo "captured status: $status"
    if [ "$status" -eq 1 ]; then return 1; fi
    exit 1
  }
  version_exists                  # prints "captured status: 0", then exits 1
  ```

  Impact: first publish of any new platform sub-package, and any retry where a sub-package wasn't yet uploaded, both fail. The idempotent skip-if-exists path that this function was supposed to enable is dead on arrival.

  Smallest fix: capture `$?` inside the body of the false branch (`else status=$?`), or inline the check without `local`. Sketch:
  ```sh
  npm_view ... >/dev/null
  status=$?
  case "$status" in
    0) return 0 ;;
    1) return 1 ;;
    *) exit 1 ;;
  esac
  ```
  Or use `||` so the exit code is captured directly:
  ```sh
  npm_view ... >/dev/null || status=$?
  ```

- **`check-sprint`** (carry-over from prior review) — PR description still has no `Closes #<issue>` reference; the process gate is red with `No linked issue found`. Add the link (or have a maintainer attach a board issue) and re-run.

### Retracted praise

- The earlier praise on `.github/workflows/npm-publish.yml:225-262` ("npm_view with E404-vs-transient discrimination is correct") was wrong **for the wrapper**. The retry/backoff inside `npm_view` itself is fine, but `version_exists` mishandles the return values, so the end-to-end behavior is broken. I should have constructed the failing input — a first-publish trace — instead of reading the helper in isolation. Lesson recorded.

### Praise (still stands)

- `npm/scripts/prepare-packages.js:114` — targeted `tar -xzf … binFile` plus the existing `lstat().isFile()` gate keeps the unpack step honest against symlink/path-traversal/missing-member archives.
- `.github/workflows/npm-publish.yml:188-210` — `awk '$2 == file'` + per-archive `subset` + `archive_count -eq 0` guard correctly closes the substring-match and empty-dist holes.
- `.github/workflows/release-publish.yml:127` — `publish-npm needs: [publish, build, release-upload]` correctly serializes npm publish behind GH Release upload.
- `npm/bin/octo-cli.js:30-66` — injectable `(platform, arch, resolver)` defaults make the shim unit-testable on a non-Windows runner; the new Windows `.exe` mapping test and missing-package test are well-scoped.

Jerry-Xin (approved):
✅ APPROVE — re-reviewed on the new head `eea14b2c9e0a`. Both prior blockers are resolved on the code side.

🔴→✅ Blocking #1 (version_exists first-publish abort) — FIXED.
`npm-publish.yml` now captures the status in an explicit `else` branch:
```
if npm_view ... version >/dev/null; then
  return 0
else
  status=$?
fi
if [ "$status" -eq 1 ]; then return 1; fi
exit 1
```
`$?` is now read inside `else` where it reflects `npm_view`'s real exit code, not the masked `if`-compound status. I re-ran my mock against the new code: rc=0 → skip-publish ✅, rc=1 (E404 / first publish) → **publish ✅** (no more abort), rc=2 (hard error) → exit 1 ✅. A regression test was also added (`npm/scripts/prepare-packages.test.js:220` "npm publish helper treats registry 404 as not published") that exercises the E404 path directly. `npm test` is green. (Credit yujiawei for catching the original bug.)

🔴→⚠️ Blocking #2 (check-sprint) — author's part done; now a maintainer action.
The PR body now includes `Closes #59`, so the "no linked issue" error is gone. `check-sprint` is still red, but the reason changed to: `Linked issue has no Sprint set. Please ask a Maintainer to assign the linked issue to Sprint W26 on the Octo Board.` That is a maintainer-only step (assigning #59 to the sprint on org project board #2) — the author cannot resolve it from the PR. Nothing further is needed from the author here.

✅ Highlights: targeted tar extraction (path-traversal fix), strict checksum verification, E404-vs-transient handling in `npm view`, sequential publish ordering, injectable shim resolver (Windows `.exe` testable), and now a regression test for the publish-helper 404 path. All code-relevant checks pass (`npm test`, plus `osv-scanner`, `actionlint`, CodeQL from prior runs).

Net: code LGTM, approving. The remaining `check-sprint` red is a maintainer sprint-assignment on #59, not a code issue — a maintainer needs to set the Sprint field, then re-run the check before merge.

OctoBoooot (dismissed):
## Re-review on `eea14b2c9e0a` (delta)

**Verdict:** Request changes.

Per the harness's CI-red precondition the verdict can't be APPROVE yet, but the substantive picture is much closer to it: both blockers from my prior review are addressed, and the remaining red is no longer author-blocking.

### Blocker #1 — resolved ✓ (`version_exists` `$?` masking)

The fix is exactly the right shape:

```sh
if npm_view "$name@$VERSION" "$name@$VERSION" version >/dev/null; then
  return 0
else
  status=$?       # $? here is still the condition's exit code
fi
```

Re-traced all three paths to confirm:
- `npm_view` returns 0 (exists) → `return 0`. ✓
- `npm_view` returns 1 (E404 / first publish) → else runs, `status=1`, `-eq 1` matches, `return 1` → caller proceeds to publish. ✓ (this was the broken path)
- `npm_view` returns 2 (persistent failure) → else runs, `status=2`, falls through to `exit 1`. ✓

Symmetric check on the sibling `dist-tags.latest` block (`if reg_latest=$(npm_view …); then :; else status=$?; …`): the same shape applies — `status` is captured inside the else, so it carries the substitution's exit code correctly. No latent twin bug.

The new test in `prepare-packages.test.js` is particularly well-designed: it extracts the `npm_view` + `version_exists` helpers directly out of `npm-publish.yml` by string-slicing the workflow source, then runs them under bash with a mocked `npm` covering exists / E404 / hard-fail. That pins the contract against future regressions of the same shape and keeps the test in lockstep with the workflow without duplicating the code. The subshell wrapper `(version_exists …)` around the hard-fail assertion correctly isolates the `exit 1` from the test driver.

### Blocker #2 — partially resolved (`check-sprint`)

PR body now has `Closes #59` — the author-side fix is in. The gate has moved to its next state: `Linked issue has no Sprint set. Please ask a Maintainer to assign the linked issue to Sprint W26 on the Octo Board`. That's a maintainer-only action; nothing further from the author on this. Once #59 is assigned to Sprint W26 and `check-sprint` is re-run green, this PR is APPROVE-ready from my side — no further code review needed.

### Praise (new)

- `npm/scripts/prepare-packages.test.js:219-292` — the YAML-string-extract pattern for testing inline workflow bash is a clever way to avoid the "tested helper drifted from the deployed helper" failure mode. Cheaper than factoring the helper into a separate script and gives the same regression-guard signal.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-cli PR#58 Review Report
**PR:** https://github.com/Mininglamp-OSS/octo-cli/pull/58
**Head SHA:** `eea14b2c9e0affa9ee5cd6b11b426e35686ce9e7`
**Title:** fix(npm): harden bundled package publishing
**Changed files:** 6 (+307 / -65)
**Classification:** security_sensitive (automated review)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Publish ordering (release-upload → publish-npm) | ✅ | `release-publish.yml` L132: `needs: [publish, build, release-upload]` |
| Checksum exact-match validation | ✅ | `npm-publish.yml` L191–210: awk `$2 == file` + subset file + archive_count guard |
| Selective archive extraction | ✅ | `prepare-packages.js` L113: `tar -xzf archive -C workDir binFile` |
| Argument allowlist | ✅ | `prepare-packages.js` L69–71: `ARG_KEYS` Set + unknown/duplicate rejection |
| Shim refactor (testable, injectable) | ✅ | `octo-cli.js` L29: `resolver = require.resolve` default param; L80–87: `module.exports` |
| LICENSE in sub-packages | ✅ | `prepare-packages.js` L126: `copyProjectDoc(npmDir, pkgDir, "LICENSE")`; L136: `files` includes `"LICENSE"` |
| npm view retry + 404 detection | ✅ | `npm-publish.yml` L227–249: 3-attempt retry with E404/404 Not Found regex discrimination |
| Test coverage | ✅ | `prepare-packages.test.js`: 4 new tests (unknown/dup args, Windows .exe, missing pkg, npm helper) |
| README accuracy | ✅ | `README.md` L59: npm installs use platform sub-packages; L83–84: Windows .tar.gz documented |

## 2. Findings

### No P0 or P1 issues found.

### P2 — `npm_view` error message includes redundant arguments

- **File:** `.github/workflows/npm-publish.yml` L255–256
- **Diff-scope:** new (introduced by this PR)
- **Detail:** `version_exists` calls `npm_view "$name@$VERSION" "$name@$VERSION" version`. The first arg is the label, the rest are npm args. On hard failure, the error message becomes: `npm view failed for @mininglamp-oss/octo-cli@9.9.9 after 3 attempts.` — the label already contains the full specifier, so the `$@` expansion in the error path is redundant. Cosmetic only; no functional impact.
- **Suggestion:** Consider `npm_view "$name@$VERSION" "$name" version` for a cleaner label, or just accept the minor redundancy.

## 3. Recommendations

1. **(optional)** The test `"npm publish helper treats registry 404 as not published"` extracts bash helpers from the workflow YAML using `indexOf` markers. This is inherently brittle — any reformatting of the YAML or relocation of the helper block will silently break extraction. The test already has `assert.notStrictEqual(start, -1)` guards, which is good. Consider extracting the helpers to a separate `.sh` file in a future cleanup if the inline block grows.

## 4. Additional Observations

- The `release-publish.yml` job graph change from parallel (`release-upload ‖ publish-npm`) to serial (`release-upload → publish-npm`) adds latency to the release pipeline. This is an explicit trade-off documented in the comments (L14–15). Acceptable for correctness.
- The `workflow_dispatch` path of `npm-publish.yml` (manual rerun) still works independently without `release-upload` — it downloads archives from the GitHub Release directly. This is correct and intentional.
- The `NODE_AUTH_TOKEN` conditional (`steps.version.outputs.DRY_RUN == '' && secrets.NPM_TOKEN || ''`) correctly prevents token leakage in dry-run mode.

## 5. Data-Flow Tracing

### 5a. npm_view → version_exists → publish_one chain

**Upstream:** `npm view <pkg>@<version> version` (npm CLI, exit code semantics: 0=found, 1=error)

**Flow:**
1. `npm_view` captures stdout+stderr via `output=$(npm view "$@" 2>&1)`
2. On npm success → `printf '%s\n' "$output"` + `return 0`
3. On npm failure → regex checks stderr for E404/404 Not Found → `return 1` (not published) or retries → `return 2` (hard failure)

**Consumer:** `version_exists` calls `npm_view`, checks exit code:
- 0 → `return 0` (version exists → skip publish)
- 1 → `return 1` (not published → proceed)
- 2 → `exit 1` (abort entire script)

**Consumer:** `publish_one` calls `version_exists` inside `if`:
- return 0 → skip (already published)
- return 1 → proceed to publish
- exit 1 → terminates script (hard failure aborts entire publish step)

**Verdict:** Data flows correctly. All three states (exists/not-exists/hard-error) are properly discriminated. The `exit 1` on hard failure is intentional and tested.

### 5b. Checksum validation chain

**Upstream:** `checksums.txt` from goreleaser (format: `<sha256hex>  <filename>`)

**Flow:**
1. `for f in *.tar.gz` iterates all archives in `dist/`
2. `awk -v file="$f" '$2 == file { print }' checksums.txt` extracts exact match (field 2 = filename)
3. Matched lines appended to `$subset` temp file
4. `archive_count` guards against empty dist
5. `sha256sum --check "$subset"` verifies exactly the extracted subset

**Verdict:** Data flows correctly. No `--ignore-missing` means any archive not in checksums.txt fails (caught by the `awk` empty-line check). No substring matching — `$2 == file` is exact equality. Archive filenames are goreleaser-generated (no spaces, no special chars), so awk field splitting is safe.

### 5c. Selective tar extraction

**Upstream:** GoReleaser archives contain `octo-cli` (or `octo-cli.exe`) at archive root

**Flow:**
1. `execFileSync("tar", ["-xzf", archive, "-C", workDir, binFile])` — only extracts the named binary
2. `fs.existsSync(extracted)` verifies extraction
3. `fs.lstatSync(extracted)` + `isFile()` rejects symlinks/non-regular files
4. Binary copied to package dir, workDir cleaned up

**Verdict:** Data flows correctly. Selective extraction is strictly more secure than full extraction — prevents malicious archive members (symlinks, unexpected files) from reaching the npm package. The symlink guard still functions correctly with selective extraction.

### 5d. Argument parsing

**Upstream:** `process.argv.slice(2)` from CLI invocation

**Flow:**
1. `parseArgs` iterates pairs, validates each key starts with `--`
2. `ARG_KEYS.has(name)` rejects unknown arguments (allowlist: version, dist, out)
3. `hasOwnProperty` check rejects duplicates
4. Only after validation does `main()` proceed to `fs.rmSync(outDir, ...)`

**Verdict:** Data flows correctly. Validation runs before any filesystem mutation. The allowlist approach prevents injection of unexpected flags.

## 6. Blind-Spot Checklist (R5 — security_sensitive, all items mandatory)

**C1 — Dual-path parity:** Clear. All dual paths are covered:
- `resolveBinary` both error branches (unknown platform + missing package) are tested
- `publish_one` handles both already-published and not-published paths
- `version_exists` handles all three npm_view return codes
- `copyProjectDoc` fallback path (npmDir vs npmDir/..) used consistently for both sub-packages and main package

**C2 — Control-flow ordering / nesting reuse:** Clear. No double-application risk:
- `npm_view` retry loop has clean exit on each outcome (success/404/hard-fail)
- `set -euo pipefail` + conditional `if` contexts properly suppress premature exit
- `copyProjectDoc` is a simple copy with no ordering dependency
- `sha256sum --check` runs on a clean subset file, not piped from the loop

**C3 — Authorization boundary ≠ capability boundary:** N/A. No new endpoints or tools exposed. The workflow permissions remain minimal (`contents: read`, `actions: read`, `id-token: write`). NPM_TOKEN only flows to the publish step.

**C4 — Authorization lifecycle / container-member cascade:** N/A. No auth/authorization logic changes.

**C5 — Build/note ≠ runtime correctness:** Clear. This PR modifies CI workflows and npm packaging scripts:
- Workflow changes are validated by `actionlint` (per PR description)
- `prepare-packages.js` changes are validated by the test suite
- The shim refactor preserves identical runtime behavior (the `main()` function is gated by `require.main === module`; `module.exports` does not affect CLI execution)
- No relative-URL or base-path issues introduced

**C6 — Governance / policy / security document consistency:** N/A. No governance or policy documents modified. README changes are factual documentation updates (npm install mechanism, Windows archive format) consistent with the code changes.

## 7. Cross-Round Blocker Recheck (R6)

N/A — this is the first review of this PR.

---

**[Octo-Q] verdict: APPROVE**

All reviewer-requested hardening items from PR #56 are properly addressed. The checksum validation is materially more secure (exact match replaces substring match), the publish ordering prevents immutable-version-before-assets races, selective tar extraction reduces attack surface, and the shim refactor significantly improves testability without changing runtime behavior. Test coverage is thorough and covers the key error paths. No P0 or P1 issues found.

OctoBoooot (approved):
## Re-review on `eea14b2c9e0a` — APPROVE on code

**Verdict:** Approve, with one external gate to clear.

Code-side, both blockers from prior rounds are resolved on this SHA:

- **`version_exists` first-publish abort** — fixed by capturing `status=$?` inside the `else` branch where `$?` is still the condition's exit code. Re-traced all three paths (exists/E404/hard-fail) and they now route correctly; the sibling `dist-tags.latest` block uses the same `else status=$?` shape so no latent twin bug. The regression test (`npm publish helper treats registry 404 as not published`) extracts the helpers directly from `npm-publish.yml` and exercises them under bash with a mocked `npm` — that string-slice pattern keeps the test in lockstep with the deployed bash without duplicating it, which is a nice piece of work.
- **`check-sprint` linked-issue** — author-side is done (`Closes #59` added). The gate has advanced to its next state: `Linked issue has no Sprint set. Please ask a Maintainer to assign the linked issue to Sprint W26 on the Octo Board`. That's a maintainer-only action; nothing left for the author.

Dismissing my prior CR because the only remaining red is the Sprint-W26 assignment for #59, which is outside the author's control. GitHub `reviewDecision` will stay `CHANGES_REQUESTED` until that gate clears and the prior CR-on-old-SHA from peers is sorted, so this APPROVE doesn't enable a premature merge.

Once #59 is assigned to Sprint W26 and `check-sprint` re-runs green, this is ready to merge from my side.

yujiawei (approved):
# Code Review — PR #58 (octo-cli)

**Head SHA reviewed:** `eea14b2c9e0affa9ee5cd6b11b426e35686ce9e7`
**Base:** `main` (merge-base `4da7dc0`, follow-up to #56)
**Classification:** security-sensitive (release / npm publish pipeline, supply-chain surface)

This is a re-review on the current head `eea14b2`, which adds the `fix(npm): preserve registry lookup status` commit on top of the hardening commit. I reviewed against the current head with the correct merge-base and did not inherit any prior approval.

## 1. Spec compliance

**Spec: ✅**

The PR sets out five hardening items; all are implemented, and nothing out-of-scope was added.

- **Publish ordering** — ✅ `release-publish.yml`: `publish-npm` now `needs: [publish, build, release-upload]`, so an immutable npm version can no longer appear before the matching GitHub Release assets/checksums are attached. The job-graph comment was updated to match.
- **Checksum validation** — ✅ `npm-publish.yml`: `--ignore-missing` removed; each `*.tar.gz` in `dist/` is matched against `checksums.txt` by exact field (`awk '$2 == file'`), accumulated into a subset, and `sha256sum --check` runs against that subset. An empty-`dist/` guard (`archive_count == 0`) was added.
- **npm view retry/error handling** — ✅ `npm_view()` retries transient failures (3 attempts, backoff) and distinguishes a real registry 404 (`return 1`) from a hard failure (`return 2`); `version_exists()` and the `reg_latest` block consume those codes.
- **Archive extraction** — ✅ `prepare-packages.js` now extracts only the expected binary member (`tar -xzf archive -C workDir binFile`) and still rejects non-regular members via `lstatSync().isFile()`. `parseArgs()` rejects unknown and duplicate arguments before any output path is touched.
- **Platform-package metadata / LICENSE** — ✅ Every platform sub-package now ships `LICENSE` (`files: ["bin/<bin>", "LICENSE"]` + `copyProjectDoc`).
- **Shim test coverage** — ✅ `octo-cli.js` resolver is now injectable; Windows `.exe` path coverage and the missing-platform-package error path are locked in. README updated to reflect npm sub-package distribution and tar.gz-for-all-platforms.

No additions beyond the stated scope were found (no extra flags, fields, or interfaces).

## 2. Code quality

**Quality: Approved**

I verified the most failure-prone changes by hand, and cross-checked with an independent second-opinion reviewer:

- **`version_exists()` exit-status logic** (the prior round's blocker) is now correct. Traced under `set -euo pipefail`:
  - First publish (registry 404): `npm_view` returns 1 → captured in the `else` branch → `version_exists` returns 1 → publish proceeds. ✅
  - Already published: returns 0 → skip. ✅
  - Hard/transient failure after retries: returns 2 → `exit 1`, failing the job closed rather than silently publishing or skipping. ✅
  The `$?` masking bug from the earlier head is resolved by capturing status inside an explicit `else`.
- **Checksum subset matching** — confirmed an archive present in `dist/` but absent from (or only a substring of) `checksums.txt` produces no `awk` match → hard error before any npm package is generated. No fail-open path.
- **Selective tar extraction does not weaken the symlink defense** — confirmed that extracting a single named member that is a symlink still yields a symlink on disk, which `lstatSync().isFile()` rejects. Supply-chain posture is preserved.
- **Shim refactor preserves behavior** — error messages, exit code 1, and the signal re-raise (`128+signum`) path are unchanged; the `require.main === module` guard keeps `require()` side-effect-free.
- Packaging test suite passes locally (9/9).

### Non-blocking suggestions (P2 — not required to merge)

- **`npm/scripts/prepare-packages.test.js:141-167`** — the "rejects non-regular members" test uses a single-member archive, so it proves the `isFile()` guard but not that extraction is actually *selective*. Removing the `binFile` argument from the `tar` call (line ~114) would not be caught. Consider an archive with multiple members asserting only the binary lands in `workDir`.
- **`npm/scripts/prepare-packages.test.js:196`** — the Windows `.exe` test injects an identity resolver, so it validates string construction only, not real resolution. The non-Windows integration test covers `require.resolve`; a Windows-specific name typo would slip through. Minor, since names are hardcoded constants.
- **`npm/bin/octo-cli.js:85-91`** — the refactor exports `PLATFORM_PACKAGES`, `resolveBinary`, etc. from the shipped shim purely for tests. This becomes an accidental import surface. Consider a separate test-support module if you want to keep these internal. Functionally harmless today.

## 3. Overall verdict

**APPROVE**

Spec ✅ and no P0/P1 code-quality findings. The prior `version_exists` first-publish blocker is fixed and verified; checksum, ordering, and extraction changes all fail closed. The three remaining items are P2 test/maintainability suggestions and do not block merge.

## 4. Reviewer coverage note

- Verified by hand: workflow shell logic under `set -euo pipefail` (npm_view/version_exists/reg_latest), checksum subset matching, selective-extraction symlink behavior, shim behavior parity, local test run.
- An independent second-opinion pass over the same diff reported no blocking correctness, fail-open, ordering, checksum, or packaging issues and agreed the changes fail closed.
- A long-context whole-repo advisory pass was attempted but did not return in time; it is recorded as absent (no signal), not as agreement. Its absence does not affect this verdict, which rests on the verified analysis above.
- Not exercised at runtime: live npm registry behavior, GoReleaser archive production, and Windows-host execution of the produced binary — these depend on CI/release infrastructure outside the diff.
- Note: the only red status on the PR is the `check-sprint` process gate; all code checks (npm test, actionlint, osv-scanner, CodeQL, build & test) are green. The sprint gate is a process/linkage matter, not a code defect, and is outside the scope of this code review.

PR #71, opened 2026-07-09
docs: sync README with docs domain (8 domains, 71 operations)

boris-clark:
## Summary

The docs domain (`octo-docs` skill + 24 operations) landed in #68, but `README.md` was never updated and still describes the pre-docs surface. This PR syncs the README with the actual embedded specs and CLI surface. Docs-only change — no business logic touched.

## Verified counts

All numbers were verified against the command-registration source, not copied. Sources cross-checked:

- `internal/registry/specs/*.json` — counted unique `operationId`s per spec.
- `internal/registry/loader_test.go` (`TestAllDomainOperationCounts`, `TestListServices`) — the authoritative per-domain assertion; passes on `main`.
- Running binary registry (`schema --list`) and `skills/octo-docs/SKILL.md`.

| Domain | Ops |
|--------|-----|
| docs | 24 |
| matter | 14 |
| group | 9 |
| thread | 8 |
| bot | 6 |
| message | 4 |
| file | 4 |
| event | 2 |
| **Total** | **71 across 8 domains** |

The README's existing convention counts the withheld `matter` domain in both the domain count and the operation total (its previous "47 across 7" = the same sum without docs), so this keeps that convention and simply adds docs: **47 → 71 operations, 7 → 8 domains**. (Note: `schema --list` shows 57/7 because `matter` is withheld from the CLI command tree via `x-octo-disabled`, but the README documents the full embedded surface, matching how it already lists the withheld `matter` row.)

## Changes

- **Architecture blurb** — `47 operations across 7 domains` → `71 operations across 8 domains`.
- **Domains table** — add the `docs` row (24 ops), ordered by op count.
- **Quick Start** — add representative docs commands (`create`, `list`, `get`, `content get`, `members set`, `comments add`), each verified against the actual command help.
- **Agent Skills** — list `octo-docs` (it is active and returned by `octo-cli skills`, unlike the withheld `octo-matter`).

## Notes

Opened as a **draft** intentionally — holding for review before it is marked ready.

While verifying, I noticed `CLAUDE.md`'s "Command Structure" line reads "8 domains, 69 operations" and its command tree omits `docs content get|edit` (2 ops that do exist). I left `CLAUDE.md` out of this PR to keep it README-focused; happy to fix it here or in a follow-up if preferred.

PR #68, opened 2026-07-08
Add spec-driven docs commands and octo-docs skill

boris-clark:
## What

Adds a spec-driven `docs` service and an `octo-docs` skill so the merged bot docs API (the surface under `/v1/bot/docs`) is drivable from the CLI. No hand-written command code — the subtree auto-generates from the dotted operationIds in the embedded spec. This includes the doc-body content endpoints: reading a document's live body and applying incremental block-op edits.

## Scope

24 operations across:

- **lifecycle** — `create`, `list`, `get`, `rename`, `delete`
- **body content** — `content get` (read live body + base version), `content edit` (incremental block-op edit under an If-Match base-version guard)
- **members & sharing** — `members list|set|remove`, `forward-grant`
- **comments** — `comments list|add|edit|delete`
- **versions** — `versions list|create|state|rename|delete|restore`
- **attachments (metadata only)** — `attachments presign|get|resolve`

## Doc-body content editing

`docs.content.get` / `docs.content.edit` map to `GET` / `PATCH /v1/bot/docs/{docId}/content`, verified against `octo-docs-backend@main` (`src/api/routes/docContent.ts`, `src/collab/docBodyEdit.ts`, `src/api/services/editDocBody.ts`):

- `content get` (reader) returns `{ docId, doc, schemaVersion, baseVersion }` — the live ProseMirror body plus the base-version token that pins the read.
- `content edit` (writer) applies a batch of `insert`/`replace`/`delete` block ops (the `DocEditOp` union), addressed by block path, under a mandatory optimistic-concurrency guard. The ops batch is passed through the generic `--data` escape hatch; the base version is passed via `--base-version` and sent as the `If-Match` header. A stale token is rejected with `412 base_version_stale`. Only `doc_type: doc` is accepted (whiteboard/board return `409 unsupported_doc_type`).

### Spec-declared header parameters (general engine feature)

Rather than a docs-only `If-Match` carve-out, the request engine now binds any spec parameter declared with `"in": "header"` to a CLI flag and emits it as a per-request header (the transport already supported `Request.Headers`). An optional `x-octo-flag` alias on a parameter lets an awkward wire name like `If-Match` surface as a clean first-class flag (`--base-version`):

- `loader.go` reads `x-octo-flag` into `ParamInfo.FlagName`.
- `flags.go` `registerHeaderFlags` binds `"in":"header"` params; it runs before body-flag registration, and the body-flag collision guard now excludes header-flag names so an alias can't be shadowed by a same-named body field.
[...]

PR #73, opened 2026-07-09
Add docs sheet get/edit commands with paged reads

boris-clark:
## What

Adds the spreadsheet content surface to the `docs` domain, mirroring the existing `docs content get/edit` shape:

- **`docs sheet get <docId>`** — read the live cell map + dims + `baseVersion`.
- **`docs sheet edit <docId> --base-version <token> --data '<cells JSON>'`** — batch cell set/delete under the `If-Match` optimistic-concurrency guard. The cells batch is a JSON object `{ "sheetId!row:col": {v,f,s} | null }` (a `null` value deletes the cell).

Part of octo-cli#69 Stage3 (A). Depends on the backend paginated-read protocol added in octo-docs-backend#42 — the two were designed together so the CLI pagination flags match the wire protocol exactly.

## Reading a large sheet in pages

A whole-sheet `docs sheet get` of a grid over the backend's ~1MB read cap returns `413 sheet_too_large`. Passing **`--limit <n>`** opts into paged reads: each response carries `hasMore` + an opaque `nextCursor`; follow it via **`--cursor`** until `hasMore` is false. `sheetDims` is returned on the first page only. A `--cursor` whose snapshot moved returns `409 sheet_changed` (restart from page one for a consistent snapshot).

## Spec-driven, no command code

These are **spec-only** additions to `internal/registry/specs/docs.json`. The command tree, `--limit`/`--cursor` query flags, and the `--base-version → If-Match` header wiring all generate from the metadata — no hand-written Go, same as the rest of the domain.

## Error mapping

`409` / `412` / `422` now map to clean CLI codes (`CONFLICT` / `PRECONDITION_FAILED` / `UNPROCESSABLE_ENTITY`) with actionable hints, so the backend's `412 base_version_stale`, `409 sheet_changed` / `unsupported_doc_type`, and `422 sheet_snapshot_invalid` surface clearly instead of as bare `HTTP_4xx`.

## Docs & tests

- README domain table + examples and the `octo-docs` skill gain a spreadsheet section (incl. the paged-read loop).
- Operation-count expectations updated (docs `24 → 26`, total `71 → 73`).
- New functional tests: whole-sheet GET (no pagination query), paged GET (`--limit`/`--cursor` reach the wire), edit sends the cells batch + `If-Match` and requires `--base-version`.

`go build`, `go vet`, `gofmt`, and `go test ./...` all clean.


OctoBoooot (commented):
## Review: Add docs sheet get/edit commands with paged reads (#73)

**Verdict:** Approve-quality with one **minor** — holding at COMMENT because the PR is marked **draft** (and the minor is worth folding in first). No red CI (`build & test`/`npm test` are *skipping* on the draft, not failing; `osv-scan` passes).

Adds `docs.sheet.get` / `docs.sheet.edit` (24→26 ops) plus a taxonomy extension for three HTTP statuses. First substantive touch to `internal/output` in this series — reviewed it closely. Well-executed: the sheet edit reuses the `If-Match`/`x-octo-flag: base-version` header mechanism from #68, and `docs.sheet.get` correctly declares cursor pagination without hacking the `--page-all` engine (whole-sheet read omits `limit`/`cursor`, paged read passes them).

### Minor

- **`internal/output/errors.go:108-110,199-204`** — the three new status mappings (409→CONFLICT, 412→PRECONDITION_FAILED, 422→UNPROCESSABLE_ENTITY, all →`validation`) are added to both `backendErrorMapping` and `codeFromStatus`, but the existing table test `TestParseBackendError_EmptyBody_StatusInference` (`errors_test.go:162-171`) wasn't extended to cover them — it still lists only 401/403/404/413/429/500/502/503/504/418. These three are exactly the operationally-relevant codes for the sheet/content edit path (412 = stale base-version, the optimistic-concurrency reject; 409 = conflict; 422 = semantic), so they deserve a row each. Three lines: `{409,"validation","CONFLICT"}`, `{412,"validation","PRECONDITION_FAILED"}`, `{422,"validation","UNPROCESSABLE_ENTITY"}`. The mappings themselves are correct — verified both tables resolve for each status.

### Praise

- `docs.sheet.get` pagination modeling is the right call: cursor-based (`limit`/`cursor` → `{hasMore, nextCursor}`) declared as plain query params rather than forced through the `{data, pagination}` `--page-all` engine, consistent with the deliberate no-`--page-all` decision for the other docs list ops. `TestDocsSheetGet_WholeSheet` asserts a bare read sends *no* `limit`/`cursor` (so it can't accidentally trigger the 413 a whole-sheet read would hit), and `TestDocsSheetGet_Paginated` asserts both reach the wire — the two modes are pinned.
- `TestDocsSheetEdit_SendsCellsBatchAndIfMatch` + `TestDocsSheetEdit_RequiresBaseVersion` mirror the content-edit test pattern exactly: cells batch via `--data`, base-version as the `If-Match` header (not body), required-flag guard fails before any HTTP call. Consistent optimistic-concurrency contract across `content` and `sheet`.
- Extending the taxonomy so 412 carries `base version stale; re-read to get the current base version, then retry` gives the agent an actionable recovery hint that matches the sheet/content edit flow — good operability.

### Out of scope (informational)

- `CLAUDE.md:25` count (`71 operations / 74 commands`) will need bumping to `73 / 76` once this lands — same doc-sync class from #71. Not this PR's diff (it updates README + loader_test correctly), just flagging so the count trail stays consistent.

Flip to non-draft with the three error-table rows added and this is an APPROVE.

Jerry-Xin (dismissed):
The PR is in scope for `octo-cli`, but the new error-mapping behavior does not actually cover the backend codes described by the PR.

**🔴 Blocking**
- 🔴 Critical — `internal/output/errors.go:125` preserves backend envelope codes before considering HTTP status, but the new mappings at `internal/output/errors.go:108` only add status-derived codes like `CONFLICT`, `PRECONDITION_FAILED`, and `UNPROCESSABLE_ENTITY`. If the backend returns the documented envelope codes such as `base_version_stale`, `sheet_changed`, `unsupported_doc_type`, or `sheet_snapshot_invalid`, they bypass the new mappings and fall back to `api_error` with no hint. This contradicts the PR’s stated error-mapping goal. Add direct mappings/tests for those backend codes, or make unknown envelope codes fall back through the recognized HTTP status mapping.

**💬 Non-blocking**
- 🟡 Warning — `skills/octo-docs/SKILL.md:140` documents `--output json`, but the CLI global flag is `--format`, so the paged-read loop fails with `unknown flag: --output`. Use `--format json` or omit it since JSON is already the default.

**✅ Highlights**
- The spec-only command additions align with the existing generated command architecture.
- The new service tests cover command-tree registration, query pagination flags, request body shape, and `If-Match` header wiring.
- `go test ./...` passes locally.

ploy-elison (approved):
**APPROVED**

Reviewed the head branch (`feat/docs-sheet-commands`) diff and the generated command shape via the functional tests.

### What holds up

- **Endpoint wiring.** `docs.sheet.get` → `GET /v1/bot/docs/{docId}/sheet`, `docs.sheet.edit` → `PATCH .../sheet`, both spec-driven with no hand-written Go — consistent with the rest of the metadata-driven `docs` domain. `TestDocs_RegistryShape` / `TestDocs_TreeShape` and the `24→26` / `71→73` op-count expectations are updated in lockstep.
- **Pagination flags match the backend protocol.** `--limit`/`--cursor` reach the wire as query params (`TestDocsSheetGet_Paginated`), and with neither flag set no `limit`/`cursor` appears on the wire — the whole-sheet read (`TestDocsSheetGet_WholeSheet`). This matches PR#42's opt-in contract exactly.
- **Guarded write.** `--base-version` is wired to the `If-Match` header (not duplicated into the JSON body), the cells batch travels via `--data`, an explicit `null` value round-trips as a delete, and `--base-version` is required — the command fails before any request when it is missing (`TestDocsSheetEdit_RequiresBaseVersion`), mirroring the backend's mandatory optimistic-concurrency guard.
- **Error passthrough.** `409/412/422` now map to `CONFLICT`/`PRECONDITION_FAILED`/`UNPROCESSABLE_ENTITY` with actionable hints. The `412` hint ("re-read to get the current base version, then retry") reaches the user even for the backend's flat `{"error":"base_version_stale"}` body via the `hintFromStatus` fallback, and the specific 409 string (`sheet_changed` vs `unsupported_doc_type`) is preserved in the raw message/detail. Clear enough to act on.
- **Docs.** README table/examples and the `octo-docs` SKILL gain a spreadsheet section with a correct paged-read loop; section renumbering is consistent.

### One non-blocking note (not a blocker)

- `internal/output/errors.go`: for a **read** `413 sheet_too_large`, the static hint is `"request body exceeds 1MB limit"`, which is write/request-oriented — the actual remedy on the read path is `--limit`/`--cursor`. In practice the backend's own `hint: "retry with ?limit=<n>"` surfaces in the raw message, so a caller still sees the right guidance; the static CLI hint is just slightly off for this one code. Cosmetic only.

No blocking issues. Matches the backend wire protocol, well-tested, English, no AI attribution.

OctoBoooot (commented):
## On Jerry-Xin's two findings (#73 @d73450ae7778)

Traced both through `ParseBackendError`. One confirmed minor, one legitimate question (not a confirmed 🔴).

### 🟡 #2 — CONFIRMED (minor): `--output json` is not a real flag

`SKILL.md:435` uses `octo-cli docs sheet get … --output json`. The registered global output flag is **`--format`** (`cmd/root.go:40`: `json|table|csv|ndjson`), plus `--jq`. There is no `--output` flag, so an agent copy-pasting that line gets `unknown flag: --output`. Fix: `--format json`. Same class as the `--resolved true` doc bug from #68 — embedded agent docs that fail on paste. Minor (docs-only), real.

### #1 — legitimate QUESTION, severity unverifiable from this repo

Jerry-Xin is pointing at something real about *when* the new mappings fire. `ParseBackendError` resolves in layers:
- **Layer 1 (matters envelope `{"error":{"code":"…"}}`):** looks up `backendErrorMapping[envelope.code]` directly. The status→code path is NOT consulted here.
- **Layer 2 (dmworkim `{"msg","status"}`) + fallback (empty/raw body):** uses `codeFromStatus(status)`, so the new 409/412/422 → CONFLICT/PRECONDITION_FAILED/UNPROCESSABLE_ENTITY mappings *do* fire.

So the new mappings help the dmworkim/empty-body paths. Whether they help the **docs edit error path** depends on the docs backend's envelope shape, which I can't see from this repo:
- If docs returns a matters-style envelope with a **generic** code (`CONFLICT`/`PRECONDITION_FAILED`/`UNPROCESSABLE_ENTITY`) → the new mappings fire perfectly. ✓
- If docs returns **domain-specific** codes (`base_version_stale`, `sheet_changed`, `unsupported_doc_type`, `sheet_snapshot_invalid` — the ones the spec prose names) → Layer 1 looks those up, misses (they're not keys), and falls back to `api_error` + no hint. The friendly "base version stale; re-read…" hint the PR intends for the 412 flow would then never show.

Note every *existing* `backendErrorMapping` key is UPPER_SNAKE and domain-specific where needed (`MATTER_NOT_FOUND`, `ASSIGNEE_NOT_FOUND`, `DUPLICATE_ASSIGNEE`), which suggests the matters backend returns specific UPPER codes in the envelope. If docs follows that convention, the effective keys would be `BASE_VERSION_STALE` / `SHEET_CHANGED` / etc., not the generic three this PR added.

**So: is this a 🔴?** Not a confirmed one — it's a question the author/backend can settle in one line: *what `error.code` values does the docs bot mount actually return for 409/412/422?* If generic, the PR is correct as-is. If domain-specific, the mappings should key on those codes (or add them alongside) so the hint fires on the real envelope. Either way the change doesn't make anything *worse* (worst case the new keys are inert for docs and you get today's behavior), so it's not a regression blocker — but it may not deliver the operability win it's aiming for.

### My own finding stands (minor)

Separately from the above: the three new status rows aren't covered by `TestParseBackendError_EmptyBody_StatusInference` (`errors_test.go:162`). Adding `{409,…CONFLICT}`/`{412,…PRECONDITION_FAILED}`/`{422,…UNPROCESSABLE_ENTITY}` there would at least lock the status-inference path — and writing that test would surface the Layer-1-vs-status question above concretely.

### Net

No confirmed blocker. Two confirmed minors (`--output`→`--format`; missing status-table rows) + one question (docs envelope codes → does the friendly hint actually fire on the edit path?). Verdict stays **COMMENT** (PR is draft). If the author confirms docs returns generic 409/412/422 envelope codes, #1 closes; if domain-specific, key the mappings on those.

yujiawei (dismissed):
# Code Review — PR #73 (octo-cli)

**Verdict: APPROVED** — no blocking (P0/P1) issues. Three non-blocking test-coverage suggestions (P2) below.

Reviewed at head SHA `d73450ae7778f8444bff16b2e94dd9344f24a0f1`, merge-base `0f8a944` (current `main` HEAD — no sibling-PR conflict). `go build`, `go vet`, `gofmt -l`, and `go test ./...` all clean locally.

## Spec compliance ✅

The change is spec-only (`docs.json`) plus the error map, docs, and tests — no hand-written Go command code, consistent with the metadata-driven engine. I traced the full generation path (`loader.go` → `flags.registerHeaderFlags` → `run.buildHeaders` → `client` `Header.Set`) and confirmed at runtime via `--dry-run`:

- `docs sheet get` → `GET /v1/bot/docs/{docId}/sheet`, risk `read`; `docs sheet edit` → `PATCH …/sheet`, risk `write`. Both register correctly.
- `--base-version` is **required** and reaches the wire as the `If-Match` header, **not** duplicated into the JSON body. Omitting it fails before any request is sent.
- `--limit`/`--cursor` reach the wire as query params; a whole-sheet `get` (neither flag) sends no pagination query.
- `sheet.get` correctly does **not** declare `x-octo-pagination`, so the generic `--page-all` walker (which expects a `{data, pagination}` envelope) is not wired to this endpoint — avoiding a shape mismatch with the top-level `hasMore`/`nextCursor` response.
- Operation counts are correct: `docs` 24→26, total 71→73 (verified against all eight specs).

No missing, extra, or divergent behavior vs. the PR description. This mirrors the established `docs.content.get/edit` pattern exactly.

## Code quality

**Security (security-sensitive path).** The `If-Match`/base-version optimistic-concurrency wiring is sound: the token travels only in the header, is emitted only when the flag is set, and the generic header mechanism is already in production for `content.edit`. No token leakage into body or logs (dry-run redacts `Authorization`). The new `409/412/422` mappings classify as `validation` (client-recoverable), which is correct — a stale base version (`412 base_version_stale`) surfaces as `PRECONDITION_FAILED` with an actionable re-read hint. No auth/space-header regression.

**Correctness.** The error map is internally consistent: each new code has both a `codeFromStatus` switch case and a `backendErrorMapping` table entry (type + hint), and `typeFromStatus` routes 409/412/422 to `validation` via the 4xx branch, agreeing with the table (no drift). The dmworkim `{msg,status}` layer resolves the new codes via status; the matters `{error:{code}}` layer resolves them by canonical code — both established conventions.

### Non-blocking suggestions (P2)

1. **`internal/output/errors_test.go:68-106`** — `TestParseBackendError_MattersCodes` iterates a hard-coded `cases` slice ending at `PAYLOAD_TOO_LARGE`, **not** `backendErrorMapping` itself. The three new codes (`CONFLICT`, `PRECONDITION_FAILED`, `UNPROCESSABLE_ENTITY`) are therefore unasserted; a mistyped or removed new entry would not fail this test, and the test's doc comment ("exercises every entry in the backend mapping table") is now inaccurate. Consider ranging over `backendErrorMapping` directly so the table and its test cannot drift, or add the three codes to the slice.

2. **`internal/output/errors_test.go:156-172`** — `TestParseBackendError_EmptyBody_StatusInference` covers 401/403/404/413/429/5xx/418 but not 409/412/422, so the new `codeFromStatus` cases on the empty/unparseable-body fallback path are untested. A bare `412` with no JSON body must still yield `PRECONDITION_FAILED`/`validation` for a client to detect a stale base version; adding those three status cases would lock that in.

3. **`cmd/service/docs_test.go:656-692`** — the new functional tests cover the 200 happy path only. Since the whole point of the new error mappings on a concurrency-sensitive edit is that callers branch on the returned code to re-read and retry, one end-to-end test where the stub returns `412`/`409` and the CLI surfaces `PRECONDITION_FAILED`/`CONFLICT` would assert the highest-value contract for this endpoint.

None of the above blocks merge — the mappings and wiring are correct as written; these are coverage hardening for a security-sensitive path.

### For human verification

Given the `needs-human-review` label, one item worth a human eye: this CLI change was co-designed with the backend paginated-read protocol (octo-docs-backend#42). The CLI faithfully sends `--limit`/`--cursor` as query params and `--base-version` as `If-Match`; the actual wire contract (413 caps, cursor snapshot semantics, `sheet_changed` on a moved cursor) lives on the backend and is out of scope for this repo's tests. Worth confirming the two were merged/deployed together so the flags match the live protocol.

boris-clark:
Pushed 6aee6d7 addressing the two confirmed minors, plus an answer to the envelope-code question.

**Minors fixed**
- `errors_test.go`: added the 409/412/422 rows to the `TestParseBackendError_EmptyBody_StatusInference` table so the CONFLICT / PRECONDITION_FAILED / UNPROCESSABLE_ENTITY fallback path is locked. `go test ./internal/output/...` is green (full `go test ./...` also passes).
- `skills/octo-docs/SKILL.md`: `--output json` → `--format json` in both examples (the `docs content get` base-version snippet and the paged `docs sheet get` loop). `--format` is the registered global flag (`cmd/root.go`).

**Envelope-code question — docs backend does NOT use `error.code`**

I checked octo-docs-backend source. Every route serializes errors as a **flat** `{"error":"<string>"}` body, not the matters `{"error":{"code":...}}` shape. For the sheet edit path specifically (`src/api/services/editDocSheet.ts`, `src/api/routes/docSheet.ts`):

- 412 → `{"error":"base_version_stale"}`
- 422 → `{"error":"sheet_cell_invalid"}`
- 409 → `{"error":"unsupported_doc_type"}`

So the codes are domain-specific, but the important part is the envelope *shape*: because `error` is a string, `ParseBackendError` Layer 1 (`json.Unmarshal` into `struct{ Error struct{ Code string } }`) fails to unmarshal and is skipped entirely; Layer 2 (`{msg,status}`) has no `msg` and is skipped too. The request therefore falls through to the **Layer 3 status-inference fallback**, which keys on the HTTP status via `codeFromStatus`/`hintFromStatus` — not on `error.code`. Since docs-backend sets the correct 409/412/422 statuses, the generic CLI code and the friendly hint (e.g. 412 → "base version stale; re-read to get the current base version, then retry") **do surface** for the docs edit path. The raw domain code is preserved in the message body (the sub-2KB body is echoed).

Conclusion: the mappings added here are correct and effective as-is via the status path — **no code change needed**. Keying `backendErrorMapping` on `base_version_stale` etc. would be dead weight, since Layer 1 can never populate `.code` from a flat-string envelope regardless of the table contents.

yujiawei (approved):
# Code Review — PR #73 (octo-cli)

**Verdict: APPROVED** — no blocking (P0/P1) issues. Two non-blocking observations (P2) and one refuted concern documented below.

Reviewed at head `d73450ae7778f8444bff16b2e94dd9344f24a0f1` against merge-base `0f8a9444` (`origin/main`). `go build ./...`, `go vet ./...`, and `go test ./...` all pass (11 packages green); the new `docs sheet get|edit` operations register correctly and carry the right risk classification (`read` / `write`).

## 1. Spec compliance ✅

The change is spec-only, exactly as the commit message claims, and mirrors the existing `docs content get/edit` pair:

- Two operations added to `internal/registry/specs/docs.json`: `docs.sheet.get` (`GET /v1/bot/docs/{docId}/sheet`) and `docs.sheet.edit` (`PATCH` same path). Confirmed registered via `TestDocs_RegistryShape` / `TestDocs_TreeShape` and the built binary (`schema --list docs`).
- `--base-version` is wired to the `If-Match` header via `x-octo-flag: base-version` with `required: true`, matching `docs.content.edit` exactly (`internal/registry/specs/docs.json`). `TestDocsSheetEdit_SendsCellsBatchAndIfMatch` verifies the token reaches the wire as `If-Match` and is **not** duplicated into the JSON body; `TestDocsSheetEdit_RequiresBaseVersion` verifies the command fails before any request when the flag is omitted.
- `--limit` / `--cursor` reach the wire only when set (`TestDocsSheetGet_Paginated`), and a plain read sends neither (`TestDocsSheetGet_WholeSheet`).
- Op counts bumped in step: `loader_test.go` docs `24 → 26`; README `71 → 73`; `SKILL.md` section renumbered §3–§7 with no dangling cross-references (the remaining `§2` refs are still valid).

No missing scope (漏建), no extra endpoints/flags/fields beyond the commit message (超建), and no deviation from the established convention (偏离).

## 2. Code quality — Approved

Clean, spec-driven, and consistent with the domain. The new tests are non-vacuous and assert the real wire contract (method, path, `If-Match` placement, `null`-as-delete semantics, no base-version body leak, pagination query params).

### Non-blocking observations (P2)

**P2 — `TestParseBackendError_MattersCodes` does not exercise the three new codes.** `internal/output/errors_test.go` iterates a hard-coded `cases` slice, not `backendErrorMapping`, so `CONFLICT` / `PRECONDITION_FAILED` / `UNPROCESSABLE_ENTITY` (added at `internal/output/errors.go:108-110`) are unasserted, and `TestParseBackendError_EmptyBody_StatusInference` similarly omits 409/412/422. The mappings themselves are correct and CI is green — this is a test-coverage/maintainability gap (a future typo in Type/Hint wouldn't fail CI), not a functional defect. Suggested fix: range `backendErrorMapping` directly, or add the three cases. Note the doc comment "exercises every entry" is now stale.

**P2 — friendly-hint delivery depends on the backend's error-body shape.** `ParseBackendError` Layer 1 matches the verbatim `error.code` against `backendErrorMapping`; the new keys are the HTTP-status names (`CONFLICT`, `PRECONDITION_FAILED`, `UNPROCESSABLE_ENTITY`). If the docs backend returns a domain code in a matters envelope (e.g. `{"error":{"code":"base_version_stale"}}`), Layer 1 misses and falls through to `api_error` with no hint; the friendly hints only fire when the backend returns the generic HTTP-status code, a `{msg,status}` (dmworkim) body, or an empty body (status inference). This is **not a regression** — these keys did not exist before, so previously-unmatched bodies behave exactly as they did. Whether the hints actually surface on the sheet edit path is answerable only against the docs backend (what `error.code` it returns for 409/412/422), which is out of scope for octo-cli.

### Refuted concern — `--base-version` in `--dry-run` output (not a blocker)

An automated pass flagged the base-version token appearing unmasked in `--dry-run` output as a P0 leak. This does not hold for this PR:

- `internal/client/client.go` (`renderDryRun`) is **not modified by this PR** — the spec-declared-header copy is pre-existing behavior, and the `If-Match`/`base-version` wiring already shipped with `docs.content.edit`.
- The base-version is an **opaque optimistic-concurrency (ETag-style) token that the user supplies themselves** on the command line — it is not a credential. Writes still require the `Authorization` bearer token (which *is* masked in dry-run) plus writer permission; a leaked/stale base-version grants nothing and is rejected with `412` if stale. Echoing user-provided input back in dry-run is not information disclosure.

## 3. Security note (PR flagged security-sensitive)

The security surface here is the optimistic-concurrency guard, and it is wired correctly: `--base-version` is `required: true` and mapped to `If-Match`, so an edit cannot silently land without the guard (verified by `TestDocsSheetEdit_RequiresBaseVersion` — server not called when the flag is missing). The CLI is a thin request-issuer; the substantive enforcement (`base_version_stale`, `sheet_snapshot_invalid`, size caps `too_many_cells` / `cell_too_large` / `sheet_too_large`) lives in the docs backend and is out of scope for this repo. **For a human verifier:** confirm on the backend that (a) the `If-Match` guard is actually enforced for the sheet edit path (not just documented), and (b) the exact `error.code` values returned for 409/412/422 so the new hints in §2 land as intended.

yujiawei (approved):
# Code Review — PR #73 (octo-cli), re-review at head `6aee6d75`

**Verdict: APPROVED.** Re-reviewed at the new head `6aee6d751f030022b2d86d2c3124749815789275` (the prior approval was at `d73450ae`; a new head gets a fresh verdict rather than inheriting the old one).

## Delta since `d73450ae` (2 files, +5/−2)

The single new commit is a clean, purely-additive response to earlier review notes — no new behavior, no new risk:

- `internal/output/errors_test.go` — adds `{409, validation, CONFLICT}`, `{412, validation, PRECONDITION_FAILED}`, `{422, validation, UNPROCESSABLE_ENTITY}` to `TestParseBackendError_EmptyBody_StatusInference`, locking the new `codeFromStatus` branches. Closes the status-inference half of the earlier P2 test-coverage note.
- `skills/octo-docs/SKILL.md` — fixes `--output json` → `--format json` in two example snippets (the paged-read loop and the content-edit example). `--format` is the real global flag (`cmd/root.go:40`); the old text would have failed on paste with `unknown flag: --output`. This resolves the docs bug flagged independently on the PR.

Verified at the new head: `go build ./...`, `go vet ./...`, `go test ./...` all pass (11 packages); no stray `--output` remains in README/SKILL.

## Standing result (unchanged from the base review)

Spec-only, spec-driven addition mirroring `docs content get/edit` exactly. `--base-version` → `If-Match` via `x-octo-flag` (`required: true`); `--limit`/`--cursor` reach the wire only when set; op counts bumped in step (24→26 / 71→73). Optimistic-concurrency guard wired correctly (edit cannot land without the base-version token). No blocking (P0/P1) issues.

### Residual non-blocking P2 (unchanged)

- `TestParseBackendError_MattersCodes` still iterates a hard-coded `cases` slice that omits the three new codes, so Layer-1 (matters-envelope) mapping of `CONFLICT`/`PRECONDITION_FAILED`/`UNPROCESSABLE_ENTITY` remains unasserted (only the status-inference path was locked this round). Maintainability only — mapping code is correct, CI green.
- Friendly-hint delivery still depends on the backend's error-body shape: the new keys are HTTP-status names, so a matters envelope carrying a domain code (`base_version_stale`, `sheet_changed`) misses Layer 1 → `api_error`, no hint. Not a regression; resolvable only against the docs backend.

Neither blocks merge.

OctoBoooot (approved):
## Re-review on `6aee6d751f03` (delta) — both minors resolved

**Verdict:** Approve. PR is non-draft and all CI is green (`build & test (1.24.x)` ✓, `npm test` ✓, `check-sprint` ✓, `osv-scan` ✓).

The delta since my COMMENT addresses both confirmed minors exactly:

- **`errors_test.go`** — the status-inference table now includes `{409,"validation","CONFLICT"}`, `{412,"validation","PRECONDITION_FAILED"}`, `{422,"validation","UNPROCESSABLE_ENTITY"}`, matching the new `codeFromStatus` / `backendErrorMapping` entries. The empty-body/status path for all three is now locked.
- **`SKILL.md`** — `--output json` → `--format json` fixed in **both** spots (line 78 content-edit example and line 137 sheet-pagination loop; I'd only flagged the one, the author caught both). The embedded agent docs no longer fail on paste.

Everything from the prior round still holds: sheet get/edit wire shape tested (whole-sheet vs paged, cells via `--data`, base-version as `If-Match`, required-flag guard before any HTTP call); the errors.go change is strictly additive (base had zero 409/412/422 handling, so no regression); op count 26/73 consistent.

On the one open **question** from last round — whether the docs backend emits generic (`CONFLICT`/`PRECONDITION_FAILED`) or domain-specific (`base_version_stale`/`sheet_changed`) `error.code` values, which determines if the friendly 412 hint fires on the *edit* path via Layer 1: that's a backend-envelope confirmation, unverifiable from this repo, and non-blocking either way (the change can't make anything worse). Worth the author/backend confirming so the operability hint actually lands, but it doesn't gate this PR.

Clean iteration. `mergeStateStatus` — confirm branch-currency + the paired backend PR (#42) before merge; approved ≠ mergeable, but no code objection from my side.

Jerry-Xin (approved):
Re-review on new head `6aee6d7` (prior APPROVE was on `d73450a`, auto-dismissed by this push). True delta is 2 files, +5/-2, no rebase (ahead 1 / behind 0).

**Both tracked open items are now resolved:**
- ✅ SKILL.md `--output json` → `--format json` fixed at both call sites (docs content get + sheet paging loop). Matches the real global flag (`root.go`).
- ✅ `TestParseBackendError_EmptyBody_StatusInference` extended with 409/CONFLICT, 412/PRECONDITION_FAILED, 422/UNPROCESSABLE_ENTITY — the table loop genuinely asserts `Type`+`Code` via `ParseBackendError(status, nil)`, so these are real assertions covering the codes added to `codeFromStatus`.

**On the "lowercase domain codes bypass the mapping" concern:** traced the backend contract in `octo-docs-backend` — it returns errors as a **bare-string body** `{"error":"base_version_stale"}` (see `test/docSheet.test.ts`, `test/docContent.test.ts`), **not** the matters envelope `{"error":{"code":...}}`. In `ParseBackendError`, `json.Unmarshal` of a string into the Layer-1 `Error{Code,Message}` struct returns a non-nil error, so the `== nil` guard fails and Layer 1 is skipped. The path falls through to **Layer 3 status inference**, where `codeFromStatus(412)=PRECONDITION_FAILED` + the "base version stale; re-read…" hint fires correctly off the HTTP status. So the new 409/412/422 mappings are **not inert** — the friendly hints DO surface on the real edit/get paths via status, exactly as intended. No lowercase-key mapping is required for this backend's wire shape.

**💬 Non-blocking**
- The `TestParseBackendError_MattersCodes` doc-comment ("exercises every entry in the backend mapping table") is now slightly stale: the 3 new uppercase entries (CONFLICT/PRECONDITION_FAILED/UNPROCESSABLE_ENTITY) aren't in *that* envelope-path table (they're only in the status-inference table). Cosmetic — the entries are exercised via status inference and used correctly at runtime. Optional: add the 3 rows there too, or soften the comment.
- Spec/backend nit (out of PR scope): docs.json prose says `sheet_snapshot_invalid` → 422, but the backend test asserts 409. Worth reconciling in the spec later.

**✅ Highlights:** command shape, `--limit`/`--cursor` query wiring, and `--base-version`→`If-Match` header wiring are covered by focused service tests; `go test ./...` passes. Strictly additive, no regressions.

APPROVE.

PR #78, opened 2026-07-10
feat(docs): add docs.scene.export CLI operation (whiteboard PNG/SVG)

boris-clark:
Closes #77.

## What
Adds the `docs.scene.export` operation so bots can export a whiteboard's live scene:

```
octo-cli docs scene export <docId> --image-format png|svg -o board.png
```

## How
- `internal/registry/specs/docs.json`: new `GET /v1/bot/docs/{docId}/export` operation (`docs.scene.export`) with a `format` enum (png/svg, default png) and `x-octo-binary-response: true`. The query param carries `x-octo-flag: image-format`, so the CLI flag is `--image-format` while the wire param stays `?format=` — the global persistent `--format` output flag (json|table|csv|ndjson) is not shadowed.
- Engine: the existing binary-response path only DESCRIBED a 2xx body (`{status, content_type, size}`). Added an `--output`/`-o` flag that makes the client WRITE the 2xx body to disk and echo the saved path in the envelope. A local write failure is a hard, non-retryable validation error.
- `--output`/`-o` is scoped to operations that return an **inline 2xx binary body** (new `OperationDetail.BinaryBody` signal, derived from whether the op declares a 2xx response with a content body). Redirect-style binary ops such as `file.download` (302-only, no consumable body) no longer register `-o` — previously the flag was accepted there but silently wrote nothing. `x-octo-binary-response` is unchanged, so the 302 Location surfacing still works.
- Hardened `runPaginated`'s next-page request clone to copy the whole struct (was rebuilding it field-by-field and dropping `OutputPath`/`BinaryResponse`/`RawBody`/`ContentType`).
- Updated the registry-count tests and the octo-docs skill (§8).

## Tests
`go build ./...`, `go vet ./...`, `go test ./...` all green. New regressions:
- `TestBinaryBodyGatingDistinguishesInlineFromRedirect` (loader): `docs.scene.export` has `BinaryBody`, `file.download` does not.
- `TestFileDownload_NoOutputFlag`: `file download` exposes no `-o`/`--output`.
- `TestDocsSceneExport_OutputFlagWritesFileToDisk`: `docs scene export -o` still lands bytes on disk.
- Plus the existing `--image-format` collision regressions and client output-to-disk tests.

Pairs with the backend endpoint (Mininglamp-OSS/octo-docs-backend#49, PR Mininglamp-OSS/octo-docs-backend#50).

PR #79, opened 2026-07-10
XIN-796: reject index-less/illegal whiteboard elements in docs scene edit

Jerry-Xin:
## What

`docs scene edit` passed each whiteboard element's fractional-index `index`
through untouched — it never produced, validated, or normalized it. An element
sent with no `index` (or a malformed one) reached the backend, where a buggy
repair path rewrote it into an invalid order key such as `r00000003`, which
broke the board when opened.

This change rejects index-less / malformed-index elements **client-side, before
the request is sent**, and stops the CLI's own docs skill from teaching callers
to omit `index`.

## Changes

- **`internal/fracindex`** — a small, dependency-free port of the
  `fractional-indexing` library's `validateOrderKey`, so the CLI recognizes the
  same valid/invalid z-order keys the frontend and backend do (the exact
  `r00000003` artifact fails as a truncated integer part).
- **Spec-driven validation** — a new `x-octo-validate-elements-index` extension
  on `docs.scene.edit` (`internal/registry/specs/docs.json`), parsed into
  `OperationDetail.ValidateElementsIndex`. `runOperation` validates every
  upserted element's `index` when the flag is set and fails with a
  `VALIDATION_ERROR` (exit code 2) plus a hint on how to produce a correct key.
  Keeping it in the spec avoids a hard-coded operationId carve-out in the engine.
- **Delete-only batches** (no `elements` to upsert) are unaffected.
- **`skills/octo-docs/SKILL.md`** — the example previously showed an index-less
  element; it now carries a valid `index` (`a0`) and the doc states that `index`
  is mandatory and must be a real fractional index (never an `r`+digits, plain
  integer, or timestamp).

## Behavior

```
# missing index      -> VALIDATION_ERROR, exit 2, not sent
# index "r00000003"  -> VALIDATION_ERROR, exit 2, not sent
# index "a0"         -> passes, request sent as before
```

## Tests

- `internal/fracindex` unit tests (valid keys, the `r00000003` artifact, empty,
  plain integer, timestamp, reserved bound, truncated integer part, bad digit).
- End-to-end `docs scene edit` tests: index-less element rejected before any
  request, invalid `index` rejected, delete-only batch still allowed, and the
  existing happy-path batch updated to carry a valid `index`.
- `go build ./...`, `go vet ./...`, and `go test ./... -count=1` all green.

## Note for reviewers

A complementary server-side guard is worth adding: the scene-edit endpoint
should return `422 board_element_invalid` for an element with a missing/invalid
`index` rather than routing it through the repair path. That is a backend change
outside this PR's scope, but recommended as a second line of defense.

Refs XIN-796.

### DIFF
--- a/internal/fracindex/fracindex.go
+++ b/internal/fracindex/fracindex.go
@@ -1,0 +1,72 @@
+// Package fracindex validates fractional-index order keys — the z-order keys
+// the whiteboard backend expects on every Excalidraw element.
+//
+// It is a faithful port of the reference `fractional-indexing` library's
+// validateOrderKey (the same rules the frontend and backend enforce). The CLI
+// uses it to reject malformed keys locally, before they reach the backend, so
+// callers can no longer send index-less or garbage-index elements (e.g. the
+// "r00000003" repair artifact from XIN-792) that corrupt a board.
+package fracindex
+
+import (
+	"errors"
+	"fmt"
+)
+
+// base62Digits is the ordered digit alphabet used for the fractional part of a
+// key. The integer part's magnitude is encoded in the head character.
+const base62Digits = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+
+// smallestInteger is the reserved lower bound the library never emits as a real
+// key; treating it as valid would let a caller pin an element below every other.
+const smallestInteger = "A00000000000000000000000000"
+
+// integerLength returns the total length the integer part must have given its
+// head character. Mirrors the reference getIntegerLength: 'a'..'z' encode
+// increasing positive magnitudes, 'A'..'Z' increasing negative magnitudes.
+func integerLength(head byte) (int, error) {
+	switch {
+	case head >= 'a' && head <= 'z':
+		return int(head-'a') + 2, nil
+	case head >= 'A' && head <= 'Z':
+		return int('Z'-head) + 2, nil
+	default:
+		return 0, fmt.Errorf("invalid order key head: %q", string(head))
+	}
+}
+
+// ValidateOrderKey reports whether key is a well-formed fractional index. It
+// returns nil for a valid key and a descriptive error otherwise. An empty key
+// is always invalid.
+func ValidateOrderKey(key string) error {
+	if key == "" {
+		return errors.New("order key is empty")
+	}
+	if key == smallestInteger {
+		return fmt.Errorf("invalid order key: %q (reserved lower bound)", key)
+	}
+	intLen, err := integerLength(key[0])
+	if err != nil {
+		return err
+	}
+	if intLen > len(key) {
+		return fmt.Errorf("invalid order key: %q (integer part truncated)", key)
+	}
+	// The fractional part (everything after the integer part) must be composed
+	// entirely of base-62 digits.
+	for i := intLen; i < len(key); i++ {
+		if !isBase62(key[i]) {
+			return fmt.Errorf("invalid order key: %q (bad fractional digit %q)", key, string(key[i]))
+		}
+	}
+	return nil
+}
+
+func isBase62(c byte) bool {
+	for i := 0; i < len(base62Digits); i++ {
+		if base62Digits[i] == c {
+			return true
+		}
+	}
+	return false
+}
--- a/cmd/service/docs_test.go
+++ b/cmd/service/docs_test.go
@@ -806,7 +806,7 @@ func TestDocsSceneEdit_SendsBatchAndIfMatch(t *testing.T) {
 		w.WriteHeader(200)
 		_, _ = w.Write([]byte(`{"docId":"d1","bytes":256,"baseVersion":"BV_NEXT==","newDocVersionSeq":7}`))
 	})
-	batch := `{"elements":[{"id":"e1","type":"rectangle","version":4}],"deletedElementIds":["e2"],"files":{}}`
+	batch := `{"elements":[{"id":"e1","type":"rectangle","version":4,"index":"a0"}],"deletedElementIds":["e2"],"files":{}}`
 	root.SetArgs([]string{"docs", "scene", "edit", "d1", "--base-version", "BV_ABC==", "--data", batch})
 	if err := root.Execute(); err != nil {
 		t.Fatalf("execute: %v", err)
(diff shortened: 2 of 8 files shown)


OctoBoooot (commented):
## Review: XIN-796: reject index-less/illegal whiteboard elements in docs scene edit (#79)

**Verdict:** Comment — one **minor** (port-completeness gap), otherwise clean and it correctly kills the target bug. Draft PR; no red CI (`build & test`/`npm test` skip on the draft; `osv-scan` green). Author is Jerry-Xin; I'm posting the formal verdict since that's a separate identity from my viewer.

Client-side defensive validation: reject whiteboard elements with a missing/malformed fractional-index `index` before they reach the backend, spec-gated by `x-octo-validate-elements-index` (only `docs.scene.edit` declares it). Verified the wiring is clean — validation runs in `run.go` *after* `resolveBody`, so both `--data` and promoted-flag body paths are covered, no bypass; the engine keys off the spec flag, not a hard-coded operationId.

**Target bug is genuinely fixed (verified by hand-trace + test):** the XIN-792 artifact `r00000003` → head `r` gives integer length 19 > `len("r00000003")` = 9 → rejected as "integer part truncated." `fracindex_test.go` pins it, plus plain-integer `3`, timestamp, empty, bad head, bad fractional digit. `validate.go` correctly distinguishes absent index (comma-ok) / non-string index / empty — all rejected with an actionable `indexHint`; delete-only batches (no `elements` key) and `elements: []` pass through. Error is `ErrValidation`→exit 2, non-retryable. No regression: validation only fires on `docs.scene.edit`, op count unchanged.

### Minor — integer-part digits are not validated (port omits `validateInteger`) _(confidence: high)_

`fracindex.go ValidateOrderKey` checks the fractional part's digits (`for i := intLen; i < len(key)`) but never validates the **integer part's body** (indices `1..intLen-1`). The reference `fractional-indexing` `validateOrderKey` calls `validateInteger` on the integer part, which requires every digit there to be base62. Hand-traced concrete false-accepts: `ValidateOrderKey("a!")` → head `a`, intLen 2, int part `a!`, the `!` is never checked → returns **nil** (should reject); same for `c00!`, `a!0`. These aren't the XIN-792 artifact (that's caught by the length check) and the backend has its own validation, so blast radius is low — but it's a faithfulness gap in a "faithful port," and the test suite has no bad-integer-digit case. Smallest fix: also loop `for i := 1; i < intLen; i++ { if !isBase62(key[i]) … }` and add a `"a!"`-style reject test. Not merge-blocking.

### Nits (cosmetic)
- `SKILL.md:192-193` — double blank line.
- The reference library is `fractional-indexing`; if a comment/name says otherwise ("jitterbug"), align it.

### Praise
- Spec-flag gating (`x-octo-validate-elements-index`) rather than hard-coding `docs.scene.edit` in the engine keeps the validation reusable and the service layer generic — consistent with how `x-octo-flag`/`BinaryBody` were handled in #78. And running the check post-`resolveBody` closes the `--data`-vs-flag bypass that a naive flag-only check would miss. _(confidence: high)_
- The `indexHint` on every rejection is good agent-operability: it tells the caller how to produce a valid key instead of just saying "invalid." _(confidence: high)_

Solid defensive fix — kills the r00000003 path cleanly with real tests. Fold in the integer-part check (or note it as intentional scope) and, once out of draft, it's APPROVE-ready.

internal/fracindex/fracindex.go
```diff
@@ -0,0 +1,72 @@
...
+	intLen, err := integerLength(key[0])
+	if err != nil {
+		return err
+	}
+	if intLen > len(key) {
+		return fmt.Errorf("invalid order key: %q (integer part truncated)", key)
+	}
+	// The fractional part (everything after the integer part) must be composed
+	// entirely of base-62 digits.
+	for i := intLen; i < len(key); i++ {
```
OctoBoooot, line 57: **[minor]** Integer-part digits aren't validated. The fractional-part loop starts at `i := intLen`, so digits `1..intLen-1` (the integer part after the head) are never base62-checked — unlike the reference `validateOrderKey`, which runs `validateInteger` over the integer part. Concrete false-accepts: `ValidateOrderKey("a!")`, `"c00!"`, `"a!0"` all return nil but should reject. Not the XIN-792 artifact (that's caught by the `intLen > len(key)` truncation check) and the backend re-validates, so low blast radius — but it's a faithfulness gap in a faithful port, and there's no bad-integer-digit test. Fix: add `for i := 1; i < intLen; i++ { if !isBase62(key[i]) { return err } }` plus a reject test. Non-blocking.

ken-anderson2006:
Re-pointing this PR to a branch on my own fork so it can be maintained under my account. Superseded by #80, which carries the identical change (same commit). Closing this one to avoid duplicates.

Jerry-Xin:
Superseded by #80 (re-pointed to the ken-anderson2006 fork, identical code, same tree 8ed41a4b). Closing this one.

yujiawei (approved):
# Code Review — PR #79 (octo-cli)

## Summary

This PR does exactly what XIN-796 asks: it rejects index-less / malformed-index whiteboard elements **client-side, before the request is sent**, so they can no longer reach the backend repair path that rewrote them into corrupt order keys (the `r00000003` artifact). The approach is clean and idiomatic for a spec-driven CLI, the wiring is complete end-to-end, and the core corruption vector is provably caught. I verified the behavior against the reference `fractional-indexing` library and by tracing the validation path.

**Verdict: Approved.** No blocking issues. A few P2 accuracy/hardening suggestions below — none should hold the merge.

## What works

- **The `r00000003` artifact is correctly rejected.** `integerLength('r') = ('r'-'a')+2 = 19`, and the key is only 9 chars, so the `intLen > len(key)` ("integer part truncated") branch fires before the fractional loop is ever reached. Confirmed by the unit test and by hand.
- **No typed-nil footgun.** `validateElementsIndex` returns a concrete `*output.ExitError`, and `run.go` compares `verr != nil` directly (not after boxing into `error`), so the classic "typed nil is non-nil as an interface" trap does not fire.
- **Spec-driven gate is the right idiom.** `x-octo-validate-elements-index` follows the established `x-octo-*` per-operation flag pattern and is consumed the whole way through (`loader.go` → `OperationDetail.ValidateElementsIndex` → `runOperation`). No dead metadata.
- **Reachability of the guard is sound.** `docs.scene.edit` is not multipart and its body is always a JSON object via `--data` (a non-object `--data` is rejected upstream in `resolveBody`), so the `map[string]any` type assertion and the non-multipart placement are correct. Delete-only / elements-absent batches are intentionally allowed, matching the test.
- **The empty-key guard is load-bearing**, not redundant: `key == ""` is checked before `integerLength(key[0])`, which would otherwise index-panic on an empty string. An element with `"index":""` reaches this path, so keep it.

## Non-blocking suggestions (P2)

1. **`internal/fracindex/fracindex.go` — the "faithful port" claim is inaccurate; please reword.**
   The package doc (and PR body) say this is a "faithful port of the reference `fractional-indexing` `validateOrderKey`," but it diverges in **two** directions:
   - It **omits** the reference's trailing-zero rule (`if (f.slice(-1) === digits[0]) throw`). So `"a00"`, `"a0V0"` are **accepted** here but **rejected** by the reference.
   - It **adds** a base-62 fractional-alphabet check the reference does not perform. So `"a0-"` is **rejected** here (and `fracindex_test.go` bakes this in as `wantErr: true`) but **accepted** by the reference.

   Neither divergence weakens the XIN-796 gate — `r00000003` and all garbage/index-less cases are still rejected, and the extra strictness only rejects more junk (real library-emitted keys are always base-62). But the "faithful port" wording is factually wrong. Recommend either (a) restore the trailing-zero rule so it truly is a port, or (b) reword the doc/comment to "a stricter-than-reference validator" and note the two intentional deviations (and adjust the `"a0-"` test comment so it isn't presented as reference-consistent).

2. **`cmd/service/docs_test.go` — assert the error *type*, not just non-nil.**
   `TestDocsSceneEdit_RejectsIndexlessElement` / `RejectsInvalidIndex` assert only `err != nil` and `!called`. Mutating the error to a non-validation type (exit 1 instead of 2) still passes these tests. The repo already has the idiom (`ee := output.AsExitError(err); ee.Type == "validation"`). Adding it pins the documented exit-code-2 contract.

3. **`cmd/service/validate.go` — three error branches are untested:** `elements` not an array, an element not an object, and `index` present-but-not-a-string. The implementation is correct; a couple of table cases (`elements:"x"`, `elements:["x"]`, `elements:[{"index":123}]`) would prevent silent regressions on exactly the defensive branches this PR is about.

4. **Optional: multi-element / empty-array coverage.** No test exercises a batch where element `[1]` carries the bad key (the `elementLabel` already emits `elements[1] (id ...)` — worth asserting), and no test proves a present-but-empty `elements:[]` passes with a valid `--base-version`. Nice-to-have, not required.

## Note on the server-side follow-up

Agreed with the PR's own note: a server-side `422 board_element_invalid` guard for missing/invalid `index` (instead of routing through the repair path) is the right second line of defense. Out of scope here; worth a follow-up ticket.

OctoBoooot (commented):
## Correction — retracting my integer-part finding (A); trailing-zero (B) stands. Credit: Steve

I checked the canonical reference source (`rocicorp/fractional-indexing` `src/index.js`) and **my finding A was wrong**. Retracting it on both #79 and #80.

### ✗ Retract — "integer-part digits not base62-validated" (my earlier minor, posted `confidence: high`)

I claimed the reference's `validateInteger` charset-checks the integer-part body digits, so `ValidateOrderKey("a!")` / `"c00!"` should reject. **That is false.** The reference `validateInteger` is:
```js
function validateInteger(int) {
  if (int.length !== getIntegerLength(int[0])) throw ...;  // LENGTH ONLY
}
```
It only checks the integer part's **length**, never its digit charset. So `"a!"` (length 2, correct for head `a`) is accepted by the reference too. This port matches the reference here — there is **no** divergence, and adding an integer-part charset check would make the CLI *stricter* than what it ports, not more faithful. My "high confidence" was misplaced; I asserted the reference's behavior from memory instead of reading it. My mistake — thanks to Steve for byte-verifying and refuting it.

### ✓ Stands — trailing-zero fractional part (finding B)

The reference `validateOrderKey` **does** reject a fractional part ending in `digits[0]` (`'0'`):
```js
const f = key.slice(i.length);
if (f.slice(-1) === digits[0]) throw ...;
```
The port has no such check → `"a00"`, `"a0V0"`, `"a10"` are accepted by the CLI but rejected by the reference/backend. Real faithfulness divergence, byte-verified. This is the one genuine gap.

### Net corrected verdict
- **One real minor**, not two: the missing trailing-zero check (B). Fix: reject `key[len-1]=='0'` after the integer part, add a `"a00"` reject test.
- Finding A withdrawn — the port is faithful on integer-part digits.
- Severity: still **minor**, not blocker — backend re-validates authoritatively (CLI wrong-accept → backend 4xx, no corruption), and the XIN-792 target `r00000003` is correctly rejected regardless. But it does contradict the "faithful port" docstring, so worth fixing before merge.
- yujiawei's "reword to *stricter than reference*" is doubly wrong now: on integer-part the port is *equal* to the reference, and on trailing-zero it's *looser* — never stricter.

Verdict remains **COMMENT** (draft; one minor; backend authoritative). Same commit as #79 — applies to both. Correcting the record matters more than defending my earlier finding.

PR #80, opened 2026-07-10
XIN-796: reject index-less/illegal whiteboard elements in docs scene edit

ken-anderson2006:
## What

`docs scene edit` passed each whiteboard element's fractional-index `index`
through untouched — it never produced, validated, or normalized it. An element
sent with no `index` (or a malformed one) reached the backend, where a buggy
repair path rewrote it into an invalid order key such as `r00000003`, which
broke the board when opened.

This change rejects index-less / malformed-index elements **client-side, before
the request is sent**, and stops the CLI's own docs skill from teaching callers
to omit `index`.

## Changes

- **`internal/fracindex`** — a small, dependency-free port of the
  `fractional-indexing` library's `validateOrderKey`, so the CLI recognizes the
  same valid/invalid z-order keys the frontend and backend do (the exact
  `r00000003` artifact fails as a truncated integer part).
- **Spec-driven validation** — a new `x-octo-validate-elements-index` extension
  on `docs.scene.edit` (`internal/registry/specs/docs.json`), parsed into
  `OperationDetail.ValidateElementsIndex`. `runOperation` validates every
  upserted element's `index` when the flag is set and fails with a
  `VALIDATION_ERROR` (exit code 2) plus a hint on how to produce a correct key.
  Keeping it in the spec avoids a hard-coded operationId carve-out in the engine.
- **Delete-only batches** (no `elements` to upsert) are unaffected.
- **`skills/octo-docs/SKILL.md`** — the example previously showed an index-less
  element; it now carries a valid `index` (`a0`) and the doc states that `index`
  is mandatory and must be a real fractional index (never an `r`+digits, plain
  integer, or timestamp).

## Behavior

```
# missing index      -> VALIDATION_ERROR, exit 2, not sent
# index "r00000003"  -> VALIDATION_ERROR, exit 2, not sent
# index "a0"         -> passes, request sent as before
```

## Tests

- `internal/fracindex` unit tests (valid keys, the `r00000003` artifact, empty,
  plain integer, timestamp, reserved bound, truncated integer part, bad digit).
- End-to-end `docs scene edit` tests: index-less element rejected before any
  request, invalid `index` rejected, delete-only batch still allowed, and the
  existing happy-path batch updated to carry a valid `index`.
- `go build ./...`, `go vet ./...`, and `go test ./... -count=1` all green.

## Note for reviewers

A complementary server-side guard is worth adding: the scene-edit endpoint
should return `422 board_element_invalid` for an element with a missing/invalid
`index` rather than routing it through the repair path. That is a backend change
outside this PR's scope, but recommended as a second line of defense.

Refs XIN-796.

### DIFF
--- a/internal/fracindex/fracindex.go
+++ b/internal/fracindex/fracindex.go
@@ -1,11 +1,19 @@
 // Package fracindex validates fractional-index order keys — the z-order keys
 // the whiteboard backend expects on every Excalidraw element.
 //
-// It is a faithful port of the reference `fractional-indexing` library's
-// validateOrderKey (the same rules the frontend and backend enforce). The CLI
-// uses it to reject malformed keys locally, before they reach the backend, so
-// callers can no longer send index-less or garbage-index elements (e.g. the
-// "r00000003" repair artifact from XIN-792) that corrupt a board.
+// A key is accepted only if it satisfies every rule the frontend and backend
+// enforce, so the CLI rejects locally exactly what they reject server-side:
+//   - a well-formed integer part (correct length for its head character), from
+//     the reference `fractional-indexing` library's validateOrderKey;
+//   - a full-key base-62 charset, from the frontend/backend INDEX_RE
+//     (/^[A-Za-z0-9]+$/) — the integer part is charset-checked, not just the
+//     fraction;
+//   - no trailing '0' in the fractional part, from the reference
+//     validateOrderKey (a non-canonical key).
+//
+// The CLI uses it to reject malformed keys locally, before they reach the
+// backend, so callers can no longer send index-less or garbage-index elements
+// (e.g. the "r00000003" repair artifact from XIN-792) that corrupt a board.
 package fracindex
 
 import (
@@ -52,12 +60,21 @@ func ValidateOrderKey(key string) error {
 	if intLen > len(key) {
 		return fmt.Errorf("invalid order key: %q (integer part truncated)", key)
 	}
-	// The fractional part (everything after the integer part) must be composed
-	// entirely of base-62 digits.
-	for i := intLen; i < len(key); i++ {
+	// Every byte of the key — integer part and fractional part alike — must be a
+	// base-62 digit. The frontend (INDEX_RE = /^[A-Za-z0-9]+$/) and backend
+	// (identical regex) apply this gate to the whole key, so a non-base62
+	// integer-part byte such as the '!' in "a!" must be rejected here too.
+	for i := 0; i < len(key); i++ {
 		if !isBase62(key[i]) {
-			return fmt.Errorf("invalid order key: %q (bad fractional digit %q)", key, string(key[i]))
+			return fmt.Errorf("invalid order key: %q (bad digit %q)", key, string(key[i]))
 		}
+	}
+	// The fractional part (everything after the integer part) must not end in the
+	// first digit '0': a trailing zero is non-canonical and equivalent to the
+	// shorter key, so the reference validateOrderKey and the frontend/backend all
+	// reject it (e.g. "a00", "a10").
+	if len(key) > intLen && key[len(key)-1] == base62Digits[0] {
+		return fmt.Errorf("invalid order key: %q (fractional part ends in zero)", key)
 	}
 	return nil
 }
--- a/internal/fracindex/fracindex.go
+++ b/internal/fracindex/fracindex.go
@@ -1,0 +1,72 @@
+// Package fracindex validates fractional-index order keys — the z-order keys
+// the whiteboard backend expects on every Excalidraw element.
+//
+// It is a faithful port of the reference `fractional-indexing` library's
+// validateOrderKey (the same rules the frontend and backend enforce). The CLI
+// uses it to reject malformed keys locally, before they reach the backend, so
+// callers can no longer send index-less or garbage-index elements (e.g. the
+// "r00000003" repair artifact from XIN-792) that corrupt a board.
+package fracindex
+
+import (
+	"errors"
+	"fmt"
+)
+
+// base62Digits is the ordered digit alphabet used for the fractional part of a
+// key. The integer part's magnitude is encoded in the head character.
+const base62Digits = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+
+// smallestInteger is the reserved lower bound the library never emits as a real
+// key; treating it as valid would let a caller pin an element below every other.
+const smallestInteger = "A00000000000000000000000000"
+
+// integerLength returns the total length the integer part must have given its
+// head character. Mirrors the reference getIntegerLength: 'a'..'z' encode
+// increasing positive magnitudes, 'A'..'Z' increasing negative magnitudes.
+func integerLength(head byte) (int, error) {
+	switch {
+	case head >= 'a' && head <= 'z':
+		return int(head-'a') + 2, nil
+	case head >= 'A' && head <= 'Z':
+		return int('Z'-head) + 2, nil
+	default:
+		return 0, fmt.Errorf("invalid order key head: %q", string(head))
+	}
+}
+
+// ValidateOrderKey reports whether key is a well-formed fractional index. It
+// returns nil for a valid key and a descriptive error otherwise. An empty key
+// is always invalid.
+func ValidateOrderKey(key string) error {
+	if key == "" {
+		return errors.New("order key is empty")
+	}
+	if key == smallestInteger {
+		return fmt.Errorf("invalid order key: %q (reserved lower bound)", key)
+	}
+	intLen, err := integerLength(key[0])
+	if err != nil {
+		return err
+	}
+	if intLen > len(key) {
+		return fmt.Errorf("invalid order key: %q (integer part truncated)", key)
+	}
+	// The fractional part (everything after the integer part) must be composed
+	// entirely of base-62 digits.
+	for i := intLen; i < len(key); i++ {
+		if !isBase62(key[i]) {
+			return fmt.Errorf("invalid order key: %q (bad fractional digit %q)", key, string(key[i]))
+		}
+	}
+	return nil
+}
+
+func isBase62(c byte) bool {
+	for i := 0; i < len(base62Digits); i++ {
+		if base62Digits[i] == c {
+			return true
+		}
+	}
+	return false
+}
--- a/cmd/service/docs_test.go
+++ b/cmd/service/docs_test.go
@@ -806,7 +806,7 @@ func TestDocsSceneEdit_SendsBatchAndIfMatch(t *testing.T) {
 		w.WriteHeader(200)
 		_, _ = w.Write([]byte(`{"docId":"d1","bytes":256,"baseVersion":"BV_NEXT==","newDocVersionSeq":7}`))
 	})
-	batch := `{"elements":[{"id":"e1","type":"rectangle","version":4}],"deletedElementIds":["e2"],"files":{}}`
+	batch := `{"elements":[{"id":"e1","type":"rectangle","version":4,"index":"a0"}],"deletedElementIds":["e2"],"files":{}}`
 	root.SetArgs([]string{"docs", "scene", "edit", "d1", "--base-version", "BV_ABC==", "--data", batch})
 	if err := root.Execute(); err != nil {
 		t.Fatalf("execute: %v", err)
@@ -849,5 +849,59 @@ func TestDocsSceneEdit_RequiresBaseVersion(t *testing.T) {
 	}
 	if called {
 		t.Error("server must not be called when required --base-version is missing")
+	}
+}
+
+// TestDocsSceneEdit_RejectsIndexlessElement confirms an upsert element with no
+// `index` is rejected locally (non-zero exit) before any request is sent, so
+// index-less whiteboard elements can no longer reach the backend (XIN-792).
+func TestDocsSceneEdit_RejectsIndexlessElement(t *testing.T) {
+	called := false
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		called = true
+	})
+	batch := `{"elements":[{"id":"e1","type":"rectangle","version":4}]}`
+	root.SetArgs([]string{"docs", "scene", "edit", "d1", "--base-version", "BV_ABC==", "--data", batch})
+	if err := root.Execute(); err == nil {
+		t.Fatal("expected error for an element missing `index`")
+	}
+	if called {
+		t.Error("server must not be called when an element is missing `index`")
+	}
+}
+
+// TestDocsSceneEdit_RejectsInvalidIndex confirms the exact repair artifact from
+// XIN-792 (`r00000003`) — a malformed fractional-index — is rejected locally.
+func TestDocsSceneEdit_RejectsInvalidIndex(t *testing.T) {
+	called := false
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		called = true
+	})
+	batch := `{"elements":[{"id":"e1","type":"rectangle","version":4,"index":"r00000003"}]}`
+	root.SetArgs([]string{"docs", "scene", "edit", "d1", "--base-version", "BV_ABC==", "--data", batch})
+	if err := root.Execute(); err == nil {
+		t.Fatal("expected error for an element with an invalid `index`")
+	}
+	if called {
+		t.Error("server must not be called when an element carries an invalid `index`")
+	}
+}
+
+// TestDocsSceneEdit_DeleteOnlyBatchAllowed confirms a batch with no `elements`
+// to upsert (only soft-deletes) still passes — index validation applies only to
+// upserted elements.
+func TestDocsSceneEdit_DeleteOnlyBatchAllowed(t *testing.T) {
+	called := false
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		called = true
+		w.WriteHeader(200)
+		_, _ = w.Write([]byte(`{"docId":"d1","bytes":8,"baseVersion":"BV_NEXT==","newDocVersionSeq":8}`))
+	})
+	root.SetArgs([]string{"docs", "scene", "edit", "d1", "--base-version", "BV_ABC==", "--data", `{"deletedElementIds":["e2"]}`})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("delete-only batch should pass validation: %v", err)
+	}
+	if !called {
+		t.Error("server should be called for a valid delete-only batch")
 	}
 }
--- a/cmd/service/run.go
+++ b/cmd/service/run.go
@@ -72,6 +72,14 @@ func runOperation(cobraCmd *cobra.Command, f *cmdutil.Factory, rt *operationRunt
 			_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
 			return err
 		}
+		// Reject index-less / malformed-index whiteboard elements locally, before
+		// sending, for operations that declare it in the spec (docs.scene.edit).
+		if d.ValidateElementsIndex {
+			if verr := validateElementsIndex(body); verr != nil {
+				_ = f.EmitError(verr) //nolint:errcheck // best-effort emit before returning err
+				return verr
+			}
+		}
 		req.Body = body
 	}
 
--- a/cmd/service/validate.go
+++ b/cmd/service/validate.go
@@ -1,0 +1,65 @@
+package service
+
+import (
+	"fmt"
+
+	"github.com/Mininglamp-OSS/octo-cli/internal/fracindex"
+	"github.com/Mininglamp-OSS/octo-cli/internal/output"
+)
+
+// indexHint is the guidance attached to every fractional-index validation
+// error so a caller (typically an agent) learns the correct way to produce a
+// key instead of inventing one.
+const indexHint = "every whiteboard element must carry a valid fractional-index `index` (z-order key) — generate it with the fractional-indexing / Excalidraw rules (e.g. \"a0\", \"a1\"); never omit it or fabricate an \"r\"+digits, plain-integer, or timestamp value"
+
+// validateElementsIndex enforces that every element in a scene-edit body carries
+// a valid fractional-index `index`. It is invoked only for operations that
+// declare x-octo-validate-elements-index in the spec. Rejecting here — before
+// the request is sent — stops index-less / garbage-index elements from reaching
+// the backend, where a buggy repair path would rewrite them into an invalid key
+// (XIN-792) and break the board.
+//
+// A body with no `elements` key (e.g. a delete-only batch) is left untouched.
+func validateElementsIndex(body any) *output.ExitError {
+	m, ok := body.(map[string]any)
+	if !ok {
+		// No structured body (or --data was not an object); nothing to check.
+		return nil
+	}
+	raw, present := m["elements"]
+	if !present {
+		return nil
+	}
+	elems, ok := raw.([]any)
+	if !ok {
+		return output.ErrValidation("`elements` must be a JSON array", indexHint)
+	}
+	for i, e := range elems {
+		el, ok := e.(map[string]any)
+		if !ok {
+			return output.ErrValidation(fmt.Sprintf("elements[%d] must be a JSON object", i), indexHint)
+		}
+		label := elementLabel(el, i)
+		idxRaw, has := el["index"]
+		if !has {
+			return output.ErrValidation(fmt.Sprintf("%s is missing `index`", label), indexHint)
+		}
+		idx, ok := idxRaw.(string)
+		if !ok {
+			return output.ErrValidation(fmt.Sprintf("%s `index` must be a string", label), indexHint)
+		}
+		if err := fracindex.ValidateOrderKey(idx); err != nil {
+			return output.ErrValidation(fmt.Sprintf("%s has an invalid `index`: %v", label, err), indexHint)
+		}
+	}
+	return nil
+}
+
+// elementLabel builds a human-readable reference to an element for error
+// messages, preferring its id when present.
+func elementLabel(el map[string]any, i int) string {
+	if id, ok := el["id"].(string); ok && id != "" {
+		return fmt.Sprintf("elements[%d] (id %q)", i, id)
+	}
+	return fmt.Sprintf("elements[%d]", i)
+}
--- a/internal/fracindex/fracindex_test.go
+++ b/internal/fracindex/fracindex_test.go
@@ -27,6 +27,20 @@ func TestValidateOrderKey(t *testing.T) {
 		{"bad head", "!0", true},
 		{"truncated integer part", "b0", true}, // head 'b' requires a 3-char integer part
 		{"bad fractional digit", "a0-", true},
+
+		// Integer-part charset: every byte of the key must be base62, not just
+		// the fractional part. The frontend (INDEX_RE = /^[A-Za-z0-9]+$/) and
+		// backend (identical regex) reject a non-base62 integer-part byte, so
+		// the CLI must too.
+		{"integer-part non-base62", "a!", true},
+		{"integer-part non-base62 with fraction", "c00!", true},
+
+		// Trailing-zero rule: the fractional part must not end in the first
+		// digit '0' (a non-canonical key). The reference validateOrderKey and
+		// the frontend/backend all reject these.
+		{"trailing zero", "a00", true},
+		{"trailing zero short", "a10", true},
+		{"trailing zero deep", "a0V0", true},
 	}
 
 	for _, tc := range tests {
--- a/internal/fracindex/fracindex_test.go
+++ b/internal/fracindex/fracindex_test.go
@@ -1,0 +1,43 @@
+package fracindex
+
+import "testing"
+
+func TestValidateOrderKey(t *testing.T) {
+	tests := []struct {
+		name    string
+		key     string
+		wantErr bool
+	}{
+		// Valid keys produced by the fractional-indexing library.
+		{"integer zero", "a0", false},
+		{"integer one", "a1", false},
+		{"integer with fractional part", "a0V", false},
+		{"upper-head integer", "Zz", false},
+		{"deep fractional", "a0zzzzzzzzzz", false},
+
+		// The exact repair artifact from XIN-792 must be rejected: head 'r'
+		// demands a 19-char integer part but the key is only 9 chars.
+		{"repair artifact r00000003", "r00000003", true},
+
+		// Other malformed shapes.
+		{"empty", "", true},
+		{"plain integer", "3", true},
+		{"timestamp", "1720598400000", true},
+		{"reserved smallest integer", smallestInteger, true},
+		{"bad head", "!0", true},
+		{"truncated integer part", "b0", true}, // head 'b' requires a 3-char integer part
+		{"bad fractional digit", "a0-", true},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			err := ValidateOrderKey(tc.key)
+			if tc.wantErr && err == nil {
+				t.Errorf("ValidateOrderKey(%q) = nil, want error", tc.key)
+			}
+			if !tc.wantErr && err != nil {
+				t.Errorf("ValidateOrderKey(%q) = %v, want nil", tc.key, err)
+			}
+		})
+	}
+}
--- a/internal/registry/loader.go
+++ b/internal/registry/loader.go
@@ -212,6 +212,12 @@ type OperationDetail struct {
 	// redirect-style binary ops such as file.download, which return a 302 with
 	// no consumable body — offering -o there is a silent no-op footgun.
 	BinaryBody bool `json:"binary_body,omitempty"`
+	// ValidateElementsIndex captures the x-octo-validate-elements-index
+	// extension. When true, the CLI rejects a request whose body carries any
+	// `elements[]` entry missing a valid fractional-index `index` before it is
+	// sent — so index-less / garbage-index whiteboard elements (XIN-792) can no
+	// longer reach the backend and corrupt a board.
+	ValidateElementsIndex bool `json:"validate_elements_index,omitempty"`
 }
 
 // ListOperations returns every operation for a service, sorted by operationId.
@@ -319,6 +325,7 @@ func buildDetail(service string, doc map[string]any, pathStr, method string, op 
 
 	d.Multipart = boolOf(op["x-octo-multipart"])
 	d.BinaryResponse = boolOf(op["x-octo-binary-response"])
+	d.ValidateElementsIndex = boolOf(op["x-octo-validate-elements-index"])
 	if d.BinaryResponse {
 		if resps, ok := op["responses"].(map[string]any); ok {
 			d.BinaryBody = hasSuccessBody(doc, resps)
--- a/internal/registry/specs/docs.json
+++ b/internal/registry/specs/docs.json
@@ -447,8 +447,9 @@
       "patch": {
         "operationId": "docs.scene.edit",
         "summary": "Apply an element upsert/delete batch to a whiteboard scene (needs writer)",
-        "description": "Applies a batch of element upserts and soft-deletes to a doc_type 'board' scene under a mandatory optimistic-concurrency guard. Pass the base-version token from `docs scene get` via --base-version (sent as the If-Match header); the server rejects with 412 base_version_stale if the live scene moved since that read. `elements` are full elements to upsert (CAS: the higher `version` wins), `deletedElementIds` are soft-delete tombstones, and `files` upserts file refs; the batch is a structured object, so pass it as JSON through --data. Only doc_type 'board' is accepted (doc / sheet return 409 unsupported_doc_type). Gates: element count over the max → 413 too_many_elements, a single element too large → 413 element_too_large, whole scene over the cap → 413 doc_too_large, an element failing the whitelist → 422 board_element_invalid, a file ref with no usable attachId → 422 board_file_invalid, missing base version or malformed shape → 400 invalid_body.",
+        "description": "Applies a batch of element upserts and soft-deletes to a doc_type 'board' scene under a mandatory optimistic-concurrency guard. Pass the base-version token from `docs scene get` via --base-version (sent as the If-Match header); the server rejects with 412 base_version_stale if the live scene moved since that read. `elements` are full elements to upsert (CAS: the higher `version` wins), `deletedElementIds` are soft-delete tombstones, and `files` upserts file refs; the batch is a structured object, so pass it as JSON through --data. Every element in `elements` must carry a valid fractional-index `index` (z-order key); the CLI rejects an index-less or malformed-index element locally before sending. Only doc_type 'board' is accepted (doc / sheet return 409 unsupported_doc_type). Gates: element count over the max → 413 too_many_elements, a single element too large → 413 element_too_large, whole scene over the cap → 413 doc_too_large, an element failing the whitelist → 422 board_element_invalid, a file ref with no usable attachId → 422 board_file_invalid, missing base version or malformed shape → 400 invalid_body.",
         "x-octo-risk": "write",
+        "x-octo-validate-elements-index": true,
         "parameters": [
           { "name": "docId", "in": "path", "required": true, "schema": { "type": "string" } },
           {
@@ -470,7 +471,7 @@
                   "elements": {
                     "type": "array",
                     "items": { "type": "object" },
-                    "description": "full Excalidraw elements to upsert; on a key collision the element with the higher `version` (then smaller `versionNonce`) wins the CAS. Pass via --data."
+                    "description": "full Excalidraw elements to upsert; on a key collision the element with the higher `version` (then smaller `versionNonce`) wins the CAS. Each element MUST carry a valid fractional-index `index` (z-order key produced by the fractional-indexing / Excalidraw rules, e.g. `a0`); an index-less or malformed-index element is rejected locally before sending. Pass via --data."
                   },
                   "deletedElementIds": {
                     "type": "array",
--- a/skills/octo-docs/SKILL.md
+++ b/skills/octo-docs/SKILL.md
@@ -175,9 +175,21 @@ octo-cli docs scene get <docId>
 #   elements          — full elements to upsert (CAS: higher `version` wins)
 #   deletedElementIds — element ids to soft-delete (tombstone)
 #   files             — file refs to upsert
+# Every element MUST carry a valid fractional-index `index` (z-order key); the
+# CLI rejects an index-less or malformed-index element locally before sending.
 octo-cli docs scene edit <docId> --base-version "<token>" \
-  --data '{"elements":[{"id":"e1","type":"rectangle","version":4}],"deletedElementIds":["e2"],"files":{}}'
+  --data '{"elements":[{"id":"e1","type":"rectangle","version":4,"index":"a0"}],"deletedElementIds":["e2"],"files":{}}'
 ```
+
+> **`index` is mandatory on every upserted element.** It must be a valid
+> fractional-index (z-order) key — the same jitterbug / Excalidraw key format the
+> board uses, generated with the `fractional-indexing` rules (e.g. `a0`, `a1`,
+> `a0V`). Never omit `index`, and never fabricate one as an `r`+digits string, a
+> plain integer, or a timestamp: an index-less element used to slip through to a
+> buggy backend repair path that rewrote it into an invalid key (like
+> `r00000003`) and broke the board. The CLI now rejects such elements with a
+> non-zero exit before the request is sent.
+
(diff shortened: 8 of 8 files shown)


internal/fracindex/fracindex.go
```diff
@@ -0,0 +1,72 @@
...
+	intLen, err := integerLength(key[0])
+	if err != nil {
+		return err
+	}
+	if intLen > len(key) {
+		return fmt.Errorf("invalid order key: %q (integer part truncated)", key)
+	}
+	// The fractional part (everything after the integer part) must be composed
+	// entirely of base-62 digits.
+	for i := intLen; i < len(key); i++ {
```
OctoBoooot: **[minor]** Integer-part digits aren't validated. The fractional-part loop starts at `i := intLen`, so digits `1..intLen-1` (the integer part after the head) are never base62-checked — unlike the reference `validateOrderKey`, which runs `validateInteger` over the integer part. Concrete false-accepts: `ValidateOrderKey("a!")`, `"c00!"`, `"a!0"` all return nil but should reject. Not the XIN-792 artifact (that's caught by the `intLen > len(key)` truncation check) and the backend re-validates, so low blast radius — but it's a faithfulness gap in a faithful port, and there's no bad-integer-digit test. Fix: add `for i := 1; i < intLen; i++ { if !isBase62(key[i]) { return err } }` plus a reject test. Non-blocking. (Same finding as #79 — this PR is the identical commit.)

OctoBoooot (commented):
## Review: XIN-796: reject index-less/illegal whiteboard elements in docs scene edit (#80)

**Verdict:** Comment — one **minor** (port-completeness gap), otherwise clean and it correctly kills the target bug. Draft PR; no red CI (`build & test`/`npm test` skip on the draft; `osv-scan` green).

**Note — duplicate of #79:** this PR is the **same commit** (`843d8770ce48`) as #79, opened as a separate cross-fork PR by a different author. Byte-identical diff (same 8 files, same blobs, same CI runs). The two are redundant — maintainers likely want to land one and close the other. My review below is the same verdict I posted on #79, verified at this exact SHA.

Client-side defensive validation: reject whiteboard elements with a missing/malformed fractional-index `index` before they reach the backend, spec-gated by `x-octo-validate-elements-index` (only `docs.scene.edit` declares it). Wiring verified clean — validation runs in `run.go` *after* `resolveBody`, so both `--data` and promoted-flag body paths are covered; the engine keys off the spec flag, not a hard-coded operationId.

**Target bug genuinely fixed (hand-trace + test):** the XIN-792 artifact `r00000003` → head `r` gives integer length 19 > `len` 9 → rejected as "integer part truncated," pinned by a test alongside plain-integer/timestamp/empty/bad-head/bad-fractional cases. `validate.go` distinguishes absent / non-string / empty index (all rejected with an actionable `indexHint`); delete-only batches and `elements: []` pass. `ErrValidation`→exit 2, non-retryable. No regression: fires only on `docs.scene.edit`, op count unchanged.

### Minor — integer-part digits not validated (port omits `validateInteger`) _(confidence: high)_

`fracindex.go ValidateOrderKey` validates the fractional part (`for i := intLen; i < len(key)`) but never the integer part's body (indices `1..intLen-1`). The reference `fractional-indexing` `validateOrderKey` runs `validateInteger` over the integer part, requiring base62 there too. Hand-traced false-accepts: `ValidateOrderKey("a!")`, `"c00!"`, `"a!0"` return nil but should reject. Not the XIN-792 artifact (caught by the length check) and the backend re-validates, so low blast radius — but it's a faithfulness gap in a "faithful port," with no bad-integer-digit test. Fix: `for i := 1; i < intLen; i++ { if !isBase62(key[i]) { return err } }` + a reject test. Non-blocking.

### Nits (cosmetic)
- `SKILL.md:192-193` — double blank line.
- Reference library is `fractional-indexing`; align any comment/name that says otherwise.

### Praise
- Spec-flag gating (`x-octo-validate-elements-index`) rather than hard-coding `docs.scene.edit` keeps the engine generic; running the check post-`resolveBody` closes the `--data`-vs-flag bypass a flag-only check would miss. _(confidence: high)_
- The `indexHint` on every rejection tells the caller how to produce a valid key — good agent operability. _(confidence: high)_

Solid defensive fix. Fold in the integer-part check (or note it as intentional scope), pick one of #79/#80 to land, and once out of draft it's APPROVE-ready.

Jerry-Xin (dismissed):
**APPROVE** — Clean, well-scoped, well-tested client-side defensive validation that rejects index-less / illegal whiteboard elements in `docs.scene.edit` before they can corrupt a board. Build + `go test ./internal/fracindex/... ./cmd/service/...` both green on this SHA.

### 🔴 Blocking
None.

### 🟡 Non-blocking (completeness)
`fracindex.ValidateOrderKey` validates only the **fractional** part's base62 digits (loop from `intLen` to end) but omits the reference `fractional-indexing` impl's `validateInteger` for the **integer-part** digits. As a result some malformed keys are falsely accepted:
- `ValidateOrderKey("a!")` → head `a` ⇒ `intLen=2`, `2 > len 2` is false, fractional loop is empty ⇒ returns `nil`.
- `ValidateOrderKey("a!0")`, `ValidateOrderKey("c00!")` similarly slip through where the reference impl rejects.

Severity is low: the actual XIN-792 target artifact `r00000003` **is** correctly rejected (head `r` ⇒ `intLen=19 > len 9` ⇒ truncation error, and it's covered by a test), so the ticket's goal is met. This is only a gap for *other* malformed integer-part shapes. Recommend a follow-up adding `validateInteger` + tests for parity with the reference library.

### 🔵 Cosmetic
- `skills/octo-docs/SKILL.md` ~L192-193: double blank line.
- "jitterbug" wording reads oddly; should reference `fractional-indexing`.

### ✅ Highlights
- **Spec-driven gating**: validation fires only when the op declares `x-octo-validate-elements-index` (only `docs.scene.edit` carries it); engine doesn't hardcode an operationId, so no leakage to other ops (e.g. `scene.get` untouched).
- **Both paths covered**: `validateElementsIndex` runs after `resolveBody()` in `run.go`, covering both `--data` and flag-merge inputs — no bypass.
- **Correct Go zero-value handling**: comma-ok distinguishes absent vs present `index`; numeric `0` → "must be a string"; `""` → "empty"; all rejected. Delete-only batch (no `elements` key) and `elements:[]` allowed; `elements:null` rejected.
- **Clean error contract**: `ErrValidation` → `VALIDATION_ERROR`, `ExitCode=2` (non-retryable), each error carries an `indexHint`.
- **The XIN-792 artifact `r00000003` is correctly killed** via the truncation path, with a test asserting it.
- **Good test coverage**: `fracindex_test` + `docs_test` cover accept/reject/delete-only cases; #78 happy-path tests updated with a valid `"index":"a0"` still pass; docs op count stays 29 (no op/doc-count drift).
- **No injection surface** — pure structured validation that narrows the attack surface.

jeff-wilson2010 (changes requested):
Reviewed at head `843d8770ce4857313127d7647cbdf8dc99d105d6` (the live CLI PR — #79 is now closed; both share this sha). Line-by-line, traced to sink. **CHANGES_REQUESTED.**

### What works
- `run.go:75-86` — validation is gated on `d.ValidateElementsIndex` (the `x-octo-validate-elements-index` spec extension), runs on the resolved `--data` body before the request is sent, emits, and returns. `ErrValidation` → `Type "validation"` → `ExitCode() == 2`, so a missing/malformed index exits 2 as required.
- `validate.go` — a body with no `elements` key is left untouched (delete-only batches pass), missing/`index`-not-a-string/invalid-index are each rejected with a helpful hint; the error carries the offending element label.
- The exact XIN-792 artifact `r00000003` and index-less elements are rejected (head `r` needs a 19-char integer part). Tests `docs_test.go` cover reject-indexless, reject-`r00000003`, and delete-only-allowed. SKILL and spec docs updated with the mandatory-`index` guidance. Scope is clean.

### Blocker — cross-repo rule divergence in `internal/fracindex/fracindex.go`
The package doc comment claims this is "a faithful port of the reference `fractional-indexing` library's validateOrderKey (the same rules the frontend and backend enforce)." It is not, and its accepted-key set is a strict **superset** of the BE/FE rule, so the three repos do not agree byte-for-byte — the primary acceptance criterion for this fix.

1. **Integer-part charset is not validated** (`fracindex.go:222-228`). The base62 loop runs only over the fractional part (`for i := intLen; i < len(key)`). The integer-part characters after the head are never charset-checked. So `a!` is accepted here (head `a`, `intLen == 2 == len`, no fractional chars to scan), but FE (`BASE62_RE = /^[0-9A-Za-z]+$/`) and BE (the `BASE62_CHARS` charset gate) both reject it. The reference `validateOrderKey` also skips this, which is exactly why BE and FE add a **full-key** charset gate — this port dropped that gate for the integer part.

2. **Trailing-zero fractional rule is missing** (`fracindex.go:208-230`). The reference `validateOrderKey` throws when the fraction ends in `digits[0]` (`'0'`); FE's `isValidOrderKey` and BE's `generateKeyBetween` probe both enforce it. `ValidateOrderKey("a00")` returns nil here, but FE's own `normalize.test.ts` pins `isValidIndex("a00") === false`. So the CLI accepts a key the other two repos reject.

Net effect: an element the FE/BE consider invalid (`a00`, or any integer-part non-base62) passes the CLI's "reject before sending" gate, defeating part of the guard's purpose and breaking the byte-for-byte contract.

### Suggested fix
- Add a full-key base62 charset gate (cover the integer part, not just the fraction), mirroring FE's `BASE62_RE` / BE's `BASE62_CHARS`.
- Add the trailing-zero check: after slicing the fraction, reject when its last byte is `base62Digits[0]` (`'0'`).
- Extend `fracindex_test.go` with `a00` → invalid and an integer-part non-base62 case (e.g. `a!`) → invalid, to pin parity with FE/BE.

Once these land, re-request and I'll re-verify parity against the FE/BE rule.

OctoBoooot (commented):
## Supplement on #80 (and #79 — same commit) — a 2nd faithfulness gap (credit: jeff-wilson2010)

jeff-wilson2010's trailing-zero finding is **real and I missed it** — adding it. Also correcting a backwards characterization from yujiawei's #79 note.

### 2nd gap (verified) — fractional part ending in `0` not rejected

The reference `validateOrderKey` rejects a key whose fractional part ends in `0` (a non-canonical key — trailing zero is equivalent to the shorter key). This port has no such check. Hand-traced false-accepts: `ValidateOrderKey("a00")` and `"a0V0"` return nil, but the frontend/backend reject them. So together with the integer-part gap I flagged, the port is looser than the reference on **two** axes:
1. integer-part digits not base62-checked (`"a!"`, `"c00!"` accepted) — my original finding
2. fractional part ending in `0` not rejected (`"a00"`, `"a0V0"` accepted) — jeff's finding

Both mean the docstring's claim — *"a faithful port … the same rules the frontend and backend enforce"* — is inaccurate: the CLI accepts a class of keys the backend rejects.

### Severity: these are **minor**, not 🔴 (honest calibration)

jeff tagged both 🔴; I'd hold them at **minor**. Reasoning: the CLI validator is a *client-side pre-reject* layer; the backend re-validates and is authoritative. When the CLI wrongly accepts one of these, the user gets a backend 4xx instead of a local fast-fail — no corruption, no data loss, and it's still strictly better than the pre-PR state (no validation at all). The actual XIN-792 target `r00000003` **is** correctly rejected (length-truncation path, tested). So this is a completeness/faithfulness gap with a narrow blast radius, not a production-risk blocker.

That said — since the whole point of the change is "pre-reject what the backend would reject," two demonstrable holes plus a false "faithful/same rules" docstring are worth fixing before merge. Recommended in one pass:
- add `for i := 1; i < intLen; i++ { if !isBase62(key[i]) … }` (integer-part digits),
- reject a fractional part ending in `0` (`key[len-1]=='0'` after the int part),
- add reject tests for `"a!"`, `"a00"`, `"a0V0"`,
- and soften the docstring, or make it truly faithful.

### Correction to yujiawei's #79 note
yujiawei suggested rewording the docstring to *"stricter than reference."* That's backwards — the port is **looser**, not stricter (it accepts keys the reference rejects, per both gaps above). If the doc is reworded, it should say "less complete than the reference (integer-part + trailing-zero unchecked)," not "stricter." Editing it to "stricter" would ship an inaccurate claim.

### Verdict
Unchanged: **COMMENT** (draft PR; two minors, no defensible blocker/major — backend is the authoritative gate). Both #79 and #80 are the identical commit `843d8770ce48`, so this applies to both; land one, close the other, and fold both fracindex gaps into the fix. Credit jeff-wilson2010 for the trailing-zero catch.

Jerry-Xin (dismissed):
Revising my earlier APPROVE to REQUEST_CHANGES: after a byte-verify triggered by @jeff-wilson2010's review, I confirmed one real trailing-zero fractional-index parity gap against the reference lib. My previous approval treated the validation gap as non-blocking, but the divergence violates the PR's own "faithful port" contract, so it needs to be addressed.

## 🔴 Blocking: trailing-zero fractional part is not rejected

`internal/fracindex/fracindex.go` `ValidateOrderKey` has no check that the fractional part does not end in `'0'`. The reference it claims to faithfully port — `rocicorp/fractional-indexing` `validateOrderKey` — rejects any key whose fractional part ends in the zero digit:

```js
if (f.slice(-1) === digits[0]) {
  throw new Error("invalid order key: " + key);
}
```

Byte-verified divergence (both harnesses run on identical inputs):

| input  | this port | reference (front/backend) |
|--------|-----------|---------------------------|
| `a10`  | ACCEPTED  | REJECTED                  |
| `a00`  | ACCEPTED  | REJECTED                  |
| `a1V0` | ACCEPTED  | REJECTED                  |

This directly violates the PR's stated contract that it is a "faithful port that recognizes the same valid/invalid z-order keys as frontend and backend." Right now these three key classes are accepted here but rejected everywhere else.

Fix is one line — after the base62 loop, guard the (non-empty) fractional part:

```go
if len(key) > intLen && key[len(key)-1] == base62Digits[0] {
    return <ErrValidation with indexHint>
}
```

Please also add a test asserting `a10` / `a00` are rejected.

## 🟡 Non-blocking (NOT a parity mismatch): integer-part body charset not validated

`ValidateOrderKey`'s base62 loop starts at `i := intLen`, so it only charset-checks the fractional part. As a result `"a!"` / `"c00!"` are accepted. However — and this corrects the framing in @jeff-wilson2010's review that labeled it 🔴 — the reference impl is **equally lax** here: it also does not charset-validate the integer-part body digits. So this is **not** a CLI ↔ front/backend divergence. Adding a check would make the CLI *stricter* than the lib it ports. Optional to tighten; not required for parity.

## ✅ What's right

- The primary XIN-792 target artifact `r00000003` is correctly rejected either way.
- Spec-driven gating is clean.
- Both `--data` and flag input paths are covered.
- The exit-2 non-retryable `ErrValidation` contract is well-defined.
- The backend independently re-validates (422 `board_element_invalid`), so a slipped key is caught server-side — the real-world blast radius of the trailing-zero gap is degraded UX, not data corruption. This keeps the issue scoped to faithful-port correctness/parity rather than a security/data hole.

Credit to @jeff-wilson2010 — the trailing-zero finding aligns with their review.

OctoBoooot (commented):
## Correction — retracting my integer-part finding (A); trailing-zero (B) stands. Credit: Steve

I checked the canonical reference source (`rocicorp/fractional-indexing` `src/index.js`) and **my finding A was wrong**. Retracting it on both #79 and #80.

### ✗ Retract — "integer-part digits not base62-validated" (my earlier minor, posted `confidence: high`)

I claimed the reference's `validateInteger` charset-checks the integer-part body digits, so `ValidateOrderKey("a!")` / `"c00!"` should reject. **That is false.** The reference `validateInteger` is:
```js
function validateInteger(int) {
  if (int.length !== getIntegerLength(int[0])) throw ...;  // LENGTH ONLY
}
```
It only checks the integer part's **length**, never its digit charset. So `"a!"` (length 2, correct for head `a`) is accepted by the reference too. This port matches the reference here — there is **no** divergence, and adding an integer-part charset check would make the CLI *stricter* than what it ports, not more faithful. My "high confidence" was misplaced; I asserted the reference's behavior from memory instead of reading it. My mistake — thanks to Steve for byte-verifying and refuting it.

### ✓ Stands — trailing-zero fractional part (finding B)

The reference `validateOrderKey` **does** reject a fractional part ending in `digits[0]` (`'0'`):
```js
const f = key.slice(i.length);
if (f.slice(-1) === digits[0]) throw ...;
```
The port has no such check → `"a00"`, `"a0V0"`, `"a10"` are accepted by the CLI but rejected by the reference/backend. Real faithfulness divergence, byte-verified. This is the one genuine gap.

### Net corrected verdict
- **One real minor**, not two: the missing trailing-zero check (B). Fix: reject `key[len-1]=='0'` after the integer part, add a `"a00"` reject test.
- Finding A withdrawn — the port is faithful on integer-part digits.
- Severity: still **minor**, not blocker — backend re-validates authoritatively (CLI wrong-accept → backend 4xx, no corruption), and the XIN-792 target `r00000003` is correctly rejected regardless. But it does contradict the "faithful port" docstring, so worth fixing before merge.
- yujiawei's "reword to *stricter than reference*" is doubly wrong now: on integer-part the port is *equal* to the reference, and on trailing-zero it's *looser* — never stricter.

Verdict remains **COMMENT** (draft; one minor; backend authoritative). Same commit as #79 — applies to both. Correcting the record matters more than defending my earlier finding.

yujiawei (approved):
# Code Review — PR #80 (octo-cli)

**Head SHA:** `843d8770ce4857313127d7647cbdf8dc99d105d6`

## Summary

This PR closes a real board-corruption path: `docs scene edit` used to pass each whiteboard element's fractional `index` through untouched, so an index-less or malformed element reached the backend and got rewritten by a buggy repair path into an invalid order key (`r00000003`), breaking the board. The change rejects such elements client-side before the request is sent, driven by a new `x-octo-validate-elements-index` spec extension rather than a hard-coded operationId carve-out. The approach is clean and the scoping is tight.

Verified locally at the head SHA: `go build ./...`, `go vet ./...`, and `go test ./...` for the touched packages all pass.

## Spec compliance

Implements exactly what the linked issue asks and no more:

- **Rejection is client-side, pre-send** — `validateElementsIndex` runs in `runOperation` (`cmd/service/run.go:75-82`) after body resolution and before dispatch, gated on the spec flag. Confirmed by the new tests, which assert the server is never called on a bad batch (`cmd/service/docs_test.go:858-895`).
- **Spec-driven, not operationId-hardcoded** — `x-octo-validate-elements-index: true` on `docs.scene.edit` (`internal/registry/specs/docs.json`), parsed into `OperationDetail.ValidateElementsIndex` (`internal/registry/loader.go:215-217, 328`).
- **Delete-only batches unaffected** — a body with no `elements` key returns early (`cmd/service/validate.go:31-34`), covered by `TestDocsSceneEdit_DeleteOnlyBatchAllowed`.
- **No over-build** — no extra flags or unrelated behavior; the `elements` array is array-of-object, which `promotableKind` never promotes to a flag, so it can only arrive via `--data` as `[]any` and the type assertion in `validateElementsIndex` has no bypass.
- **Docs kept honest** — `skills/octo-docs/SKILL.md` no longer teaches an index-less element and states `index` is mandatory.

## Findings

### P2 — `fracindex` port omits the reference library's trailing-zero rule (`internal/fracindex/fracindex.go:41-63`)

The package doc calls this "a faithful port of the reference `fractional-indexing` library's `validateOrderKey` (the same rules the frontend and backend enforce)". It is close but not exact:

- The reference `validateOrderKey` ends with a check that rejects any key whose **fractional part ends in the zero digit** (`const f = key.slice(i.length); if (f.slice(-1) === digits[0]) throw`). This port has no such check — it only validates that each fractional char is base-62. As a result it **accepts** keys the reference rejects: `ValidateOrderKey("a0V0")`, `("a00")`, `("Zz0")` all return `nil` (verified by compiling and running the port).
- Conversely the port **adds** a per-digit base-62 check on the fractional part that the reference does not perform.

Impact is low and this is **not** a merge blocker: a well-formed, library-generated key never carries a trailing-zero fractional part, so legitimate callers are unaffected, and every headline artifact from the issue (missing `index`, `r00000003`, plain integer, timestamp, empty, reserved lower bound) is still correctly rejected. But it is a genuine fidelity gap against the explicitly-referenced algorithm, and it is untested. Suggested resolution — pick one:

1. Add the trailing-zero check after the fractional loop and a test case (e.g. `{"trailing fractional zero", "a1V0", true}`); or
2. Soften the doc wording from "faithful port … same rules the backend enforces" to something like "validates the malformed shapes this issue produced", so the code doesn't contradict its own contract.

### nit — no CHANGELOG entry (`CHANGELOG.md`)

The `[Unreleased]` section documents the binary rename but not this new user-visible behavior (`docs scene edit` now rejects index-less/malformed-index elements locally with a non-zero exit). A one-line Added/Changed bullet would keep the changelog complete; the file declares it follows Keep a Changelog.

## Notes (no action required)

- The `api` passthrough command deliberately ignores the registry/spec, so validation does not apply there — this is consistent with its documented escape-hatch design, not a gap.
- `boolOf` (vs. the stricter `truthy`) is used for the new flag, matching the convention of every other `x-octo-*` boolean except `x-octo-disabled`. Correct as written since the spec uses a real JSON boolean.
- The PR's own note recommending a complementary server-side `422 board_element_invalid` guard is a sound defense-in-depth suggestion; agreed, and appropriately out of scope here.

## Verdict

**APPROVED.** Spec compliance is complete with no over-build; no P0/P1 defects. The trailing-zero fidelity gap (P2) and the CHANGELOG entry (nit) are worth addressing but do not block merge.

yujiawei (changes requested):
# Code Review — PR #80 (octo-cli)

**Head SHA:** `843d8770ce4857313127d7647cbdf8dc99d105d6`

> **Revising my earlier verdict from APPROVED to CHANGES_REQUESTED.** My first pass rated the fractional-index parity gap as a non-blocking P2 because the backend re-validates. On re-examination that was the wrong bar: this PR's *stated acceptance criterion* is byte-for-byte parity with the frontend/backend, and it provably fails that for one key class. Independently re-verified below.

## Summary

The design is sound and the scoping is tight: `docs scene edit` previously passed each element's fractional `index` through untouched, so an index-less/malformed element reached the backend's buggy repair path and got rewritten into an invalid order key (`r00000003`), corrupting the board. This PR rejects such elements client-side before sending, driven by a spec extension (`x-octo-validate-elements-index`) rather than a hard-coded operationId carve-out. `go build`, `go vet`, and the touched-package tests all pass at the head SHA.

But the core mechanism does not meet its own contract.

## Blocker — `fracindex` port is not the byte-for-byte parity it claims

Both the PR description and the package doc state the port "recognizes the **same valid/invalid z-order keys the frontend and backend do**" / is "a faithful port … the same rules the frontend and backend enforce." That parity *is* the acceptance criterion for this fix, and the port violates it.

**Missing trailing-zero rule** (`internal/fracindex/fracindex.go:41-63`). The reference `validateOrderKey` ends by rejecting any key whose fractional part ends in the zero digit:

```js
const f = key.slice(i.length);
if (f.slice(-1) === digits[0]) throw new Error("invalid order key: " + key);
```

The Go port has no equivalent — its loop (lines 57-61) only checks each fractional char is base-62. I compiled and ran the actual port to confirm the divergence:

| input  | this port | reference / FE / BE |
|--------|-----------|---------------------|
| `a00`  | accepted  | rejected            |
| `a10`  | accepted  | rejected            |
| `a1V0` | accepted  | rejected            |

These are exactly the "malformed key that should be caught before it reaches the backend" class this guard exists to stop, yet they pass through it. This is a genuine functional hole in the guard's stated purpose, not a cosmetic doc mismatch.

**Fix (one line + test):** after the fractional loop, reject a non-empty fractional part ending in `base62Digits[0]`:

```go
if len(key) > intLen && key[len(key)-1] == base62Digits[0] {
    return fmt.Errorf("invalid order key: %q (fractional part ends in zero)", key)
}
```

Add a test pinning `a00` / `a10` → invalid (the current suite has no coverage for this rule, which is why the gap was invisible).

## Secondary (non-blocking) — integer-part charset

The base-62 loop starts at `i := intLen`, so the integer-part body digits are never charset-checked; `a!` and `c00!` are accepted. Note: the **reference library** `validateOrderKey`/`validateInteger` is equally lax here (it only length-checks the integer part), so this is *not* a divergence from the reference. Whether it diverges from the FE/BE full-key charset gate depends on FE/BE source I can't verify from this repo, so I'm not treating it as blocking. If FE/BE do apply a full-key base-62 regex, tightening to a full-key charset check (and a matching test) would be worth doing while you're in here; otherwise it's optional.

## What's right

- Validation is gated on `d.ValidateElementsIndex`, runs on the resolved body before send, and returns a `VALIDATION_ERROR` → exit 2 (`cmd/service/run.go:75-82`, `cmd/service/validate.go`).
- Delete-only batches (no `elements`) pass untouched; missing / non-string / invalid `index` each rejected with a helpful hint and element label.
- The headline XIN-792 artifact `r00000003` and index-less elements are correctly rejected (head `r` requires a 19-char integer part).
- Spec-driven gating avoids an engine carve-out; both `--data` and flag input paths are covered; `elements` is array-of-object so it can't be promoted to a flag and bypass the `[]any` assertion.
- SKILL.md and the spec description updated to teach the mandatory-`index` rule.

## Nits (optional)

- No `CHANGELOG.md` `[Unreleased]` entry for this new user-visible behavior (scene edit now rejects bad-index elements locally with a non-zero exit).
- The PR's own note recommending a server-side `422 board_element_invalid` guard is a sound defense-in-depth follow-up; agreed, and correctly out of scope here.

## Verdict

**CHANGES_REQUESTED.** The port claims byte-for-byte parity with the frontend/backend as its correctness contract but accepts a class of keys (`a00`, `a10`, `a1V0`) they reject — a real hole in the guard this PR is built to provide. The fix is one line plus a test. Once the trailing-zero check lands, re-request and this becomes a straightforward approve.

OctoBoooot (approved):
## Re-review on `aa84abae98ac` (delta) — both fracindex gaps fixed; APPROVE

**Verdict:** Approve. The one real blocker (trailing-zero) and the integer-part item are both resolved, verified by trace + new tests. Non-draft.

CI note: only `check-sprint` (pass) and `label` (pass) have run — the code CI jobs (`build & test`, `npm test`) haven't executed on this cross-fork PR yet (typically gated pending maintainer approval of the fork's workflow run). Nothing is red; but since I can't see a green `build & test` here, a maintainer should confirm it runs green before merge. `reviewDecision` shows CHANGES_REQUESTED from prior-SHA change-requests that need refreshing on this head.

### ✓ Fixed — trailing-zero fractional part (the real blocker)
`ValidateOrderKey` now rejects a key whose fractional part ends in `base62Digits[0]` (`if len(key) > intLen && key[len(key)-1] == '0'`). Traced: `"a00"`, `"a10"`, `"a0V0"` now reject; `"a0V0z"` (trailing non-zero) still accepts. Tests added for all three. This closes the parity gap jeff-wilson2010 caught and I confirmed against the reference source.

### ✓ Fixed — integer-part charset, with the *correct* authority cited
The charset loop now runs over the whole key (`for i := 0`), so `"a!"` / `"c00!"` reject. Importantly, the new docstring cites the right source: the **frontend/backend `INDEX_RE = /^[A-Za-z0-9]+$/`** gate the entire key — so this *is* genuine CLI↔backend parity, even though the reference `validateInteger` is length-only (which is what my retraction keyed on). The author reconciled both authorities correctly; my earlier "would make CLI stricter" concern is resolved by the regex being the operative backend rule. Tests added.

### Verified no over-rejection
Traced all prior valid keys still pass (`a0`, `a1`, `a0V`, `Zz`, `a0zzzzzzzzzz`) and the XIN-792 target `r00000003` still rejects (truncation path). The full-key charset + trailing-zero additions don't break any canonical key.

### Praise
- The docstring rewrite is exactly right: it now enumerates the three rules with their *sources* (reference `validateOrderKey` for integer-length + trailing-zero; frontend/backend `INDEX_RE` for full-key charset) rather than the vague "faithful port" claim that started this thread. That precision is what lets the next reader verify parity instead of trusting it. _(confidence: high)_

All findings across #79/#80 (same change, two PRs) are now resolved. From my side this is APPROVE; land one of the duplicate PRs and confirm code CI runs green on the fork. Nice close-out — credit jeff-wilson2010 for the trailing-zero catch that made the fix complete.

Jerry-Xin (approved):
Re-review at head `aa84abae` — the blocking trailing-zero parity gap is fixed. ✅ APPROVE.

## 🔴→✅ Blocking finding resolved: trailing-zero fractional part now rejected

The delta (`843d8770...aa84abae`, ahead_by=1, behind_by=0 — a clean single commit, not a rebase) touches exactly two files (`internal/fracindex/fracindex.go` +26/-9, `fracindex_test.go` +14/-0) and adds the guard I asked for, matching the reference `rocicorp/fractional-indexing` semantics:

```go
// The fractional part must not end in the first digit '0' ...
if len(key) > intLen && key[len(key)-1] == base62Digits[0] {
    return fmt.Errorf("invalid order key: %q (fractional part ends in zero)", key)
}
```

`base62Digits[0] == '0'`. Byte-traced the exact port logic on the concrete cases:

| key | before | now | correct? |
|---|---|---|---|
| `a10` | accepted (gap) | **rejected** (trailing zero) | ✅ |
| `a00` | accepted (gap) | **rejected** (trailing zero) | ✅ |
| `a1V0` | accepted (gap) | **rejected** (trailing zero) | ✅ |
| `a1` | accepted | accepted | ✅ valid, no trailing zero |
| `a0` | accepted | accepted | ✅ see note below |
| `Zz`, `a0V`, `a0zzzzzzzzzz` | accepted | accepted | ✅ no false-reject of valid keys |

**`a0` edge (verified correct):** head `'a'` → `intLen=2`, `len("a0")=2`. The guard is `len(key) > intLen`, so `2 > 2` is false → `a0` is NOT hit by the trailing-zero rule. That is correct: in `a0` the `'0'` is the *integer* part (integer value 0), not a fractional trailing zero. The reference keeps the classic `"a0"`/`"Zz"` form as canonical valid keys, and its trailing-zero throw is on the fractional midpoint (`a.slice(-1) === zero`). So the port matches the reference exactly. The #78 happy-path fixture (`docs_test.go:809` uses `"index":"a0"`) and the `{"integer zero","a0",false}` unit case both remain valid — **no regression**, tests are consistent.

## Integer-part charset check (optional, not required) — added, and it's fine

They also tightened the charset loop to start at `i:=0`, so the integer-part body is now base62-checked too (`a!`, `c00!` rejected). I noted in my prior review this is *not* a divergence-required fix (the reference `validateInteger` checks length only). Here it's justified against the frontend/backend `INDEX_RE = /^[A-Za-z0-9]+$/`, which does gate the whole key — so it aligns the CLI with the server regex rather than over-shooting it. Harmless and defensible; does not false-reject any legitimate base62 key.

## Regression checks — all intact
- **XIN-792 artifact** `r00000003` → still rejected (head `'r'` demands 19-char integer part, key is 9 → "integer part truncated"). ✅
- Delta scope is only the two fracindex files — `scene.get`, `docs.go`, and the spec-gating (`x-octo-validate-elements-index` in `docs.json` + `validate.go` caller) are **untouched**. scene.edit validation stays spec-gated. ✅
- New tests assert `a00`/`a10`/`a0V0` rejected and `a!`/`c00!` rejected; valid keys retained. ✅

## Build & tests (at `aa84abae`)
- `go build ./...` → **OK**
- `go test ./internal/fracindex/...` → **ok**
- `go test ./cmd/service/...` → **ok**

Docstring was also rewritten to accurately describe the three rules instead of the loose "faithful port" wording — cosmetic notes from before are addressed. Nothing blocking remains.

jeff-wilson2010 (changes requested):
## Re-review — head `aa84abae98ac4273151a08c27d73d0f57083625a`

Verdict: **CHANGES_REQUESTED**

Old CR is superseded; this is a fresh review at the new head. Verified diff `843d8770..aa84abae` line-by-line to sink, built/vetted/tested locally (all green), and compared byte-for-byte against the FE and BE gates.

### What the rework got right

- **Full-key base62 gate** (`fracindex.go:65`): the loop now runs `i := 0` over the whole key, so a non-base62 byte in the integer part (`a!`, `c00!`) is rejected. Correct.
- **Trailing-zero rule** (`fracindex.go:76`): `len(key) > intLen && key[len-1] == '0'` rejects a non-canonical fractional tail (`a00`, `a10`, `a0V0`) without touching an integer-only key that ends in `0` (`b00`, `c000`) or a mid-fraction zero (`a0V` stays valid). Correct.
- Tests cover the new reject cases plus the kept regressions; `go build/vet/test` all pass. Rework scope is confined to `fracindex.go` + its test.
- Against the concrete verdict table this task enumerated, the CLI matches **11/11** (rejects `r00000003`/`a!`/`c00!`/`a00`/`a10`/`a0V0`/sentinel; accepts `a0`/`a1`/`Zz`/`a0V`).

### Why this is still CHANGES_REQUESTED — the core acceptance point is not met

The stated pass condition is **byte-for-byte parity with the FE and BE gates**. That is not true, and I verified it against the actual source, not the claim:

- FE `isValidIndex` — `octo-web/packages/whiteboard-schema/src/normalize.ts:46-48`
- BE `isValidIndex` — `octo-docs-backend/src/whiteboard/schema/normalize.ts:46-48`

Both gate **only** on `INDEX_RE = /^[A-Za-z0-9]+$/` and non-empty. Empirically they **accept** `r00000003`, `a00`, `a10`, `a0V0`, and the sentinel `A0…0`; they reject only `a!` / `c00!` / empty. The CLI rejects all seven. So the CLI is a **strict superset**, diverging on 5 of 11 cases — it is not byte-aligned with either.

| key | CLI @aa84abae | FE `isValidIndex` | BE `isValidIndex` |
|---|---|---|---|
| r00000003 | REJECT | ACCEPT | ACCEPT |
| a00 / a10 / a0V0 | REJECT | ACCEPT | ACCEPT |
| sentinel `A0…0` | REJECT | ACCEPT | ACCEPT |
| a! / c00! | REJECT | REJECT | REJECT |
| a0 / a1 / Zz / a0V | ACCEPT | ACCEPT | ACCEPT |

Two concrete consequences:

1. **Behavioral over-rejection.** The guard fires on `docs.scene.edit`, which upserts *full* elements. An agent that reads an existing element whose index the backend already accepts (`a00`, or the `r00000003` artifact itself), edits another field, and re-upserts, now gets the whole batch rejected at the CLI though the backend would accept it. The loose FE/BE regex never imposes that.
2. **The package comment now makes a new inaccurate claim.** It correctly dropped the old "faithful port" line, but the opening now reads: *"so the CLI rejects locally exactly what they reject server-side."* Given the CLI rejects strictly *more* than FE/BE, this is false in the opposite direction — the same class of comment inaccuracy the previous CR flagged, just inverted.

### What I want before approval

Do **not** loosen the CLI to match the regex — that would re-admit the `r00000003` artifact this whole effort exists to stop. The strict validator is the *right* behavior for a write-side guard. The real inconsistency is that FE/BE `isValidIndex` is too loose. So:

1. Fix the package-comment sentence so it stops asserting exact server-side parity. State instead that the CLI is an intentional **strict superset** of the FE/BE `isValidIndex` regex, enforcing the reference `validateOrderKey` rules (integer-length, trailing-zero, reserved lower bound) that the loose regex omits.
2. If true three-repo parity is genuinely the goal, that has to be closed on the **FE/BE side** by tightening `isValidIndex` — file that as a separate follow-up. As shipped, FE and BE still accept the exact `r00000003` corruption this PR targets.

Once the comment reflects reality and the FE/BE gap is explicitly adjudicated (kept-as-strict-superset, with a FE/BE follow-up filed), I'm happy to approve the CLI leg — the code itself is sound.

OctoBoooot (commented):
## On jeff-wilson2010's over-rejection / docstring finding (#80 @aa84abae)

Checked against the reference source. jeff raises a fair *documentation* point but the 🔴 severity doesn't hold from repo-visible evidence. Verdict stays APPROVE.

### What's verifiable
The reference `rocicorp/fractional-indexing` `validateOrderKey` does **not** charset-check the fractional part — it only enforces: not-smallest-integer, integer part length/head, and no trailing `'0'`. So the CLI's new full-key base62 loop **is stricter than the reference library** on one axis: e.g. `"a0!"` — reference *accepts* (fractional `!` unchecked), CLI *rejects*.

### What's NOT verifiable from this repo (the crux of jeff's 🔴)
Whether that stricter-than-the-lib behavior is *inconsistent with FE/BE*. The docstring asserts FE/BE apply `INDEX_RE = /^[A-Za-z0-9]+$/` to the whole key — if that's accurate, the CLI **matches** FE/BE and there is no over-rejection. jeff asserts FE/BE are laxer (so the CLI over-rejects). Both are claims about an external system neither of us can see from this repo; they directly conflict. I won't assert either as fact.

### Severity calibration
- **"CLI over-rejects vs FE/BE" — question, not blocker.** It depends on unverified FE/BE behavior, and the *only* keys it could wrongly reject are those with a **non-base62 char in the fractional part** — malformed keys `fractional-indexing` never emits, which the PR is arguably right to reject locally. Real-world over-reject surface ≈ zero. The action is to *confirm with the author/backend* whether FE/BE truly enforce `INDEX_RE`, not to block. _(confidence: high that the impact is negligible; the FE/BE parity itself is unverifiable from here.)_
- **Docstring "rejects locally exactly what they reject server-side" — nit.** jeff's underlying point is fair: an *exact* three-way-parity claim is stronger than can be byte-confirmed from this repo. If FE/BE parity isn't verified to the byte, soften to "at least as strict as FE/BE (a superset of the reference library's rules)." Documentation-precision nit, not a code defect.

### Net
No sustainable blocker on repo-visible evidence, so **APPROVE stands**. The trailing-zero fix is verified real; the integer-part/full-key charset is correct *if* `INDEX_RE` is the true FE/BE rule (author's cited authority) and harmless even if slightly stricter (only rejects non-base62 garbage). Two concrete asks for the author, neither merge-blocking from my side: (1) confirm FE/BE enforce `INDEX_RE` on the whole key; (2) if you can't confirm exact parity, reword the docstring's "exactly" to "superset / at least as strict." Credit jeff-wilson2010 for pushing on the parity-direction precision — it's the right instinct even where I can't sustain it as a blocker.

yujiawei (approved):
# Code Review — PR #80 (octo-cli)

**Verdict: APPROVE.** Correct, well-scoped fix. All findings below are advisory (P2/nits); none block merge.

This review was done against head `aa84aba`, diffed against `merge-base(HEAD, main)`. `go build ./...`, `go vet ./...`, and `go test ./... -count=1` are all green locally.

## Spec compliance

The change does exactly what XIN-796 asks and nothing more:

- **Implemented as required:** a dependency-free `validateOrderKey` port (`internal/fracindex/fracindex.go`), spec-gated validation via the new `x-octo-validate-elements-index` extension on `docs.scene.edit` (`internal/registry/specs/docs.json:452`, parsed in `internal/registry/loader.go:328`), the `runOperation` hook (`cmd/service/run.go:75-82`), and the `octo-docs` SKILL example/doc update.
- **No over-build:** validation is gated by a spec flag rather than a hard-coded operationId carve-out, so it only fires for `docs.scene.edit`. Delete-only batches are untouched. The recommended server-side `422 board_element_invalid` guard is correctly left out of scope and only noted for reviewers.
- **No deviation:** the approach matches the PR description.

## Code quality

### Correctness of the fractional-index port — verified sound

The port mirrors the reference `fractional-indexing` `validateOrderKey`:
- integer-part length is derived from the head byte (`a..z` → 2..27, `A..Z` → 27..2) and truncation is rejected (`fracindex.go:60-66`);
- the reserved smallest integer `A00…0` (27 chars, matching head `A`) is rejected (`fracindex.go:56-58`);
- the trailing-`0` rule is applied only to the fractional part (`fracindex.go:80-82`), so `a0` (integer part only) stays valid while `a00`/`a0V0` are rejected.

The one intentional divergence — checking base-62 charset across the **whole** key rather than only the fractional part (`fracindex.go:71-75`) — is justified: it matches the frontend/backend `INDEX_RE` (`/^[A-Za-z0-9]+$/`). Since real keys are always produced by `generateKeyBetween` (base-62 only), this can only reject hand-crafted garbage, never a legitimately generated key. I spot-checked head/length boundaries (`A0`, `zz`, `a0z`, `Zz`, the `r00000003` artifact) and behavior is correct.

### Validation wiring — no bypass

I traced the body path. `elements` is an array-of-object property, so `promotableKind` (`cmd/service/flags.go:230-243`) returns `(0,false)` and it is **never** registered as a promotable body flag — it can only arrive through `--data`. `--data` is unmarshalled into a `map[string]any` in `resolveBody`, which yields `elements` as `[]any` of `map[string]any`. A top-level JSON array (`--data '[...]'`) is rejected earlier in `resolveBody` (`--data is not a JSON object`) before `validateElementsIndex` ever runs. Validation also runs before the dry-run and `--page-all` branches, so neither path skips it.

### Advisory findings (P2 — non-blocking)

- **[P2] `cmd/service/validate.go:24-27` — fail-open on non-object body is defensive but unreachable.** If the parsed body is not a `map[string]any`, the validator returns `nil`. For the realistic case (a JSON array via `--data`), `resolveBody` already rejects it upstream, so this branch can't actually let a bad element through today. It's fine as defense-in-depth; worth a one-line comment noting the upstream guard so a future refactor of `resolveBody` doesn't silently open a hole.

- **[P2] `cmd/service/validate.go:30-36` — `{"elements": null}` is rejected locally as `elements must be a JSON array`.** This is acceptable (the spec types `elements` as an array, and a delete-only batch should omit the key, as the new `TestDocsSceneEdit_DeleteOnlyBatchAllowed` does). Flagging only so it's a conscious choice: if any caller currently sends `elements: null` to mean "no upserts", they'll now get a client-side error instead of the previous pass-through.

- **[P2] test coverage — add a direct loader assertion + integer-length boundary cases.** The e2e tests transitively prove `ValidateElementsIndex` loads as `true` (the reject tests would fail otherwise), but an explicit `OperationDetail.ValidateElementsIndex == true` assertion in a loader test would guard against a future spec-serialization change silently disabling the gate. Similarly, `fracindex_test.go` covers the important shapes but not the exact head/length boundaries (`A`/`Z`, `a`/`z`, exact-length vs one-short) that are easiest to regress.

## Coverage note (what this review checked beyond the diff)

I verified `resolveBody`'s concrete output types, that `elements` is not a promotable flag, that a top-level array is rejected before validation, `boolOf` handling of the native-bool spec value (transitively via the passing e2e reject tests), and that dry-run / `--page-all` do not bypass validation. Not independently re-derived: the exact backend `INDEX_RE`/`validateOrderKey` source (validated against the reference `fractional-indexing` library and the PR's own citations).

Nice fix — closing a real board-corruption path client-side with a clean, spec-driven gate.

jeff-wilson2010 (approved):
## Re-review (corrected) — head `aa84abae98ac4273151a08c27d73d0f57083625a`

Verdict: **APPROVED**

This supersedes my earlier CHANGES_REQUESTED on this same head. That verdict was wrong, and the error was mine: I compared this PR against FE/BE **`main`**, which still carries the old loose `INDEX_RE = /^[A-Za-z0-9]+$/` gate. The coordinated hardening lives in the in-flight PRs, and I should have reviewed against those heads. Re-verified against the correct sources:

- FE `octo-web` PR #622 @ `35c7661c` — `packages/whiteboard-schema/src/normalize.ts:59` `isValidOrderKey` = full-key `BASE62_RE` + `orderKeyIntegerLength` head/length check (L52) + `SMALLEST_INTEGER` reject (L61) + trailing-zero reject (L65); `isValidIndex` (L89) delegates to it.
- BE `octo-docs-backend` PR #52 @ `e8791ebb` — `src/whiteboard/schema/normalize.ts:71` `isValidIndex` = `BASE62_CHARS` charset gate + `generateKeyBetween(v, null)` structural probe (throws on illegal jitterbug structure).

I ran all 11 cases through the *actual* FE function and the *actual* `fractional-indexing@4` probe the BE imports (not the descriptive comments). Every repo agrees:

| key | CLI @aa84abae | FE @35c7661c | BE @e8791ebb |
|---|---|---|---|
| r00000003 | REJECT | REJECT | REJECT |
| a! | REJECT | REJECT | REJECT |
| c00! | REJECT | REJECT | REJECT |
| a00 | REJECT | REJECT | REJECT |
| a10 | REJECT | REJECT | REJECT |
| a0V0 | REJECT | REJECT | REJECT |
| A0…0 (sentinel) | REJECT | REJECT | REJECT |
| a0 | ACCEPT | ACCEPT | ACCEPT |
| a1 | ACCEPT | ACCEPT | ACCEPT |
| Zz | ACCEPT | ACCEPT | ACCEPT |
| a0V | ACCEPT | ACCEPT | ACCEPT |

**11/11 identical across all three repos** — byte-for-byte parity is met. The three implementations reach it by different routes (Go port / TS port / `generateKeyBetween` probe), but the accept/reject boundary is the same, including the intended reject of the `r00000003` XIN-792 artifact.

Consequently the two objections in my earlier review are withdrawn:
- The package-comment line "the CLI rejects locally exactly what they reject server-side" is **accurate** against the coordinated FE/BE heads — not a false claim.
- There is no over-rejection relative to the backend: FE/BE PR heads reject the same keys, so a `docs.scene.edit` batch the CLI blocks would also be stripped/rejected server-side.

The CLI leg itself: three rework changes correct and verified to sink, tests cover the new reject cases + regressions, `go build/vet/test` green, scope confined to `fracindex.go` + test. Approving.

PR #64, opened 2026-06-27
ci: decommission multica triage workflows

lml2468:
Removes the multica triage chain caller workflows: `octo-issue-feed.yml`, `octo-pr-review-feed.yml`, `octo-pr-result-notify.yml`, `octo-issue-notify.yml` (whichever exist in this repo). The triage path is being replaced with a loop engineering pattern; multica autopilots, skills, and secrets are being removed in parallel. Sister change: org-wide reusable workflow PR in `Mininglamp-OSS/.github`.

PR #85, opened 2026-07-15
feat(docs): add docs.share get/set for programmatic share scope

boris-clark:
## What

Adds programmatic control of a document's **space-level share scope** to `octo-cli`, so a bot can read and set who in a space can reach a doc — not just read it. Two new spec-driven operations under the existing docs surface:

- `docs share get <docId>` → `{docId, shareScope, shareRole}` (needs reader)
- `docs share set <docId> --scope restricted|anyone_in_space [--role read|edit]` (needs admin)

Until now `octo-cli` could manage explicit members (`docs members set/remove/list`) but had no share-scope operation.

## Contract

Byte-aligned with the docs-backend handler on `GET/PUT /v1/bot/docs/{docId}/share`:

- Scope enum: `restricted` | `anyone_in_space`; role enum: `read` | `edit`.
- `--role` is **required** with `--scope anyone_in_space` (missing/invalid → `400 invalid_role`).
- `--role` is **ignored** with `--scope restricted` and normalized to `read` by the backend, so it may be omitted.
- The CLI stays a thin transport: scope/role validation and `invalid_scope` / `invalid_role` / `403` / `404` / `409` all remain server-side, matching the members pattern.
- The endpoint is served on the bot mount (`/v1/bot/docs/*`) because docs-backend re-mounts the same `docsRouter` (which owns `/:docId/share`) behind the bot identity middleware — no human endpoint or auth bypass involved.

## Implementation notes

`octo-cli` is metadata-driven, so the commands are added purely in `internal/registry/specs/docs.json`. One small enabling change: `x-octo-flag` — which already aliases query/header params — now also applies to **request-body properties**, so the clean caller-facing flags `--scope` / `--role` front the exact wire body keys `shareScope` / `shareRole`. The property name always stays the JSON body key; only the flag label changes.

## Tests

- Registry: operation counts, method/path shape, and the new body-property flag alias.
- Service engine: `get` read path; `set` with `--scope restricted` (body omits `shareRole`); `set` with `--scope anyone_in_space --role edit`; and that the leaf exposes `--scope`/`--role` (not the camelCase property names) with the aliases never leaking into the body.
- Full suite green (`go build ./... && go vet ./... && go test ./...`), `gofmt` clean.

## Related

- octo-docs #64 (space-scoped document share permissions)
- Depends on the docs-backend share endpoint (`feat/space-scoped-share-permissions`); enums/normalization/error codes are matched to that handler.

Draft — opened to carry review before it goes in.

### DIFF
--- a/internal/registry/specs/docs.json
+++ b/internal/registry/specs/docs.json
@@ -593,6 +593,79 @@
         "responses": { "200": { "description": "Removed" } }
       }
     },
+    "/v1/bot/docs/{docId}/share": {
+      "get": {
+        "operationId": "docs.share.get",
+        "summary": "Read a document's space-level share scope (needs reader)",
+        "description": "Returns the current share settings any viewer can see: {docId, shareScope, shareRole}. shareScope is 'restricted' (only owner/members) or 'anyone_in_space' (every member of the doc's space gets the share role). shareRole is meaningful only when shareScope is anyone_in_space; a restricted doc always reports shareRole 'read'.",
+        "x-octo-risk": "read",
+        "parameters": [
+          { "name": "docId", "in": "path", "required": true, "schema": { "type": "string" } }
+        ],
+        "responses": {
+          "200": {
+            "description": "Current share settings",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "docId": { "type": "string" },
+                    "shareScope": { "type": "string", "enum": ["restricted", "anyone_in_space"] },
+                    "shareRole": { "type": "string", "enum": ["read", "edit"] }
+                  }
+                }
+              }
+            }
+          }
+        }
+      },
+      "put": {
+        "operationId": "docs.share.set",
+        "summary": "Set a document's space-level share scope (needs admin)",
+        "description": "Change the share scope so either only owner/members can access (--scope restricted) or every member of the doc's space gets a share role (--scope anyone_in_space --role read|edit). Normalization matches the backend: with --scope anyone_in_space a valid --role is required (a missing/invalid role is 400 invalid_role); with --scope restricted the role is ignored and stored as read, so --role may be omitted. An unknown scope is 400 invalid_scope; a non-admin caller is 403; a missing/cross-space doc is 404; an archived doc is 409. Share only ever RAISES access (owner/members are never downgraded).",
+        "x-octo-risk": "write",
+        "parameters": [
+          { "name": "docId", "in": "path", "required": true, "schema": { "type": "string" } }
+        ],
+        "requestBody": {
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "object",
+                "required": ["shareScope"],
+                "properties": {
+                  "shareScope": { "type": "string", "x-octo-flag": "scope", "enum": ["restricted", "anyone_in_space"], "description": "share scope to set" },
+                  "shareRole": { "type": "string", "x-octo-flag": "role", "enum": ["read", "edit"], "description": "share role for anyone_in_space (required when scope=anyone_in_space; ignored and stored as read when scope=restricted)" }
+                }
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "Updated share settings",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "docId": { "type": "string" },
+                    "shareScope": { "type": "string", "enum": ["restricted", "anyone_in_space"] },
+                    "shareRole": { "type": "string", "enum": ["read", "edit"] }
+                  }
+                }
+              }
+            }
+          },
+          "400": { "description": "invalid_scope / invalid_role" },
+          "403": { "description": "forbidden (not admin/owner)" },
+          "404": { "description": "not_found (missing or cross-space doc)" },
+          "409": { "description": "conflict (archived doc)" }
+        }
+      }
+    },
     "/v1/bot/docs/{docId}/forward-grant": {
       "post": {
         "operationId": "docs.forward-grant",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **`octo-cli docs share get|set`** — program the space-level share scope of a
+  document. `docs share get <docId>` reads the current `{docId, shareScope,
+  shareRole}` (needs reader); `docs share set <docId> --scope
+  restricted|anyone_in_space [--role read|edit]` changes it (needs admin). The
+  scope enum is `restricted` / `anyone_in_space` and the role enum is `read` /
+  `edit`; a valid `--role` is required with `--scope anyone_in_space` and is
+  ignored (stored as `read`) with `--scope restricted`, matching the
+  docs-backend `/v1/bot/docs/{docId}/share` contract byte-for-byte. Backed by a
+  new `x-octo-flag` alias on request-body properties, so the clean `--scope` /
+  `--role` flags front the `shareScope` / `shareRole` wire keys.
+
 ### Changed
 - **Renamed the binary and CLI command from `octo` to `octo-cli`** for
   consistency with the repository and Go module name. This affects every
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   restricted|anyone_in_space [--role read|edit]` changes it (needs admin). The
   scope enum is `restricted` / `anyone_in_space` and the role enum is `read` /
   `edit`; a valid `--role` is required with `--scope anyone_in_space` and is
-  ignored (stored as `read`) with `--scope restricted`, matching the
-  docs-backend `/v1/bot/docs/{docId}/share` contract byte-for-byte. Backed by a
+  ignored (stored as `read`) with `--scope restricted`. This targets the
+  planned `/v1/bot/docs/{docId}/share` endpoint (backend lands in
+  octo-docs-backend#68); merge this after that endpoint ships. Backed by a
   new `x-octo-flag` alias on request-body properties, so the clean `--scope` /
   `--role` flags front the `shareScope` / `shareRole` wire keys.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -22,7 +22,7 @@
 - Each Bot has an **owner**; operations are attributed to the Bot identity. For LLM-backed paths (`matter extract`) the bot acts on behalf of its owner — pass `owner_uid` as `creator_uid`.
 - `OCTO_SPACE_ID` (or `--space`) supplies space context for platform-scoped bots. Space-scoped bots resolve their space server-side.
 
-## Command Structure (8 domains, 75 operations / 78 commands incl. 3 matter transition aliases)
+## Command Structure (8 domains, 77 operations / 80 commands incl. 3 matter transition aliases)
 
 Service commands are auto-registered. The hand-written leaves are `schema`, `version`, `api` (generic passthrough), `config`, `auth`, and the cobra-generated `completion`.
 
@@ -51,6 +51,7 @@ octo-cli docs      create | list | get | rename | delete | forward-grant
                sheet    get|edit
                scene    get|edit
                members  list|set|remove
+               share    get|set
                comments list|add|edit|delete
                versions list|create|state|rename|delete|restore
                attachments presign|get|resolve
--- a/cmd/service/docs_test.go
+++ b/cmd/service/docs_test.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"encoding/json"
+	"io"
 	"net/http"
 	"net/url"
 	"strings"
@@ -377,11 +378,14 @@ func TestDocsMembersRemove_TwoPathArgs(t *testing.T) {
 }
 
 // TestDocsShareGet_ReadPath checks docs.share.get is a plain GET on the
-// /share sub-resource with the docId in the path and no body.
+// /share sub-resource with the docId in the path and no request body — a read
+// must not ship a payload, so the handler asserts the received body is empty.
 func TestDocsShareGet_ReadPath(t *testing.T) {
 	var gotMethod, gotPath string
+	var gotBody []byte
 	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
 		gotMethod, gotPath = r.Method, r.URL.Path
+		gotBody, _ = io.ReadAll(r.Body)
 		w.WriteHeader(200)
 		_, _ = w.Write([]byte(`{"docId":"d1","shareScope":"restricted","shareRole":"read"}`))
 	})
@@ -391,6 +395,9 @@ func TestDocsShareGet_ReadPath(t *testing.T) {
 	}
 	if gotMethod != "GET" || gotPath != "/v1/bot/docs/d1/share" {
 		t.Errorf("got %s %s, want GET /v1/bot/docs/d1/share", gotMethod, gotPath)
+	}
+	if len(gotBody) != 0 {
+		t.Errorf("GET must send an empty request body; got %q", gotBody)
 	}
 }
 
--- a/cmd/service/docs_test.go
+++ b/cmd/service/docs_test.go
@@ -38,6 +38,7 @@ func TestDocs_TreeShape(t *testing.T) {
 		"sheet":       {"get", "edit"},
 		"scene":       {"get", "edit", "export"},
 		"members":     {"list", "set", "remove"},
+		"share":       {"get", "set"},
 		"comments":    {"list", "add", "edit", "delete"},
 		"versions":    {"list", "create", "state", "rename", "delete", "restore"},
 		"attachments": {"presign", "get", "resolve"},
@@ -79,6 +80,8 @@ func TestDocs_RegistryShape(t *testing.T) {
 		"docs.members.list":        {"GET", "/v1/bot/docs/{docId}/members"},
 		"docs.members.set":         {"PUT", "/v1/bot/docs/{docId}/members"},
 		"docs.members.remove":      {"DELETE", "/v1/bot/docs/{docId}/members/{uid}"},
+		"docs.share.get":           {"GET", "/v1/bot/docs/{docId}/share"},
+		"docs.share.set":           {"PUT", "/v1/bot/docs/{docId}/share"},
 		"docs.forward-grant":       {"POST", "/v1/bot/docs/{docId}/forward-grant"},
 		"docs.comments.list":       {"GET", "/v1/bot/docs/{docId}/comments"},
 		"docs.comments.add":        {"POST", "/v1/bot/docs/{docId}/comments"},
@@ -370,6 +373,103 @@ func TestDocsMembersRemove_TwoPathArgs(t *testing.T) {
 	}
 	if gotMethod != "DELETE" || gotPath != "/v1/bot/docs/d1/members/u9" {
 		t.Errorf("got %s %s, want DELETE /v1/bot/docs/d1/members/u9", gotMethod, gotPath)
+	}
+}
+
+// TestDocsShareGet_ReadPath checks docs.share.get is a plain GET on the
+// /share sub-resource with the docId in the path and no body.
+func TestDocsShareGet_ReadPath(t *testing.T) {
+	var gotMethod, gotPath string
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotMethod, gotPath = r.Method, r.URL.Path
+		w.WriteHeader(200)
+		_, _ = w.Write([]byte(`{"docId":"d1","shareScope":"restricted","shareRole":"read"}`))
+	})
+	root.SetArgs([]string{"docs", "share", "get", "d1"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if gotMethod != "GET" || gotPath != "/v1/bot/docs/d1/share" {
+		t.Errorf("got %s %s, want GET /v1/bot/docs/d1/share", gotMethod, gotPath)
+	}
+}
+
+// TestDocsShareSet_RestrictedBody checks --scope restricted maps to a PUT on
+// /share with the byte-exact {shareScope} wire key and no shareRole (a
+// restricted doc ignores the role; the backend normalizes it to read).
+func TestDocsShareSet_RestrictedBody(t *testing.T) {
+	var gotMethod, gotPath string
+	var gotBody map[string]any
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotMethod, gotPath = r.Method, r.URL.Path
+		_ = json.NewDecoder(r.Body).Decode(&gotBody)
+		w.WriteHeader(200)
+		_, _ = w.Write([]byte(`{"docId":"d1","shareScope":"restricted","shareRole":"read"}`))
+	})
+	root.SetArgs([]string{"docs", "share", "set", "d1", "--scope", "restricted"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if gotMethod != "PUT" || gotPath != "/v1/bot/docs/d1/share" {
+		t.Errorf("got %s %s, want PUT /v1/bot/docs/d1/share", gotMethod, gotPath)
+	}
+	if gotBody["shareScope"] != "restricted" {
+		t.Errorf("body shareScope = %v, want restricted", gotBody["shareScope"])
+	}
+	if _, present := gotBody["shareRole"]; present {
+		t.Errorf("body must omit shareRole when --role is not passed; got %v", gotBody)
+	}
+}
+
+// TestDocsShareSet_AnyoneEditBody checks --scope anyone_in_space --role edit
+// maps both flags to their wire keys (shareScope / shareRole), proving the
+// x-octo-flag aliases (--scope, --role) do not leak into the body.
+func TestDocsShareSet_AnyoneEditBody(t *testing.T) {
+	var gotBody map[string]any
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		_ = json.NewDecoder(r.Body).Decode(&gotBody)
+		w.WriteHeader(200)
+		_, _ = w.Write([]byte(`{"docId":"d1","shareScope":"anyone_in_space","shareRole":"edit"}`))
+	})
+	root.SetArgs([]string{"docs", "share", "set", "d1", "--scope", "anyone_in_space", "--role", "edit"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if gotBody["shareScope"] != "anyone_in_space" || gotBody["shareRole"] != "edit" {
+		t.Errorf("body = %v, want {shareScope:anyone_in_space, shareRole:edit}", gotBody)
+	}
+	// The flag aliases must never appear as body keys.
+	if _, bad := gotBody["scope"]; bad {
+		t.Errorf("body leaked the --scope flag alias as a wire key: %v", gotBody)
+	}
+	if _, bad := gotBody["role"]; bad {
+		t.Errorf("body leaked the --role flag alias as a wire key: %v", gotBody)
+	}
+}
+
+// TestDocsShareSet_FlagAliases confirms the leaf exposes --scope / --role (from
+// x-octo-flag) and NOT the camelCase property names, while --data stays
+// available for the raw-body escape hatch.
+func TestDocsShareSet_FlagAliases(t *testing.T) {
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {})
+	docs := findCmd(root, "docs")
+	share := findCmd(docs, "share")
+	if share == nil {
+		t.Fatal("missing docs share resource group")
+	}
+	set := findCmd(share, "set")
+	if set == nil {
+		t.Fatal("missing docs share set leaf")
+	}
+	for _, want := range []string{"scope", "role", "data"} {
+		if set.Flags().Lookup(want) == nil {
+			t.Errorf("docs share set: expected --%s flag", want)
+		}
+	}
+	for _, bad := range []string{"shareScope", "shareRole", "share-scope", "share-role"} {
+		if set.Flags().Lookup(bad) != nil {
+			t.Errorf("docs share set: --%s must NOT exist (aliased to --scope/--role)", bad)
+		}
 	}
 }
 
--- a/cmd/service/flags.go
+++ b/cmd/service/flags.go
@@ -187,7 +187,14 @@ func registerBodyFlags(cmd *cobra.Command, rt *operationRuntime, d *registry.Ope
 		if !ok {
 			continue
 		}
-		flagName := strings.ReplaceAll(name, "_", "-")
+		// Flag name: an explicit x-octo-flag override on the property wins (so a
+		// clean --scope can front a shareScope wire key); otherwise derive it from
+		// the property name with underscores turned into dashes. The wire body key
+		// (bf.apiName) always stays the property name regardless of the flag alias.
+		flagName := prop.FlagName
+		if flagName == "" {
+			flagName = strings.ReplaceAll(name, "_", "-")
+		}
 		// Avoid collisions with --data, --file, a query param, or a spec-declared
 		// header flag of the same name (e.g. an If-Match header exposed as
 		// --base-version takes precedence over a body baseVersion mirror).
--- a/internal/registry/loader.go
+++ b/internal/registry/loader.go
@@ -172,6 +172,13 @@ type SchemaInfo struct {
 	MaxLength   int                   `json:"max_length,omitempty"`
 	MaxItems    int                   `json:"max_items,omitempty"`
 	Ref         string                `json:"$ref,omitempty"`
+	// FlagName is the optional CLI flag override from the x-octo-flag extension
+	// on a request-body property, mirroring ParamInfo.FlagName for query/header
+	// params. It lets a promoted body field expose a clean flag name (e.g.
+	// --scope) while the wire body key stays the property name (e.g. shareScope),
+	// so the CLI can honour a caller-facing flag without diverging from the
+	// byte-exact backend contract. Empty means "derive from the property name".
+	FlagName string `json:"flag_name,omitempty"`
 }
 
 // PaginationInfo captures the `x-octo-pagination` extension. It tells the
@@ -506,6 +513,7 @@ func resolveSchemaWithDepth(doc, s map[string]any, depth int) SchemaInfo { //nol
 		Type:        stringOf(s["type"]),
 		Format:      stringOf(s["format"]),
 		Description: stringOf(s["description"]),
+		FlagName:    stringOf(s["x-octo-flag"]),
 	}
 	if req, ok := s["required"].([]any); ok {
 		for _, x := range req {
--- a/internal/registry/loader_test.go
+++ b/internal/registry/loader_test.go
@@ -41,7 +41,7 @@ func TestAllDomainOperationCounts(t *testing.T) {
 		"file":    4,
 		"bot":     6,
 		"event":   2,
-		"docs":    29,
+		"docs":    31,
 	}
 	totalWant := 0
 	for svc, want := range expected {
@@ -244,6 +244,33 @@ func TestQueryParamFlagAliasAvoidsGlobalCollision(t *testing.T) {
 	}
 	if found.Name != "format" {
 		t.Errorf("wire query param name = %q, want format (must be preserved)", found.Name)
+	}
+}
+
+// TestBodyPropertyFlagAlias pins docs.share.set: its shareScope/shareRole body
+// properties carry x-octo-flag scope/role so the CLI exposes clean --scope /
+// --role flags while the wire body keys stay the byte-exact backend contract
+// (shareScope / shareRole). The property name is never renamed — only the flag
+// alias is added.
+func TestBodyPropertyFlagAlias(t *testing.T) {
+	r := MustNew()
+	op, ok := r.GetOperation("docs.share.set")
+	if !ok {
+		t.Fatal("docs.share.set not found")
+	}
+	if op.RequestBody == nil || op.RequestBody.Properties == nil {
+		t.Fatal("docs.share.set: expected a request body with properties")
+	}
+	cases := map[string]string{"shareScope": "scope", "shareRole": "role"}
+	for prop, wantFlag := range cases {
+		p, ok := op.RequestBody.Properties[prop]
+		if !ok {
+			t.Errorf("docs.share.set: missing body property %q", prop)
+			continue
+		}
+		if p.FlagName != wantFlag {
+			t.Errorf("%s flag alias = %q, want %q (from x-octo-flag)", prop, p.FlagName, wantFlag)
+		}
 	}
 }
 
--- a/skills/octo-docs/common.md
+++ b/skills/octo-docs/common.md
@@ -116,11 +116,29 @@ octo-cli docs members remove <docId> <uid>                       # admin; owner 
 
 # Grant-only forward access (never downgrades). Only reader|writer are grantable.
 octo-cli docs forward-grant  <docId> --uid <uid> --role reader
+
+# Space-level share scope — read/set who in the space can reach the doc.
+octo-cli docs share get <docId>                                       # reader; {docId, shareScope, shareRole}
+octo-cli docs share set <docId> --scope restricted                    # admin; only owner/members can access
+octo-cli docs share set <docId> --scope anyone_in_space --role read   # admin; every space member gets read
+octo-cli docs share set <docId> --scope anyone_in_space --role edit   # admin; every space member gets edit
 ```
 
 `docs members set` is a PUT-upsert: it adds the member if absent or changes the
 role if already present. The target uid must be a real Octo user — a miss returns
 `404 user_not_found` and writes no ghost member. Schema: `octo-cli schema docs.members.set`.
+
+`docs share set` changes the space-level share scope. Two scopes: `restricted`
+(only the owner and explicit members reach the doc — the default) and
+`anyone_in_space` (every member of the doc's own space is granted the share
+role). The `--role` (`read` | `edit`) is **required** with `--scope
+anyone_in_space` and is **ignored** with `--scope restricted` (the backend
+normalizes it to `read`), so omit `--role` when restricting. Sharing only ever
+*raises* access — an owner or member is never downgraded by a share setting.
+Error codes: `400 invalid_scope` (unknown scope), `400 invalid_role`
+(anyone_in_space with a missing/invalid role), `403` (caller not admin/owner),
+`404` (missing or cross-space doc), `409` (archived doc). Schema:
+`octo-cli schema docs.share.set`.
 
 ---
 


Jerry-Xin (dismissed):
This PR is in scope for `octo-cli` and cleanly adds the docs share operations through the existing metadata-driven command path.

💬 Non-blocking
- 🔵 Suggestion: [cmd/service/docs_test.go](/Users/mlamp/.openclaw/tmp/pr-review-og_q5sl0/repo/cmd/service/docs_test.go:381) says the GET path sends no body, but the assertion only checks method/path. Consider asserting an empty request body if you want that contract pinned explicitly.

✅ Highlights
- The new `docs.share.get` / `docs.share.set` spec entries match the existing docs registry pattern and use the bot-mounted `/v1/bot/docs/{docId}/share` route.
- The body-property `x-octo-flag` implementation preserves wire keys via `bf.apiName`, so `--scope` and `--role` correctly serialize as `shareScope` and `shareRole`.
- Test coverage covers registry shape, alias loading, exposed flags, and request body serialization for both restricted and space-wide share cases.
- Focused tests passed: `go test ./internal/registry ./cmd/service`.

mochashanyao (approved):
[Octo-Q · automated review]

---

---

## Data Flow Traceback

| Consumer | Upstream source | Verified |
|---|---|---|
| `--scope` flag value → body `shareScope` | `flags.go:211` sets `bf.apiName = "shareScope"` (property name); `run.go:168` writes `base[bf.apiName] = *bf.strVal` | ✅ Wire key is always the spec property name, never the flag alias |
| `--role` flag value → body `shareRole` | Same path: `bf.apiName = "shareRole"`, written via `run.go:168` | ✅ |
| `x-octo-flag` JSON extension → `SchemaInfo.FlagName` | `loader.go:516` calls `stringOf(s["x-octo-flag"])`; `stringOf` returns `""` for nil/non-string, so a missing extension falls through to the default name derivation at `flags.go:195` | ✅ |
| `SchemaInfo.FlagName` → cobra flag name | `flags.go:190` reads `prop.FlagName`; if non-empty, uses it directly; otherwise derives from property name | ✅ |

## Blind Spot Checklist (C1–C6)

- **C1 (Dual-path parity)** — Clear. `x-octo-flag` now read for body properties (new) and query/header params (existing `ParamInfo.FlagName` at `loader.go:353`). Symmetric pattern, same collision avoidance in both paths.
- **C2 (Control-flow ordering)** — Clear. Flag name resolution happens once at registration time, not at request time. No ordering or re-entry concern.
- **C3 (Authorization boundary)** — Clear. `docs share set` is `x-octo-risk: "write"`. Auth is server-side (403 for non-admin). CLI does not expose any client-side auth bypass.
- **C4 (Authorization lifecycle / container-member cascade)** — N/A. Share scope is a flat property on a document, not a nested container-member relationship.
- **C5 (Build ≠ runtime correct)** — Clear. No build artifacts, packaging, or environment-dependent behavior changed.
- **C6 (Governance/docs self-consistency)** — Clear. CHANGELOG, CLAUDE.md, skills doc, and spec all agree on enums, error codes, and normalization behavior.

N/A — first review round for this PR.

---

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Data flow from `x-octo-flag` spec extension through loader → flag registration → body serialization is correct and fully tested. Flag collision avoidance is preserved. Authorization is server-side. Documentation is consistent across all surfaces.

Jerry-Xin (dismissed):
## Review: `docs.share get/set` — **CHANGES_REQUESTED**

Reviewed at head `c1aa4328`. The CLI code is clean, follows the metadata-driven registry pattern, and its tests are green. The blocker is not in the Go code — it's the backend endpoint this feature targets, which does **not** exist. Details below, point by point.

### 1. New operations — OK
`docs.share.get <docId>` and `docs.share.set <docId> --scope … [--role …]` are registered purely via `internal/registry/specs/docs.json` and follow the same shape as `docs.members.*` (path-arg `docId`, JSON envelope, GET/PUT). Tree/registry shape is pinned by `cmd/service/docs_test.go` (`TestDocs_TreeShape`, `TestDocs_RegistryShape`) and the domain count bump in `loader_test.go` (`docs: 29 → 31`). No hand-written leaf added — correct for this codebase.

### 2. `x-octo-flag` body-property alias — OK, verified end-to-end
The new `--scope`/`--role` flags front the `shareScope`/`shareRole` wire keys correctly:
- `internal/registry/loader.go` — `SchemaInfo.FlagName` is populated from `x-octo-flag` (`resolveSchemaWithDepth`).
- `cmd/service/flags.go:190-197` — `flagName := prop.FlagName` (fallback to the dashed property name); the `bodyFlag{apiName: name}` at line 211 keeps the **property name** as the wire key.
- `cmd/service/run.go:156-169` — body is assembled as `base[bf.apiName] = *bf.strVal` and only when `Flags().Changed(flagName)`. So `--role` is omitted from the body when not passed, and the flag alias never leaks as a wire key.

This is well tested (`TestDocsShareSet_RestrictedBody`, `TestDocsShareSet_AnyoneEditBody`, `TestDocsShareSet_FlagAliases`, `TestBodyPropertyFlagAlias`). Nice, self-consistent implementation.

### 3. Auth — OK (client side)
No new auth. The endpoints sit under `/v1/bot/docs/…` and would ride the existing bot-token mechanism. Nothing to flag in the CLI.

### 4. Tests — green
`go build ./...` clean; `go test ./...` all pass (including the new `docs_test.go` / `loader_test.go` cases). Coverage matches the request: get, set restricted (no `shareRole`), anyone_in_space+edit, and flag-alias assertions.

### 5. Skill docs — synced
`skills/octo-docs/common.md`, `CLAUDE.md` command tree, and `CHANGELOG.md` are consistent with the actual command form (`docs share get|set`, `--scope`/`--role`, error-code notes). Domain count in `CLAUDE.md` updated (75→77 ops).

### 6. AI traces — clean
Single commit authored/committed by a human account, no `Co-Authored-By`, no "Generated by / AI / Claude" markers.

---

## ⛔ Blocking issue — backend endpoint does not exist

The PR (CHANGELOG, commit message, and `docs.json` descriptions) states the contract is **"byte-for-byte" aligned with the docs-backend `/v1/bot/docs/{docId}/share` handler**, and the upstream task write-back claimed the bot side already exposes this endpoint via a `docsRouter` dual-mount. I verified against `Mininglamp-OSS/octo-docs-backend` (all branches) and **that endpoint does not exist**:

- No `/share` route anywhere. `src/api/routes/docs.ts` `docsRouter` only registers `POST /`, `GET /`, `GET /:docId`, `PATCH /:docId`, `DELETE /:docId` — no `/:docId/share`. There is no separate share router (`grep -rni "shareRouter\|/:docId/share"` → nothing).
- The dual-mount in `src/api/app.ts` is real, but `botApi` re-mounts exactly the existing routers (`docsRouter`, `membersRouter`, … `boardExportRouter`) — none of which serves `/share`. Re-mounting cannot expose a route that was never defined.
- No `shareScope` / `shareRole` / `anyone_in_space` / `restricted` scope enum, and no normalization/error-code logic (`400 invalid_role`, `restricted → read`, `409 archived`), exists in `src/`, `test/`, or `migrations/`. The only "share" code is `src/util/docShareLink.ts`, an unrelated share-**URL** builder.

**Consequences:**
1. `docs share get/set` will return **404** against the current backend — the command is dead on arrival.
2. The "byte-for-byte contract alignment" claim is currently **unverifiable and unfounded**: there is no handler, enum, normalization, or status-code table on the backend to align against. The enums (`restricted`/`anyone_in_space`, `read`/`edit`), the `anyone_in_space`-requires-role rule, the `restricted`→`read` normalization, and the `400/403/404/409` table are all asserted by this PR alone.

## Requested changes
1. Do not merge until the docs-backend share endpoint (XIN-1113 / backend plan B) actually lands on `/v1/bot/docs/{docId}/share` and is verifiable. As a draft this is fine to keep open, but it must not `--approve`/merge ahead of the backend.
2. Once the backend exists, re-verify the wire contract against the real handler — enum spellings, the required-role rule for `anyone_in_space`, the `restricted`→`read` normalization, and the `400 invalid_scope` / `400 invalid_role` / `403` / `404` / `409` status codes — and adjust the spec if anything diverges. The current descriptions are aspirational, not confirmed.
3. Soften the "matches the docs-backend contract byte-for-byte" wording in `CHANGELOG.md` / `docs.json` until it can be backed by shipped backend code, so the claim isn't misleading to consumers.

The CLI implementation itself is solid; the block is purely the missing/unverified backend dependency.

yujiawei (approved):
# Code Review — PR #85 (octo-cli)

**Scope:** `feat(docs): add docs.share get/set for programmatic share scope` — head `c1aa4328`. Two new operations (`docs.share.get` GET/reader, `docs.share.set` PUT/admin) plus a new `x-octo-flag` alias on request-**body** properties, mirroring the pre-existing alias for query/header params. 8 files, +249/-3.

Reviewed against merge-base `0eb40d16`. Build clean (`go build ./...`), `go vet` clean, `gofmt` clean, full `go test ./...` green. Verified end-to-end behavior via `--dry-run`, not just by reading the diff.

## Verdict: Approve ✅

No blocking (P0/P1) issues. This is a clean, well-tested, minimal-surface change that stays byte-exact with the backend contract. The findings below are all non-blocking polish/hardening suggestions.

---

## 1. Contract fidelity — correct

- `docs.share.get`: `GET /v1/bot/docs/{docId}/share`, `x-octo-risk: read`, no request body → no `--data` flag is exposed on the leaf (verified: `docs share get --help` shows zero body flags). `docId` is a path param, not a body field.
- `docs.share.set`: `PUT /v1/bot/docs/{docId}/share`, `x-octo-risk: write`, `required: [shareScope]`, enums `restricted|anyone_in_space` / `read|edit`, response codes `200/400/403/404/409` all documented.
- The `x-octo-flag` aliases (`--scope`, `--role`) never leak into the JSON body. The wire body always uses the property name (`shareScope`/`shareRole`) because `run.go` writes `base[bf.apiName]` where `apiName` is the property name, while `Changed()` is keyed on the flag alias. Confirmed by dry-run:
  - `share set d1 --scope restricted` → body `{"shareScope":"restricted"}` (no `shareRole` injected — omitted flags stay off the wire).
  - `share set d1 --scope anyone_in_space --role edit` → body `{"shareScope":"anyone_in_space","shareRole":"edit"}`.
  - No `scope`/`role` keys ever appear in the body.
- The token stays masked in `--dry-run` output (`Bearer app_***`); the body is user-supplied share config, not a secret.

## 2. Implementation — sound and minimal

- `SchemaInfo.FlagName` (`loader.go`) is populated from `x-octo-flag` on body-property schemas, symmetric with the existing `ParamInfo.FlagName` path for query/header params. The empty-`FlagName` fallback (`strings.ReplaceAll(name, "_", "-")`) preserves behavior for every existing op — verified `docs members set` still derives `--uid`/`--role` from property names.
- Flag/`--data` precedence is correct and unchanged: `--data` seeds the base map, explicit flags override it.

## 3. Non-blocking suggestions (P2)

1. **Collision guard doesn't cover global persistent flags.** `registerBodyFlags` (`cmd/service/flags.go:201`) skips a body flag that collides with `--data`/`--file`/a query flag/a header flag, but not with the root persistent flags (`--format`, `--jq`, `--space`, `--dry-run`, etc. in `cmd/root.go:39-48`). A future body property whose `x-octo-flag` is, say, `format` would silently shadow the global `--format` on that leaf (cobra allows it with the local flag winning — no panic). Note this is **not a regression and does not affect this PR**: `--scope`/`--role` collide with nothing, and the pre-existing `registerQueryFlags` has the *same* gap, so body handling is already at least as strict as query handling. Worth a small future hardening (a shared helper that also checks `cmd.Root().PersistentFlags()`), ideally paired with a guard test — mirroring the existing `TestDocsSceneExport_ImageFormatFlagNoGlobalShadow` intent for the body path.
2. **No client-side validation that `--role` is required when `--scope anyone_in_space`.** Omitting `--role` there sends `{"shareScope":"anyone_in_space"}` and relies on the backend `400 invalid_role`. This is consistent with the CLI's thin-passthrough design (enums aren't validated client-side anywhere), so it's acceptable — just a UX round-trip rather than an early local error.
3. **Test coverage gaps (nice-to-have, not required):** no negative test for `--scope anyone_in_space` without `--role`, no test for an unknown `--scope` value, and no `--data` + flag-override test specific to the new aliased flags (the generic `TestMatterCreate_DataThenFlagOverride` covers the mechanism). The 4 new tests plus `TestBodyPropertyFlagAlias` do exercise the real spec/loader/flag path (no mock seam) and cover the important positive cases.

## 4. Security note (this PR was flagged security-sensitive)

Share scope is access control, so I checked the trust boundary specifically: the CLI makes **no** authz decision locally — admin/reader enforcement, scope/role validation, and the "share only ever raises access" invariant are all backend contracts. The CLI only marshals flags to a byte-exact request body and masks the credential in dry-run. Nothing here weakens the security boundary; the substantive checks live server-side and should be confirmed there (out of scope for this CLI diff).

OctoBoooot (approved):
## Review: feat(docs): add `docs share get|set` for programmatic share scope (#85)

**Verdict:** Approve — the change is spec-driven, self-consistent, and well tested; all code CI is green (`build & test (1.24.x)`, `npm test`, `osv-scan`) and `check-sprint` governance also passes. No blocker or major.

Risk tier: normal. I read the diff plus the loader/flags mechanism it depends on, and verified the PR's own contract claims against the embedded spec rather than trusting the CHANGELOG prose.

**What it does:** two new operations (`docs.share.get` GET, `docs.share.set` PUT) on `/v1/bot/docs/{docId}/share`, generated from `internal/registry/specs/docs.json`. The one genuinely new mechanism is honoring `x-octo-flag` on **request-body properties** (previously only query/header params), so `--scope`/`--role` front the `shareScope`/`shareRole` wire keys.

**Claims checked:**
- *"the flag alias never leaks into the wire body"* → verified in code + test: `flags.go` sets `flagName = prop.FlagName` for the CLI flag but `bf.apiName` (the wire key) stays the property name. `TestDocsShareSet_AnyoneEditBody` asserts `shareScope`/`shareRole` are sent and `scope`/`role` are absent. ✓
- *"`--role` omitted with `--scope restricted` sends no `shareRole`"* → `TestDocsShareSet_RestrictedBody` asserts the key is absent (not sent as empty). ✓
- *"op count docs 29→31, and the tree exposes `share {get,set}`"* → matches `loader_test.go` (docs 31) and `docs_test.go` tree/registry shapes. ✓
- *"clean flags only; no `--shareScope`/`--share-scope`"* → `TestDocsShareSet_FlagAliases` pins both the presence of `--scope/--role/--data` and the absence of the camelCase/dashed variants. ✓

### Parity checked (loader change is non-regressing)
The new `FlagName: stringOf(s["x-octo-flag"])` line in `resolveSchemaWithDepth` reads `x-octo-flag` from inside a **schema** object. I grepped every `x-octo-flag` in all specs: the only other uses are the `format` query param and three `If-Match` **header** params — all at the *parameter* level (read via `ParamInfo.FlagName`), none inside a body-property schema. So no existing operation's flag name changes; the new behavior is isolated to the two new `share.set` body properties. Confirmed by `go build ./...` + `go test ./cmd/service/ ./internal/registry/` green locally.

### Security
CLI is a thin transport (consistent with the `members` pattern): no client-side authz, `docId` passed through the path, scope/role enum + `invalid_scope`/`invalid_role`/`403`/`404`/`409` all enforced server-side. No new client-side privilege path; nothing in the diff bypasses a server check.

### Question (non-blocking — rollout ordering, not a code defect)
- The spec asserts the `/v1/bot/docs/{docId}/share` contract "byte-for-byte." Whether the **backend** service actually serves this route lives in a separate repo (`octo-docs-backend`) and can't be verified from here. If the endpoint isn't deployed yet, `docs share get/set` would 404 at runtime until it ships. That's a release-sequencing question for the maintainer (CLIs commonly land alongside/ahead of the server), not a defect in this CLI change — flagging so merge timing is a conscious call.

### Nits (≤3)
- `cmd/service/docs_test.go` `TestDocsShareGet_ReadPath` — the comment says "no body" but the test only asserts method/path. A one-line `r.ContentLength == 0` (or empty-body) assertion would pin the GET-has-no-body contract. _(confidence: high)_
- The "`--role` required when `--scope anyone_in_space`" rule is server-side only (`required:["shareScope"]` in the spec). That matches the thin-transport design, but a client-side pre-check would turn a round-trip `400 invalid_role` into an instant local error. Optional. _(confidence: high)_

### Praise
- Generalizing `x-octo-flag` from params to body properties is the right shape: it lets the CLI present clean, caller-facing flags (`--scope`/`--role`) without renaming the wire keys, so the byte-exact backend contract is preserved *and* the flag surface stays ergonomic. And it shipped with a test (`TestBodyPropertyFlagAlias`) that pins exactly that separation — flag alias vs. wire key — which is the property most likely to silently regress later. _(confidence: high)_

Jerry-Xin (changes requested):
Summary: The CLI wiring for `docs share get|set` is clean, follows the metadata-driven registry pattern, and is well-tested — but the feature is dead-on-arrival because the backend endpoint it targets does not exist yet. Requesting changes until the backend contract is real.

🔴 Blocking
- **Target endpoint does not exist.** `docs.share.get` / `docs.share.set` are modeled on `GET|PUT /v1/bot/docs/{docId}/share` (`internal/registry/specs/docs.json`, `cmd/service/docs_test.go`), but `octo-docs-backend` `docsRouter` (`src/api/routes/docs.ts`) defines only the root `POST`/`GET` and `GET|PATCH|DELETE /:docId` — there is no `/share` route anywhere in that repo, and a code search for `shareScope` returns zero hits. As shipped, every `docs share get|set` call will 404. The green test suite does not catch this: every test mocks the HTTP server locally, so it verifies method/path/body shaping against a stub, not against a real backend route.
- **No backend enum / normalization / error-code support to align against.** The PR encodes semantics — `shareScope` (`restricted` / `anyone_in_space`), `shareRole` (`read` / `edit`), `--role` required with `anyone_in_space`, `restricted` normalized to `read`, and the implied 400/403/404/409 behaviors — but none of that exists server-side. There is nothing to align the CLI contract to, so these choices are currently unverifiable assumptions.
- **Unfounded "byte-for-byte contract" claim.** `CHANGELOG.md` (and the docs entry) states the behavior matches "the docs-backend `/v1/bot/docs/{docId}/share` contract byte-for-byte." Since that route/contract is not shipped in the backend, this wording is aspirational, not factual. Please soften it (e.g. "targets the planned `/v1/bot/docs/{docId}/share` endpoint") until it is backed by merged backend code.

💬 Non-blocking
- Once the backend lands, add an integration-level check (or at least a note) so the CLI contract is validated against the real route, not only a local stub.

✅ Highlights
- The `x-octo-flag` body-property alias is a clean way to front `shareScope` / `shareRole` wire keys with `--scope` / `--role` flags without leaking flag names into the JSON body.
- The new operations reuse the existing registry-driven command path rather than hand-rolling a new leaf — good consistency with the rest of the `docs` surface.
- Test coverage is thorough for what it targets: tree shape, registry method/path, restricted-body omits `shareRole`, anyone+edit maps both keys, and exposed flag names (`cmd/service/docs_test.go`, `internal/registry/loader_test.go`). This is exactly the wiring to keep once the backend exists.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #85 (octo-cli)

## Summary

This PR adds two new CLI operations — `docs share get` and `docs share set` — for reading and setting a document's space-level share scope. It introduces a new spec entry in `docs.json` for `/v1/bot/docs/{docId}/share` (GET + PUT), extends the registry schema resolver to capture `x-octo-flag` on request-body properties (mirroring the existing param-level capability), and wires the flag-name override into `registerBodyFlags` so `--scope` / `--role` front the `shareScope` / `shareRole` wire keys. Tests, docs, and operation counts are all updated consistently.

## Verification

Local checks at head `c1aa4328`:
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./...` — pass (all packages)

- ✅ **Wire-key isolation** — `resolveBody` in `run.go:156` uses `bf.apiName` (property name, e.g. `shareScope`) as the JSON key regardless of the flag alias. `TestDocsShareSet_AnyoneEditBody` asserts `scope` / `role` never leak as body keys.
- ✅ **Flag alias correctness** — `TestDocsShareSet_FlagAliases` proves `--scope` / `--role` exist and `--shareScope` / `--shareRole` / `--share-scope` / `--share-role` do not.
- ✅ **Restricted-scope body shape** — `TestDocsShareSet_RestrictedBody` asserts `shareRole` is absent when `--role` is not passed, matching the backend contract.
- ✅ **Schema parity** — `SchemaInfo.FlagName` mirrors `ParamInfo.FlagName`; both read `x-octo-flag`, both fall back to the derived name. `TestBodyPropertyFlagAlias` pins the alias values.
- ✅ **Collision guard** — `registerBodyFlags` checks `data`, `file`, query, and header flags before registration. No collision for `scope` / `role` in this operation (only path param `docId`).
- ✅ **Operation counts** — `TestAllDomainOperationCounts` bumps docs from 29→31; CLAUDE.md updates 75→77 ops / 78→80 cmds. Consistent.

## Findings

### P2 — Missing body-flag-to-body-flag collision guard (`cmd/service/flags.go:197`)

The collision check in `registerBodyFlags` guards against `data`, `file`, query params, and header flags — but does not check `rt.bodyFlags[flagName]`. With the new `x-octo-flag` override on body properties, a future spec could alias two body properties to the same flag name (or an alias could collide with another property's derived name), causing a pflag panic at CLI startup (duplicate flag registration). Not triggered by this PR — `scope` and `role` are unique within `docs.share.set` and across all current body properties. Adding `if rt.bodyFlags[flagName] != nil { continue }` (matching the existing skip pattern) would close the gap for future specs.

**Diff-scope:** new (introduced by this PR's `x-octo-flag`-on-body-property feature).

## Data Flow Tracing

- `x-octo-flag` in `docs.json:631-632` → `resolveSchemaWithDepth` reads `s["x-octo-flag"]` at `loader.go:516` → populates `SchemaInfo.FlagName` → `registerBodyFlags` reads `prop.FlagName` at `flags.go:193` → uses it as cobra flag name (`--scope` / `--role`) → `bodyFlag.apiName` set to property name (`shareScope` / `shareRole`) at `flags.go:211` → `resolveBody` writes `base[bf.apiName]` at `run.go:162` → wire body uses property name, never the flag alias. ✅
- `stringOf(s["x-octo-flag"])` at `loader.go:516` safely handles nil, absent, wrong-type → returns `""` → fallback to `strings.ReplaceAll(name, "_", "-")` at `flags.go:197`. No crash on specs without the extension. ✅

## Blind Spot Checklist (C1–C6, security_sensitive)

- **C1 — Dual-path parity:** ✅ Clear. `ParamInfo.FlagName` (loader.go:353) already handles `x-octo-flag` for query/header params. This PR adds the same capability for body properties via `SchemaInfo.FlagName` (loader.go:516). Both follow identical pattern: explicit override wins, else derive from name.
- **C2 — Control-flow ordering:** ✅ Clear. `registerBodyFlags` is called once per operation during command setup (`service.go:138`). The `FlagName` override is a simple if-empty fallback — no double-application risk.
- **C3 — Authorization boundary:** ✅ Clear. CLI is a passthrough; server enforces auth. Spec declares `x-octo-risk: write` for PUT, `x-octo-risk: read` for GET.
- **C4 — Authorization lifecycle:** N/A — CLI spec definition; server-side share cascade and archived-doc handling are out of scope.
- **C5 — Build/runtime:** ✅ Build passes, tests pass, vet clean. Spec JSON consumed at runtime by registry loader, well-tested by added tests.
- **C6 — Governance/docs self-consistency:** ✅ CHANGELOG, CLAUDE.md, and `skills/octo-docs/common.md` all accurately describe the feature. No contradictions.

## Cross-round Blocker Re-review (R6)

- `resolveSchemaWithDepth` reads `x-octo-flag` via `stringOf`, which safely returns `""` for nil, wrong type, or absent keys — no crash on specs that don't carry the extension.
- The spec declares `x-octo-risk: read` for GET and `x-octo-risk: write` for PUT, consistent with existing risk annotations. Server enforces auth; CLI is a passthrough.
- CHANGELOG accurately describes the feature, scope enum, role behavior, and backend contract alignment.
- `skills/octo-docs/common.md` documentation matches the spec, including error codes (400/403/404/409) and the "share only raises access" invariant.
- No `x-octo-flag` collision between body aliases and existing query/header flags on any operation in the current spec set.

## Verdict: APPROVED

Well-structured addition: new spec entry, schema resolver extension, and flag wiring are all consistent with the existing `x-octo-flag` pattern for params. Tests cover the key contracts (wire-key isolation, flag alias correctness, restricted-scope body shape). The one P2 (body-flag collision guard) is a hardening suggestion for the new feature's future use, not a blocker for this PR.

[Octo-Q] verdict: APPROVE — No P0/P1 found. One P2 (body-flag collision guard) is non-blocking. PR is well-tested and consistent with existing patterns.

yujiawei (approved):
# Code Review — PR #85 (octo-cli)

Reviewed at head SHA `c1aa4328fb94b298a95f812ada8274c19fc2d2a4`. Verified against a fresh checkout: `go build ./...`, `go vet ./...`, `go test ./...` all green, `gofmt` clean on the changed Go files.

## Verdict: **APPROVED**

No blocking (P0/P1) issues. Two optional P2 nits below. This is a clean, well-scoped, well-tested change.

## Spec compliance
The diff does exactly what the PR describes and nothing more:
- Adds `docs share get` (GET `/v1/bot/docs/{docId}/share`, read) and `docs share set` (PUT, write) via the metadata-driven spec (`internal/registry/specs/docs.json:588-642`).
- Generalizes the existing `x-octo-flag` extension — previously only on query/header params — to request-body properties (`internal/registry/loader.go:513-517`, `cmd/service/flags.go:190-197`), so `--scope`/`--role` front the wire keys `shareScope`/`shareRole`.
- Docs (CHANGELOG, CLAUDE.md command tree + op count, skills/octo-docs/common.md) all updated consistently. `loader_test.go` bumps `docs` 29 → 31 and the shape tests pin the new routes.

No over-build, no missing pieces, no deviation from the stated contract.

## What I verified closely (correctness / security)

**The flag alias never leaks into the wire body.** In `registerBodyFlags`, the alias only changes the *cobra flag name*; the body key is bound separately: `bf := &bodyFlag{apiName: name, ...}` (`cmd/service/flags.go:211`), where `name` is the property key, independent of `flagName`. `resolveBody` writes `base[bf.apiName]` (`cmd/service/run.go:156-169`), so the body always carries `shareScope`/`shareRole`. `TestDocsShareSet_AnyoneEditBody` asserts the aliases (`scope`/`role`) are absent from the body — good, this is the load-bearing invariant.

**No regression to existing body properties.** `x-octo-flag` on a body property is opt-in: when the extension is absent, `prop.FlagName == ""` and the code falls back to the historical `strings.ReplaceAll(name, "_", "-")` derivation (`flags.go:194-197`). A grep of all specs confirms only the two new `docs.share.set` properties use a body-level `x-octo-flag` today, so no existing operation's flags or wire body change.

**Collision handling is intact.** The existing guard (`flags.go:201`) still runs on the resolved `flagName`, so an aliased body flag colliding with `--data`/`--file`/a query/header flag is still skipped. Header flags are registered before body flags, preserving precedence.

**Thin-transport security posture is acceptable and consistent.** This CLI does no client-side validation of scope/role enums, no `--role`-required enforcement, and no auth — all of that (enum, `invalid_scope`/`invalid_role`, admin-for-PUT `403`, `404`, `409`, and role normalization) stays server-side, matching the established `docs members set` pattern. The `(required)` label on `--scope` in `--help` is descriptive text only (cobra does not enforce body flags), which is the same behavior `docs members set` already exhibits — not a new footgun. The `--data` escape hatch justifies not hard-requiring promoted body flags.

## Nits (P2 — non-blocking, optional)

- **`cmd/service/flags.go:194-197`** — an explicit `x-octo-flag` override bypasses the underscore→dash normalization applied to derived names. If a future spec author wrote `"x-octo-flag": "some_flag"`, the CLI would expose a non-idiomatic `--some_flag`. Not triggered by this PR (both new overrides are single words) and identical to the pre-existing `paramFlagName` behavior for query/header params, so purely a future-proofing suggestion.
- **`cmd/service/docs_test.go` (`TestDocsShareSet_AnyoneEditBody`)** — unlike its sibling `TestDocsShareSet_RestrictedBody`, this test does not assert method/path. Low impact: `TestDocsShareSet_RestrictedBody` and `TestDocs_RegistryShape` both already pin `PUT /v1/bot/docs/{docId}/share`. Adding the two assertions here would just make this test self-contained.

## For a human to verify (out of this repo's scope)
Because the CLI is a thin transport and this is access-control-adjacent, the actual security guarantees live in docs-backend, which I cannot inspect from here:
- Confirm the docs-backend `GET/PUT /v1/bot/docs/{docId}/share` handler is deployed and enforces **admin for PUT**, the scope/role enums, `--role`-required-for-`anyone_in_space`, restricted→`read` normalization, and the "share only ever raises access" invariant. The CLI grants nothing itself; every guarantee is server-side.
- Confirm the bot mount (`/v1/bot/docs/*` re-mounting `docsRouter`) applies bot-identity middleware to `/:docId/share` and does not bypass the admin check — the PR asserts this but it is not verifiable from octo-cli.

## Advisory coverage disclosure
This review absorbed a cross-family long-context advisory pass (surfaced the two P2 nits above, both verified against the code). A second line-level adversarial advisory leg was **absent this round due to a gateway infrastructure error** unrelated to the PR — noting it so the coverage is not overstated as full consensus. My own independent review covers the flag-alias mechanism, wire-body invariant, regression surface, and the security posture above.

yujiawei (changes requested):
# Code Review (revised) — PR #85 (octo-cli)

**This supersedes my earlier Approve on this PR.** After posting it I verified the cross-repo dependency this feature rests on, and that check changes the verdict.

Reviewed at head SHA `c1aa4328fb94b298a95f812ada8274c19fc2d2a4`.

## Verdict: **CHANGES REQUESTED**

The CLI code itself is clean, isolated, non-regressing, and well-tested — that part of my earlier review stands. But the feature depends on a backend contract that does not exist, and the PR ships factual claims about that contract that are untrue.

## Blocking

**1. The target backend endpoint does not exist.** I searched `Mininglamp-OSS/octo-docs-backend` directly:
- `shareScope` → **0 code hits**; `shareRole` → **0 hits**; `anyone_in_space` → **0 hits**.
- `src/api/routes/docs.ts` `docsRouter` defines only `POST /`, `GET /`, `GET|PATCH|DELETE /:docId` — there is **no `/:docId/share` route** on `main`.
- The three repo-wide `/share` matches are all unrelated (`Dockerfile`, `src/whiteboard/exportScene.ts`, `test/awareness.test.ts`).

As shipped, every `docs share get|set` call will `404`. The green Go test suite does not catch this because each test mocks a local HTTP stub — it verifies method/path/body *shaping*, not that the route exists server-side.

**2. The stated dependency branch does not exist.** The PR body says it "Depends on the docs-backend share endpoint (`feat/space-scoped-share-permissions`)." That branch is **absent** from `octo-docs-backend` (GitHub returns "Branch not found"). So the enums, `read`-normalization, and `400/403/404/409` error codes the CLI encodes have no server counterpart to align to — they are currently unverifiable assumptions, not a matched contract.

**3. `CHANGELOG.md` states a false fact.** It claims the behavior matches "the docs-backend `/v1/bot/docs/{docId}/share` contract byte-for-byte." Since that route/contract is not merged anywhere in the backend, this wording is aspirational, not factual. Please soften it (e.g. "targets the planned `/v1/bot/docs/{docId}/share` endpoint") until it is backed by merged backend code. This is the concrete spec-side defect: a user-facing artifact asserts alignment with a contract that does not exist.

The PR body itself notes "Draft — opened to carry review before it goes in," which is consistent with this not being merge-ready yet.

## What is genuinely good (unchanged from my first pass)

- **The `x-octo-flag` generalization to body properties is correct and isolated.** The flag alias only renames the cobra flag; the wire key is bound separately via `bf.apiName` (`cmd/service/flags.go:211`) and written as `base[bf.apiName]` (`cmd/service/run.go:156-169`), so the body always carries `shareScope`/`shareRole`, never `scope`/`role`. `TestDocsShareSet_AnyoneEditBody` pins exactly that invariant.
- **No regression.** `x-octo-flag` on a body property is opt-in; absent it, the historical `strings.ReplaceAll(name, "_", "-")` derivation is used (`flags.go:194-197`). A grep of all specs confirms only the two new `docs.share.set` properties use a body-level `x-octo-flag` today, so no existing operation changes.
- Build/vet/test all green, `gofmt` clean.

## To unblock

1. Land (or confirm-merged) the `octo-docs-backend` `GET/PUT /v1/bot/docs/{docId}/share` handler, then re-verify the CLI's enums / role-normalization / error codes against the real route; **or** explicitly sequence this PR to merge after the backend and say so.
2. Fix the `CHANGELOG.md` "byte-for-byte" wording so it does not claim alignment with an unshipped contract.
3. (Non-blocking) Once the backend exists, add an integration-level check so the contract is validated against the real route, not only a local stub.

## Non-blocking nits (carried over)
- `cmd/service/flags.go:194-197` — an explicit `x-octo-flag` override bypasses underscore→dash normalization; a future `"x-octo-flag": "some_flag"` would surface as `--some_flag`. Not triggered here; matches pre-existing param-level behavior.
- `TestDocsShareSet_AnyoneEditBody` does not assert method/path (its sibling `TestDocsShareSet_RestrictedBody` and `TestDocs_RegistryShape` already do); adding the assertions would make it self-contained.

## Note on my earlier verdict
My first review deferred backend existence as "for a human to verify." That was the wrong call — the dependency was verifiable from here, and verifying it is precisely what a security-sensitive, cross-repo change requires. Corrected accordingly.

internal/registry/specs/docs.json
```diff
@@ -585,6 +585,79 @@
         "responses": { "200": { "description": "Removed" } }
       }
     },
+    "/v1/bot/docs/{docId}/share": {
```
OctoBoooot, line 596: **[blocker]** This models `GET|PUT /v1/bot/docs/{docId}/share` and the PR claims it is "byte-aligned with the docs-backend handler … byte-for-byte," but that endpoint does not exist in the backend. Verified against `Mininglamp-OSS/octo-docs-backend@main` (public, cloned fresh): `src/api/routes/docs.ts` `docsRouter` defines only `POST /`, `GET /`, `GET /:docId`, `PATCH /:docId`, `DELETE /:docId` — no `/share` route; no `share.ts` route file; and zero occurrences of `shareScope`/`shareRole`/`anyone_in_space` anywhere in `src/`. So the route, the enums, the required-role rule, the `restricted→read` normalization, and the 400/403/404/409 codes are all CLI-side assertions with no backend counterpart — `docs share get|set` will 404 on every call. Green CI misses this because all tests mock a local server and assert only method/path/body shaping. Fix: land the backend `/share` handler first (or same coordinated release), or soften the wording to "targets the planned endpoint (backend not yet shipped)" in CHANGELOG/spec/docs so the repo doesn't assert a contract that isn't there.

OctoBoooot (changes requested):
## Re-review → REQUEST_CHANGES (correcting my earlier APPROVE) — `c1aa4328fb94`

**I'm retracting my prior APPROVE.** In it I filed the "backend endpoint may not exist" concern as a *non-blocking rollout-ordering question*, on the stated grounds that it "isn't verifiable from this repo." That was wrong: `Mininglamp-OSS/octo-docs-backend` is **public**, so it *is* verifiable — I just hadn't done the check before approving. I've now done it, and the PR's central contract claim is falsified.

**Verdict:** Request changes — one **blocker**. The CLI code is clean and its tests are green, but the feature is dead-on-arrival and the PR asserts a backend contract that does not exist.

### Blocker
- **The target endpoint does not exist in the backend, so `docs share get|set` 404s on every call — and the PR's "byte-for-byte" contract claim is false.** This PR (CHANGELOG + `docs.json` + `common.md`) states it is *"byte-aligned with the docs-backend handler on `GET/PUT /v1/bot/docs/{docId}/share`."* Verified against `octo-docs-backend@main` (public repo, cloned fresh):
  - `src/api/routes/docs.ts` `docsRouter` defines **only**: `POST /`, `GET /`, `GET /:docId`, `PATCH /:docId`, `DELETE /:docId` (lines 129–203). There is **no `/share` route**.
  - No `share.ts` route file exists under `src/api/routes/` (a sibling `members.ts` does, but no share).
  - **Zero** occurrences of `shareScope`, `shareRole`, or `anyone_in_space` anywhere in the backend `src/` tree.
  So the modeled `GET|PUT /v1/bot/docs/{docId}/share`, the `restricted`/`anyone_in_space` + `read`/`edit` enums, the "`--role` required for anyone_in_space", the `restricted→read` normalization, and the `400/403/404/409` codes are all **unilateral CLI-side assertions with no backend counterpart** to be byte-aligned with. This is a Principle-3 failure: a PR that claims parity with an existing contract must have that contract exist. The green CI doesn't catch it because every test mocks a local HTTP server and asserts only method/path/body *shaping* — none exercises a real backend route.
  **Fix (either):** (a) land the backend `/share` handler first (or in the same coordinated release) so the contract is real and re-verifiable; or (b) if the CLI is intentionally landing ahead of the server, soften the wording — drop "byte-aligned … byte-for-byte" and mark it *targets the planned `/share` endpoint (backend not yet shipped)* in CHANGELOG/spec/docs — so the repo doesn't assert a contract that isn't there.

### What is genuinely good (unchanged from my first pass, and reusable the moment the backend lands)
- Generalizing `x-octo-flag` from params to **request-body properties** is well done: `--scope`/`--role` front the `shareScope`/`shareRole` wire keys without renaming them, and `flags.go` keeps `bf.apiName` = the property name. Isolated and non-regressing — I grepped every spec: the only other `x-octo-flag` uses are param/header-level (`format`, `If-Match`), so no existing operation's flags change.
- Tests are mutation-meaningful *for the CLI's own behavior*: wire keys don't leak (`TestDocsShareSet_AnyoneEditBody`), restricted omits `shareRole`, camelCase flags rejected, op count 29→31 pinned. `go build ./...` + `go test ./cmd/service/ ./internal/registry/` green locally.

The wiring is right; the missing piece is real backend support (or honest wording about its absence). Once the `/share` handler exists in `octo-docs-backend`, this is re-review-ready and most of it will re-verify unchanged.

### Correction note
My earlier APPROVE (review 4702836131) should be treated as superseded by this one. I got the disposition wrong by assuming the cross-repo fact was out of reach instead of checking — the fact was one clone away, and it's decisive.

boris-clark:
## Re: the `/v1/bot/docs/{docId}/share` "endpoint does not exist" blocker

Thanks for the careful cross-repo checks. Two of the three blocking points share one root premise ("the `/share` route does not exist, so every call 404s"), and I want to lay out the ground truth on that, because it was verified against the wrong tree. The remaining point (the CHANGELOG wording) is valid and I address it separately below.

### The `/share` route exists — it lands in `octo-docs-backend#68`, not on `main`

The searches that returned "0 hits for `shareScope`" and "no `/:docId/share` route" were run against `octo-docs-backend@main` (pre-#68). The route lives on the open, **APPROVED**, CI-green PR `octo-docs-backend#68` (`feat/space-scoped-share-permissions`, head `beb7009c`), which is the backend contract this CLI targets. On that head:

- `src/api/routes/docs.ts:311-312`
  ```ts
  docsRouter.get('/:docId/share', getShareHandler)
  docsRouter.put('/:docId/share', putShareHandler)
  ```
- The handlers implement exactly the semantics this PR encodes: `shareScope` ∈ `restricted | anyone_in_space`, `shareRole` ∈ `read | edit`, `400 invalid_scope`, `400 invalid_role` (`anyone_in_space` requires a valid role), and the `restricted → read` normalization (`docs.ts:257-304`).
- `docsRouter` is **double-mounted** in `src/api/app.ts`: `app.use('/api/v1/docs', api)` (line 127, `api.use(docsRouter)`) **and** `app.use('/v1/bot/docs', botApi)` (line ~159, `botApi.use(docsRouter)`). So `/v1/bot/docs/{docId}/share` resolves.
- Deployment check confirms this — `GET /v1/bot/docs/<id>/share` against the running instance on `:8091` returns **HTTP 401 (auth required), not 404**. A non-existent route would 404; a 401 means the route matched and hit the auth middleware.

So the enums, role-normalization, and `400/403/404/409` codes are not unilateral CLI assumptions — they mirror the merged-pending backend handler in #68. The green Go tests mock a local stub by design (unit-level wire-shaping), but the contract they shape to is real in #68.

### On the CHANGELOG "byte-for-byte" wording — this one is fair

The blocker about `CHANGELOG.md:18` claiming the behavior matches the backend contract "byte-for-byte" is a legitimate, separate point, and it stands independently of the endpoint question: **#68 has not merged to `main` yet**, so asserting present-tense parity with a shipped contract is overstated. That wording will be softened to something like *"targets the planned `/v1/bot/docs/{docId}/share` endpoint (backend lands in `octo-docs-backend#68`)"* so the repo does not claim alignment with a contract that isn't on `main` yet.

### Merge ordering

To keep the claim honest and the feature non-dead-on-arrival, **#85 merges after `octo-docs-backend#68`** — backend contract lands first, then the CLI. I'll sequence it that way.

### Non-blocking: `docs_test.go` empty-body assertion

Good call — worth pinning. `TestDocsShareGet` currently asserts method/path only; adding an explicit empty-request-body assertion makes the GET contract self-documented. It's non-blocking, so I'll fold it into the same follow-up that softens the CHANGELOG wording rather than churn the head now.

Re-requesting review so this can be re-evaluated against #68's actual head rather than pre-#68 `main`.

boris-clark:
Pushed a follow-up commit (`8c19100`) addressing the CHANGELOG blocker; requesting a re-review on the new head.

**What changed since `c1aa432`** (docs/tests only — no CLI command logic touched):

- **CHANGELOG** — softened the "matches the docs-backend `/v1/bot/docs/{docId}/share` contract byte-for-byte" wording. It now reads as targeting the *planned* endpoint (`backend lands in octo-docs-backend#68`) and states the merge order (this PR merges after #68 ships). This removes the aspirational-contract claim while the backend route is unmerged.
- **`cmd/service/docs_test.go`** — added the GET-empty-body assertion suggested in review: `TestDocsShareGet_ReadPath` now reads the request body in the stub handler and asserts it is empty, pinning the "GET read sends no payload" contract.

CI on the new head is green (build & test / vet / gofmt / npm / osv-scan all pass).

The spec/flags/loader wiring is unchanged from the version already accepted at the code layer. @Jerry-Xin @yujiawei @OctoBoooot — could you re-review on `8c19100`? Note this still depends on and should merge after Mininglamp-OSS/octo-docs-backend#68.

OctoBoooot (approved):
## Re-review → APPROVE (`8c19100d4f76`) — prior blocker resolved

**Verdict:** Approve. The blocker from my last review (CHANGES_REQUESTED on `c1aa4328`) is fixed, and the contract is now grounded in a real backend PR. All code CI is green (`build & test (1.24.x)`, `npm test`, `osv-scan`) and `check-sprint` governance passes. One nit; no blocker/major.

### What changed since `c1aa4328` (delta reviewed)
Two-file delta, both addressing my findings:
- **CHANGELOG.md** — the false claim *"matching the docs-backend contract byte-for-byte"* is **removed**, replaced with *"This targets the **planned** `/v1/bot/docs/{docId}/share` endpoint (backend lands in octo-docs-backend#68); **merge this after that endpoint ships**."* That's honest about the not-yet-shipped backend and self-gates merge timing.
- **docs_test.go** — `TestDocsShareGet_ReadPath` now reads the request body and asserts `len(gotBody)==0`, pinning the GET-has-no-body contract (my nit from the first pass).

### Blocker resolved — verified against the backend this time
My prior blocker had two parts; both are now addressed:
1. *False "byte-for-byte" parity claim* → **removed** from CHANGELOG (see above). The one present-tense residue is in the spec description (see nit).
2. *Contract had no backend counterpart* → **now grounded.** I verified `Mininglamp-OSS/octo-docs-backend#68` (`feat(share): space-scoped document share permissions`, OPEN) implements exactly this contract: `GET|PUT .../{docId}/share`, `share_scope` `0=restricted / 1=anyone_in_space`, `share_role` `1=read / 2=edit`, `restricted`→role-ignored-stored-as-read normalization, and the identical error set — `400 invalid_scope`, `400 invalid_role` (anyone_in_space w/ missing role), `403` (not admin/owner), `404` (missing/cross-space), `409` (archived). The CLI's spec is a faithful client of a real backend contract, not an invented one.

**Merge readiness (the honest caveat):** backend#68 is still **OPEN**, so the route isn't live yet — a merge *today* would 404 until #68 ships. The PR now states this explicitly and instructs merging afterward, so this is a conscious maintainer merge-ordering decision, not a code defect. APPROVE reflects the code; sequencing is the maintainer's call, and the PR documents it correctly.

### Nit (≤3)
- `internal/registry/specs/docs.json:618` — the `docs.share.set` description still says *"Normalization **matches** the backend"* (present tense), the same class as the CHANGELOG line that was just softened. Since #68 isn't merged, *"matches the planned backend contract (octo-docs-backend#68)"* would be fully consistent with the new CHANGELOG wording. Purely cosmetic — the described normalization does match #68. _(confidence: high)_

### Praise
- Clean, minimal response to the review: the author took the exact fix offered (soften the wording + cite the concrete backend PR) rather than over-rotating, and tightened the GET-body test in the same push. And the underlying mechanism remains the strong part — generalizing `x-octo-flag` to request-body properties so `--scope`/`--role` front the `shareScope`/`shareRole` wire keys without renaming them, pinned by `TestBodyPropertyFlagAlias`. The moment #68 merges, this wiring is ready as-is. _(confidence: high)_

### Correction note
This supersedes both my APPROVE on `c1aa4328` (which under-weighted the then-false parity claim) and my subsequent CHANGES_REQUESTED (correct for that commit). This commit fixes the substance the CR flagged.

Jerry-Xin (approved):
The repo-local defect from the prior review is fixed and the CLI wiring is clean and well-tested; approving with an explicit merge-ordering caveat.

**✅ What the new commit fixed**
- `CHANGELOG.md`: removed the earlier unfounded "matching the docs-backend contract byte-for-byte" claim and replaced it with honest wording — the feature "targets the planned `/v1/bot/docs/{docId}/share` endpoint (backend lands in octo-docs-backend#68); merge this after that endpoint ships." This resolves the repo-local blocker (the false parity claim) and self-documents the cross-repo merge dependency.
- `cmd/service/docs_test.go`: `TestDocsShareGet_ReadPath` now reads the request body (`io.ReadAll`) and asserts it is empty, pinning the "a read must not ship a payload" contract. This closes the earlier non-blocking nit.

**⚠️ Merge-ordering caveat (non-code, governance)**
- The backend `/v1/bot/docs/{docId}/share` route still does not exist as of this review (the docs router exposes only the root `POST`/`GET` and `/{docId}` `GET`/`PATCH`/`DELETE`; no share-scope handler). The CLI calls would 404 until the backend lands. This is a cross-repo rollout-ordering item, not a defect in this repo — and the PR now explicitly discloses and defers it. **This APPROVE is contingent on not merging ahead of octo-docs-backend#68 shipping the `/share` route.** The mergeable state is blocked anyway.

**✅ Highlights**
- Clean `x-octo-flag` body-property alias: `--scope`/`--role` front the `shareScope`/`shareRole` wire keys without renaming the property, so the on-the-wire JSON stays exact (`internal/registry/loader.go`, `cmd/service/flags.go`).
- Registry-driven: the operations are declared in `internal/registry/specs/docs.json` and auto-register through the existing metadata path — no hand-written command wiring.
- Thorough, mutation-meaningful tests: registry shape, alias extraction, flag exposure (and negative assertions that `--shareScope`/`--shareRole` do NOT exist), and actual request-body assertions for both `restricted` and `anyone_in_space` cases.

Verification: `go test ./...` passes.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #85 (octo-cli)

## Summary

This PR adds two new CLI commands — `docs share get` and `docs share set` — for reading and writing a document's space-level share scope. It introduces the `x-octo-flag` extension on request-body properties (previously only used on query/header params) so that the `shareScope`/`shareRole` wire keys can be exposed as clean `--scope`/`--role` flags. The spec JSON, loader, flag registration, wire-body construction, tests, skill docs, CHANGELOG, and CLAUDE.md are all updated consistently.

## Verification

Local checks at head `8c19100d`:
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./...` — pass (all 12 packages)
- JSON validation of `internal/registry/specs/docs.json` — valid

## Findings

No P0/P1 issues. The implementation is correct and well-tested.

### Things I checked that are fine

- **Data flow trace (spec → wire key)**: `x-octo-flag` is parsed in `internal/registry/loader.go:516` into `SchemaInfo.FlagName`. Flag registration in `cmd/service/flags.go:194-196` uses `FlagName` for the CLI flag name while `bodyFlag.apiName` (line 211) stays the property name. Body construction in `cmd/service/run.go:156-168` writes `base[bf.apiName]` — wire keys are always byte-exact backend contract (`shareScope`/`shareRole`). No alias leakage.
- **GET sends no body**: The `docs.share.get` spec has no `requestBody`, so `registerBodyFlags` is not called and `resolveBody` returns `nil`. Test `TestDocsShareGet_ReadPath` asserts `len(gotBody) == 0`.
- **Flag alias leakage**: Test `TestDocsShareSet_AnyoneEditBody` explicitly asserts that `scope` and `role` do not appear as wire body keys.
- **Flag collision checks**: The code at `cmd/service/flags.go:201` checks for collisions with `--data`, `--file`, query flags, and header flags. No existing flag named `scope` or `role` at any command level.
- **Operation count parity**: `loader_test.go` bumps docs from 29→31; `CLAUDE.md` bumps from 75→77 operations / 78→80 commands. Consistent with adding 2 new operations.
- **Spec JSON valid**: Parsed successfully.
- **x-octo-flag pattern consistency**: The new body-property `x-octo-flag` usage mirrors the existing `ParamInfo.FlagName` pattern (already used 4 times in the spec for `base-version` and `image-format`).
- **Docs consistency**: `skills/octo-docs/common.md` examples, error codes, and behavior description match the spec. CHANGELOG accurately notes the backend dependency.

## Blindsight Checklist (C1–C6)

- **C1 (dual-path parity)**: Get and set both added on the same path. No asymmetric pair. Clear.
- **C2 (control-flow ordering)**: `FlagName` is parsed at every schema depth but only consumed at the top level in `registerBodyFlags`. No double-application risk. Clear.
- **C3 (authorization boundary)**: CLI is a pass-through; auth enforcement is server-side. N/A for CLI.
- **C4 (authorization lifecycle)**: N/A — no backend state changes in this PR.
- **C5 (build ≠ runtime)**: Build and tests pass. N/A.
- **C6 (governance self-consistency)**: CHANGELOG, common.md, CLAUDE.md, and spec all consistent. Clear.

## Verdict: APPROVED

Clean, well-structured PR. The `x-octo-flag` body-property extension is a natural generalization of the existing param-level pattern, the data flow from spec through to wire body key is correct, and the test coverage is thorough — including explicit assertions that flag aliases never leak as wire keys. No blocking issues.

yujiawei (approved):
# Code Review — PR #85 (octo-cli)

Independent review at head SHA `8c19100d4f76fab34d4578dbc74a1e9629965b15`. This PR is access-control sensitive (it programs a document's space-level share scope — who in a space can reach a doc), so it got extra scrutiny on the transport path and blast radius. Verified locally: `go build ./...`, `go vet ./...`, and the full `go test ./...` suite are green, `gofmt` clean.

## Verdict: Approved

No correctness, security, or data-loss blockers. Four non-blocking nits below (all FYI / future-robustness), plus one release-ordering note.

---

## 1. Spec / contract compliance — pass

- New leaves `docs.share.get` (GET) and `docs.share.set` (PUT) both mount on `/v1/bot/docs/{docId}/share`, matching the members pattern and the stated backend contract. No human endpoint or auth-bypass path is introduced.
- Enums are byte-aligned: `shareScope` ∈ `{restricted, anyone_in_space}`, `shareRole` ∈ `{read, edit}`; `shareScope` is `required`, `shareRole` optional. Error codes documented (`invalid_scope`/`invalid_role` 400, 403, 404, 409).
- No over-build: the change is purely additive spec + one generic enabling change (`x-octo-flag` now honored on request-body properties). No existing operation's behavior is touched.
- Operation counts and tree-shape tests updated consistently (docs 29→31; CLAUDE.md 75→77 ops).

## 2. Code quality — Approved

**Thin-transport contract holds (the key security property).** Traced the full data flow: spec property → `loader.SchemaInfo.FlagName` (loader.go:516, from `x-octo-flag`) → flag registration (flags.go:194) → `resolveBody` (run.go:156-170) → wire body.
- The flag alias never leaks into the JSON body: `bf.apiName` is always the property name (`shareScope`/`shareRole`), while the alias (`--scope`/`--role`) is only the cobra flag label and the `bodyFlags` map key (flags.go:211, 226). `TestDocsShareSet_AnyoneEditBody` explicitly asserts `scope`/`role` never appear as body keys.
- `resolveBody` sets each key only when `Flags().Changed(flagName)` is true, so `--scope restricted` with `--role` omitted correctly ships `{shareScope: restricted}` with no `shareRole` (`TestDocsShareSet_RestrictedBody`).
- No client-side auth or validation is introduced: body flags are never `MarkFlagRequired` (that call is query/header/`--file` only, flags.go:104/139/156). Auth (403), scope/role validation, and the "restricted normalizes role to read" rule all stay server-side — consistent with `docs.members.set`.
- GET carries no body, asserted by `TestDocsShareGet_ReadPath`.

**Blast radius of the generic `x-octo-flag`-on-body change — verified contained.** The change now applies to every spec's body properties, so I enumerated all request-body properties across all 8 domains: no existing property carries `x-octo-flag`, and no two properties in any single operation derive/alias to the same flag name. The only body-level `x-octo-flag` in the tree is the new `scope`/`role` in the share block. No other operation's flag surface changes.

### Nits (non-blocking)

- **[nit] Body-vs-body flag-name collision guard is absent** — `cmd/service/flags.go:201`. The collision guard skips a body flag when it clashes with `--data`/`--file`/a query flag/a header flag, but does not check `rt.bodyFlags[flagName]`. Harmless today (share aliases are distinct, and any future collision would panic loudly in pflag at root construction — caught by the first tree-shape test, never a silent ship). But since this PR is what makes body-property aliasing possible, a future spec author who aliases one body property onto another's flag name would get an opaque pflag panic instead of a clean loader-level skip/error. Consider a one-line `rt.bodyFlags[flagName] != nil` check symmetric with the existing query/header guards.
- **[nit / FYI] `--role` is forwarded verbatim even with `--scope restricted`** — `cmd/service/run.go:156-170`. `docs share set d1 --scope restricted --role edit` sends `{shareScope: restricted, shareRole: edit}` on the wire; there is no client-side coupling that strips `shareRole`. This is intentional under thin-transport (the backend ignores/normalizes it to `read`), and has no security impact since `shareRole` cannot widen access beyond the server-enforced scope. Flagged only so it's clear the "restricted ignores role" guarantee is a backend contract, not a CLI guarantee.

## 3. Release-ordering note (merge gate, not a code defect)

The CHANGELOG states this depends on the backend share endpoint (octo-docs-backend#68) and should merge after that ships. Because the CLI is a thin metadata-driven transport, the new leaves are additive and cannot regress existing commands even if the route 404s until the backend lands — so this is a release-ordering constraint to enforce at the merge gate, not a defect in the diff. The PR is also marked Draft.

## For the human reviewer (security-sensitive)

Nothing in the CLI diff requires manual security sign-off beyond confirming the backend side: worth a human eye on octo-docs-backend#68 to confirm that (a) `PUT /share` genuinely enforces admin (403) and rejects `invalid_scope`/`invalid_role`, and (b) `restricted` truly normalizes/ignores `shareRole` server-side, since the CLI faithfully forwards whatever pair it's given. Those are the load-bearing guarantees this thin client depends on.

Jerry-Xin (approved):
Re-affirming APPROVE on the rebased head `ae627d62c56f` (previously approved `8c19100d4f76`).

Verified this is a pure rebase, not a content change to the share feature:
- Blob-SHA comparison of every #85 share-scope file between the old and new head is **byte-identical**: `CHANGELOG.md`, `cmd/service/docs_test.go`, `cmd/service/flags.go`, `internal/registry/loader.go`, `internal/registry/loader_test.go`, `skills/octo-docs/common.md`, `CLAUDE.md` — all SAME.
- The only files that differ (`internal/registry/specs/docs.json`, `skills/octo-docs/sheet.md`) differ **only** in the sheet hyperlinks content that landed from #86 via the rebase; a diff of all share/scope/role lines in `docs.json` between the two heads is empty. The `/v1/bot/docs/{docId}/share` spec (docs.share.get/set, shareScope/shareRole, `x-octo-flag` aliases) is unchanged.

So my prior review basis carries forward intact: the CHANGELOG uses honest "targets the planned endpoint … merge after that endpoint ships" wording (no false byte-for-byte claim), the GET empty-body assertion is present, and the contract is grounded in the real backend PR.

⚠️ Merge-ordering (unchanged): the backend `/v1/bot/docs/{docId}/share` route (octo-docs-backend#68) is still OPEN/unmerged, so this APPROVE remains contingent on not merging ahead of that route shipping. `mergeable_state` is blocked, and `check-sprint` governance should be cleared by a maintainer as well.

Verdict: APPROVE (carried forward to the rebased head).

OctoBoooot (approved):
## Re-review → APPROVE (`ae627d62c56f`) — delta is the already-reviewed #86 content

**Verdict:** Approve. All code CI is green (`build & test (1.24.x)`, `npm test`, `osv-scan`) and `check-sprint` governance now **passes** too. No blocker/major.

### What changed since my last APPROVE (`8c19100d`)
The delta is exactly two files — `internal/registry/specs/docs.json` and `skills/octo-docs/sheet.md` — and it is **the #86 sheet-hyperlinks + math-formula content merged into this branch**, nothing else:
- **#85's own share feature is byte-identical to what I already approved.** Verified: the `docs.share.get/set` spec, the "targets the **planned** endpoint (backend lands in octo-docs-backend#68); merge this after that endpoint ships" CHANGELOG wording, and `flags.go`/`loader.go`/`docs_test.go`/`loader_test.go` are unchanged (0-line diff on those paths). The share blocker fix from last round stands.
- The new-to-this-head lines are the `hyperlinks` batch + `sheetHyperLinks` response property and the math-formula/hyperlink skill docs — **content I already reviewed and APPROVED on #86** (`d7ca3c41`): hyperlinks grounded in `octo-docs-backend#84`, math formula riding the existing `drawings` surface, spec valid, op count unchanged.

### Merge-ordering caveat (unchanged, two cross-repo deps now)
Both backend surfaces this branch documents are still **OPEN**: `octo-docs-backend#68` (share) and `#84` (sheetHyperLinks). APPROVE reflects the code; a maintainer should sequence the merge after those land so `docs share get/set` and the `sheetHyperLinks` read don't 404. The PR is honest about this (no false "already live" claim). Purely timing.

### Carried-over nits (from the #86 content, non-blocking)
- `docs.json` `docs.sheet.edit` `Gates:` clause still enumerates "cell/dim/drawing" for 413/422 — not extended to name hyperlinks.
- `sheet.md` paged-read line names only `sheetDims` as first-page-only; should add `sheetHyperLinks`. And the xlsx-export line ("returns cells and dims") doesn't mention hyperlinks.
All cosmetic; the authoritative schemas are correct and complete.

### Praise
- Clean convergence: folding #86's content in without disturbing a byte of the already-approved share code made this re-review fast and low-risk — the blob-level stability is exactly what you want when two feature branches merge. _(confidence: high)_

Supersedes my APPROVE on `8c19100d`; the substance is unchanged plus the already-reviewed #86 content.

mochashanyao (approved):
[Octo-Q · automated review]

---

## Review: octo-cli PR#85 — `feat(docs): add docs.share get/set for programmatic share scope`
**Head SHA:** `ae627d62c56ff35c51452d4d6a7c4807e4aaf0d0`
**Files:** 8 changed (+257 / -3)
**Tests:** All 12 packages pass (`go test ./...`)

---

### Summary

Adds two new CLI operations for managing document share scope:
- `docs.share.get <docId>` — GET `/v1/bot/docs/{docId}/share` (read)
- `docs.share.set <docId> --scope <enum> [--role <enum>]` — PUT `/v1/bot/docs/{docId}/share` (write)

Introduces a new `x-octo-flag` extension on request-body properties, enabling clean flag aliases (`--scope` / `--role`) while wire body keys remain the property names (`shareScope` / `shareRole`).

---

### Verification ✅

| Check | Status | Evidence |
|-------|--------|----------|
| Spec → Loader → Flag → Wire body data flow | ✅ | Traced below |
| GET sends no body | ✅ | `TestDocsShareGet_ReadPath` asserts `len(gotBody) == 0` |
| Flag aliases don't leak into wire body | ✅ | `TestDocsShareSet_AnyoneEditBody` checks `gotBody["scope"]` / `gotBody["role"]` absent |
| Wire keys match backend contract | ✅ | `bf.apiName` always = property name (`shareScope`/`shareRole`), never the flag alias |
| Existing operations unaffected | ✅ | `FlagName == ""` falls back to `strings.ReplaceAll(name, "_", "-")` — all 12 test packages green |
| Operation counts correct | ✅ | `TestAllDomainOperationCounts` bumped docs 29→31 |
| CLAUDE.md counts updated | ✅ | 75→77 ops, 78→80 commands |

---

### Data-Flow Trace

**`--scope restricted` (set, no role):**
1. `docs.json`: `shareScope` property → `"x-octo-flag": "scope"`
2. `loader.go:516`: `resolveSchemaWithDepth` → `SchemaInfo.FlagName = stringOf(s["x-octo-flag"])` = `"scope"`
3. `flags.go:193`: `flagName = prop.FlagName` = `"scope"` (override wins)
4. `flags.go:218`: `bf := &bodyFlag{apiName: "shareScope", ...}` — wire key stays property name
5. `flags.go:222`: `cmd.Flags().StringVar(bf.strVal, "scope", "", desc)` — CLI flag is `--scope`
6. `run.go:156-166`: `resolveBody` → `cmd.Flags().Changed("scope")` → true → `base["shareScope"] = "restricted"`
7. Wire body: `{"shareScope": "restricted"}` — `shareRole` absent (not Changed). ✅

**`--scope anyone_in_space --role edit`:**
Same path for both flags. Wire body: `{"shareScope": "anyone_in_space", "shareRole": "edit"}`. ✅

**Collision guard:** `flags.go:200` — checks `"data"`, `"file"`, `queryFlags`, `headerFlags`. Neither `scope` nor `role` collides (no query/header params on this op). ✅

---

### Findings

**No P0/P1 issues found.**

#### P2 — Notes (non-blocking)

1. **Body-body flag alias collision is unguarded** (pre-existing, not introduced by this PR): If two body properties in the same operation declared the same `x-octo-flag` value, cobra would panic at startup with a duplicate-flag error. This is loud enough to catch at development time, but a graceful error in `registerBodyFlags` (e.g., "property X and Y both claim flag --Z") would be more helpful for spec authors. **Pre-existing design characteristic — not a diff-scope issue.**

2. **No client-side enum validation**: The spec declares `enum: ["restricted", "anyone_in_space"]` for `shareScope` and `enum: ["read", "edit"]` for `shareRole`. The CLI adds these to the flag description but does not reject invalid values client-side. This is consistent with all other enum flags in the codebase (server-side validation is the contract). **Consistent with existing pattern — not a gap.**

---

### Blind-Spot Checklist (R5)

| Check | Result | Evidence |
|-------|--------|----------|
| **C1 — Dual-path parity** | ✅ Clear | `get`/`set` pair both implemented; `share` resource correctly added to command tree (`TestDocs_TreeShape`) |
| **C2 — Control-flow ordering** | ✅ Clear | New `FlagName` path is additive; empty `FlagName` falls back to existing derivation — no behavioral change for existing ops |
| **C3 — Auth boundary** | N/A | Authorization is server-side; CLI only documents "needs admin"/"needs reader" |
| **C4 — Auth lifecycle** | N/A | Server-side |
| **C5 — Build/runtime parity** | ✅ Clear | Spec embedded at build time → parsed by loader → consumed by flag registration → wire body verified by tests |
| **C6 — Governance consistency** | ✅ Clear | CHANGELOG, CLAUDE.md, `skills/octo-docs/common.md` all consistent in command names, scope/role enums, and error codes |

---

### Cross-Round Blocker Recheck (R6)

N/A — first review.

---

### Extra Findings

- CHANGELOG correctly notes the backend dependency: "merge this after that endpoint ships" (octo-docs-backend#68). Good merge-ordering awareness.
- The `x-octo-flag` on body properties is a reusable capability (mirrors existing `ParamInfo.FlagName` for query/header params) — well-designed generalization.
- 5 new tests in `docs_test.go` + 1 in `loader_test.go` provide thorough coverage of the new operations and the new flag-alias mechanism.

---

[Octo-Q] verdict: **APPROVE** — Clean implementation with thorough test coverage. The `x-octo-flag` extension on body properties is a well-designed generalization of the existing param-level capability. Data flow from spec → loader → flag registration → wire body is verified correct. No P0/P1 issues. All tests pass.

yujiawei (approved):
# Code Review — PR #85 (octo-cli)

Independent review at head `ae627d62c56ff35c51452d4d6a7c4807e4aaf0d0`. Repo checked out at that exact SHA; `go build ./...`, `go vet ./...`, `go test ./...` all green, `gofmt` clean. This PR was flagged security-sensitive, so the auth boundary and the new body-flag aliasing mechanism were traced end-to-end.

## Summary

Adds two spec-driven operations to the metadata-driven CLI:

- `docs share get <docId>` → `GET /v1/bot/docs/{docId}/share`
- `docs share set <docId> --scope restricted|anyone_in_space [--role read|edit]` → `PUT /v1/bot/docs/{docId}/share`

The only engine change is a new `x-octo-flag` alias on **request-body properties** (previously only query/header params supported it), letting the clean `--scope`/`--role` flags front the exact wire keys `shareScope`/`shareRole`.

## Spec compliance — ✅

- **No under-build:** both ops are wired correctly (method/path/operationId), the docs count is bumped consistently everywhere (`29→31` in `loader_test.go:44`; registry-shape and tree-shape tests pin 31; `CLAUDE.md` header `75→77` ops / `78→80` cmds), and CHANGELOG/skills docs match the spec enums, normalization, and error-code table.
- **No over-build:** the `x-octo-flag` body extension reuses the *same* extension key already used for query (`format→--image-format`) and header (`If-Match→--base-version`) params — it is the minimal enabling change, not scope creep. Only the flag-name derivation is touched; the wire body key (`bf.apiName` at `cmd/service/flags.go:211`) stays the property name, and `resolveBody` in `run.go` is unchanged.
- **No deviation / no auth bypass:** `/share` is a pure spec-data addition on the pre-existing bot-authenticated mount (`x-octo-space-header:false` unchanged at `docs.json:10`; no mount/auth/token override). The CLI stays a thin transport — scope/role validation and `invalid_scope`/`invalid_role`/`403`/`404`/`409` are all server-side. The `--scope`/`--role` aliases provably never leak into the JSON body and the property names are never renamed (`TestDocsShareSet_AnyoneEditBody`, `TestDocsShareSet_FlagAliases`, `TestBodyPropertyFlagAlias`).

## Code quality — Approved

No P0/P1 issues. The following are non-blocking suggestions.

### P2 — Body-flag alias can shadow a global persistent flag (latent; does not affect this PR)

`cmd/service/flags.go:201` skips a promoted body flag only when it collides with `--data`, `--file`, a query flag, or a header flag — it does **not** check the root persistent flags in `cmd/root.go:39-48` (`--format`, `--jq`, `--space`, `--bot-id`, etc.). Because body flags are registered on `cmd.Flags()` (local), a body property whose `x-octo-flag` or derived name equals a persistent flag would silently shadow the global on that leaf and misroute the user's value into the body.

This PR is **not** affected — `--scope`/`--role` collide with nothing, and no body property across any spec currently derives to a persistent-flag name. But the `x-octo-flag` body extension is the first mechanism letting a spec author freely name a promoted body flag, so it newly *enables* the hazard. The project already treats this exact risk as real for query params (`docs.json:174` remaps `format→--image-format` "so it does not shadow the global --format output flag", pinned by `TestDocsSceneExport_ImageFormatFlagNoGlobalShadow`). Consider extending the line-201 guard (e.g. also skip / fail-loud when `cmd.Root().PersistentFlags().Lookup(flagName) != nil`) so the body path gets the same protection.

### P2 — Test coverage gaps (optional hardening)

- `docs.share.set` marks `shareScope` required in the spec, but the CLI does not `MarkFlagRequired` it (consistent with thin-transport — the server returns `400`). No test pins this intentional pass-through, so a future reader can't tell it's deliberate.
- `TestDocsShareGet_ReadPath` asserts method/path/empty-body but never inspects the output envelope, unlike sibling reads (`TestDocsContentGet…`, `TestDocsSheetGet…`) which assert `{ok, data.*}`.
- No positive test that an invalid `--scope`/`--role` value is transmitted verbatim (rather than rejected client-side), which would document the thin-transport enum contract and guard against someone later bolting on client-side enum validation.

## Verdict

Spec ✅ and Quality Approved → **APPROVED**. Clean, minimal, well-tested feature addition on the existing authenticated transport; the one latent footgun is out of this diff's scope. Note the PR description marks it Draft pending the backend `/share` endpoint (octo-docs-backend #68) — merge ordering is the human's call.

PR #87, opened 2026-07-15
feat: install skills from marketplace

kense-lab:
## Summary
- add `octo-cli marketplace skills <skill-id> --install <skills-root>`
- authenticate Marketplace downloads with a `bf_*` User Bot credential
- verify archive digest and safely install Skill files
- support same-domain Marketplace routing through `OCTO_MARKETPLACE_API_PREFIX`

## Verification
- `go test ./...`
- verified real Bot authentication, download, digest validation, and local installation
- verified cross-Space installation fails without creating a target directory

### DIFF
--- a/internal/marketplace/client.go
+++ b/internal/marketplace/client.go
@@ -1,0 +1,93 @@
+package marketplace
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+
+	apiClient "github.com/Mininglamp-OSS/octo-cli/internal/client"
+	"github.com/Mininglamp-OSS/octo-cli/internal/output"
+)
+
+const MaxArchiveBytes int64 = 64 << 20
+
+type APIClient interface {
+	Do(context.Context, *apiClient.Request) ([]byte, error)
+}
+
+type Skill struct {
+	ID         string `json:"id"`
+	Name       string `json:"name"`
+	FileSHA256 string `json:"file_sha256"`
+}
+
+type Client struct {
+	api  APIClient
+	http *http.Client
+}
+
+func NewClient(api APIClient, httpClient *http.Client) *Client {
+	if httpClient == nil {
+		httpClient = http.DefaultClient
+	}
+	return &Client{api: api, http: httpClient}
+}
+
+func (c *Client) GetSkill(ctx context.Context, id string) (Skill, error) {
+	body, err := c.api.Do(ctx, &apiClient.Request{
+		Service: "marketplace",
+		Method:  http.MethodGet,
+		Path:    "/skill/" + url.PathEscape(id),
+	})
+	if err != nil {
+		return Skill{}, err
+	}
+	var envelope struct {
+		Data Skill `json:"data"`
+	}
+	if err := json.Unmarshal(body, &envelope); err != nil || envelope.Data.ID == "" || envelope.Data.Name == "" || envelope.Data.FileSHA256 == "" {
+		return Skill{}, output.ErrAPI("INVALID_RESPONSE", "marketplace skill metadata is incomplete", "retry or contact the marketplace operator")
+	}
+	return envelope.Data, nil
+}
+
+func (c *Client) DownloadSkill(ctx context.Context, id string) ([]byte, error) {
+	body, err := c.api.Do(ctx, &apiClient.Request{
+		Service:        "marketplace",
+		Method:         http.MethodGet,
+		Path:           "/skill/" + url.PathEscape(id) + "/download",
+		BinaryResponse: true,
+	})
+	if err != nil {
+		return nil, err
+	}
+	var redirect struct {
+		URL string `json:"url"`
+	}
+	if err := json.Unmarshal(body, &redirect); err != nil || redirect.URL == "" {
+		return nil, output.ErrAPI("INVALID_RESPONSE", "marketplace download response is missing a URL", "retry or contact the marketplace operator")
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, redirect.URL, nil)
+	if err != nil {
+		return nil, output.ErrAPI("INVALID_RESPONSE", "marketplace returned an invalid download URL", "contact the marketplace operator")
+	}
+	resp, err := c.http.Do(req)
+	if err != nil {
+		return nil, output.ErrNetwork(fmt.Sprintf("download skill archive: %v", err), "retry the download")
+	}
+	defer resp.Body.Close()
+	body, err = io.ReadAll(io.LimitReader(resp.Body, MaxArchiveBytes+1))
+	if err != nil {
+		return nil, output.ErrNetwork(fmt.Sprintf("read skill archive: %v", err), "retry the download")
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return nil, output.ParseBackendError(resp.StatusCode, body)
+	}
+	if int64(len(body)) > MaxArchiveBytes {
+		return nil, output.ErrValidation("skill archive exceeds download limit", "reduce the skill archive below 64 MiB")
+	}
+	return body, nil
+}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -56,6 +56,7 @@ octo-cli docs      create | list | get | rename | delete | forward-grant
                attachments presign|get|resolve
 
 octo-cli auth      login | status | logout | list
+octo-cli marketplace skills <skill-id> --install <skills-root>  (top-level alias: market)
 octo-cli schema [--list [domain] | <operation-id>]
 octo-cli api <METHOD> <PATH> [--params ...] [--data ...] [--service ...]
 octo-cli config show
@@ -63,7 +64,7 @@ octo-cli completion bash|zsh|fish|powershell
 octo-cli version
 ```
 
-`octo-cli auth login` stores a bot token (read from a hidden prompt, `--with-token` stdin, or `--token-file` — never argv) under a profile keyed by `--bot-id`/`--profile`. `status`/`list` show metadata only (tokens always masked); `logout` removes a profile.
+`octo-cli auth login` stores a bot token (read from a hidden prompt, `--with-token` stdin, or `--token-file` — never argv) under a profile keyed by `--bot-id`/`--profile`. Per-environment profiles may store `--api-base-url`; explicit environment variables still win. `status`/`list` show metadata only (tokens always masked); `logout` removes a profile.
 
 Bot-type capability and per-command flags are in `docs/octo-cli-design.md`. Agent-facing usage lives under `skills/` (`octo-shared`, `octo-matter` (withheld — see above), `octo-messaging`, `octo-files`, `octo-docs`) — keep those in sync when command shapes change.
 
@@ -75,6 +76,7 @@ Bot-type capability and per-command flags are in `docs/octo-cli-design.md`. Agen
 | `OCTO_BOT_ID`       | Robot id selecting a stored profile (env form of `--bot-id`). Selector, not a secret. |
 | `OCTO_CONFIG_DIR`   | Override the credential dir (default `~/.octo-cli`).     |
 | `OCTO_API_BASE_URL`  | Unified API base URL for all services. Required.          |
+| `OCTO_MARKETPLACE_API_PREFIX` | Marketplace gateway prefix on the same host (default `/market`; empty means direct `/api/v1`). |
 | `OCTO_SPACE_ID`     | Space context for platform-scoped bots.                  |
 | `OCTO_FORMAT`       | Default output format (`json` | `table` | `csv` | `ndjson`). |
 
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -72,14 +72,15 @@ func newConfigShowCmd(f *cmdutil.Factory) *cobra.Command {
 			}
 
 			payload := map[string]any{
-				"api_base_url":     cfg.APIBaseURL,
-				"space_id":         nullableString(space),
-				"format":           cfg.Format,
-				"bot_token":        maskToken(token),
-				"bot_token_source": nullableString(source),
-				"bot_kind":         botKind(token),
-				"profile":          nullableString(profileName),
-				"robot_id":         nullableString(robotID),
+				"api_base_url":           cfg.APIBaseURL,
+				"marketplace_api_prefix": cfg.MarketplaceAPIPrefix,
+				"space_id":               nullableString(space),
+				"format":                 cfg.Format,
+				"bot_token":              maskToken(token),
+				"bot_token_source":       nullableString(source),
+				"bot_kind":               botKind(token),
+				"profile":                nullableString(profileName),
+				"robot_id":               nullableString(robotID),
 			}
 			raw, err := json.Marshal(payload)
 			if err != nil {
--- a/cmd/marketplace.go
+++ b/cmd/marketplace.go
@@ -1,0 +1,84 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Mininglamp-OSS/octo-cli/internal/cmdutil"
+	"github.com/Mininglamp-OSS/octo-cli/internal/credential"
+	"github.com/Mininglamp-OSS/octo-cli/internal/marketplace"
+	"github.com/Mininglamp-OSS/octo-cli/internal/output"
+	"github.com/Mininglamp-OSS/octo-cli/internal/skillinstall"
+)
+
+func newMarketplaceCmd(f *cmdutil.Factory) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "marketplace",
+		Aliases: []string{"market"},
+		Short:   "Install agent resources from Octo Marketplace",
+	}
+	cmd.AddCommand(newMarketplaceSkillCmd(f))
+	return cmd
+}
+
+func newMarketplaceSkillCmd(f *cmdutil.Factory) *cobra.Command {
+	var installRoot string
+	cmd := &cobra.Command{
+		Use:   "skills <skill-id>",
+		Short: "Download and install one skill from Octo Marketplace",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if strings.TrimSpace(installRoot) == "" {
+				return emitMarketplaceError(f, output.ErrValidation("--install is required", "pass the agent skills root, for example --install ~/.codex/skills"))
+			}
+			cred, err := f.Credential()
+			if err != nil {
+				return emitMarketplaceError(f, err)
+			}
+			if credential.TokenKind(cred.Token) != "user_bot" {
+				return emitMarketplaceError(f, output.ErrAuth("marketplace skill installation requires a bf_* User Bot token", "select a User Bot profile or set OCTO_BOT_TOKEN to a bf_* token"))
+			}
+			api, err := f.Client()
+			if err != nil {
+				return emitMarketplaceError(f, err)
+			}
+			client := marketplace.NewClient(api, &http.Client{Timeout: 30 * time.Second})
+			skill, err := client.GetSkill(cmd.Context(), args[0])
+			if err != nil {
+				return emitMarketplaceError(f, err)
+			}
+			archive, err := client.DownloadSkill(cmd.Context(), args[0])
+			if err != nil {
+				return emitMarketplaceError(f, err)
+			}
+			installed, err := skillinstall.Install(installRoot, skill.Name, archive, skill.FileSHA256)
+			if err != nil {
+				return emitMarketplaceError(f, output.ErrValidation(fmt.Sprintf("install skill: %v", err), "verify the marketplace archive and target directory"))
+			}
+			payload, err := json.Marshal(map[string]any{
+				"source":       "marketplace",
+				"id":           skill.ID,
+				"name":         skill.Name,
+				"installed_to": installed.InstalledTo,
+				"sha256":       skill.FileSHA256,
+				"files":        installed.Files,
+			})
+			if err != nil {
+				return emitMarketplaceError(f, err)
+			}
+			return f.EmitSuccess(payload)
+		},
+	}
+	cmd.Flags().StringVar(&installRoot, "install", "", "install into this agent skills root")
+	return cmd
+}
+
+func emitMarketplaceError(f *cmdutil.Factory, err error) error {
+	_ = f.EmitError(err)
+	return err
+}
--- a/cmd/marketplace_test.go
+++ b/cmd/marketplace_test.go
@@ -1,0 +1,125 @@
+package cmd
+
+import (
+	"archive/zip"
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	apiClient "github.com/Mininglamp-OSS/octo-cli/internal/client"
+	"github.com/Mininglamp-OSS/octo-cli/internal/config"
+	"github.com/Mininglamp-OSS/octo-cli/internal/credential"
+)
+
+func marketplaceArchive(t *testing.T) ([]byte, string) {
+	t.Helper()
+	var buf bytes.Buffer
+	zw := zip.NewWriter(&buf)
+	body := []byte("---\nname: demo\n---\n")
+	w, err := zw.Create("SKILL.md")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if _, err := w.Write(body); err != nil {
+		t.Fatal(err)
+	}
+	if err := zw.Close(); err != nil {
+		t.Fatal(err)
+	}
+	sum := sha256.Sum256(buf.Bytes())
+	return buf.Bytes(), hex.EncodeToString(sum[:])
+}
+
+func TestMarketplaceSkillInstallAndAlias(t *testing.T) {
+	archive, digest := marketplaceArchive(t)
+	var gotAuth string
+	var srv *httptest.Server
+	srv = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.Path {
+		case "/gateway/api/v1/skill/skill-1":
+			gotAuth = r.Header.Get("Authorization")
+			_, _ = w.Write([]byte(`{"code":0,"data":{"id":"skill-1","name":"demo","file_sha256":"` + digest + `"}}`))
+		case "/gateway/api/v1/skill/skill-1/download":
+			gotAuth = r.Header.Get("Authorization")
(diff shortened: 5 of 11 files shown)


Jerry-Xin (dismissed):
The PR is in scope for `Mininglamp-OSS/octo-cli`: it extends the CLI with Marketplace skill installation and related config/install support. I found no blocking correctness or security issues.

**💬 Non-blocking**
- 🟡 Warning: `internal/marketplace/client.go:57` only supports a redirect-style download response. If the Marketplace endpoint ever returns the archive directly with `200`, the shared `BinaryResponse` path converts it to metadata, so installation fails with "missing a URL." Either document the endpoint as redirect-only or add direct-archive coverage.
- 🟡 Warning: `cmd/marketplace.go:50` hard-codes a separate `30s` HTTP client for artifact downloads, so the global `--timeout` behavior is not honored for the largest part of the install flow. This can surprise users installing larger skills or using slower artifact storage.
- 🔵 Suggestion: `internal/marketplace/client.go:48` parses `data` but ignores the Marketplace envelope `code`. If this API uses `{code:0,data:...}` as the success contract, validating `code == 0` would make backend failures clearer.
- 🔵 Suggestion: the download URL returned by `/skill/<id>/download` (`internal/marketplace/client.go:73`) is fetched without scheme restriction. Content is SHA-256 verified after download and the URL originates from the authenticated backend, so this is not exploitable under the current trust model, but restricting to `https` would harden against a compromised/misconfigured backend.

**✅ Highlights**
- `internal/skillinstall/installer.go` verifies the SHA-256 digest before extraction and rejects traversal paths (multi-layer: `filepath.Clean` + `IsAbs` + `..` prefix + `filepath.Rel` check), symlinks, non-regular entry types, oversized files/archives (64 MiB download, 32 MiB total, 16 MiB/file, 1024-file cap via bounded `io.CopyN`), and archives missing `SKILL.md`.
- Install is atomic and target-scoped: extraction happens in a temp dir under the install root, then rename with a backup/restore path; unrelated skill directories are preserved. Files are created with `O_EXCL` and `0644`, dirs `0755` (no world-writable perms).
- `bf_*` User Bot token is required; marketplace routing reuses the trusted API host (only the path prefix changes), and `bot_token` remains masked in `config show`.
- Tests are meaningful and cover the security-relevant cases: path traversal, symlink rejection, missing `SKILL.md`, checksum mismatch, and target-scoped replacement — not just the happy path.

Verification run: `go test ./...` passed.

internal/marketplace/client.go
```diff
@@ -0,0 +1,93 @@
...
+		URL string `json:"url"`
+	}
+	if err := json.Unmarshal(body, &redirect); err != nil || redirect.URL == "" {
+		return nil, output.ErrAPI("INVALID_RESPONSE", "marketplace download response is missing a URL", "retry or contact the marketplace operator")
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, redirect.URL, nil)
+	if err != nil {
+		return nil, output.ErrAPI("INVALID_RESPONSE", "marketplace returned an invalid download URL", "contact the marketplace operator")
+	}
+	resp, err := c.http.Do(req)
```
OctoBoooot, line 77: **[minor]** SSRF defense-in-depth: `redirect.URL` (from the `/download` response) is followed here with no scheme/host allowlist, and the client sets no `CheckRedirect`, so the default policy chases up to 10 further redirects. `file://` is safe (Go's transport rejects it), but an `http(s)://` URL to an internal address (169.254.169.254 metadata, localhost admin ports) would be fetched blindly. It's gated by the trusted marketplace backend so it's not attacker-controllable without compromising that API — defense-in-depth, not a live hole — but on an install path this sensitive, restrict to https and/or set a CheckRedirect that re-validates each hop (or refuses redirects, since the backend returns a final URL). The checksum protects what gets installed, not where the CLI is pointed.

OctoBoooot (commented):
## Review: feat: install skills from marketplace (#87)

**Verdict:** Comment — the security-critical logic (download → verify → extract → install) is genuinely well built, no blocker or major. Holding at COMMENT rather than APPROVE for one reason: the repo's **code CI has not run** on this PR (see below), so I can't confirm the build/test/osv gate is green on a security-sensitive path — plus one minor hardening item and one nit.

Risk tier: **high** (this is a network-fetch-then-write-to-disk feature — the classic path-traversal / SSRF / arbitrary-write surface). I read the installer, the marketplace client, the command wiring, and the tests, and built + ran the targeted tests locally.

Only `check-sprint` (governance, red), `add-to-project`, and `label` ran; `code-review` is pending. The code gates that ran on #85/#86 — `build & test (1.24.x)`, `npm test`, `scan-pr/osv-scan` — are **absent**, consistent with GitHub holding a first-time contributor's workflows for maintainer approval. I verified locally: `go build ./...` clean; `go test ./internal/skillinstall/ ./internal/marketplace/ ./internal/config/` pass. (The `cmd` package's `TestAuth_*` failures are **pre-existing and environmental** — they fail identically on `main`, from an unavailable keyring/machine-id in the sandbox, not from this PR.) A maintainer should let code CI run before merge; I can't attest the repo gate from here.

### Minor — SSRF defense-in-depth on the artifact redirect
`internal/marketplace/client.go:73-77` — `DownloadSkill` takes the `{url}` from the authenticated `/download` response and issues `c.http.Do` against it with **no scheme/host allowlist**, and the client (`cmd/marketplace.go:50`) sets only a timeout, so Go's default `CheckRedirect` follows up to 10 further redirects. `file://` is already safe (Go's transport rejects it), but an `http(s)://` URL pointing at an internal address (`169.254.169.254` cloud metadata, `localhost` admin ports) would be fetched blindly. This is **gated by the trusted marketplace backend** (the URL isn't attacker-controllable without compromising that authenticated API), so it's defense-in-depth rather than a live hole — hence minor. On an install path this sensitive it's still worth: restrict to `https`, and/or set a `CheckRedirect` that re-validates each hop (or refuses redirects, since the backend already returns a final URL). The checksum protects *integrity* of what's installed, but not *where the CLI is pointed*. _(confidence: high that it's unguarded; medium on exploitability given the trust boundary.)_

### Nit (≤3)
- `CLAUDE.md:27` — the "hand-written leaves are `schema`, `version`, `api`, `config`, `auth`, and `completion`" sentence wasn't extended to include `marketplace` (the command tree above it was). One-word consistency fix. _(confidence: high)_

### Praise (specific — this is the right way to build a skill installer)
- **Checksum-before-extract is the linchpin and it's correct**: `Install` verifies `sha256(archive) == skill.FileSHA256` (from the *separate authenticated* `GetSkill` call) *before* touching the zip — so a tampered artifact from the redirect can never be written. _(confidence: high)_
- **Path-traversal is blocked twice and tested**: `filepath.Clean` + explicit `..`/abs rejection, then a `filepath.Rel` escape re-check; symlinks and non-regular entries refused; zip-bomb bounds (file-count + per-file + total, `io.CopyN` capped to the declared size); `O_EXCL` create. `installer_test.go` exercises traversal, symlink, missing-SKILL.md, and checksum-mismatch — mutation-meaningful. _(confidence: high)_
- **The redirect download carries no credentials** — a bare `http.NewRequestWithContext` with no auth header, so the bot token is never leaked to the (possibly third-party) artifact host. And the atomic temp→backup→rename install with rollback means a failed install never corrupts an existing skill. _(confidence: high)_

Net: strong, security-conscious implementation. Once code CI runs green, the only substantive item is the SSRF hardening (minor) — address or consciously accept it given the trust boundary, and this is APPROVE-ready from my side.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #87 (octo-cli)

## Summary

This PR adds a `marketplace skills <skill-id> --install <root>` command (with `market` alias) that downloads a skill archive from the Octo Marketplace API, validates its SHA-256 checksum, and extracts it into the target directory with atomic replace semantics. It also introduces a `MarketplaceAPIPrefix` config field and `ServiceURL("marketplace")` routing so marketplace requests go through a separate gateway prefix on the same host. The zip extraction in `internal/skillinstall/installer.go` has solid defense-in-depth (path traversal, symlink rejection, file-count and size budgets, O_EXCL writes). One security-critical issue in the download path requires attention before merge.

## Verification

Static analysis only at head `b8934bf30b9110cb9dd984f1824f374185d0f03b`; build and tests not executed in this environment.

- ✅ **Zip path traversal** — dual-layer defense at `internal/skillinstall/installer.go:91-101`: syntactic `filepath.Clean` + `filepath.Rel` re-verification. Both `../escape` and `foo/../../etc/passwd` are correctly rejected.
- ✅ **Symlink rejection** — `installer.go:102-104` checks `os.ModeSymlink`; file creation uses `os.O_EXCL` at `:129` preventing symlink-to-regular-file attacks regardless of zip header mode bits.
- ✅ **Auth token isolation** — the download request at `internal/marketplace/client.go:73-77` uses the plain `http.Client` without setting an `Authorization` header; the bearer token is only attached to the API client request to the trusted backend.
- ✅ **Token type gate** — `cmd/marketplace.go:44-46` correctly restricts to `user_bot` (`bf_*`) tokens via `credential.TokenKind()`, rejecting `app_bot` and `unknown`.
- ✅ **overlayProfileBaseURL** — behavior-preserving refactor: truth table verified (env var wins over profile in all four profile×env combinations).
- ✅ **ServiceURL marketplace routing** — `config.go:69-80` only modifies `u.Path`; the host always comes from `APIBaseURL`. No scheme/host injection via `OCTO_MARKETPLACE_API_PREFIX`.
- ✅ **SHA-256 integrity gate** — `installer.go:33-37` validates checksum before extraction; mismatch short-circuits to error.
- ✅ **Archive size budgets** — `MaxArchiveBytes` (64 MiB) enforced via `LimitReader` + post-read check; per-file (16 MiB) and total (32 MiB) limits in extraction.

## Findings

One P1 blocker and one P2 item.

### P1 — SSRF via unvalidated redirect URL (`internal/marketplace/client.go:77`)

`DownloadSkill` receives a download URL from the API response (`redirect.URL` at `:68`) and follows it at `:77` with `c.http.Do(req)`. The `http.Client` created at `cmd/marketplace.go:50` has no `CheckRedirect` function, so Go's default policy applies — follow up to 10 redirects automatically. Three defenses are missing:

1. **No scheme validation.** The URL could be `http://`, `file://`, or any scheme Go accepts.
2. **No private/internal IP blocking.** The URL is not resolved and checked against RFC 1918 ranges, link-local (`169.254.169.254` — cloud metadata), or loopback before connecting. A malicious API response `"url": "http://169.254.169.254/latest/meta-data/"` would cause the CLI to fetch cloud instance credentials from the operator's machine.
3. **No redirect-chain suppression.** Even if the initial URL looks external, the server can chain redirects to internal endpoints. Contrast with the API client at `internal/client/client.go:109` which correctly sets `CheckRedirect: func(...) error { return http.ErrUseLastResponse }`.

The response body (up to 64 MiB + 1 byte) is fully read into memory at `:82` before the SHA-256 check in `skillinstall.Install` rejects it — so data IS fetched from the internal target before any validation occurs.

**Diff-scope:** Entirely new code introduced by this PR.

**Fix direction:** (1) Require `https` scheme, (2) resolve hostname and reject RFC 1918 / link-local / loopback / unspecified ranges, (3) set `CheckRedirect` to `http.ErrUseLastResponse` on the download `http.Client`.

```go
// minimal fix sketch for cmd/marketplace.go
httpClient := &http.Client{
    Timeout: 30 * time.Second,
    CheckRedirect: func(req *http.Request, via []*http.Request) error {
        return http.ErrUseLastResponse
    },
}
```

### P2 — Full body read before status-code check (`internal/marketplace/client.go:81-86`)

`io.ReadAll(io.LimitReader(resp.Body, MaxArchiveBytes+1))` at `:82` executes before the status-code check at `:85`. On error responses from the redirect target (4xx/5xx), up to 64 MiB is read into memory only to be passed to `ParseBackendError` and discarded. Move the status-code check before the body read:

```go
if resp.StatusCode < 200 || resp.StatusCode >= 300 {
    body, _ := io.ReadAll(io.LimitReader(resp.Body, 4096))
    return nil, output.ParseBackendError(resp.StatusCode, body)
}
body, err = io.ReadAll(io.LimitReader(resp.Body, MaxArchiveBytes+1))
```

## Data Flow Trace

| Consumer | Upstream Source | Data Flows? |
|----------|----------------|-------------|
| `redirect.URL` → `http.NewRequestWithContext` | API response JSON `{"url":"..."}` | Yes — untrusted server data used as network destination (P1) |
| `skill.FileSHA256` → `sha256.Sum256` comparison | `GetSkill` API response `data.file_sha256` | Yes — validated before extraction |
| `skill.Name` → `namePattern` regex + `filepath.Join` | `GetSkill` API response `data.name` | Yes — validated by `^[a-z0-9][a-z0-9._-]{0,127}$` regex |
| `archive []byte` → `zip.NewReader` | Download redirect target response body | Yes — bounded by `MaxArchiveBytes+1` LimitReader |
| `cfg.MarketplaceAPIPrefix` → `u.Path` assignment | `os.Getenv("OCTO_MARKETPLACE_API_PREFIX")` or default `/market/api/v1` | Yes — path-only, cannot override host |

## Blindspot Checklist (C1–C6, security_sensitive)

- **C1 — Dual-path parity:** N/A. No symmetric add/remove/subscribe/unsubscribe pairs in this PR.
- **C2 — Control-flow ordering / safety control bypass:** `namePattern` regex tested with valid names only in tests. The regex `^[a-z0-9][a-z0-9._-]{0,127}$` correctly rejects empty strings, leading dots/hyphens, and names >128 chars. No unicode bypass (ASCII-only charset). Clear.
- **C3 — Authorization boundary ≠ capability boundary:** The `user_bot` token check at `cmd/marketplace.go:44-46` gates the entire marketplace command. The download request does NOT carry the auth token (verified). No additional endpoint exposure. Clear.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. No hierarchical auth entities in this PR.
- **C5 — Build/note pass ≠ runtime correctness:** N/A. This is a CLI tool, not a browser extension or packaged artifact.
- **C6 — Governance/strategy/document self-consistency:** `CLAUDE.md` docs updated to match new command shape and env var. The documented default `/market` vs the code's default `/market/api/v1` is slightly inconsistent — `CLAUDE.md` says "default `/market`" but `marketplaceAPIRoot()` returns `/market/api/v1`. Not a blocker, but the doc should say "default `/market/api/v1`" or "prefix default `/market`" with the `/api/v1` suffix explained.

## Verdict

One P1 blocker (SSRF via unvalidated redirect URL) must be addressed before merge. The P2 and C6 doc nit are non-blocking but recommended.

[Octo-Q] verdict: REQUEST_CHANGES — P1 SSRF in `DownloadSkill` is a production-reachable security vulnerability: a compromised marketplace API can redirect the CLI to fetch cloud metadata credentials or probe internal services from the operator's machine.

Jerry-Xin (commented):
Revising my earlier APPROVE to **COMMENT**. The install path is genuinely well-built, but I should not have APPROVED it: this repo's code CI never ran on this PR, and APPROVE on a security-sensitive install path requires a green code gate.

⛔ Why not APPROVE (process)
- The only checks on this PR are `check-sprint` (governance, failing), `label`, `add-to-project`, and a pending `code-review`. The code CI — `build & test`, `npm test`, `osv-scan` — **did not run at all** (consistent with first-time-contributor workflow gating awaiting maintainer approval). `mergeStateStatus` is BLOCKED. Unrun ≠ green, and on a network-fetch-then-write-to-disk feature I won't attest the repo gate from a local build. Once maintainers approve the workflows and code CI is green, this is APPROVE-ready from my side.

✅ Security design (byte-verified, genuinely strong — not a blocker)
- **Checksum before extract**: the archive is SHA-256-verified against `expectedSHA` from the authenticated metadata API *before* anything touches disk (fail-closed on mismatch). A tampered artifact can never be written.
- **Path traversal blocked twice**: `filepath.Clean(FromSlash(name))` + reject abs/`.`/`..`/`../`-prefix, then a `filepath.Rel(dst, path)` escape check. Symlinks and non-regular entries rejected. Zip-bomb bounds (1024 files / 16 MiB per file / 32 MiB total, with `io.CopyN` bounded by declared size). `O_EXCL|0644` create (no clobber), atomic temp→rename with backup/restore. Tests cover traversal, symlink, missing SKILL.md, and checksum mismatch.
- **Token not leaked** to the presigned/redirect artifact host.

💬 Minor / hardening (non-blocking)
- **SSRF defense-in-depth on the download redirect** (`internal/marketplace/client.go`): the `/download` redirect URL is followed with no scheme/host allowlist, no private-IP (e.g. `169.254.169.254`, `localhost`) block, and default redirect-chasing on. Go's default client does reject `file://`, and the URL comes from the *authenticated* marketplace API (so it's not attacker-controllable without a compromised backend, and the checksum still gates what installs) — which keeps this minor, not a blocker. But on a security-critical install path, restricting to `https`, rejecting private/link-local hosts, and capping redirect hops are cheap and worth adding.
- **`DownloadSkill` handles only the redirect-`{url}` response shape**, not a direct-archive body — fine if the backend contract always returns the JSON `{url}` envelope, but it's an unstated assumption worth a comment.
- **Download timeout is hardcoded 30s** (`cmd/marketplace.go`), ignoring the global `--timeout` flag (`cmd/root.go`); a user passing `--timeout 5m` for a large skill still gets cut at 30s.
- Doc nit in `CLAUDE.md` alongside the new command.

Bottom line: solid, security-conscious implementation; the operative gate is that code CI hasn't run (SKILL requires green to APPROVE). Get the workflows approved + green, optionally address the SSRF hardening, and I'm happy to flip to APPROVE.

kense-lab:
先在 fork 仓库维护，官方仓库暂不合并

yujiawei (changes requested):
# Code Review — PR #87 (octo-cli)

**Scope:** single commit `b8934bf` (`feat: install skills from marketplace`) against merge-base `0eb40d1`. 11 files, +649/-24. This is a new feature: `octo-cli marketplace skills <id> --install <root>` (alias `market`) fetches skill metadata + a signed download URL from the marketplace API, downloads a zip, verifies its sha256, and extracts it into the agent skills root. Reviewed with an adversarial security pass (this PR touches untrusted-archive extraction, download integrity, and a token gate).

## 1. Spec compliance

**Spec: ✅**

The change is a coherent, self-contained feature. No linked spec/plan to diff against, so this is judged on internal consistency:

- **No under-build:** metadata fetch, signed-URL download, size caps, checksum verification, safe extraction, atomic swap-with-backup, and a `config show` field for the new prefix are all present and tested.
- **No over-build:** surface stays scoped to one `skills` subcommand + one env var (`OCTO_MARKETPLACE_API_PREFIX`) + one config field. The `overlayProfileBaseURL` refactor is behavior-preserving for `APIBaseURL` (env still wins; net result identical).
- **No deviation:** routing reuses the existing unified-client/`ServiceURL` mechanism; the presigned-download flow mirrors the established `file.download` 302→URL pattern.

## 2. Code quality

### P1 — blocks merge

**1. CI lint gate will fail: 20 net-new `golangci-lint` violations (all in this PR's files).**
The required `build & test` job runs `golangci-lint` v2.12.2 (`.golangci.yml`, `errcheck.check-blank: true`). On a clean checkout of this branch it reports **20 issues, 0 on the base branch**. `go build/vet/test ./...` pass locally but do not run this linter, so this is easy to miss. Full list:

- `errcheck` (7) — blank-assigned errors are still flagged because `check-blank: true`. The repo convention is an explicit directive, e.g. `//nolint:errcheck // best-effort cleanup` (see `internal/client/client.go:234,333`):
  - `cmd/marketplace.go:82` `f.EmitError`
  - `internal/marketplace/client.go:81` `resp.Body.Close`
  - `internal/skillinstall/installer.go:50,68,85` `os.RemoveAll`; `:80` `os.Rename`; `:133` `rc.Close`
- `revive` exported-doc (8) — every exported symbol needs a doc comment: `internal/marketplace/client.go` `MaxArchiveBytes`(15), `APIClient`(17), `Skill`(21), `Client`(27), `NewClient`(32), `GetSkill`(39), `DownloadSkill`(57); `internal/skillinstall/installer.go` `Result`(25).
- `gocyclo` (2) — `Install` (complexity 17) and `extract` (26) exceed the `min-complexity: 15` gate. These need a small refactor (e.g. split the extract loop's per-entry validation into a helper).
- `gocritic` (3) — `internal/marketplace/client.go:73` `httpNoBody` (use `http.NoBody` not `nil`); `cmd/marketplace_test.go:20` and `internal/skillinstall/installer_test.go:20` `unnamedResult` (name the `([]byte, string)` returns).

None are logic bugs, but a red required lint check means the PR can't merge as-is, so this is a blocker rather than a nit.

### P2 — hardening / should-fix (not merge-blocking, but bundle into the same round)

**2. Validate the download URL scheme before fetching it (`internal/marketplace/client.go:73-77`).** The URL from the API response is passed straight to `http.NewRequestWithContext` + `c.http.Do` with no scheme check. Go's default transport does reject `file://`/`gopher://`/`ftp://` ("unsupported protocol scheme"), so this is not RCE/LFI; the residual risk is `http(s)://internal-host:port` probing **if** the marketplace API itself is compromised or spoofed (the API is the integrity root here). Recommend enforcing `https://` (or an explicit http/https allowlist) as defense-in-depth. `security_sensitive` — worth a human eye on the intended trust boundary.

**3. Token-kind gate is client-side only and bypassable (`cmd/marketplace.go:43` vs `cmd/api.go`).** The `bf_*` User-Bot check guards the `marketplace skills` command, but the generic `api` passthrough accepts any `--service`, and this PR newly makes `--service marketplace` route to the marketplace prefix. So `octo-cli api GET /skill/<id> --service marketplace` reaches the same endpoints with an `app_*` token, skipping the gate. This is fine **iff** the backend enforces the `bf_*` requirement server-side — a client-side gate is never an authz boundary. **Please confirm the backend enforces this**; if the CLI check is the only enforcement, it's a real gap. (Flagging for the human reviewer given the `needs-human-review` label.)

**4. Artifact download drops the bearer token (`internal/marketplace/client.go:73-77`).** The redirect target is fetched with a bare client and no `Authorization` header. This is the correct presigned-URL model (matches `file.download`), and forwarding the token to an arbitrary artifact host would be worse. Calling it out only so it's a conscious decision: the security of the download rests entirely on the URL being a short-lived signed URL. Worth a one-line comment documenting that assumption.

**5. `ServiceURL("marketplace")` erases any base path (`internal/config/config.go:74-75`).** `u.Path = c.MarketplaceAPIPrefix` overwrites the whole path, so a path-based reverse proxy base (`OCTO_API_BASE_URL=https://proxy/octo-backend`) loses `/octo-backend` and becomes `https://proxy/market/api/v1`. Consider `url.JoinPath`/appending to preserve an existing base path.

**6. Download timeout is a hardcoded 30s and ignores `--timeout` (`cmd/marketplace.go:50`).** `GetSkill` uses the timeout-aware `api` client, but the artifact fetch uses `&http.Client{Timeout: 30*time.Second}`, and that 30s covers reading up to 64 MiB (~>2 MB/s sustained). A user who sets `--timeout 5m` for a slow link still hits 30s on the download. Thread the configured timeout (or a context deadline) into the download client.

**7. Silent error discard in the backup/restore swap (`internal/skillinstall/installer.go:80`).** If `os.Rename(tmp, target)` fails and the restore `os.Rename(backup, target)` also fails, the restore error is dropped. Data isn't truly lost (the backup dir survives and the `defer` cleans up), but the user gets the original error with no hint the old skill wasn't restored. Log/wrap the restore failure. (This is also one of the `errcheck` hits in P1.)

### What's solid (reviewed, no action)

- **Zip extraction is well-defended (`installer.go:extract`)**: `filepath.Clean` + `IsAbs` + `..`-prefix + `Rel`-escape checks, symlink/non-regular rejection, `O_EXCL` create, `maxFiles`/`maxFileSize`/`maxTotalSize` caps, and `io.CopyN` bounded to the declared size. The zip-slip and symlink test cases cover the traversal paths. Two adversarial findings (using `os.Stat` vs `os.Lstat` for the `SKILL.md` check; reading the body before the status check) were checked and dropped — symlinks are fully rejected at extraction, and a non-2xx download returns via `ParseBackendError` before the body is ever treated as archive data.
- **Integrity gate ordering is correct**: sha256 is verified against `file_sha256` before `extract` runs. Note the checksum and the download URL both come from the same API, so this is same-origin/transit integrity (protects against a tampered/corrupt artifact in object storage), not protection against a malicious API — consistent with the trust model above.
- Download size guard: `io.LimitReader(resp.Body, MaxArchiveBytes+1)` + post-read length check correctly catches the boundary (the `+1` is the standard over-read trick).

## 3. Overall verdict

**REQUEST_CHANGES** — Spec ✅, but Quality is Changes-Requested on the P1 CI lint failure (20 net-new `golangci-lint` violations that will fail the required `build & test` job). The P2 items (URL-scheme validation, the client-side-only token gate + backend-enforcement question, path-preserving routing, download timeout, restore-error logging) are non-blocking but should ride along in the same fix round given the security-sensitive classification.

## 4. Suggested fix direction

1. Clear all 20 lint issues: add `//nolint:errcheck // <reason>` on the intentional best-effort/blank error drops (matching existing repo style), add doc comments to the exported marketplace/installer symbols, switch `nil`→`http.NoBody`, name the test helpers' returns, and split `Install`/`extract` to get under complexity 15. Run `golangci-lint run ./...` locally before pushing — `go vet` won't catch these.
2. Add an `https://` (or http/https allowlist) scheme check on the download URL.
3. Confirm and, ideally, document that the backend enforces the `bf_*` requirement for marketplace endpoints (the CLI gate alone is bypassable via `api --service marketplace`).
4. Thread `--timeout` into the download client; use `url.JoinPath` in `ServiceURL`; log the restore-Rename failure.

## 5. Additional notes

- CI context: `check-sprint` is red, but that's a project-board field check, not a code issue. The `build & test` (CI) and `OSV-Scanner` jobs show `action_required` (awaiting maintainer approval to run for this contributor) rather than a result — the lint failure above is from running the pinned `golangci-lint` v2.12.2 locally against this head.
- Reviewer transparency: an independent line-level second-opinion tool was unavailable this run (persistent gateway error on every attempt) and is recorded as absent, not as agreement. Findings here come from a manual pass plus an adversarial multi-lens analysis and one long-context whole-repo advisory pass; a build/vet/test run (all green locally) and a full `golangci-lint` run backed the P1.

yujiawei:
Follow-up on the download-URL finding from my review — one concrete addition to the fix direction.

Enforcing an `https://` scheme on the URL returned by the API is necessary but not sufficient on its own: the download `http.Client` (`cmd/marketplace.go:50`) has no `CheckRedirect`, so Go's default policy follows up to 10 redirects automatically. An `https` first hop can still 3xx-chain to an internal target (e.g. `http://169.254.169.254/...` or an RFC-1918 host). The sibling API client already guards against exactly this at `internal/client/client.go:109`:

```go
CheckRedirect: func(req *http.Request, via []*http.Request) error {
    return http.ErrUseLastResponse
},
```

Recommend the download client mirror that pattern (scheme allowlist + `CheckRedirect: http.ErrUseLastResponse`, and ideally reject loopback/link-local/RFC-1918 resolved hosts).

Scope note on impact: the archive body is SHA-256-checked locally and discarded on mismatch — it is never sent back to the marketplace API — so this is a blind SSRF (internal reachability/port-probing and GET-side-effects), not direct exfiltration of a fetched response. Still worth closing before merge given the cheap fix and the existing in-repo pattern to follow.

PR #89, opened 2026-07-16
feat: install Skills from Octo Marketplace

kense-lab:
## Summary

- add `octo-cli marketplace skills <skill-id> --install <dir>` with the `market` alias
- download and verify Marketplace Skill archives before atomic installation
- harden artifact downloads with HTTPS-only, private-address rejection, size limits, and redirect blocking
- align the CLI client with the Marketplace OpenAPI contract: plural `/skills/{skill_id}` routes and `skill_id` response fields

## Verification

- `go test ./...`

## API contract

- `GET /skills/{skill_id}`
- `GET /skills/{skill_id}/download`
- response resource identifier: `skill_id`

### DIFF
--- a/internal/marketplace/client.go
+++ b/internal/marketplace/client.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 
@@ -13,6 +14,7 @@ import (
 )
 
 const MaxArchiveBytes int64 = 64 << 20
+const maxErrorBodyBytes int64 = 4 << 10
 
 type APIClient interface {
 	Do(context.Context, *apiClient.Request) ([]byte, error)
@@ -70,6 +72,9 @@ func (c *Client) DownloadSkill(ctx context.Context, id string) ([]byte, error) {
 	if err := json.Unmarshal(body, &redirect); err != nil || redirect.URL == "" {
 		return nil, output.ErrAPI("INVALID_RESPONSE", "marketplace download response is missing a URL", "retry or contact the marketplace operator")
 	}
+	if err := validateDownloadURL(ctx, redirect.URL); err != nil {
+		return nil, err
+	}
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, redirect.URL, nil)
 	if err != nil {
 		return nil, output.ErrAPI("INVALID_RESPONSE", "marketplace returned an invalid download URL", "contact the marketplace operator")
@@ -79,15 +84,45 @@ func (c *Client) DownloadSkill(ctx context.Context, id string) ([]byte, error) {
 		return nil, output.ErrNetwork(fmt.Sprintf("download skill archive: %v", err), "retry the download")
 	}
 	defer resp.Body.Close()
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		errorBody, readErr := io.ReadAll(io.LimitReader(resp.Body, maxErrorBodyBytes))
+		if readErr != nil {
+			return nil, output.ErrNetwork(fmt.Sprintf("read skill download error response: %v", readErr), "retry the download")
+		}
+		return nil, output.ParseBackendError(resp.StatusCode, errorBody)
+	}
 	body, err = io.ReadAll(io.LimitReader(resp.Body, MaxArchiveBytes+1))
 	if err != nil {
 		return nil, output.ErrNetwork(fmt.Sprintf("read skill archive: %v", err), "retry the download")
-	}
-	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		return nil, output.ParseBackendError(resp.StatusCode, body)
 	}
 	if int64(len(body)) > MaxArchiveBytes {
 		return nil, output.ErrValidation("skill archive exceeds download limit", "reduce the skill archive below 64 MiB")
 	}
 	return body, nil
+}
+
+func validateDownloadURL(ctx context.Context, rawURL string) error {
+	parsed, err := url.Parse(rawURL)
+	if err != nil || parsed.Scheme != "https" || parsed.Hostname() == "" {
+		return output.ErrValidation("marketplace download URL must use https", "contact the marketplace operator")
+	}
+
+	var addresses []net.IP
+	if literal := net.ParseIP(parsed.Hostname()); literal != nil {
+		addresses = []net.IP{literal}
+	} else {
+		resolved, err := net.DefaultResolver.LookupIPAddr(ctx, parsed.Hostname())
+		if err != nil {
+			return output.ErrNetwork(fmt.Sprintf("resolve marketplace download host: %v", err), "retry the download")
+		}
+		for _, address := range resolved {
+			addresses = append(addresses, address.IP)
+		}
+	}
+	for _, address := range addresses {
+		if address.IsPrivate() || address.IsLoopback() || address.IsLinkLocalUnicast() || address.IsUnspecified() {
+			return output.ErrValidation("marketplace download URL resolves to a private or local address", "contact the marketplace operator")
+		}
+	}
+	return nil
 }
--- a/internal/marketplace/client.go
+++ b/internal/marketplace/client.go
@@ -1,0 +1,93 @@
+package marketplace
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+
+	apiClient "github.com/Mininglamp-OSS/octo-cli/internal/client"
+	"github.com/Mininglamp-OSS/octo-cli/internal/output"
+)
+
+const MaxArchiveBytes int64 = 64 << 20
+
+type APIClient interface {
+	Do(context.Context, *apiClient.Request) ([]byte, error)
+}
+
+type Skill struct {
+	ID         string `json:"id"`
+	Name       string `json:"name"`
+	FileSHA256 string `json:"file_sha256"`
+}
+
+type Client struct {
+	api  APIClient
+	http *http.Client
+}
+
+func NewClient(api APIClient, httpClient *http.Client) *Client {
+	if httpClient == nil {
+		httpClient = http.DefaultClient
+	}
+	return &Client{api: api, http: httpClient}
+}
+
+func (c *Client) GetSkill(ctx context.Context, id string) (Skill, error) {
+	body, err := c.api.Do(ctx, &apiClient.Request{
+		Service: "marketplace",
+		Method:  http.MethodGet,
+		Path:    "/skill/" + url.PathEscape(id),
+	})
+	if err != nil {
+		return Skill{}, err
+	}
+	var envelope struct {
+		Data Skill `json:"data"`
+	}
+	if err := json.Unmarshal(body, &envelope); err != nil || envelope.Data.ID == "" || envelope.Data.Name == "" || envelope.Data.FileSHA256 == "" {
+		return Skill{}, output.ErrAPI("INVALID_RESPONSE", "marketplace skill metadata is incomplete", "retry or contact the marketplace operator")
+	}
+	return envelope.Data, nil
+}
+
+func (c *Client) DownloadSkill(ctx context.Context, id string) ([]byte, error) {
+	body, err := c.api.Do(ctx, &apiClient.Request{
+		Service:        "marketplace",
+		Method:         http.MethodGet,
+		Path:           "/skill/" + url.PathEscape(id) + "/download",
+		BinaryResponse: true,
+	})
+	if err != nil {
+		return nil, err
+	}
+	var redirect struct {
+		URL string `json:"url"`
+	}
+	if err := json.Unmarshal(body, &redirect); err != nil || redirect.URL == "" {
+		return nil, output.ErrAPI("INVALID_RESPONSE", "marketplace download response is missing a URL", "retry or contact the marketplace operator")
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, redirect.URL, nil)
+	if err != nil {
+		return nil, output.ErrAPI("INVALID_RESPONSE", "marketplace returned an invalid download URL", "contact the marketplace operator")
+	}
+	resp, err := c.http.Do(req)
+	if err != nil {
+		return nil, output.ErrNetwork(fmt.Sprintf("download skill archive: %v", err), "retry the download")
+	}
+	defer resp.Body.Close()
+	body, err = io.ReadAll(io.LimitReader(resp.Body, MaxArchiveBytes+1))
+	if err != nil {
+		return nil, output.ErrNetwork(fmt.Sprintf("read skill archive: %v", err), "retry the download")
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return nil, output.ParseBackendError(resp.StatusCode, body)
+	}
+	if int64(len(body)) > MaxArchiveBytes {
+		return nil, output.ErrValidation("skill archive exceeds download limit", "reduce the skill archive below 64 MiB")
+	}
+	return body, nil
+}
--- a/internal/marketplace/client.go
+++ b/internal/marketplace/client.go
@@ -21,7 +21,7 @@ type APIClient interface {
 }
 
 type Skill struct {
-	ID         string `json:"id"`
+	ID         string `json:"skill_id"`
 	Name       string `json:"name"`
 	FileSHA256 string `json:"file_sha256"`
 }
@@ -42,7 +42,7 @@ func (c *Client) GetSkill(ctx context.Context, id string) (Skill, error) {
 	body, err := c.api.Do(ctx, &apiClient.Request{
 		Service: "marketplace",
 		Method:  http.MethodGet,
-		Path:    "/skill/" + url.PathEscape(id),
+		Path:    "/skills/" + url.PathEscape(id),
 	})
 	if err != nil {
 		return Skill{}, err
@@ -60,7 +60,7 @@ func (c *Client) DownloadSkill(ctx context.Context, id string) ([]byte, error) {
 	body, err := c.api.Do(ctx, &apiClient.Request{
 		Service:        "marketplace",
 		Method:         http.MethodGet,
-		Path:           "/skill/" + url.PathEscape(id) + "/download",
+		Path:           "/skills/" + url.PathEscape(id) + "/download",
 		BinaryResponse: true,
 	})
 	if err != nil {
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -56,6 +56,7 @@ octo-cli docs      create | list | get | rename | delete | forward-grant
                attachments presign|get|resolve
 
 octo-cli auth      login | status | logout | list
+octo-cli marketplace skills <skill-id> --install <skills-root>  (top-level alias: market)
 octo-cli schema [--list [domain] | <operation-id>]
 octo-cli api <METHOD> <PATH> [--params ...] [--data ...] [--service ...]
 octo-cli config show
@@ -63,7 +64,7 @@ octo-cli completion bash|zsh|fish|powershell
 octo-cli version
 ```
 
-`octo-cli auth login` stores a bot token (read from a hidden prompt, `--with-token` stdin, or `--token-file` — never argv) under a profile keyed by `--bot-id`/`--profile`. `status`/`list` show metadata only (tokens always masked); `logout` removes a profile.
+`octo-cli auth login` stores a bot token (read from a hidden prompt, `--with-token` stdin, or `--token-file` — never argv) under a profile keyed by `--bot-id`/`--profile`. Per-environment profiles may store `--api-base-url`; explicit environment variables still win. `status`/`list` show metadata only (tokens always masked); `logout` removes a profile.
 
 Bot-type capability and per-command flags are in `docs/octo-cli-design.md`. Agent-facing usage lives under `skills/` (`octo-shared`, `octo-matter` (withheld — see above), `octo-messaging`, `octo-files`, `octo-docs`) — keep those in sync when command shapes change.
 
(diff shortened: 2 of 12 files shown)


internal/marketplace/client.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil, output.ErrAPI("INVALID_RESPONSE", "marketplace download response is missing a URL", "retry or contact the marketplace operator")
+	}
+	if err := validateDownloadURL(ctx, redirect.URL); err != nil {
+		return nil, err
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, redirect.URL, nil)
+	if err != nil {
+		return nil, output.ErrAPI("INVALID_RESPONSE", "marketplace returned an invalid download URL", "contact the marketplace operator")
+	}
+	resp, err := c.http.Do(req)
```
OctoBoooot, line 82: **[minor]** TOCTOU / DNS-rebinding: `validateDownloadURL` resolves the hostname and checks the IPs, but `c.http.Do` here re-resolves the hostname independently — the validated address isn't pinned. A hostile resolver could return a public IP at validation time and a private one at fetch time, bypassing the private-address filter. It's gated by the trusted backend (URL from the authenticated `/download` response, redirects blocked via CheckRedirect), so defense-in-depth rather than a live hole — but to fully close it, pin the validated IP through a custom DialContext (dial the exact IP you validated) or re-check the connection's peer address. Separately, the private-range set omits CGNAT 100.64.0.0/10 (net.IP.IsPrivate returns false for it), so add that CIDR if the host could ever sit behind carrier-grade NAT.

OctoBoooot (commented):
## Review: feat: install Skills from Octo Marketplace (#89)

**Verdict:** Comment — this is the hardened successor to #87 and it directly resolves the SSRF minor I raised there; the security design is now strong across the board, no blocker or major. Holding at COMMENT (not APPROVE) for the same reason as #87: the repo's **code CI has not run** on this PR, so I can't attest the build/test/lint gate on a security-sensitive download-and-install path.

Risk tier: **high** (remote fetch → write to disk). I read the marketplace client, installer, command wiring, and both test files; built and ran the targeted tests locally (all green).

### The #87 SSRF finding is fixed — verified
My #87 COMMENT flagged that the artifact-download redirect URL was followed with no scheme/host allowlist and default redirect-chasing. This PR closes that vector, and tests each part:
- **HTTPS-only + private-address rejection** (`validateDownloadURL`): rejects non-`https`, and rejects when the host (literal IP *or* resolved) is private/loopback/link-local/unspecified. Tested: `TestDownloadSkillRejectsHTTP`, `...RejectsPrivateIPHost`, `...RejectsMetadataIP` (169.254.169.254). ✓
- **Redirect blocking**: the client sets `CheckRedirect → http.ErrUseLastResponse`, so a 3xx is returned as-is (→ `ParseBackendError`) rather than followed to an unchecked target. Tested: `TestDownloadSkillDoesNotFollowRedirect` asserts the redirect target is never requested. This is the clean fix — it removes the "validate initial URL, then bounce to internal" bypass entirely. ✓
- IPv4-mapped IPv6 is handled correctly (Go's `IsPrivate`/`IsLoopback` unwrap `::ffff:…`), so `::ffff:127.0.0.1` → loopback-rejected. ✓ (I checked this empirically rather than assuming.)

Only `check-sprint` (governance, red), `label`, and `add-to-project` ran; `code-review` pending. `build & test (1.24.x)`, `npm test`, and `osv-scan` are **absent** — same first-time-contributor workflow gating as #87/#88. I verified locally: `go build ./...` clean; `go test ./internal/skillinstall/ ./internal/marketplace/ ./internal/config/` all pass. A maintainer should let code CI run before merge; `unrun ≠ green`, and I won't attest the repo gate from a local run on a high-risk path.

### Minor — residual SSRF hardening (defense-in-depth)
- `internal/marketplace/client.go:75-82` — **TOCTOU / DNS-rebinding**: `validateDownloadURL` resolves the hostname and checks the IPs, then `c.http.Do` re-resolves the hostname independently — the validated address isn't pinned. A hostile resolver could return a public IP at validation and a private one at fetch, bypassing the check. Exploitation is gated by the trusted backend (the URL comes from the authenticated `/download` response, and redirects are blocked), so this is defense-in-depth, not a live hole — but if you want to fully close it, pin the validated IP via a custom `DialContext` (validate → dial that exact IP), or re-check the peer address after connect. _(confidence: high it's unpinned; low on exploitability given the trust boundary + redirect block.)_
- Same function — the private-range set omits **CGNAT `100.64.0.0/10`** (`net.IP.IsPrivate` returns false for it, verified). Not reachable in most deployments, but if the marketplace host could ever sit behind carrier-grade NAT it'd pass the filter. One extra CIDR check closes it. Multicast/`0.0.0.0/8` beyond `IsUnspecified` are similarly uncovered but lower-stakes. _(confidence: high.)_

### Nit (≤3)
- `CLAUDE.md:27` — the "hand-written leaves" sentence still doesn't list `marketplace` (carried over from #87). One-word consistency fix.

### Praise
- The security response to #87 is exactly right: rather than a token gesture, it adds scheme validation, resolved-IP filtering, **and** redirect blocking — the three things that together make the "backend returns a hostile URL" path safe — each with a dedicated test. Combined with the retained checksum-before-extract, path-traversal/symlink/zip-bomb defenses, and atomic temp→rename install, this is now a genuinely well-hardened remote-install feature. And the API-contract realignment (plural `/skills/{skill_id}`, `skill_id` response field) is cleanly reflected in both client and tests. _(confidence: high.)_

Net: strong, security-conscious implementation that closed the prior finding. Once code CI runs green, the only remaining items are the two defense-in-depth minors (TOCTOU pin + CGNAT) — address or consciously accept given the trust boundary, and this is APPROVE-ready from my side.

Jerry-Xin (approved):
Summary: The PR is in scope for `octo-cli`, implements Marketplace skill installation coherently, and `go test ./...` passes.

💬 Non-blocking
- 🟡 Warning: `internal/marketplace/client.go:34-37` defaults to `http.DefaultClient`, which follows redirects. The production command passes a no-redirect client in `cmd/marketplace.go:50-55`, but the hardened download behavior is package-level security logic and can be bypassed by any future caller using `NewClient(api, nil)`. Consider making `NewClient` enforce `CheckRedirect: http.ErrUseLastResponse` by default.
- 🟡 Warning: `internal/marketplace/client.go:104-127` validates DNS results before `http.Do`, but the actual dial uses the transport’s resolver again. This leaves a DNS rebinding gap between validation and connection. For stronger private-address rejection, pin validated IPs in a custom `DialContext` or re-check the connected remote address.

✅ Highlights
- ✅ Good archive safety checks in `internal/skillinstall/installer.go`: checksum verification, path traversal rejection, symlink rejection, file count/size limits, and atomic replacement are all appropriate.
- ✅ The command correctly uses the plural Marketplace route shape and `skill_id` field, matching the described API contract.
- ✅ Test coverage is solid across command aliasing, token-kind enforcement, install behavior, download URL validation, redirect blocking, and unsafe archive rejection.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #89 (octo-cli)

## Summary

This PR adds a `marketplace skills <skill-id> --install <root>` command that downloads and installs agent skills from the Octo Marketplace. The implementation spans three layers: a CLI command (`cmd/marketplace.go`), a marketplace HTTP client (`internal/marketplace/client.go`) with SSRF hardening (HTTPS-only, private-IP block, no redirect follow), and a zip-based installer (`internal/skillinstall/installer.go`) with path traversal protection, checksum verification, and atomic swap with rollback. Config gains a `MarketplaceAPIPrefix` field with sensible defaults and normalization. Tests cover the full integration flow, SSRF edge cases, installer safety, and command aliasing.

Overall this is a well-structured, security-conscious feature addition. Two P2 findings below relate to defense-in-depth gaps and a minor error-classification concern — neither is blocking.

## Verification

Static analysis only at head `45296424df3e013002d044ad4694532484c26c70`; build and tests not executed in this environment.

- ✅ **SSRF hardening** — `validateDownloadURL` (`internal/marketplace/client.go:100–128`) enforces HTTPS scheme, resolves DNS and rejects private/loopback/link-local/unspecified IPs before the download request. The download `http.Client` (`cmd/marketplace.go:52–56`) has `CheckRedirect` returning `http.ErrUseLastResponse`, preventing redirect-based SSRF bypass.
- ✅ **Zip safety** — `extract` (`internal/skillinstall/installer.go:97–150`) applies `filepath.Clean` + `filepath.Rel` double-check against traversal, rejects symlinks and non-regular entries, enforces file count (1024), per-file size (16 MiB), and total size (32 MiB) limits. Files created with `O_EXCL` to prevent clobber.
- ✅ **Checksum integrity** — `Install` verifies SHA-256 before any extraction (`internal/skillinstall/installer.go:38–41`). `strings.EqualFold` handles case variations.
- ✅ **Atomic swap** — Existing skill directory is renamed to a backup, new content is renamed into place, and backup is restored on failure (`internal/skillinstall/installer.go:65–87`). Deferred `os.RemoveAll` cleans temp and backup paths.
- ✅ **Auth gate** — `cmd/marketplace.go:41–43` requires `bf_*` (User Bot) token via `credential.TokenKind`. App Bot tokens are rejected with a clear error.
- ✅ **Config normalization** — `marketplaceAPIRoot` (`internal/config/config.go:82–96`) handles unset env (default `/market/api/v1`), empty env (`/api/v1`), and prefix with/without trailing `/api/v1`. `ServiceURL("marketplace")` correctly composes host + prefix, replacing any existing API path.
- ✅ **`overlayProfileBaseURL` refactor** — `internal/cmdutil/factory.go:206–225` restructured the env-var check from early-return to inner-conditional. Functionally equivalent: profile URL only applies when env is not set and profile has a URL.
- ✅ **Test coverage** — Integration test (`cmd/marketplace_test.go:47–99`) exercises the full Get→Download→Install flow with auth header verification. SSRF tests cover HTTP rejection, private IP, link-local metadata IP, and redirect non-following. Installer tests cover success, path traversal, symlink, missing SKILL.md, and checksum mismatch.

## Findings

### P2 — SSRF DNS rebinding TOCTOU in `validateDownloadURL` (`internal/marketplace/client.go:100–128`)

`validateDownloadURL` resolves DNS via `net.DefaultResolver.LookupIPAddr` and checks all resolved IPs against private/loopback/link-local ranges. However, the subsequent `c.http.Do(req)` at `:85` triggers an independent DNS resolution through the default resolver. Between the validation lookup and the actual connection, a DNS rebinding attack could cause the first resolution to return a public IP (passing validation) while the second returns a private IP (connecting to an internal service).

This is a known SSRF bypass pattern. The fix is to use a custom `DialContext` on the download `http.Client` that either pins the already-validated IP or re-checks resolved IPs against the same private-range predicate:

```go
dialer := &net.Dialer{Timeout: 10 * time.Second}
transport := &http.Transport{
    DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
        host, port, _ := net.SplitHostPort(addr)
        ips, _ := net.DefaultResolver.LookupIPAddr(ctx, host)
        for _, ip := range ips {
            if ip.IP.IsPrivate() || ip.IP.IsLoopback() || ip.IP.IsLinkLocalUnicast() || ip.IP.IsUnspecified() {
                return nil, fmt.Errorf("download URL resolves to blocked address %s", ip.IP)
            }
        }
        return dialer.DialContext(ctx, network, net.JoinHostPort(ips[0].IP.String(), port))
    },
}
```

Diff-scope: new (introduced by commit `6976172` SSRF hardening). Real-world exploitability is low — the marketplace API is authenticated, the CLI is short-lived, and the attack requires controlling DNS for the download hostname — but the gap is real in the hardening this PR explicitly claims.

### P2 — `DownloadSkill` 2xx response produces misleading error due to `BinaryResponse` interaction (`internal/marketplace/client.go:66–75`)

`DownloadSkill` sets `BinaryResponse: true` on the API request at `:67`. The internal client (`internal/client/client.go:237–249`) handles 2xx + `BinaryResponse` by returning a metadata envelope `{"status": 200, "content_type": "...", "size": N}` — not the raw response body. If the marketplace download endpoint returns 200 with `{"url": "..."}` instead of the expected 302 redirect, the marketplace client's `json.Unmarshal(body, &redirect)` at `:75` will parse the metadata envelope (which has no `url` field), yielding an empty URL and the error "marketplace download response is missing a URL".

The code works correctly for the intended flow (302 → internal client returns `{"url": "..."}` envelope from the `Location` header). However, if the API contract ever shifts to return a 200 with a URL body, or if an operator misconfigures the gateway to absorb redirects, the resulting error message is misleading. Consider either: (a) detecting the metadata envelope shape and producing a more specific error ("expected redirect but received 200"), or (b) not using `BinaryResponse` for this call and instead parsing the raw 302 response manually.

Diff-scope: new (introduced by this PR). Low immediate risk — the marketplace API contract is 302 redirect — but worth hardening the error path.

- `namePattern` (`internal/skillinstall/installer.go:23`) regex `^[a-z0-9][a-z0-9._-]{0,127}$` prevents path components starting with `.` or `-`, blocks `/` and `\`, and caps at 128 chars.
- `emitMarketplaceError` (`cmd/marketplace.go:86–89`) calls `f.EmitError` (sets `ErrorEmitted=true`, writes error envelope) then returns the original `*ExitError` for cobra exit-code handling — consistent with the codebase pattern.
- `io.CopyN(f, rc, size)` at `internal/skillinstall/installer.go:143` reads exactly the declared `UncompressedSize64` bytes; `size < 0` guard at `:133` catches uint64→int64 overflow.
- `normalizePathPrefix` (`internal/config/config.go:76–82`) correctly trims whitespace and slashes, returns empty for blank input, and prepends `/`.
- Integration test verifies auth header propagation (`gotAuth != "Bearer bf_test"` at `cmd/marketplace_test.go:87`), confirming the internal client sends credentials on marketplace API calls.
- `cmd/root.go:55` registers `newMarketplaceCmd` alongside existing commands — no ordering conflicts.

## Verdict: APPROVED

No P0/P1 blockers. The two P2 findings are defense-in-depth improvements and error-path hardening — neither affects correctness of the intended flow. The SSRF DNS rebinding TOCTOU (P2) is a known gap in the hardening that should be addressed in a follow-up with a custom `DialContext`, but the realistic exploitability for a CLI tool downloading from an authenticated marketplace is low. The PR is well-structured, security-conscious, and thoroughly tested.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Two P2 defense-in-depth items (DNS rebinding TOCTOU, BinaryResponse error classification) are non-blocking. Feature is well-implemented with comprehensive SSRF hardening, zip safety, and test coverage.

kense-lab:
Closing this version while the Marketplace client is refactored to consume the OpenAPI registry instead of duplicating endpoint contracts in handwritten code. A replacement PR will follow after the refactor and verification.

yujiawei (changes requested):
# Code Review — PR #89 (octo-cli)

Reviewed at head SHA `45296424df3e013002d044ad4694532484c26c70` against merge-base `0eb40d16` with `main`. `go build ./...`, `go vet`, and the new test suites all pass locally.

Overall this is a well-structured feature with a genuinely security-conscious download/install path (HTTPS-only, redirect blocking, size caps, zip-slip/symlink rejection, SHA256 integrity gate, atomic install). The two blockers below are both in the security-critical code the PR set out to harden, and both have small, well-understood fixes.

## 1. Spec compliance

**Spec: ✅**

The diff implements what the PR summary describes and nothing out of scope:
- `octo-cli marketplace skills <id> --install <dir>` with `market` alias — `cmd/marketplace.go`, `cmd/root.go:55`.
- Plural `/skills/{skill_id}` routes and `skill_id`/`file_sha256` response fields — `internal/marketplace/client.go:35-77`.
- Download hardening (HTTPS-only, private-address rejection, 64 MiB size cap, redirect blocking) — present in `internal/marketplace/client.go`.
- `OCTO_MARKETPLACE_API_PREFIX` config + `ServiceURL("marketplace")` routing — `internal/config/config.go`.

No missing scope, no unrequested features, no deviation from the stated approach. The *effectiveness* of the private-address-rejection hardening is a code-quality concern, addressed below.

## 2. Code quality

### P1 — SSRF `validateDownloadURL` is a validate-then-reconnect TOCTOU (DNS-rebinding bypass)

`internal/marketplace/client.go:104-128` resolves the host once via `net.DefaultResolver.LookupIPAddr` (line 114) and rejects private/loopback/link-local/unspecified IPs (line 123), then returns. The actual fetch at `client.go:82` (`c.http.Do(req)`, request built line 78) uses the plain `*http.Client` from `cmd/marketplace.go:50-55`, which sets only `Timeout` and `CheckRedirect` — no custom `Transport.DialContext`. The stdlib transport therefore performs its **own independent DNS resolution at dial time**.

```go
// client.go:114 — validation lookup
resolved, err := net.DefaultResolver.LookupIPAddr(ctx, parsed.Hostname())
...
// client.go:82 — the dial re-resolves the hostname independently
resp, err := c.http.Do(req)
```

The IP that is validated and the IP that is dialed come from two separate lookups. A hostname whose DNS returns a public IP on the first lookup and `127.0.0.1` / `169.254.169.254` / an internal IP on the second bypasses the check entirely. Since `validateDownloadURL` exists solely to prevent SSRF, this is a security control that is silently ineffective.

The trust model does bound the impact: the URL comes from the authenticated marketplace `/download` 302 `Location`, the flow requires a `bf_*` User-Bot token, the response body is discarded (blind), and content is SHA256-gated before install — so exfiltration and content injection are not viable. The residual, real risk is a **blind internal-GET primitive** from the CLI's network position (localhost admin ports, cloud metadata, private IPs). That is above P2 because a purpose-built defense provably does not hold; it is below "immediately exploitable" because of the gating.

**Fix:** resolve once and dial that exact IP — a custom `Transport.DialContext` (or `net.Dialer.Control`) that re-checks the concrete IP it is about to connect to, so validation happens at connect time on the address actually used.

### P1 — Double-failure during atomic swap can permanently delete the previously-installed skill (silent data loss)

`internal/skillinstall/installer.go`:

```go
// line 68 — registered unconditionally, before activation
defer func() { _ = os.RemoveAll(backup) }()
...
// line 71 — on upgrade, the ONLY copy of the old skill is moved to backup
if err := os.Rename(target, backup); err != nil { ... }
...
// line 78-82 — activate; on failure, attempt restore but discard its error
if err := os.Rename(tmp, target); err != nil {
    if hadTarget {
        _ = os.Rename(backup, target) // line 80 — error ignored
    }
    return Result{}, fmt.Errorf("activate installed skill: %w", err)
}
```

If activation (`tmp→target`, line 78) fails **and** the restore (`backup→target`, line 80) also fails, `backup` still holds the only copy of the user's prior skill — and the deferred `os.RemoveAll(backup)` (line 68) then unconditionally deletes it. The user's previously-working skill is gone from both `target` and `backup`, with no surviving copy and no reported path.

A correlated double-failure is uncommon (all three paths are same-filesystem) but plausible: `target` recreated between line 71 and 78 (ENOTEMPTY/EEXIST on both renames), or the filesystem going read-only / ENOSPC mid-operation. This exactly defeats the crash-safe-upgrade property the backup dance was written to provide, and it is unrecoverable and silent.

**Fix:** check the restore rename's error; on restore failure, skip the `RemoveAll(backup)` and surface the backup path so the user can recover manually.

### P2 — hardening and robustness gaps (non-blocking)

- **SSRF blocklist coverage** (`client.go:123`): `IsPrivate()` does not cover RFC6598 CGNAT `100.64.0.0/10`, and the guard omits link-local multicast and the non-zero part of `0.0.0.0/8`. (Cloud metadata `169.254.169.254`, IPv6 ULA, and IPv4-mapped loopback *are* covered.) Add explicit CIDR checks for `100.64.0.0/10` (and optionally `198.18.0.0/15`) once the TOCTOU above is fixed — otherwise this range remains reachable.
- **Declared-size trust** (`installer.go:120,136`): `io.CopyN(f, rc, entry.UncompressedSize64)` trusts the central-directory size. If an entry's actual content exceeds the declared size, `io.CopyN` copies exactly `size` bytes with a nil error and `rc.Close()` does not fail (CRC is only checked at EOF), so the file is silently truncated and `Install` reports success. Consider copying via `io.LimitReader` and rejecting if bytes remain. (Bounded by the whole-archive SHA gate, so not a new capability — a robustness issue against a malformed backend archive.)
- **Directory entries bypass caps** (`installer.go:111-115`): directory entries `continue` before the `maxFiles`/size accounting, so the entry-count and total-size limits do not bound directory creation. A crafted archive under the 64 MiB wire cap can still create a very large number of directories. Count all entries (or add a directory cap).
- **`ServiceURL("marketplace")` path handling** (`config.go:74`): `u.Path = c.MarketplaceAPIPrefix` replaces any path already present in `APIBaseURL`, unlike every other service which composes base+path. If `APIBaseURL` carries a gateway path prefix, marketplace silently rebuilds from the host root. Intentional and tested, but a latent misconfiguration trap worth a doc note.
- **Checksum-mismatch error taxonomy** (`cmd/marketplace.go:64-66`): the security-relevant `skill archive checksum mismatch` is wrapped into a generic `ErrValidation`, same code as "missing SKILL.md". The message text is preserved (`%v`), so it is only the machine-readable code that is coarse. Minor.
- **`overlayProfileBaseURL` refactor** (`internal/cmdutil/factory.go:206-224`): the result is identical to before, but the early-return on `OCTO_API_BASE_URL` being set was removed, so the profile store is now loaded from disk on every command even when the env var wins. Harmless but a small unnecessary I/O per invocation.

### Notes / verified-not-issues

- The `binaryResp` redirect-envelope contract (`DownloadSkill` expects a 302 `Location`, surfaced as `{"url":...}` by `internal/client/client.go:253-264`) is coupled to the documented "302 → presigned URL" contract; a direct 200-body would fail with a clean `INVALID_RESPONSE`. This matches the stated design — a forward-compat note, not a defect.
- Redirect blocking is correct: both the shared client and the marketplace client set `CheckRedirect` to `http.ErrUseLastResponse`, and `TestDownloadSkillDoesNotFollowRedirect` covers it.
- Zip-slip, symlink rejection, and path-escape handling (`installer.go:98-108`) are correct and tested.

## 3. Overall verdict

**REQUEST_CHANGES** — Spec ✅ but two P1 defects in the security-critical download/install path (SSRF TOCTOU bypass; silent data loss on correlated atomic-swap failure). Both fixes are small and localized. The P2 items are good follow-ups but do not individually block.

## 4. Suggested direction

1. Pin the validated IP to the actual dial (custom `DialContext` / `Dialer.Control`) so SSRF validation runs on the connected address.
2. Handle the restore-rename error and preserve/report `backup` on double failure.
3. Extend the IP blocklist (CGNAT + multicast) and count directory entries toward the archive limits.

## 5. Additional observations

- Test coverage for the new packages is strong (SSRF rejections, redirect non-follow, zip-slip, symlink, checksum mismatch, alias resolution). Adding a DNS-rebinding regression test and a double-failure restore test would lock in the two fixes above.

PR #90, opened 2026-07-16
feat(skills): install Marketplace Skills from OpenAPI contract

kense-lab:
## Summary

- add `octo-cli skills install <skill_id> --from marketplace --dir <skills_root>`
- embed only the `skill.get` and `skill.download` Marketplace OpenAPI operations
- resolve Marketplace request methods and paths by operationId instead of duplicating endpoint routes in command code
- securely download, SHA256-verify, extract, and atomically install Skill archives
- support common ZIP packaging directories and existing mixed-case Skill names
- keep local HTTP downloads disabled by default, with an explicit loopback-only development switch

## Verification

- `go test ./...`
- real local integration against Marketplace on `127.0.0.1:8092`
- installed Skill `f015b994-eeea-401e-8328-ce68340bc59b` with `SKILL.md`, references, and scripts

## Marketplace dependency

`GET /skills/{skill_id}/download?format=json` returns both `download_url` and `file_sha256` so the CLI can verify archive integrity before installation.

## Supersedes

Supersedes closed PR #89 after replacing the handwritten top-level Marketplace command with the existing `skills` command and OpenAPI registry.


internal/marketplace/installer.go
```diff
@@ -0,0 +1,182 @@
...
+		return Result{}, fmt.Errorf("inspect existing skill: %w", statErr)
+	}
+	if err := os.Rename(tmp, target); err != nil {
+		if hadTarget {
+			_ = os.Rename(backup, target)
+		}
+		return Result{}, fmt.Errorf("activate installed skill: %w", err)
+	}
+	if hadTarget {
+		_ = os.RemoveAll(backup)
```
OctoBoooot, line 90: **[minor]** Rollback double-fault data loss (carried from #89): when `os.Rename(tmp, target)` fails AND the restore `os.Rename(backup, target)` on line 80 also fails — both swallowed with `_ =` — this deferred `os.RemoveAll(backup)` then deletes the backed-up old skill. Narrow (two consecutive same-dir/same-fs renames failing is near-catastrophic-FS territory) and the artifact is a re-installable marketplace skill, so minor — but cheap to close: check the restore rename's error, and when it failed, skip the RemoveAll and surface the backup path so the user can recover. No test covers the rollback path; worth adding one.

OctoBoooot (commented):
## Review: feat(skills): install Marketplace Skills from OpenAPI contract (#90)

**Verdict:** Comment — the most mature iteration of this feature (#87 → #89 → #90): the SSRF hardening I asked for is intact, the architecture is now cleaner (proper embedded OpenAPI contract, hidden from the command tree), and no blocker or major. Holding at COMMENT (not APPROVE) for the same reason as #87/#89: the repo's **code CI has not run**, so I can't attest the build/test/lint gate on a high-risk download-and-install path.

Risk tier: **high** (remote fetch → write to disk). I read the client, installer, spec, command wiring, and tests; built and ran the marketplace/config/registry tests locally (green).

### What improved since #89 (delta reviewed)
- **The feature is now contract-driven.** `internal/registry/specs/marketplace.json` is a real embedded OpenAPI spec marked **`x-octo-disabled: true`** — loaded so `skill.get`/`skill.download` are available to the installer *internally*, but the raw `marketplace.*` API commands are **hidden** from the CLI tree (the same pattern as withheld `matter`). `TestMarketplaceOpenAPIContract` pins exactly this ("commands must remain hidden; skills install uses its operations internally") plus the two operations' method/path. Clean design — users go through `skills install --from marketplace`, not raw endpoints. ✓
- **SSRF hardening survived the refactor and got a safe escape hatch.** `validateDownloadURL` still enforces https-only + rejects private/loopback/link-local/unspecified, and the command still sets `CheckRedirect → ErrUseLastResponse`. The new `AllowInsecureLocalhost` is **not a footgun**: env-gated (`OCTO_MARKETPLACE_ALLOW_INSECURE_LOCALHOST`), default false, and even when true only permits `http` for **loopback** hosts (not arbitrary private IPs) — a deliberate local-dev/test escape. ✓
- Token gate (`bf_*` User Bot required, `ErrAuth` otherwise), checksum-before-extract, and atomic temp→backup→rename install all retained. ✓

Only `check-sprint` (governance, red), `label`, `add-to-project` ran; `code-review` pending. `build & test (1.24.x)`, `npm test`, `osv-scan` are **absent** — same first-time-contributor workflow gating as #87/#88/#89. Verified locally: `go build ./...` clean; `go test ./internal/marketplace/ ./internal/config/ ./internal/registry/` pass. (The `cmd` package `TestAuth_*` failures are pre-existing/environmental — they fail on `main` too.) A maintainer should let code CI run before merge; `unrun ≠ green` on a high-risk path.

### Minor — carried over from #89 (still worth fixing)
- `internal/marketplace/installer.go:78-85` — **rollback double-fault data loss.** On the path where `os.Rename(tmp, target)` fails *and* the restore `os.Rename(backup, target)` also fails (both errors swallowed with `_ =`), the deferred `os.RemoveAll(backup)` then deletes the backed-up old skill. Narrow (two same-dir/same-fs renames failing consecutively is near-catastrophic-FS territory) and the artifact is a re-installable marketplace skill, so it's minor — but the fix is cheap: check the restore error and skip the `RemoveAll` (and report the backup path) when restore failed. No test exercises the rollback path — a fair coverage gap to close alongside. _(confidence: high.)_
- `internal/marketplace/client.go` (`validateDownloadURL`) — **DNS-rebinding TOCTOU**: the host is resolved for validation, then `http.Do` re-resolves independently (no IP pinning). Gated by the authenticated backend + blocked redirects, so defense-in-depth, not a live hole; pin the validated IP via a custom `DialContext` to fully close it. (Three reviewers converged on this across #89.) _(confidence: high it's unpinned; low on exploitability.)_

### Nit (≤3)
- `internal/registry/specs/marketplace.json` is **1331 lines for 2 used operations** — it's a raw swagger dump carrying **54 unused component schemas** (machine-mangled names like `github_com_..._response.CursorList-...`) embedded into the binary. It's valid and disabled-from-the-tree so it's harmless, but trimming to the `skill.get`/`skill.download` closure (and their referenced schemas) would shrink the binary and make the contract legible. _(confidence: high.)_

### Praise
- This is exactly how to land a feature that pairs with a backend contract: rather than hand-encoding the endpoints, it embeds the marketplace's OpenAPI spec and drives the installer off it, while using `x-octo-disabled` to keep the raw API surface out of the user-facing tree — contract fidelity without command-surface bloat, pinned by a test. Combined with the now-complete security posture (SSRF + checksum + atomic install + token gate), the feature has matured cleanly across three iterations and each of my prior findings was addressed or consciously scoped. _(confidence: high.)_

Net: mature, security-conscious, contract-driven. Once code CI runs green, the remaining items are the two carried minors (rollback cleanup + TOCTOU pin) and the spec trim — address or consciously accept, and this is APPROVE-ready from my side.

Jerry-Xin (dismissed):
Project relevance gate passed: this PR extends `octo-cli`’s existing `skills` command and internal registry/client/config paths.

One-line summary: The Marketplace skill install flow is relevant, well-scoped, and covered by useful security tests; I found no merge-blocking issues.

💬 Non-blocking

🟡 Warning: `marketplace.NewClient(nil httpClient)` defaults to `http.DefaultClient`, which follows redirects. The production command passes a no-redirect client, but `DownloadSkill`’s security invariant depends on caller discipline. Consider making `NewClient` install `CheckRedirect: http.ErrUseLastResponse` by default when `httpClient == nil`, or validating redirect `Location` targets before following. See [internal/marketplace/client.go](/Users/mlamp/.openclaw/tmp/pr-review-wbkafcwf/repo/internal/marketplace/client.go:46) and [internal/marketplace/client.go](/Users/mlamp/.openclaw/tmp/pr-review-wbkafcwf/repo/internal/marketplace/client.go:118).

🟡 Warning: download host validation resolves DNS before the HTTP request, but the actual transport resolves again. This leaves a theoretical DNS rebinding gap for future or non-standard transports. A stronger approach would pin the validated IP in a custom `DialContext` or validate inside the dialer. See [internal/marketplace/client.go](/Users/mlamp/.openclaw/tmp/pr-review-wbkafcwf/repo/internal/marketplace/client.go:158).

🔵 Suggestion: archive extraction writes all files as `0644`, so executable bits from packaged scripts are lost. If Marketplace skills may ship shell/python helper scripts intended to be run directly, preserve owner executable bits from regular file entries within a constrained permission mask. See [internal/marketplace/installer.go](/Users/mlamp/.openclaw/tmp/pr-review-wbkafcwf/repo/internal/marketplace/installer.go:136).

🔵 Suggestion: add a rollback-path test for install activation failure or backup restore failure. The replacement logic is careful, but that is the riskiest filesystem path and currently only the happy replacement case is covered. See [internal/marketplace/installer.go](/Users/mlamp/.openclaw/tmp/pr-review-wbkafcwf/repo/internal/marketplace/installer.go:70).

✅ Highlights

The command is integrated under `skills install` rather than adding an unrelated top-level command, matching the PR scope and existing CLI shape.

The archive handling includes checksum verification, ZIP traversal checks, symlink rejection, file count/size limits, and atomic replacement behavior.

The tests cover the important Marketplace boundaries: auth token kind, required install directory, hidden top-level Marketplace command, HTTPS/private-address checks, redirect behavior, checksum mismatch, and unsafe archive entries.

Verification run: `go test ./...` passed.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #90 (octo-cli)

## Summary

This PR adds a `skills install <skill-id> --from marketplace --dir <root>` subcommand that fetches skill metadata and a zip archive from the Marketplace API (via an embedded OpenAPI contract in `internal/registry/specs/marketplace.json`), then extracts it into the local agent skills directory. The implementation includes SSRF hardening (HTTPS-only with explicit localhost escape, DNS resolution check against private/loopback/link-local, redirect blocking), zip-slip protection (path traversal, symlink, and size-limit checks), atomic install with backup-and-restore rollback, and `bf_*` User Bot token gating. The `marketplace` service is loaded as `x-octo-disabled: true`, so its operations are only reachable through the internal `marketplace.Client`, not through auto-generated CLI commands. The config layer gains a `MarketplaceAPIPrefix` field with sensible defaults and a `ServiceURL("marketplace")` override. A clean refactor of `overlayProfileBaseURL` preserves existing behavior.

## Verification

Local checks at head `230291e1`:
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./internal/marketplace/... ./internal/config/... ./internal/registry/...` — pass (all packages)
- `go test ./cmd/...` — pass (all sub-packages)

## Findings

No P0/P1 issues. Two P2 items below.

### P2 — `--from` flag error message is misleading when omitted (`cmd/skills.go:176`)

The `--from` flag defaults to `""` and the RunE rejects any value other than `"marketplace"`. When the user omits `--from` entirely, the error reads `"--from must be marketplace"`, which implies a wrong value was passed rather than that the flag is required. Consider using `cobra.MarkFlagRequired("from")` or emitting a distinct `"--from is required"` message when `source == ""`.

- **diff-scope**: new (this PR introduces the `install` subcommand).
- **severity rationale**: UX friction, not a broken production path. Users can still pass `--from marketplace` and succeed. P2 per R1/R4.

### P2 — Download timeout shares 30 s budget with metadata calls (`cmd/skills.go:190`)

The `http.Client{Timeout: 30 * time.Second}` created at `:190` is used for both the Marketplace API metadata calls (small JSON, fine at 30 s) and the archive download via `c.http.Do(req)` in `DownloadSkill`. A 64 MiB zip on a slow link can exceed 30 s, causing a timeout error. Consider giving the download client a separate, larger timeout (e.g. 5 min) while keeping 30 s for metadata.

- **diff-scope**: new (this PR introduces the download HTTP client).
- **severity rationale**: Operational limitation on slow links; not a correctness or security defect. P2 per R1/R4.

## Data-Flow Tracing

Each consumed data value traced to its upstream source:

1. **`skill.ID` / `skill.Name`** — consumed at `cmd/skills.go:199` in `marketplace.Install(installRoot, skill.Name, ...)`. Source: `Client.GetSkill` → `apiClient.Do` → JSON unmarshal of `envelope.Data["skill_id"]` / `envelope.Data["name"]`. Validated non-empty at `client.go:79`. Then regex-validated by `namePattern` at `installer.go:38`. Data flows correctly. ✅
2. **`archive.Body` / `archive.SHA256`** — consumed at `cmd/skills.go:199` passed to `marketplace.Install`. Source: `Client.DownloadSkill` → `apiClient.Do` for metadata → direct `c.http.Do(req)` for binary. SHA256 verified at `installer.go:39` before extraction. Archive bytes capped at 64 MiB + 1 at `client.go:139`. ✅
3. **`cfg.MarketplaceAPIPrefix`** — consumed by `cfg.ServiceURL("marketplace")` at `config.go:90`. Source: `Load()` → `marketplaceAPIRoot()`. Defaults: unset → `/market/api/v1`; empty → `/api/v1`. Verified by `TestServiceURL_Unified` and `TestLoad_Defaults`. ✅
4. **`cfg.MarketplaceAllowInsecureLocalhost`** — consumed at `cmd/skills.go:192` → `marketplace.Options{AllowInsecureLocalhost: ...}`. Source: `Load()` → `envBool(EnvMarketplaceAllowInsecureLocalhost)`. Default `false`. Correctly propagated. ✅
5. **`cred.Token`** — consumed at `cmd/skills.go:183` for `TokenKind` check and at `client.go:209` (via `apiClient.attempt`) for `Authorization: Bearer` header. Source: `f.Credential()`. Test `TestMarketplaceSkillInstall` asserts `gotAuth == "Bearer bf_test"`. ✅

## Blind Spot Checklist (R5)

- **C1 (dual-path parity)**: N/A — no paired add/remove or subscribe/unsubscribe paths in this PR.
- **C2 (control-flow ordering)**: N/A — no reused logic called in different orderings. The `validateDownloadURL` and `extract` functions are each called once in their respective flows.
- **C3 (authorization boundary)**: ✅ Checked — `skills install` requires `bf_*` User Bot token at `cmd/skills.go:183`. No other entry point to marketplace operations exists (service is `x-octo-disabled: true`). `TestMarketplaceHasNoTopLevelCommand` guards against accidental exposure.
- **C4 (authorization lifecycle)**: N/A — no container-member state cascading in this PR.
- **C5 (build ≠ runtime)**: N/A — no build/packaging changes. `go build` and `go test` both pass cleanly.
- **C6 (governance/policy)**: N/A — no governance or policy documents changed.

## Cross-Round Blocker Review (R6)

- SSRF hardening in `validateDownloadURL` (`internal/marketplace/client.go:160`): HTTPS-only scheme check, DNS resolution with all-IP private/loopback/link-local/unspecified blocking, `AllowInsecureLocalhost` opt-in for localhost dev.
- Zip-slip prevention in `extract` (`internal/marketplace/installer.go:100`): `filepath.Clean` + `filepath.Rel` double-check, symlink rejection, `maxFiles`/`maxTotalSize`/`maxFileSize` caps, `os.O_EXCL` on file creation.
- Archive size cap: `io.ReadAll(io.LimitReader(resp.Body, MaxArchiveBytes+1))` followed by `int64(len(body)) > MaxArchiveBytes` correctly detects archives exceeding 64 MiB (`internal/marketplace/client.go:139`).
- Checksum verification: `sha256.Sum256` with `strings.EqualFold` handles case differences in hex digests (`internal/marketplace/installer.go:39`).
- Atomic install: backup → rename-in → cleanup with rollback on failure (`internal/marketplace/installer.go:62`).
- Token gating: `credential.TokenKind(cred.Token) != "user_bot"` correctly restricts to `bf_*` tokens (`cmd/skills.go:183`).
- `x-octo-disabled: true` on marketplace spec ensures no auto-generated CLI commands expose the service; test `TestMarketplaceHasNoTopLevelCommand` guards regression.
- `overlayProfileBaseURL` refactor (`internal/cmdutil/factory.go:206`) is behavior-preserving: env-wins-over-profile logic is identical before and after.
- `marketplaceAPIRoot()` defaults: unset → `/market/api/v1`; empty → `/api/v1`; explicit prefix gets `/api/v1` suffix. Consistent with `CLAUDE.md` docs.
- Redirect blocking: `http.ErrUseLastResponse` in both production (`cmd/skills.go:191`) and test, verified by `TestDownloadSkillDoesNotFollowRedirect`.

## Verdict: APPROVED

No correctness, security, or build-breaking issues. The SSRF hardening, zip extraction safety, and auth gating are well-implemented. The two P2 items are non-blocking usability/operational improvements.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; two P2 nits (flag UX, download timeout budget). Safe to land after optional P2 address.

yujiawei (changes requested):
# Code Review — PR #90 (octo-cli)

_Reviewed at head SHA `230291e1eb4b1bd9212148384996b706a108d580`. Build (`go build ./...`) and the affected test packages (`internal/marketplace`, `internal/config`, `internal/registry`, `cmd/...`) are green locally._

## Summary

This PR adds `octo-cli skills install <skill-id> --from marketplace --dir <root>`, driving the install off the embedded Marketplace OpenAPI contract (`skill.get` + `skill.download`), then securely downloading, SHA256-verifying, and atomically installing the Skill archive. The overall design is solid: SSRF hardening on the download URL, checksum-gated install, a bounded archive extractor with path-traversal/symlink defenses, and a bf_* User-Bot token gate. Scope is disciplined — the marketplace service is kept hidden (`x-octo-disabled`) so it never surfaces as a public command.

There is **one blocking correctness defect** in the atomic-install rollback: a double-failure path can permanently destroy the user's pre-existing skill. Everything else is verified-correct or minor hardening suggestions.

## 1. Spec compliance

**Spec: ✅**

All summary claims verified against the diff:

- ✅ `skills install <id> --from marketplace --dir <root>` command exists (`cmd/skills.go:158-222`), with `--from` restricted to `marketplace` and `--dir` required.
- ✅ Spec embeds only `skill.get` and `skill.download` (`internal/registry/specs/marketplace.json`); registry test asserts exactly 2 marketplace operations (`internal/registry/loader_test.go`).
- ✅ Method/path resolved by `operationId`, not hardcoded routes (`internal/marketplace/client.go:57-96`, `operation()`/`operationPath()`).
- ✅ SHA256-verify → extract → atomic install (`internal/marketplace/installer.go:30-88`).
- ✅ Packaging-directory strip + mixed-case names (`archiveRootPrefix`, `namePattern`).
- ✅ Local HTTP disabled by default; explicit loopback-only dev switch via `OCTO_MARKETPLACE_ALLOW_INSECURE_LOCALHOST` (`internal/config/config.go:36-52`, `client.go:158-196`).
- ✅ Marketplace service stays hidden (`x-octo-disabled: true`); no top-level `marketplace`/`market` command (`cmd/skills_marketplace_test.go:120-128`).
- ✅ bf_* User-Bot token gate enforced (`cmd/skills.go:176-178`).

The added config surface (`MarketplaceAPIPrefix`, `MarketplaceAllowInsecureLocalhost`, `config show` exposure) and the `overlayProfileBaseURL` cleanup are in direct support of these claims and behavior-preserving — not out-of-scope over-building.

- Missing: none.
- Over-build: none.
- Deviation: none.

## 2. Code quality

### P1 — Double-failure in atomic install permanently deletes the user's existing skill

`internal/marketplace/installer.go:68, 78-83`

```go
68  defer func() { _ = os.RemoveAll(backup) }()   // registered unconditionally
...
78  if err := os.Rename(tmp, target); err != nil {
79      if hadTarget {
80          _ = os.Rename(backup, target)         // restore error swallowed
81      }
82      return Result{}, fmt.Errorf("activate installed skill: %w", err)
83  }
```

Trace of the failure path:

1. When a skill already exists, `os.Rename(target, backup)` (line 71) moves the user's **only** copy to `backup`; `target` no longer exists.
2. `os.Rename(tmp, target)` (line 78) fails.
3. The restore `os.Rename(backup, target)` (line 80) *also* fails and its error is discarded — the user's data is still at `backup`, `target` is empty.
4. Function returns at line 82.
5. The unconditional `defer` at line 68 fires and `os.RemoveAll(backup)` deletes the last surviving copy.

Net result: both `target` and `backup` are gone — the pre-existing skill is permanently lost, on an operation the caller treats as atomic/rollback-safe. The second rename is a same-directory operation so `EXDEV` is not the concern, but it can genuinely fail (a concurrent process re-creating a non-empty `target` → `ENOTEMPTY`, `EACCES`/`EBUSY` on the directory, or `ENOSPC` on the metadata op). The trigger is a filesystem transient during a legitimate install, not an attacker path — local data integrity is in scope.

**Fix (small):** capture the restore result and gate the backup cleanup on it, or move backup deletion out of the `defer` into the two success branches (line 84-86 already does the happy-path delete). On restore failure, leave `backup` on disk and surface its path in the returned error so the user can recover, e.g.:

```go
if err := os.Rename(tmp, target); err != nil {
    if hadTarget {
        if rerr := os.Rename(backup, target); rerr != nil {
            return Result{}, fmt.Errorf("activate installed skill failed and rollback failed; "+
                "previous skill preserved at %s: %w", backup, err)
        }
    }
    return Result{}, fmt.Errorf("activate installed skill: %w", err)
}
```
and drop the unconditional `defer os.RemoveAll(backup)` in favor of explicit cleanup on the paths where `target` is known-good.

### P2 — Non-blocking hardening suggestions

- **DNS-rebinding / TOCTOU on the download host** (`client.go:158-188` validate vs. `:114-118` fetch). `validateDownloadURL` resolves and checks the host's IPs, but `c.http.Do` re-resolves independently — the client sets no pinned `DialContext`/`Control`. For a hostname-based `download_url`, the validation and fetch lookups can differ (rebinding / short-TTL race), bypassing the private-IP guard. Impact is bounded to SSRF-as-probe: the `download_url` comes from the authenticated first-party marketplace and the archive is SHA256-gated against the API-supplied hash, so no malicious archive can be installed via this path. Consider resolving once and pinning the validated IP through `net.Dialer.Control`.
- **SSRF blocklist gaps** (`client.go:179-186`). `IsPrivate/IsLoopback/IsLinkLocalUnicast/IsUnspecified` miss CGNAT `100.64.0.0/10` and IPv6 link-local multicast/ULA edge cases. Minor, given the mitigations above.
- **Directory entries bypass extractor resource limits** (`installer.go:116-121`). Directory entries hit `MkdirAll` + `continue` before the `maxFiles`/size accounting, so a pathological archive with a very large number of directory entries is only bounded by the 64 MiB download cap, not `maxFiles`. Low risk (first-party source, size cap) but worth counting directories toward `maxFiles`.
- **Restore-rename failure is silently swallowed** (`installer.go:80`) — same area as the P1; even after the P1 fix, prefer surfacing the rollback failure rather than `_ =`.

### Test coverage gaps (P2)

The security controls are well tested (HTTP-scheme rejection, private/metadata-IP rejection, redirect non-follow, checksum mismatch, path traversal, symlink, missing SKILL.md, packaging-dir strip, bf_* gate, `--dir` required, marketplace-not-top-level). Gaps worth closing:

- No test exercises the install **rollback / double-failure** path — precisely the atomicity guarantee the P1 breaks.
- The redirect-non-follow test (`client_test.go:87-121`) builds its own `http.Client` rather than exercising the production wiring in `cmd/skills.go:187-192`, so the shipped guard is not covered end-to-end.
- `TestMarketplaceSkillInstall` mutates global `http.DefaultTransport` — latent hazard if any `cmd` test ever runs with `t.Parallel()`.
- No coverage for the archive resource-limit guards (`maxFiles`/`maxFileSize`/`maxTotalSize`) or the `MaxArchiveBytes` +1 boundary; `archiveRootPrefix` negative branches (mixed roots, single root without SKILL.md) are untested.

## 3. Overall verdict

**REQUEST_CHANGES**

Spec compliance is clean (✅). The blocker is the single P1 data-loss defect in the atomic-install rollback — it silently destroys a user's existing skill on a double rename failure, on an operation advertised as atomic, and the fix is small. The P2 items are good follow-ups but are not merge blockers on their own.

## 4. Suggested direction

1. Make the backup cleanup conditional (remove the unconditional `defer`); on restore failure, preserve `backup` and return its path in the error.
2. Add a rollback test that forces `tmp→target` to fail and asserts the prior skill's contents survive.
3. (Optional hardening) pin the validated download IP via `DialContext`; count directory entries toward `maxFiles`.

## 5. Additional notes

The PR branch appears to have been cut from an older base, so a raw `git diff origin/main..HEAD` spuriously shows unrelated `docs`-side content (docs.share, sheet hyperlinks, etc.) as "removed." These are not touched by this PR's commits — they are artifacts of the branch topology, not real deletions. No action needed beyond an eventual rebase; flagging so the phantom diff isn't mistaken for scope creep.

internal/marketplace/installer.go
kense-lab: Fixed in e7f5e16. Backup cleanup is no longer deferred unconditionally. If activation and rollback both fail, the previous Skill remains at the reported backup path. Added a fault-injection regression test covering the double-failure path; go test ./... passes.

Jerry-Xin (commented):
Re-review at new head `e7f5e16d146f` (delta from previously-reviewed `230291e1`: 1 commit, `fix(skills): preserve backup when install rollback fails`, touching only `installer.go` +9/-4 and `installer_test.go` +39).

**Verdict: COMMENT** — the fix is correct and closes the rollback data-loss finding; verdict is gated only by the unrun code CI (not by the code).

## ✅ The rollback double-fault fix is byte-verified correct
This commit resolves the data-loss finding that three reviewers independently surfaced (missing test / mechanism / data-loss). Verified against the full file at this head:

- **Removed the unconditional `defer os.RemoveAll(backup)`** — this was the root data-loss vector.
- **Backup is now cleaned up only on the success path** (`if hadTarget { _ = os.RemoveAll(backup) }` after activation succeeds) — no orphan leak on the happy path.
- **On the double-fault path** (activation `rename` fails AND rollback `rename` fails), it now returns `activate installed skill: %v; rollback failed, previous Skill preserved at %s: %w` — the backup is **not** deleted, and the user is told exactly where their previous skill survives.
- **On activation-fail + rollback-success**, the old skill is restored to target and the error is returned; the orphan tmp backup dir is harmless (no data loss).
- **New test `TestInstallPreservesBackupWhenActivationAndRollbackFail`** injects a `rename` that fails on call #2 (activation) and #3 (rollback), then asserts the error mentions preservation **and** re-reads the backup dir to confirm the old skill content (`"old"`) survived. Real coverage of the exact double-fault path via clean dependency injection (`installWithRename`).

## 💬 Residual (all non-blocking, carried over)
- 🟡 `marketplace.NewClient(nil)` falls back to `http.DefaultClient` (follows redirects); the production caller passes a no-redirect client, so only the nil-fallback is unprotected. Latent nit.
- 🟡 DNS re-resolution after validation (rebinding TOCTOU) — minor on a validated-then-fetch path.
- 🟡 30s total `http.Client.Timeout` could clip a 64MiB archive on a slow link — consider a stall/idle timeout or make it configurable.

Note: the `0o644` extraction (no execute bit) is a **deliberate, correct** security control for docs-only skills, not a deficiency — agreeing with Octo-PR's calibration; don't "fix" it by preserving archive exec bits.

## Why COMMENT (not APPROVE)
Same operative gate as the #87→#89→#90 arc: **code CI has not run** (`build & test` / `npm test` / `osv-scan` absent — first-time-contributor gating; `code-review` still pending, only `label` passed, `check-sprint` red). `unrun ≠ green` on a high-risk download-and-install path. The PR also currently **conflicts with `main`** and must be rebased before it can merge. Local build + targeted marketplace tests pass, and the implementation itself is mature — once rebased, CI runs green, and governance clears, it's APPROVE-ready from my side.

OctoBoooot (commented):
## Re-review → Comment (`e7f5e16d146f`) — rollback data-loss finding RESOLVED

**Verdict:** Comment. The delta since my last review fixes the one finding that had reviewer convergence (the rollback double-fault data-loss), and does it cleanly with a test. No blocker/major. Still holding at COMMENT (not APPROVE) only because the repo's **code CI has not run** (unchanged from #90's prior head) and the PR is still **CONFLICTING with `main`** — both maintainer/rebase actions, not code issues.

### What changed since `230291e1` (delta reviewed — 2 files)
The rollback path I (and Jerry-Xin/mochashanyao/yujiawei) flagged is fixed exactly as recommended:
- **`internal/marketplace/installer.go`** — the unconditional `defer os.RemoveAll(backup)` is **removed**, and on rollback failure the restore error is now **surfaced, not swallowed**: `"activate installed skill: %v; rollback failed, previous Skill preserved at %s: %w"`. So on the double-fault path (activation rename fails **and** the restore rename also fails), the backed-up prior Skill is **preserved on disk and its path reported**, instead of being silently deleted. This directly addresses yujiawei's "user data loss" concern — a locally-modified same-name skill in the backup now survives.
- **Testability**: extracted `installWithRename(..., rename func)` so `Install` delegates with `os.Rename` and the failure path is injectable.
- **`installer_test.go`** — `TestInstallPreservesBackupWhenActivationAndRollbackFail` drives the exact double-fault and asserts both the informative error *and* that the backup dir with the old skill's content survives. Closes the rollback test-gap Jerry-Xin noted.

I traced every backup path to confirm the `defer` removal introduced no leak or unintended deletion: success path cleans the backup (`if hadTarget { os.RemoveAll(backup) }`); restore-success consumes it (renamed back to target); restore-failure deliberately preserves it; early returns before the target→backup rename never create the dir. All marketplace tests pass locally (SSRF suite + installer + the new rollback test); `go build ./...` clean.

### Still-open (carried minors, unchanged — non-blocking)
- **DNS-rebinding TOCTOU** (`client.go` `validateDownloadURL`): validate resolves the host, `http.Do` re-resolves without IP pinning. Gated by authenticated backend + blocked redirects → defense-in-depth. Pin the validated IP via `DialContext` to fully close.
- **CGNAT `100.64.0.0/10`** not in the private-range filter (`IsPrivate` excludes it).
- **Spec bloat nit**: `marketplace.json` still carries ~54 unused component schemas for 2 used ops.
None of these are merge-blockers.

### CI / merge status (why still COMMENT)
Unchanged from the prior head: only `check-sprint` (governance, red), `label`, `add-to-project` ran; `build & test`/`npm test`/`osv-scan` **absent** (first-time-contributor gating). PR is **CONFLICTING/DIRTY** with `main`. `unrun ≠ green`, so I can't attest the code gate on this high-risk path. These are maintainer/rebase actions, not author code defects.

### Praise
- Textbook response to the finding: removed the unsafe unconditional cleanup, turned a swallowed error into an actionable message that tells the user *where their data is*, made the path testable with a minimal seam, and pinned it with a test that asserts the data actually survives. This is the right way to resolve a "rare but real data-loss" report — fix the mechanism and prove it. _(confidence: high.)_

Net: the substantive finding is resolved and verified. Once the PR is rebased and a maintainer lets code CI run green, this is APPROVE-ready from my side (the two carried minors are optional hardening).

internal/marketplace/installer.go
kense-lab: Rebased onto current main and force-pushed the branch at e847840. The rollback double-fault fix remains intact on the latest head, the regression test is still present, and the marketplace scope is unchanged: only internal OpenAPI skill.get/skill.download plus the skills install entry. Verified with git diff --check, go test ./internal/marketplace ./internal/config ./internal/registry, and go test ./cmd -run Marketplace.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #90 (octo-cli)

## Summary

This PR introduces `octo-cli skills install <skill-id> --from marketplace --dir <dir>`, a new command that fetches skill metadata from the Marketplace API, downloads the ZIP archive, validates its integrity (SHA-256), and installs it into the agent's skills root directory with atomic rename and rollback. It adds a new `internal/marketplace` package (client, installer), an embedded OpenAPI spec for the marketplace service (disabled for auto-registration), SSRF hardening on the artifact download path, and configuration for the marketplace API gateway prefix.

The security posture is thoughtful and well-tested: SSRF protection blocks private/link-local/metadata IPs, HTTPS is enforced (with an explicit localhost opt-in), ZIP path traversal is prevented, symlinks are rejected, checksums are verified before extraction, and atomic install with backup/rollback preserves the previous skill on failure. Test coverage is comprehensive across the client, installer, and command integration paths.

## Verification

Static analysis only at head `e7f5e16d146f48f28241b3e78135697d9dfc796d`; build and tests not executed in this environment.

- ✅ **SSRF protection** — `validateDownloadURL` at `internal/marketplace/client.go:155` rejects non-HTTPS schemes, resolves DNS, and blocks private/loopback/link-local/unspecified IPs. Cloud-metadata `169.254.169.254` is explicitly tested and rejected. Redirects are blocked via `CheckRedirect` returning `http.ErrUseLastResponse`.
- ✅ **Auth gating** — `cmd/skills.go:170` restricts marketplace installs to `bf_*` User Bot tokens via `credential.TokenKind`. App Bot tokens (`app_*`) are rejected with a clear error message.
- ✅ **Path traversal prevention** — `extract()` at `internal/marketplace/installer.go:109` applies `filepath.Clean`, rejects absolute paths, `..` traversal, and symlinks. A secondary `filepath.Rel` check confirms no escape from the destination directory.
- ✅ **Atomic install with rollback** — `installWithRename` at `:37` uses temp-dir + rename with backup of existing skill. If activation fails, rollback restores the backup. If rollback also fails, the error message points to the preserved backup location. Test `TestInstallPreservesBackupWhenActivationAndRollbackFail` validates this path.
- ✅ **Checksum verification** — SHA-256 of the downloaded archive is verified against the server-provided digest before extraction (`:44`).
- ✅ **overlayProfileBaseURL** — The refactor at `internal/cmdutil/factory.go:206-225` moves the `OCTO_API_BASE_URL` env-var check from outer early-return to inner guard. Functionally equivalent: if the env var is set, `cfg.APIBaseURL` is never overwritten. Verified by tracing both code paths.
- ✅ **Registry isolation** — `marketplace` is `x-octo-disabled` in the OpenAPI spec, preventing auto-registered CLI commands. The internal marketplace client uses `GetOperation` which works on disabled services. Test `TestMarketplaceHasNoTopLevelCommand` confirms no top-level marketplace command leaks.

## Findings

No P0/P1 issues; two P2 items and one nit below.

### P2 — Independent compressed and uncompressed size budgets (`internal/marketplace/client.go:18`)

`DownloadSkill` caps the HTTP response at `MaxArchiveBytes` (64 MiB, compressed), while `extract()` caps uncompressed file total at `maxTotalSize` (32 MiB) and individual files at `maxFileSize` (16 MiB). These are independent budgets. A high-compression-ratio ZIP — for example a 2 MiB archive of repeated text expanding to 32 MiB — passes both caps but still forces the extractor to write the full 32 MiB to disk. The uncompressed cap does bound disk usage, so this is not a blocker, but the two caps are not coordinated. Consider lowering `MaxArchiveBytes` or adding a compressed-to-uncompressed ratio check for tighter defense-in-depth.

### P2 — `ServiceURL("marketplace")` silently discards base-URL path (`internal/config/config.go:84`)

`ServiceURL("marketplace")` replaces `u.Path` entirely with `MarketplaceAPIPrefix`. If `APIBaseURL` carries a non-trivial path — for example `https://gateway.example.com/octo` — the resulting marketplace URL becomes `https://gateway.example.com/<prefix>`, silently dropping `/octo`. All other services return `APIBaseURL` unchanged. This is fine for the current deployment topology where the API lives at the host root, but worth either documenting as a constraint or guarding with a warning when the base URL has a non-empty path.

### Nit — DNS rebinding TOCTOU in download URL validation (`internal/marketplace/client.go:160`)

`validateDownloadURL` resolves DNS at validation time, but `c.http.Do(req)` resolves DNS independently at connection time. A DNS-rebinding attack could theoretically serve a private IP at connection time after a public IP passes validation. This is extremely difficult to exploit for a CLI one-shot download, and the main SSRF vectors (private IPs, cloud metadata) are blocked at validation time. Flagging for completeness.

- `overlayProfileBaseURL` refactor is functionally equivalent (env-var check moved from outer to inner guard).
- `operationPath` uses `url.PathEscape` for skill IDs — correct URL encoding.
- `archiveRootPrefix` correctly strips a single packaging directory only when all entries share it and `SKILL.md` is at the root of that directory.
- `envBool` helper handles all common truthy values (`1`, `true`, `yes`, `on`) case-insensitively.
- `marketplaceAPIRoot()` normalization ensures the prefix always ends with `/api/v1` and handles the empty-string edge case (returns `/api/v1`).
- Error handling in `cmd/skills.go` consistently routes through `emitSkillInstallError` for structured error output.
- `TestMarketplaceSkillInstall` uses a TLS mock server with proper transport override and validates the Authorization header propagates.
- No docs drift — `CLAUDE.md` is updated consistently with the new command, env var, and config field.

## Verdict: APPROVED

This is a well-engineered feature with strong security posture (SSRF hardening, path traversal prevention, checksum verification, atomic install with rollback, auth gating). The two P2 items are defense-in-depth and documentation concerns, not correctness blockers. No P0/P1 issues found.

[Octo-Q] verdict: APPROVE — No P0/P1 blockers. Two P2 defense-in-depth/documentation items and one nit. Strong security posture and comprehensive test coverage.

Jerry-Xin (commented):
Re-review at head `e84784031fc1`. **Verdict: COMMENT (unchanged).**

## Delta is a rebase, not new code
`e7f5e16d…e84784031` shows ahead 9 / behind 7 — that's a **rebase onto `main`** (picking up #85/#86 docs work), not new PR content. Byte-verified: the four core files are **blob-identical** to my previously-reviewed head —
- `internal/marketplace/installer.go` (sha `ff87f005`) — SAME
- `internal/marketplace/client.go` (sha `fd85b808`) — SAME
- `internal/config/config.go` (sha `573ca956`) — SAME
- `cmd/skills.go` (sha `eadffe65`) — SAME

So the rollback double-fault fix I verified at `e7f5e16d` (removed unconditional `defer os.RemoveAll(backup)`; preserves + reports backup path on double-fault; `TestInstallPreservesBackupWhenActivationAndRollbackFail`) stands unchanged. **Bonus:** the rebase resolved the merge conflict — `mergeable` is now `MERGEABLE` (was CONFLICTING).

## Adjudicating the two newest 🟡 notes
- 🟢 **"archive-size limit vs decompressed-size limit not coordinated"** — I **don't corroborate this as a gap.** `extract()` already has a complete, coordinated decompression-bomb defense: a per-file `maxFileSize` cap **and** a cumulative `total+size > maxTotalSize` cap on the **declared uncompressed** sizes, checked *before* each write, with `io.CopyN(f, rc, size)` enforcing exactly the declared size (a lying entry can't over-read). The decompressed total *is* the coordinated bound, independent of and stricter than the compressed archive size. Well-formed, not a defect.
- 🟡 **env-var refactor complexity** — style/readability opinion, non-blocking; the behavior is equivalence-preserving.

## Merge gate (unchanged — why still COMMENT)
- **Code CI has not run** — only `label` passed; `build & test`/`npm test`/`osv-scan` absent (first-time-contributor gating), `code-review` pending. `unrun ≠ green` on a high-risk download-and-install path.
- **`check-sprint` governance red** → maintainer must clear.
- **`reviewDecision: CHANGES_REQUESTED`** — yujiawei's 🔴 CR (rollback data-loss) is still active on the record; its mechanism is now **fixed** at this head, so it should be dismissed/re-graded by that reviewer.

Implementation itself is mature and now merges cleanly. Once code CI runs green and governance clears, it's APPROVE-ready from my side.

OctoBoooot (commented):
## Re-review → Comment (`e84784031fc1`) — rebase onto main; marketplace code unchanged

**Verdict:** Comment (carries forward from my `e7f5e16d` review). This head is that same reviewed commit **rebased onto main** — the conflict is resolved (`mergeable` is now MERGEABLE), but code CI still hasn't run, so I'm holding at COMMENT for the same reason as the whole #87→#90 arc.

### What changed since `e7f5e16d` (delta reviewed)
Pure rebase — the delta is entirely **already-reviewed content from main**, not new marketplace work:
- The marketplace feature files are **byte-identical** across the rebase: `internal/marketplace/**`, `internal/registry/specs/marketplace.json`, and `cmd/skills.go` show a **0-line diff**. The rollback data-loss fix I verified on `e7f5e16d` (preserve backup + report path + test) is intact.
- The rest of the delta is #85 (`docs.share` — docs.json/flags.go/loader.go/common.md) + #86 (sheet hyperlinks) + CHANGELOG folding in from main — all content I already reviewed and APPROVED on their own PRs. That's what cleared the CONFLICTING/DIRTY state.
- Verified on this head: `go build ./...` clean; `go test ./internal/marketplace/ ./internal/registry/` pass.

### On the two notes raised against this head (both non-corroborated)
- **"download cap vs decompressed cap uncoordinated (zip-bomb risk)"** — doesn't hold up: the caps *are* coordinated and defensive. Download ≤ 64 MiB (compressed), decompressed **total ≤ 32 MiB**, per-file ≤ 16 MiB, ≤ 1024 files, and `io.CopyN` bounds each extraction to its declared size. The on-disk cap (32 MiB) is enforced *during* extraction and is lower than the download cap, so a high-ratio zip bomb still can't exceed 32 MiB. That's the correct design, not a gap.
- **"envBool refactor adds complexity"** — subjective style; the refactor is functionally equivalent (config_test covers it). Not a defect.

### Still-open (carried minors, unchanged — non-blocking)
DNS-rebinding TOCTOU (pin the validated IP via `DialContext`); CGNAT `100.64.0.0/10` absent from the private-range filter; `marketplace.json` still carries ~54 unused component schemas for 2 used ops. None merge-blocking.

### CI / merge status (why still COMMENT)
`mergeable` is now MERGEABLE (rebase resolved the conflict) but `mergeStateStatus` is BLOCKED: only `check-sprint` (governance, **red**) and `label` ran; `build & test`/`npm test`/`osv-scan` are **still absent** even post-rebase (first-time-contributor gating), and `code-review` is pending. `unrun ≠ green` on a high-risk download-and-install path. A maintainer still needs to let code CI run and clear `check-sprint`.

Net: no new marketplace code to review — the rollback fix stands and the rebase cleanly integrates already-approved work. Once a maintainer lets code CI run green + clears governance, this is APPROVE-ready from my side.

kense-lab:
@yujiawei The P1 rollback data-loss issue from your review has been fixed and regression-tested in e7f5e16, then preserved through the rebase at e847840. The previous Skill backup is now retained and its path reported when activation and rollback both fail. Could you please re-review or dismiss the stale CHANGES_REQUESTED review? `go test ./...` passes.

yujiawei (approved):
# Code Review — PR #90 (octo-cli)

Reviewed at head `e8478403` (the branch was rebased onto current `main`; the marketplace/installer/client/cmd diff is byte-identical to the previously reviewed revision, and the rebase resolves the phantom-diff artifact from the earlier review — the docs-side changes no longer show as spurious deletions). Base merge-base: `1e3acf6`. `go build ./...`, `go test ./...`, and `go vet ./...` all pass locally.

This is a re-review. The single blocking issue from the prior round — a data-loss defect in the atomic-install rollback — is now fixed and covered by a test. No new blockers found.

## 1. Spec compliance

**Result: PASS ✅**

The change implements exactly what the PR describes and nothing out of scope:

- `skills install <id> --from marketplace --dir <root>` added as a subcommand of the existing `skills` command, not a new top-level command (`cmd/skills.go:154`, guarded by `TestMarketplaceHasNoTopLevelCommand`).
- Marketplace access is driven by the embedded OpenAPI contract via `operationId` (`skill.get`, `skill.download`) rather than hardcoded routes (`internal/marketplace/client.go:57,86`; `internal/registry/specs/marketplace.json`), and the marketplace service stays hidden from the command tree (`TestMarketplaceOpenAPIContract` asserts `ServiceDisabled("marketplace")`).
- Download is https-only by default with an explicit loopback-only dev switch (`OCTO_MARKETPLACE_ALLOW_INSECURE_LOCALHOST`), SHA256-verified against the API-supplied hash, size-capped, and extracted atomically.
- No unrequested flags, fields, or endpoints; the OpenAPI doc embeds only `skill.get` and `skill.download` (operation count asserted in `loader_test.go`).

## 2. Code quality

**Result: Approved (no P0/P1)**

### Resolved from prior review

**Atomic-install rollback no longer destroys the user's existing skill** — `internal/marketplace/installer.go:81-91`. The unconditional `defer os.RemoveAll(backup)` is gone. Backup cleanup now happens only on the success path (`if hadTarget { _ = os.RemoveAll(backup) }`), and on a double failure (activation rename fails, then restore rename fails) the backup is preserved and its path is surfaced in the returned error:

```go
if err := rename(tmp, target); err != nil {
    if hadTarget {
        if restoreErr := rename(backup, target); restoreErr != nil {
            return Result{}, fmt.Errorf("activate installed skill: %v; rollback failed, previous Skill preserved at %s: %w", err, backup, restoreErr)
        }
    }
    return Result{}, fmt.Errorf("activate installed skill: %w", err)
}
```

The `installWithRename` seam lets `TestInstallPreservesBackupWhenActivationAndRollbackFail` (`installer_test.go:126`) force the exact double-failure path and assert the prior skill's contents survive on disk. Good, targeted regression coverage for the defect that was reported.

### Non-blocking (P2 / advisory)

None of these block merge. The download source is the authenticated first-party marketplace and every archive is SHA256-gated against the API-supplied hash before install, which bounds the blast radius of the items below.

- **DNS-rebinding / TOCTOU on the download host** — `client.go:158-188` validates the resolved IPs, but `c.http.Do` (`client.go:118`) re-resolves independently; there is no pinned `DialContext`/`Control`. For a hostname-based `download_url`, the validation and fetch lookups can diverge (short-TTL / rebinding), so the private-IP guard is best-effort. Impact is limited to SSRF-as-probe given the SHA256 gate. Consider resolving once and pinning the validated IP via `net.Dialer.Control`.
- **SSRF blocklist gaps** — `client.go:179-186`. `IsPrivate/IsLoopback/IsLinkLocalUnicast/IsUnspecified` do not cover CGNAT `100.64.0.0/10` or some IPv6 ULA/link-local-multicast edge cases. Minor given the mitigations above.
- **Fixed 30s client timeout vs 64 MiB cap** — `cmd/skills.go:187-192`. `http.Client.Timeout` is a wall-clock deadline over the entire request including body read, so a large archive on a slow link (≈2.1 MB/s needed for 64 MiB in 30s) can fail even though the context deadline would allow more. Consider a longer timeout for the download leg, or drop `Timeout` and rely on a context deadline sized to the body.
- **Directory entries bypass extractor resource limits** — `installer.go:121-126`. Directory entries hit `MkdirAll` + `continue` before the `maxFiles` / size accounting, so a pathological archive with a huge number of directory entries is bounded only by the 64 MiB download cap, not `maxFiles`. Low risk (first-party source, size cap), but counting directories toward `maxFiles` would close it.
- **Reserved-backup temp dir can leak on an early error** — `installer.go:65-71`. If `os.Remove(backup)` fails, the empty temp dir is left behind. Cosmetic.

### Test coverage

Security controls are well covered: https-only rejection, private/metadata-IP rejection, redirect non-follow, checksum mismatch, path traversal, symlink rejection, missing SKILL.md, single packaging-dir strip, `--dir` required, App-Bot rejection, and marketplace-not-top-level. The new rollback double-failure test closes the gap called out last round.

Remaining gaps (all P2, optional): the redirect-non-follow test (`client_test.go:87-121`) builds its own `http.Client` rather than exercising the production wiring in `cmd/skills.go:187-192`, so the shipped `CheckRedirect` guard is not covered end-to-end; `TestMarketplaceSkillInstall` mutates global `http.DefaultTransport` (latent hazard if any `cmd` test ever adopts `t.Parallel()`); and the archive resource-limit guards (`maxFiles` / `maxFileSize` / `maxTotalSize`, `MaxArchiveBytes` boundary) and the `archiveRootPrefix` negative branches (mixed roots, single root without SKILL.md) are untested.

## 3. Overall verdict

**APPROVE**

Spec compliance is clean, the previously blocking data-loss defect is fixed and regression-tested, and build/test/vet pass. The remaining items are advisory hardening and follow-up test coverage, not merge blockers. I'd suggest picking up the download-timeout and DNS-pinning items as fast follow-ups.

kense-lab:
Maintainer action required before this PR can become mergeable:

1. Approve the fork workflow runs blocked as `action_required`:
   - CI: https://github.com/Mininglamp-OSS/octo-cli/actions/runs/29550545826
   - OSV-Scanner: https://github.com/Mininglamp-OSS/octo-cli/actions/runs/29550546059
   - History Check: https://github.com/Mininglamp-OSS/octo-cli/actions/runs/29550546087
2. Re-request or dismiss yujiawei’s stale CHANGES_REQUESTED review; its P1 finding was fixed and verified on the latest head.

The contributor token cannot approve workflow runs or request reviewers. Latest head is mergeable with no conflicts and `go test ./...` passes.

kense-lab:
@OctoBoooot @Jerry-Xin Please review the latest head and approve if there are no remaining blocking findings. The rollback P1 is fixed with a regression test, the branch is rebased and mergeable, and `go test ./...` passes.

PR #91, opened 2026-07-17
feat(marketplace): add OpenAPI-driven skill and MCP commands

kense-lab:
## Summary

- embed the Marketplace OpenAPI operations and generate Skill/MCP CLI commands from the registry
- preserve resource domains when one service spec exposes multiple command families
- add progressive-disclosure agent guidance for Skill and MCP workflows
- cover command shape, request construction, visibility enums, Space headers, and embedded skill references

## Validation

- `go test ./...`
- `git diff --check`
- real local Skill publish/search/download/update flow
- real local MCP create/search/detail/update flow

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -74,7 +74,7 @@ octo-cli version
 
 `octo-cli auth login` stores a bot token (read from a hidden prompt, `--with-token` stdin, or `--token-file` — never argv) under a profile keyed by `--bot-id`/`--profile`. `status`/`list` show metadata only (tokens always masked); `logout` removes a profile.
 
-Bot-type capability and per-command flags are in `docs/octo-cli-design.md`. Agent-facing usage lives under `skills/` (`octo-shared`, `octo-matter` (withheld — see above), `octo-messaging`, `octo-files`, `octo-docs`, `octo-html`) — keep those in sync when command shapes change.
+Bot-type capability and per-command flags are in `docs/octo-cli-design.md`. Agent-facing usage lives under `skills/` (`octo-shared`, `octo-matter` (withheld — see above), `octo-messaging`, `octo-files`, `octo-docs`, `octo-html`, `octo-marketplace`) — keep those in sync when command shapes change.
 
 ## Environment
 
--- a/README.md
+++ b/README.md
@@ -261,6 +261,8 @@ Machine-readable usage docs for AI Agents live under [`skills/`](./skills/):
   lives in sibling files loaded on demand: `sheet.md` (spreadsheets), `doc.md`
   (rich-text body), `board.md` (whiteboard), and `common.md` (comments,
   versions, members/sharing, attachments).
+- [`octo-marketplace`](./skills/octo-marketplace/SKILL.md) — search, install,
+  publish, and update Marketplace Skills and MCP server listings.
 
 These docs are also **embedded in the binary**, so a released `octo-cli` ships them:
 
--- a/cmd/service/flags.go
+++ b/cmd/service/flags.go
@@ -68,7 +68,7 @@ func registerQueryFlags(cmd *cobra.Command, rt *operationRuntime, d *registry.Op
 			continue
 		}
 		flagName := paramFlagName(p)
-		qf := &queryFlag{apiName: p.Name, kind: schemaTypeKind(p.Type)}
+		qf := &queryFlag{apiName: p.Name, kind: queryParamKind(p)}
 		desc := p.Description
 		if len(p.Enum) > 0 {
 			desc = fmt.Sprintf("%s (one of: %s)", desc, formatEnum(p.Enum))
@@ -246,12 +246,16 @@ func promotableKind(p *registry.SchemaInfo) (valueKind, bool) {
 	return 0, false
 }
 
-func schemaTypeKind(t string) valueKind {
-	switch t {
+func queryParamKind(p *registry.ParamInfo) valueKind {
+	switch p.Type {
 	case "integer", "number":
 		return kindInt
 	case "boolean":
 		return kindBool
+	case "array":
+		if p.Items != nil && p.Items.Type == "string" {
+			return kindStringSlice
+		}
 	}
 	return kindString
 }
--- a/cmd/service/marketplace_test.go
+++ b/cmd/service/marketplace_test.go
@@ -91,21 +91,27 @@ func TestMarketplaceSkillSearchRequest(t *testing.T) {
 }
 
 func TestMarketplaceMCPSearchFiltersRequest(t *testing.T) {
-	var gotQuery string
+	var gotQuery map[string][]string
 	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
-		gotQuery = r.URL.RawQuery
+		gotQuery = r.URL.Query()
 		w.Header().Set("Content-Type", "application/json")
 		_, _ = w.Write([]byte(`{"data":[],"pagination":{"total":0,"page":1,"page_size":20}}`))
 	})
 
-	root.SetArgs([]string{"marketplace", "mcp", "list", "--transport", "stdio", "--visibility", "public", "--source", "space", "--created-by-type", "bot", "--tag", "cli", "--sort", "relevance"})
+	root.SetArgs([]string{"marketplace", "mcp", "list", "--transport", "stdio", "--visibility", "public", "--source", "space", "--created-by-type", "bot", "--tag", "cli", "--tag", "devops", "--sort", "relevance"})
 	if err := root.Execute(); err != nil {
 		t.Fatalf("execute: %v", err)
 	}
-	for _, want := range []string{"transport=stdio", "visibility=public", "source=space", "created_by_type=bot", "tag=cli", "sort=relevance"} {
-		if !strings.Contains(gotQuery, want) {
-			t.Errorf("query = %q, want %q", gotQuery, want)
-		}
+	want := map[string][]string{
+		"transport":       {"stdio"},
+		"visibility":      {"public"},
+		"source":          {"space"},
+		"created_by_type": {"bot"},
+		"tag":             {"cli", "devops"},
+		"sort":            {"relevance"},
+	}
+	if !reflect.DeepEqual(gotQuery, want) {
+		t.Errorf("query = %#v, want %#v", gotQuery, want)
 	}
 }
 
--- a/cmd/service/marketplace_test.go
+++ b/cmd/service/marketplace_test.go
@@ -16,23 +16,32 @@ func TestMarketplaceRegistryShape(t *testing.T) {
 		method string
 		path   string
 	}{
-		"skill_category.list":   {http.MethodGet, "/market/api/v1/skill_categories"},
-		"mcp_category.list":     {http.MethodGet, "/market/api/v1/mcp_categories"},
-		"skill.list":            {http.MethodGet, "/market/api/v1/skills"},
-		"skill.create":          {http.MethodPost, "/market/api/v1/skills"},
-		"skill.get":             {http.MethodGet, "/market/api/v1/skills/{skill_id}"},
-		"skill.update":          {http.MethodPatch, "/market/api/v1/skills/{skill_id}"},
-		"skill.download":        {http.MethodGet, "/market/api/v1/skills/{skill_id}/download"},
-		"skill.reupload.create": {http.MethodPost, "/market/api/v1/skills/{skill_id}/reuploads"},
-		"skill.version.list":    {http.MethodGet, "/market/api/v1/skills/{skill_id}/versions"},
-		"skill_upload.create":   {http.MethodPost, "/market/api/v1/skill_uploads"},
-		"skill_upload.parse":    {http.MethodPost, "/market/api/v1/skill_uploads/{skill_upload_id}/parse"},
-		"skill_parse_task.get":  {http.MethodGet, "/market/api/v1/skill_parse_tasks/{skill_parse_task_id}"},
-		"mcp.list":              {http.MethodGet, "/market/api/v1/mcps"},
-		"mcp.create":            {http.MethodPost, "/market/api/v1/mcps"},
-		"mcp.probe":             {http.MethodPost, "/market/api/v1/mcps/_probe"},
-		"mcp.get":               {http.MethodGet, "/market/api/v1/mcps/{mcp_id}"},
-		"mcp.update":            {http.MethodPatch, "/market/api/v1/mcps/{mcp_id}"},
+		"skill_category.list":      {http.MethodGet, "/market/api/v1/skill_categories"},
+		"mcp_category.list":        {http.MethodGet, "/market/api/v1/mcp_categories"},
+		"skill.list":               {http.MethodGet, "/market/api/v1/skills"},
+		"skill.mine.list":          {http.MethodGet, "/market/api/v1/skills/mine"},
+		"skill.tag.list":           {http.MethodGet, "/market/api/v1/skills/tags"},
+		"skill.publish":            {http.MethodPost, "/market/api/v1/bot/skills/publish"},
+		"skill.create":             {http.MethodPost, "/market/api/v1/skills"},
+		"skill.get":                {http.MethodGet, "/market/api/v1/skills/{skill_id}"},
+		"skill.update":             {http.MethodPatch, "/market/api/v1/skills/{skill_id}"},
+		"skill.delete":             {http.MethodDelete, "/market/api/v1/skills/{skill_id}"},
+		"skill.skillmd.get":        {http.MethodGet, "/market/api/v1/skills/{skill_id}/skill_md"},
+		"skill.download":           {http.MethodGet, "/market/api/v1/skills/{skill_id}/download"},
+		"skill.reupload.create":    {http.MethodPost, "/market/api/v1/skills/{skill_id}/reuploads"},
+		"skill.version.list":       {http.MethodGet, "/market/api/v1/skills/{skill_id}/versions"},
+		"skill_upload.create":      {http.MethodPost, "/market/api/v1/skill_uploads"},
+		"skill_upload.parse":       {http.MethodPost, "/market/api/v1/skill_uploads/{skill_upload_id}/parse"},
+		"skill_parse_task.get":     {http.MethodGet, "/market/api/v1/skill_parse_tasks/{skill_parse_task_id}"},
+		"skill_icon_upload.create": {http.MethodPost, "/market/api/v1/skill_icon_uploads"},
+		"mcp.list":                 {http.MethodGet, "/market/api/v1/mcps"},
+		"mcp.mine.list":            {http.MethodGet, "/market/api/v1/mcps/mine"},
+		"mcp.create":               {http.MethodPost, "/market/api/v1/mcps"},
+		"mcp.probe":                {http.MethodPost, "/market/api/v1/mcps/_probe"},
+		"mcp.get":                  {http.MethodGet, "/market/api/v1/mcps/{mcp_id}"},
+		"mcp.update":               {http.MethodPatch, "/market/api/v1/mcps/{mcp_id}"},
+		"mcp.delete":               {http.MethodDelete, "/market/api/v1/mcps/{mcp_id}"},
+		"mcp_icon_upload.create":   {http.MethodPost, "/market/api/v1/mcp_icon_uploads"},
 	}
 
 	if got := len(r.ListOperations("marketplace")); got != len(wants) {
@@ -64,14 +73,55 @@ func TestMarketplaceSkillSearchRequest(t *testing.T) {
 		_, _ = w.Write([]byte(`{"data":[],"pagination":{"has_more":false}}`))
 	})
 
-	root.SetArgs([]string{"marketplace", "skill", "list", "--q", "deep miner", "--page-size", "20"})
+	root.SetArgs([]string{"marketplace", "skill", "list", "--q", "deep miner", "--tag", "cli", "--sort", "latest", "--page-size", "20"})
 	if err := root.Execute(); err != nil {
 		t.Fatalf("execute: %v", err)
 	}
 	if gotPath != "/market/api/v1/skills" {
 		t.Errorf("path = %q", gotPath)
 	}
-	for _, want := range []string{"q=deep+miner", "page_size=20"} {
+	for _, want := range []string{"q=deep+miner", "tag=cli", "sort=latest", "page_size=20"} {
+		if !strings.Contains(gotQuery, want) {
+			t.Errorf("query = %q, want %q", gotQuery, want)
+		}
+	}
+	if strings.Contains(gotQuery, "offset=") {
+		t.Errorf("query = %q, must not use removed offset pagination", gotQuery)
+	}
+}
+
+func TestMarketplaceMCPSearchFiltersRequest(t *testing.T) {
+	var gotQuery string
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotQuery = r.URL.RawQuery
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"data":[],"pagination":{"total":0,"page":1,"page_size":20}}`))
+	})
+
+	root.SetArgs([]string{"marketplace", "mcp", "list", "--transport", "stdio", "--visibility", "public", "--source", "space", "--created-by-type", "bot", "--tag", "cli", "--sort", "relevance"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	for _, want := range []string{"transport=stdio", "visibility=public", "source=space", "created_by_type=bot", "tag=cli", "sort=relevance"} {
+		if !strings.Contains(gotQuery, want) {
+			t.Errorf("query = %q, want %q", gotQuery, want)
+		}
+	}
+}
+
+func TestMarketplaceMCPCategoryFiltersRequest(t *testing.T) {
+	var gotQuery string
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotQuery = r.URL.RawQuery
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"data":[]}`))
+	})
+
+	root.SetArgs([]string{"marketplace", "mcp-category", "list", "--mode", "mine", "--created-by-type", "bot"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	for _, want := range []string{"mode=mine", "created_by_type=bot"} {
 		if !strings.Contains(gotQuery, want) {
 			t.Errorf("query = %q, want %q", gotQuery, want)
 		}
@@ -98,6 +148,30 @@ func TestMarketplaceSkillPublishRequest(t *testing.T) {
 		t.Errorf("path = %q", gotPath)
 	}
 	if gotBody["parse_task_id"] != "task-1" || gotBody["visibility"] != "space" {
+		t.Errorf("body = %#v", gotBody)
+	}
+}
+
+func TestMarketplaceBotSkillPublishRequest(t *testing.T) {
+	var gotPath string
+	var gotBody map[string]any
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotPath = r.URL.Path
+		if err := json.NewDecoder(r.Body).Decode(&gotBody); err != nil {
+			t.Errorf("decode body: %v", err)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"data":{"skill_id":"skill-1"}}`))
+	})
+
+	root.SetArgs([]string{"marketplace", "skill", "publish", "--skill-upload-id", "upload-1", "--visibility", "space", "--changelog", "Initial release"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if gotPath != "/market/api/v1/bot/skills/publish" {
+		t.Errorf("path = %q", gotPath)
+	}
+	if gotBody["skill_upload_id"] != "upload-1" || gotBody["visibility"] != "space" || gotBody["changelog"] != "Initial release" {
 		t.Errorf("body = %#v", gotBody)
 	}
 }
@@ -133,12 +207,14 @@ func TestMarketplaceCommandTree(t *testing.T) {
 		t.Fatal("missing marketplace command")
 	}
 	domains := map[string][]string{
-		"skill-category":   {"list"},
-		"mcp-category":     {"list"},
-		"skill":            {"list", "create", "get", "update", "download", "reupload", "version"},
-		"skill-upload":     {"create", "parse"},
-		"skill-parse-task": {"get"},
-		"mcp":              {"list", "create", "probe", "get", "update"},
+		"skill-category":    {"list"},
+		"mcp-category":      {"list"},
+		"skill":             {"list", "mine", "tag", "publish", "create", "get", "update", "delete", "download", "reupload", "version", "skillmd"},
+		"skill-upload":      {"create", "parse"},
+		"skill-parse-task":  {"get"},
+		"skill-icon-upload": {"create"},
+		"mcp":               {"list", "mine", "create", "probe", "get", "update", "delete"},
+		"mcp-icon-upload":   {"create"},
 	}
 	for domain, children := range domains {
 		domainCmd := findCmd(marketplace, domain)
@@ -159,7 +235,7 @@ func TestMarketplaceCommandTree(t *testing.T) {
 
 func TestMarketplaceSkillTagsAreStrings(t *testing.T) {
 	r := registry.MustNew()
-	for _, id := range []string{"skill.create", "skill.update"} {
+	for _, id := range []string{"skill.publish", "skill.create", "skill.update"} {
 		op, ok := r.GetOperation(id)
 		if !ok || op.RequestBody == nil {
 			t.Fatalf("%s request body not found", id)
@@ -179,6 +255,7 @@ func TestMarketplaceVisibilityEnumsMatchBackend(t *testing.T) {
 	}{
 		{"skill.create", []string{"public", "private", "space"}},
 		{"skill.update", []string{"public", "private", "space"}},
+		{"skill.publish", []string{"public", "private", "space"}},
 		{"mcp.create", []string{"public", "private"}},
 		{"mcp.update", []string{"public", "private"}},
 	} {
--- a/cmd/service/marketplace_test.go
+++ b/cmd/service/marketplace_test.go
@@ -1,0 +1,226 @@
+package service
+
+import (
+	"encoding/json"
+	"net/http"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-cli/internal/registry"
+)
+
+func TestMarketplaceRegistryShape(t *testing.T) {
+	r := registry.MustNew()
+	wants := map[string]struct {
+		method string
+		path   string
+	}{
+		"skill_category.list":   {http.MethodGet, "/market/api/v1/skill_categories"},
+		"mcp_category.list":     {http.MethodGet, "/market/api/v1/mcp_categories"},
+		"skill.list":            {http.MethodGet, "/market/api/v1/skills"},
+		"skill.create":          {http.MethodPost, "/market/api/v1/skills"},
+		"skill.get":             {http.MethodGet, "/market/api/v1/skills/{skill_id}"},
+		"skill.update":          {http.MethodPatch, "/market/api/v1/skills/{skill_id}"},
+		"skill.download":        {http.MethodGet, "/market/api/v1/skills/{skill_id}/download"},
+		"skill.reupload.create": {http.MethodPost, "/market/api/v1/skills/{skill_id}/reuploads"},
+		"skill.version.list":    {http.MethodGet, "/market/api/v1/skills/{skill_id}/versions"},
+		"skill_upload.create":   {http.MethodPost, "/market/api/v1/skill_uploads"},
+		"skill_upload.parse":    {http.MethodPost, "/market/api/v1/skill_uploads/{skill_upload_id}/parse"},
+		"skill_parse_task.get":  {http.MethodGet, "/market/api/v1/skill_parse_tasks/{skill_parse_task_id}"},
+		"mcp.list":              {http.MethodGet, "/market/api/v1/mcps"},
+		"mcp.create":            {http.MethodPost, "/market/api/v1/mcps"},
+		"mcp.probe":             {http.MethodPost, "/market/api/v1/mcps/_probe"},
+		"mcp.get":               {http.MethodGet, "/market/api/v1/mcps/{mcp_id}"},
+		"mcp.update":            {http.MethodPatch, "/market/api/v1/mcps/{mcp_id}"},
+	}
+
+	if got := len(r.ListOperations("marketplace")); got != len(wants) {
+		t.Fatalf("marketplace operation count = %d, want %d", got, len(wants))
+	}
+	for id, want := range wants {
+		op, ok := r.GetOperation(id)
+		if !ok {
+			t.Fatalf("operation %q not found", id)
+		}
+		if op.Method != want.method || op.Path != want.path {
+			t.Errorf("%s = %s %s, want %s %s", id, op.Method, op.Path, want.method, want.path)
+		}
+		if op.BaseURLEnv != "OCTO_API_BASE_URL" {
+			t.Errorf("%s base URL env = %q, want OCTO_API_BASE_URL", id, op.BaseURLEnv)
+		}
+		if !op.SpaceHeader {
+			t.Errorf("%s must send the space header", id)
+		}
+	}
+}
+
+func TestMarketplaceSkillSearchRequest(t *testing.T) {
+	var gotPath, gotQuery string
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotPath = r.URL.Path
+		gotQuery = r.URL.RawQuery
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"data":[],"pagination":{"has_more":false}}`))
+	})
+
+	root.SetArgs([]string{"marketplace", "skill", "list", "--q", "deep miner", "--page-size", "20"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if gotPath != "/market/api/v1/skills" {
+		t.Errorf("path = %q", gotPath)
+	}
+	for _, want := range []string{"q=deep+miner", "page_size=20"} {
+		if !strings.Contains(gotQuery, want) {
+			t.Errorf("query = %q, want %q", gotQuery, want)
+		}
+	}
+}
+
+func TestMarketplaceSkillPublishRequest(t *testing.T) {
+	var gotPath string
+	var gotBody map[string]any
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotPath = r.URL.Path
+		if err := json.NewDecoder(r.Body).Decode(&gotBody); err != nil {
+			t.Errorf("decode body: %v", err)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"data":{"skill_id":"skill-1"}}`))
+	})
+
+	root.SetArgs([]string{"marketplace", "skill", "create", "--parse-task-id", "task-1", "--visibility", "space"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if gotPath != "/market/api/v1/skills" {
+		t.Errorf("path = %q", gotPath)
+	}
+	if gotBody["parse_task_id"] != "task-1" || gotBody["visibility"] != "space" {
+		t.Errorf("body = %#v", gotBody)
+	}
+}
+
+func TestMarketplaceMCPUpdateRequest(t *testing.T) {
+	var gotMethod, gotPath string
+	var gotBody map[string]any
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotMethod, gotPath = r.Method, r.URL.Path
+		if err := json.NewDecoder(r.Body).Decode(&gotBody); err != nil {
+			t.Errorf("decode body: %v", err)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"data":{"mcp_id":"mcp-1"}}`))
+	})
+
+	root.SetArgs([]string{"marketplace", "mcp", "update", "mcp-1", "--slogan", "Updated"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if gotMethod != http.MethodPatch || gotPath != "/market/api/v1/mcps/mcp-1" {
+		t.Errorf("request = %s %s", gotMethod, gotPath)
+	}
+	if gotBody["slogan"] != "Updated" {
+		t.Errorf("body = %#v", gotBody)
+	}
+}
+
+func TestMarketplaceCommandTree(t *testing.T) {
+	root, _, _ := rootWithService(t, func(http.ResponseWriter, *http.Request) {})
+	marketplace := findCmd(root, "marketplace")
+	if marketplace == nil {
+		t.Fatal("missing marketplace command")
+	}
+	domains := map[string][]string{
+		"skill-category":   {"list"},
+		"mcp-category":     {"list"},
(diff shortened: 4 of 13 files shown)


Jerry-Xin (commented):
Solid, spec-driven addition — the Go surface is clean and the security-sensitive install/extract logic lives in agent-facing docs with correct mitigations; holding at COMMENT only because code CI (build&test / osv-scan) has not run yet on a high-risk install path.

## 🔴 Blocking
None. No code-level blockers found. The one gating concern is process, not code (see gate note below).

## 💬 Non-blocking
- **Gate / CI:** `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`, `check-sprint` is failing, and `code-review` is still `PENDING` — the actual code CI (build & test, npm test, osv-scan) has not run. On a download-and-execute install surface I want a green build/test + osv-scan before an approve. Locally `go build ./...` and `go test ./...` both pass at head, so I expect CI to be green once it runs.
- **`mcp.probe` risk classification:** `_probe` is tagged `x-octo-risk: read`. It performs an outbound connection attempt to a user-supplied endpoint (server-side, non-persisting). Read is defensible since nothing is stored, but it is a network side-effect driven by caller input; consider whether the risk gate should treat it as more than a plain read.
- **`skill.download` dual 200/302:** the spec advertises both a JSON `200` and a `302` redirect. The client correctly does not auto-follow redirects and `download` does not expose `-o` (good — avoids the redirect `-o` footgun that `file.download` documents). Worth a one-line spec/doc note that callers must pass `--response-format json` to get the URL+digest envelope rather than the raw redirect.
- **Install-flow safety is doc-enforced, not code-enforced:** SHA-256 match, `..`/absolute/symlink/device rejection, atomic move, and "never execute archive scripts" are specified in the skill markdown for the agent runtime to honor. That is the right layer given no extraction code ships here, but it means correctness depends on runtime adherence — no compiler/test guardrail. Fine for this PR; flagging so it is a conscious decision.

## ✅ Highlights
- Arg-binding is correct across the new surface: body-only ops (`mcp probe`, `skill create`, `mcp create`) take zero positional args + promoted flags/`--data`; path-param ops (`skill get/download/update`, `mcp get/update`) take exactly their `{id}` positional. The positional-vs-body defect class does not appear here.
- The `attachOperation` change (preserve the first operationId segment as a resource group when it differs from the service name) cleanly resolves the `skill.list` vs `mcp.list` collision under one `marketplace` service, and `_`→`-` normalization gives tidy `skill-category` / `skill-upload` / `skill-parse-task` groups with no collisions.
- No new download/extract/redirect-follow Go code is introduced; the REST client keeps `CheckRedirect => ErrUseLastResponse`, so presigned URLs are surfaced, not fetched (SSRF-via-redirect safe).
- Strong test coverage: registry shape/counts, space-header contract, visibility enums vs backend, tag-schema-is-string-array, command-tree shape, request construction (query/body/method/path), and embedded-skill reference checks. Matches the repo's per-module test convention.
- Skill docs are well-scoped: probe-vs-local-handshake split for stdio, secret-redaction rules, placeholder-not-real-secret guidance, and payload-normalization helper for envelope/paginated shapes.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #91 (octo-cli)

## Summary

This PR adds the `marketplace` service to octo-cli, an OpenAPI-driven CLI that
generates cobra command trees from embedded spec files. The marketplace is the
first service whose operationId first segments (e.g. `skill.list`, `mcp.create`)
differ from the service name (`marketplace`), so the core `attachOperation`
function gains a conditional start-index to preserve resource-group segments
that would previously have been skipped. A 545-line OpenAPI 3.1.0 spec covering
17 operations, three skill docs, and tests round out the change.

The core logic change in `cmd/service/service.go` is correct and well-tested.
Two P2 items in the spec are noted below.

## Verification

- ✅ **`attachOperation` backward compatibility** — when `segs[0] == d.Service`
  (all existing services: docs, bot, event, file, group, matter, message,
  thread), `start = 1` preserves the exact prior loop behavior. The new
  `start = 0` branch only activates for marketplace-style operationIds where
  the first segment is a resource group, not the service name
  (`cmd/service/service.go:79-82`).
- ✅ **Underscore-to-hyphen normalization** — `strings.ReplaceAll(segs[i], "_", "-")`
  at `cmd/service/service.go:84` handles `skill_category` → `skill-category`.
  Searched all existing specs: no operationId contains underscores (they use
  hyphens: `read-receipt`, `md-get`, `member-add`), so the normalization is a
  no-op for all existing services.
- ✅ **Test coverage** — `TestMarketplaceCommandTree` asserts 6 domain groups
  with all expected children. `TestRegisterServiceCommands_PreservesForeignOperationDomain`
  specifically pins that marketplace preserves "skill" and "mcp" resource
  groups while existing services keep their flat structure.
  `TestAllDomainOperationCounts` expects marketplace: 17 ops.
- ✅ **Space-header contract** — `TestServiceSpaceHeaderContract` pins
  marketplace `skill.get` = `true`, matching the spec's
  `x-octo-space-header: true`.
- ✅ **Skill docs** — `TestOctoMarketplaceReferencesEmbedded` verifies
  `SKILL.md` routes to `skills.md` and `mcp.md`, both embedded and non-empty.
- ✅ **`loader.go` change** — comment-only update adding "marketplace" to the
  domain list. No behavioral change, no async/retry logic touched.

Static analysis only at head `2456b06e`; build and tests not executed in this
environment.

## Findings

### P2 — Most operations lack response schemas (`internal/registry/specs/marketplace.json:17`)

Fourteen of 17 marketplace operations declare only `"200": { "description": "OK" }`
without a `content.application/json.schema`. The CLI runtime still works — raw
JSON passes through to the user — but `octo-cli schema` cannot describe response
shapes for agents. Only `skill.get` and `skill.download` carry full response
schemas via `$ref`.

Diff-scope: new (introduced by this PR). Not a correctness issue for CLI users,
but agents relying on schema introspection to build marketplace workflows will
see empty response shapes for most commands.

Adding `content.application/json.schema` with `$ref` to component schemas for
the main list and write operations — following the existing
`SkillEnvelope` / `DownloadEnvelope` pattern — would close this.

### P2 — MCP list has no auto-pagination path (`internal/registry/specs/marketplace.json:270`)

`mcp.list` uses offset pagination (`page`, `page_size` query params) but does
not declare `x-octo-pagination`. The CLI's `--page-all` engine only drives
cursor pagination, so MCP search cannot be auto-paginated. The skill docs at
`skills/octo-marketplace/mcp.md` note "Search is paginated" but do not show how
to advance pages.

Diff-scope: new (introduced by this PR). Users must manually increment `--page`
for large MCP searches. If the backend supports cursor pagination for MCPs,
adding `x-octo-pagination` would enable `--page-all`; otherwise, documenting
manual page advancement in `mcp.md` would help.

- The `attachOperation` change does not affect any existing service's command
  tree — `segs[0]` always equals `d.Service` for the 8 pre-existing services.
- `findChild` correctly searches `cur.Commands()` by `Name()`, so subcommand
  dedup works when multiple operations share a resource group prefix
  (e.g. `skill.list`, `skill.create`, `skill.get` all attach under one
  `skill` parent).
- The leaf command name uses `segs[len(segs)-1]` (the last segment), which is
  always the verb — correct for both `skill.list` → `list` and
  `skill_upload.create` → `create` (after underscore normalization → no
  collision with `skill.create` because they live under different parents:
  `skill-upload` vs `skill`).
- `x-octo-risk` declarations are consistent: `.list`/`.get`/`.download` =
  `read`, `.create`/`.update` = `write`, `.probe` = `read`.
- The `x-octo-binary-response` on `skill.download` correctly gates the
  `-o` / `--output` flag, and the skill docs describe SHA-256 verification
  plus path-traversal rejection for the install flow.

## Verdict: APPROVED

Solid, well-tested addition. The core `attachOperation` change correctly handles
the marketplace naming convention while preserving all existing service
behavior. Both P2 findings are spec-completeness items — not correctness or
security concerns — and can be addressed in follow-up.

yujiawei (dismissed):
# Code Review — PR #91 (octo-cli)

**Head SHA reviewed:** `2456b06ef0e25f31708fd8b6fc58d8ce783e0ba5` · merge-base `1e3acf6`
Build (`go build ./...`), `go vet`, and the full test suite (`cmd/service`, `internal/registry`, `skills`) all pass locally. `git diff --check` clean.

## 1. Spec compliance

**Spec: ✅**

The change delivers exactly what the summary describes and nothing out of scope:
- New embedded `marketplace` OpenAPI spec with 17 operations across two catalog domains (Skill + MCP), wired into the metadata-driven engine.
- `attachOperation` extended to preserve the operationId's first segment as a resource group when it differs from the service name, so `skill.list` and `mcp.list` both become addressable under `marketplace` instead of colliding as `marketplace list`.
- Underscore→dash normalization on intermediate and leaf segments, so `skill_category.list` → `marketplace skill-category list`.
- Agent-facing docs under `skills/octo-marketplace/` (SKILL.md router + skills.md + mcp.md), plus README/CLAUDE.md/loader doc updates and an embed test.

No missing, extra, or divergent behavior versus the stated scope. No linked plan applies (external-contributor PR), so this is a plain rubric review.

## 2. Code quality

**Quality: Approved**

Verified independently against the checked-out head:

- **`attachOperation` (`cmd/service/service.go:78-98`) — no regression.** Every existing multi-level operationId (`docs.*`, `matter.*`) has underscore-free intermediate segments, so the new `ReplaceAll(_, -)` is a verified no-op for them, and every existing spec's first operationId segment equals its service name, so `start` stays `1` for them. Only `marketplace` legitimately triggers `start=0`. Confirmed the generated tree by running the binary: `marketplace {skill,mcp,skill-category,mcp-category,skill-upload,skill-parse-task}` with correct leaves.
- **Pagination asymmetry is intentional and correct.** `skill.list` declares cursor `x-octo-pagination` and gets `--page-all`; `mcp.list` uses offset `page`/`page_size` and declares none, so it correctly gets no `--page-all`. Output-envelope splitting is independent of the extension, so an `{data:[],pagination:{}}` MCP response still surfaces as `.data` + `._pagination`. The docs are careful: `--page-all` is only referenced in `skills.md`, not `mcp.md`.
- **`x-octo-flag` + space-header contracts verified at runtime.** `skill download --response-format json` maps to `?format=json` (dry-run confirmed); space-header ops emit `X-Space-Id` only when a space is configured, matching the spec's `x-octo-space-header: true` and the new contract test.
- **Tests are meaningful**, asserting method/path/base-url/space-header per operation, request construction, visibility enums (`public/private/space` for skills, `public/private` for MCP), the foreign-domain tree shape, and embedding.

### Non-blocking suggestions (P2)

- **P2 — Skill install destination hardening (`skills/octo-marketplace/skills.md:38`).** Step 6 moves the archive to `<skills-root>/<skill-name>` where `skill-name` comes from remote metadata; steps 5 already reject unsafe *archive entry* paths, but not the directory name itself. Since the spec puts no pattern on `name`, a value like `../../x` could escape the Skills root. Suggest adding an explicit "reduce name to a safe basename and confirm the resolved path stays under the Skills root" step. This is agent guidance against a trusted first-party backend, so advisory, not blocking.
- **P2 — Non-injective underscore normalization (`cmd/service/service.go:84,97`).** `foo_bar.create` and `foo-bar.create` would normalize to the same cobra path. No embedded spec collides today; a duplicate-normalized-path guard in `RegisterServiceCommands` would fail loud if a future spec introduced one.
- **P2 — `--page-all` default cap (`cmd/service/run.go:228`, `skills.md:15`).** `--page-all` stops after 10 pages by default (`--page-limit`), merging silently with no truncation marker. This is pre-existing engine behavior (untouched here); the doc wording ("only when all results are needed") is reasonable, but a one-line mention of `--page-limit` would prevent an agent assuming completeness on 200+ result sets.
- **P2 — `--verbose` prints request bodies (`internal/client/client.go:220`, pre-existing, untouched).** `mcp probe`/`mcp create` carry secret-bearing `headers`/`env`, and `--verbose` logs the (truncated) request body. `mcp.md` already mandates secret placeholders and never persisting real secrets, which mitigates this for the documented flow. Worth a future redaction pass on nested secret fields, but out of scope for this PR and not introduced by it.

## 3. Overall verdict

**APPROVED** — Spec ✅ and Quality Approved. All findings are P2 advisories; none block merge. Clean, well-tested, and consistent with the existing metadata-driven engine.

kense-lab:
已 rebase 到最新 `main`，解决 `octo-html` 与 Marketplace registry 测试冲突。当前 head `65f8e4d`，`go test ./...` 与 `git diff --check` 均通过。

@Jerry-Xin @lml2468 @mochashanyao @yujiawei 请基于当前 head review。

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #91 (octo-cli)

Reviewed at head `300e8e3d`. Scope: new Marketplace OpenAPI spec (4234 lines),
command-tree registration logic for multi-domain services, agent skill
reference docs, and tests covering registry shape, request building, command
tree, and skill embedding.

## Summary

This PR adds the Marketplace service to the CLI's metadata-driven command
engine. A 26-operation OpenAPI 3.1 spec defines skill and MCP server
endpoints; the engine auto-generates the `marketplace skill …` and `marketplace
mcp …` command subtrees. The `attachOperation` function gains a conditional
first-segment skip: when the operationId's first segment differs from the
registry service name (e.g. `skill.list` under service `marketplace`), the
segment is preserved as a resource group instead of being discarded. Underscores
in operationId segments are now normalized to hyphens for CLI command names,
consistent with the existing flag-name derivation in `flags.go`. Agent skill
references (`SKILL.md`, `skills.md`, `mcp.md`) document search, install,
publish, update, and probe workflows.

Overall the implementation is well-structured: the command-tree change is
minimal and backwards-compatible with existing services, tests cover registry
shape, request building, command tree, visibility enums, and skill embedding.
Two non-blocking P2 items below.

## Verification

- ✅ **Backward compatibility** — verified that no existing spec (matter, message, group, thread, file, bot, event, docs) contains an underscore in any operationId. The `_`→`-` replacement in `attachOperation` is a no-op for all existing services. The `start` index change is also a no-op because all existing operationIds begin with their service name (`matter.create`, `message.send`, etc.), so `segs[0] == d.Service` holds and `start = 1` as before.
- ✅ **Command tree correctness** — traced all 26 marketplace operationIds through `attachOperation` logic. Each produces the expected CLI path (e.g. `skill_upload.create` → `marketplace skill-upload create`, `mcp_category.list` → `marketplace mcp-category list`). The `TestMarketplaceCommandTree` test covers all 8 domain groups and their children.
- ✅ **Spec integrity** — `marketplace.json` is valid JSON, declares `x-octo-service: marketplace`, `x-octo-base-url: OCTO_API_BASE_URL`, `x-octo-space-header: true`. `TestMarketplaceRegistryShape` validates all 26 operations' method, path, base URL env, and space header.
- ✅ **Test coverage** — 303 lines of marketplace-specific tests plus extensions to `loader_test.go` and `service_test.go`. The `TestMarketplaceVisibilityEnumsMatchBackend` test catches enum drift between spec and backend. The `TestMarketplaceSkillTagsAreStrings` test validates the tags schema. The `TestOctoMarketplacePublishFlowChecksOwnedNameBeforeMutation` doc-contract test enforces owned-name lookup ordering before upload initialization.
- ✅ **Embed wiring** — `//go:embed specs/*.json` in `loader.go` is unchanged; the new `marketplace.json` is picked up automatically. `TestNewLoadsAllServices` includes "marketplace" in the expected service list.

Static analysis only at head `300e8e3de1984520932a9f86083ac361e050f935`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Two P2 items below.

### P2 — MCP listing lacks `--page-all` auto-pagination (`internal/registry/specs/marketplace.json:326`)

`mcp.list` and `mcp.mine.list` use offset-based pagination (`page`/`page_size`
query parameters, `OffsetList` response schema) but the spec does not declare
`x-octo-pagination` for either operation. The CLI's auto-pagination engine
(`runPaginated` in `cmd/service/run.go`) requires this extension to wire up
`--page-all` and `--page-limit`. By contrast, `skill.list` and `skill.mine.list`
declare cursor-based pagination and get `--page-all` for free.

Users listing many MCPs will only see the first page (default 20) unless they
manually pass `--page 2 --page-size 100`. The spec description at `:327`
explicitly says "offset pagination," so this appears intentional — the CLI's
pagination engine currently only supports cursor-based traversal. Consider
either extending the engine to support offset-based `--page-all` or calling
out the limitation in the skill reference docs so agents don't assume parity
with skill listing.

### P2 — SKILL.md groups `mcp list` with cursor-paginated commands (`skills/octo-marketplace/SKILL.md:28`)

The shared output rule states "Do not apply that expression to paginated search
commands: `skill list` and `mcp list` expose their result array directly as CLI
`.data` and pagination as `._pagination`." While the output shape description
is accurate, grouping them together implies both support the same auto-
pagination capability. An agent following this guidance would attempt `--page-all`
on `mcp list` and find the flag absent. Clarify that MCP listing uses manual
offset pagination (`--page N --page-size N`) and does not support `--page-all`.

- `attachOperation` start-index logic: correctly distinguishes `segs[0] == d.Service` (existing services, start=1) from `segs[0] != d.Service` (marketplace, start=0). No collision or command-tree corruption possible.
- Underscore-to-hyphen normalization in both intermediate segments and leaf verb: consistent with `paramFlagName` in `cmd/service/flags.go:115`.
- `TestMarketplaceRegistryShape` validates operation count (26), method/path for each, base URL env, and space header — strong regression net.
- `TestMarketplaceCommandTree` validates the exact domain count (8) and child commands per domain — catches any registration regression.
- `TestMarketplaceVisibilityEnumsMatchBackend` catches enum drift between skill (public/private/space) and MCP (public/private) specs.
- Agent skill docs enforce safe publish flow ordering: owned-name lookup before upload, confirmation before mutation, `mktemp -d` staging, SHA-256 verification on install. The `TestOctoMarketplacePublishFlowChecksOwnedNameBeforeMutation` doc-contract test validates ordering.
- `TestRegisterServiceCommands_PreservesForeignOperationDomain` directly tests the new `start` logic for the marketplace service.

## Verdict: APPROVED

No correctness, security, or build-breaking issues. The command-tree logic
change is minimal, backwards-compatible, and well-tested. The two P2 items
are non-blocking documentation/spec gaps around MCP offset pagination.





## Data Flow Trace

- **`d.ID` (operationId)** — sourced from `marketplace.json` `operationId` field per path/method. Loaded by `loader.go:walkOperations` → `buildDetail` → `OperationInfo.ID`. Consumed by `attachOperation` via `d.ID` after `reg.GetOperation(info.ID)` in `service.go:38`. Data flows correctly from spec → registry → command tree. ✅
- **`d.Service`** — sourced from `x-octo-service` extension in `marketplace.json` (`"marketplace"`), falling back to filename. Set in `buildDetail` at `loader.go`. Consumed in `attachOperation` for the `start` index comparison. Value is always `"marketplace"` for this spec. ✅
- **`segs` derivation** — `strings.Split(d.ID, ".")` on operationIds like `skill.list` → `["skill", "list"]`, `mcp_category.list` → `["mcp_category", "list"]`. Correctly split, no edge cases with empty segments or trailing dots in the spec. ✅
- **`name` (command segment)** — derived from `segs[i]` with `_` → `-` replacement. Flows to `cobra.Command.Use` and `findChild` lookup. `findChild` does exact `Name()` match, so `skill-upload` and `skill` are distinct. ✅
- **`verb` (leaf command)** — derived from `segs[len(segs)-1]` with `_` → `-` replacement. Passed to `buildOperationCmd` as display name only; actual routing uses `d.Path` and `d.Method`. No downstream data corruption. ✅

## Blind Spot Checklist

- **C1 — Dual-path parity**: `attachOperation` modifies one symmetric path (command registration). No paired add/remove or subscribe/unsubscribe logic. N/A.
- **C2 — Control-flow ordering / nesting**: `attachOperation` is called once per operation in a flat loop in `RegisterServiceCommands`. No nested or recursive calls. The `start` index and `_` → `-` replacement are idempotent. N/A.
- **C3 — Authorization boundary**: This PR adds CLI command registration only; no auth/permission changes. All operations inherit existing bot-profile auth. N/A.
- **C4 — Authorization lifecycle**: N/A — no auth changes.
- **C5 — Build/note parity**: `//go:embed specs/*.json` glob picks up `marketplace.json` automatically. No build config changes needed. The JSON is validated by `json.Unmarshal` at startup; parse errors would surface as a `MustNew()` panic. ✅
- **C6 — Governance/docs self-consistency**: SKILL.md references `skills.md` and `mcp.md`; both are embedded and tested (`TestOctoMarketplaceReferencesEmbedded`). The P2 finding above notes the pagination doc inconsistency. ⚠️

---

[Octo-Q] verdict: APPROVE — No P0/P1 issues. The command-tree logic change is minimal, backwards-compatible with all existing services (verified: no existing operationId contains underscores or starts with a non-service-name segment), and covered by comprehensive tests. Two P2 items around MCP offset-pagination documentation gaps are non-blocking.

Jerry-Xin (dismissed):
The PR is relevant to octo-cli, but array-valued Marketplace filters are generated incorrectly.

### 🔴 Blocking

- 🔴 Critical — Repeatable query filters silently discard values.

  [flags.go](/Users/mlamp/.openclaw/tmp/pr-review-n91uxra6/repo/cmd/service/flags.go:71) determines flag type solely from `p.Type`. Since `schemaTypeKind` does not handle `"array"`, Marketplace array parameters become scalar `StringVar` flags rather than `StringSliceVar` flags.

  This affects `transport`, `visibility`, `source`, `created-by-type`, and `tag`. Repeating a flag overwrites earlier values instead of producing repeated query parameters, despite the OpenAPI contract and [mcp.md](/Users/mlamp/.openclaw/tmp/pr-review-n91uxra6/repo/skills/octo-marketplace/mcp.md:20) explicitly advertising repeatable filters. Skill `tag` parameters also specifically require repetition.

  Preserve parameter item information in the registry and map arrays of strings to `kindStringSlice`. Add a request test using at least two values and assert both appear in `r.URL.Query()[name]`.

### ✅ Highlights

- Marketplace operations are appropriately grouped under distinct Skill and MCP command families.
- Space-header behavior and visibility enums have focused coverage.
- Embedded progressive-disclosure references are tested.
- `go test ./...` and `git diff --check` pass.

Jerry-Xin (dismissed):
The PR is relevant to octo-cli, but array-valued Marketplace filters are generated incorrectly.

### 🔴 Blocking

- 🔴 Critical — Repeatable query filters silently discard values.

  `cmd/service/flags.go:71` determines flag type solely from `p.Type` via `schemaTypeKind`. Since `schemaTypeKind` does not handle `"array"`, Marketplace array query parameters fall through to a scalar `StringVar` flag rather than a `StringSliceVar` flag. Note the body-property path already does this correctly (`promotableKind` at `flags.go:186` maps `array` → `kindStringSlice`), so the query-side path is inconsistent.

  This affects `transport`, `visibility`, `source`, `created-by-type`, and `tag`. Repeating a flag overwrites earlier values instead of producing repeated query parameters, despite the OpenAPI contract and `skills/octo-marketplace/mcp.md:20` explicitly advertising repeatable filters. Skill `tag` parameters also specifically require repetition.

  Fix: preserve parameter item information in the registry and map arrays of strings to `kindStringSlice` in the query-flag builder. Add a request test that passes at least two values for one of these flags and asserts both appear in `r.URL.Query()[name]`.

### ✅ Highlights

- Marketplace operations are appropriately grouped under distinct Skill and MCP command families.
- Space-header behavior and visibility enums have focused coverage.
- Embedded progressive-disclosure references are tested.
- `go test ./...` and `git diff --check` pass.

yujiawei (changes requested):
# Code Review — PR #91 (octo-cli)

Reviewed at head SHA `300e8e3`, diff scoped to merge-base `1e3acf6..300e8e3`. Build, `go vet`, and the full `go test ./...` suite pass locally; `git diff --check` is clean. This review draws on two independent automated reviewer passes plus manual verification of the load-bearing findings.

The design is sound: keeping the whole marketplace surface in the OpenAPI spec and letting the metadata engine generate commands is exactly the right pattern for this codebase, and the `attachOperation` change to preserve a foreign first segment (`skill`/`mcp` under service `marketplace`) is well-reasoned and correctly isolated. The problems below are in how the generic engine handles two OpenAPI shapes this spec is the **first** to use — array query params and a redirect-download without the binary flag — so they slipped through because no prior spec exercised those paths.

## 1. Spec compliance

The generated command tree matches the spec: 26 operations, correct service (`marketplace`), `OCTO_API_BASE_URL`, and space-header contract; visibility enums and string-array `tags` line up with the backend per the new tests. No missing or out-of-scope commands. **Spec compliance: OK.**

## 2. Correctness

### P1 — Multi-value array query filters are silently dropped

`marketplace.json` is the first spec to declare `type: array` query parameters (`tag`, `transport`, `visibility`, `source`, `created_by_type` on `mcp list` / `mcp mine list`, and `tag` on `skill list`). But `schemaTypeKind` (`cmd/service/flags.go:249`) has no `array` case, so it falls through to `kindString`. Every array query param therefore registers as a scalar `StringVar`, and the `kindStringSlice` branch in `registerQueryFlags` (`cmd/service/flags.go:91`) is dead code for query params.

Observed behavior (verified with a throwaway test against the generated command):

- `mcp list --tag cli --tag devops` → sends `tag=devops` only (the first value is silently discarded).
- `mcp list --tag cli,devops` → sends `tag=cli%2Cdevops`, one literal joined string, not two values.

`mcp.md:20` explicitly documents these filters as "repeatable/comma-separated", and `marketplace_test.go` only ever passes a single `--tag`, so the gap isn't caught. This is silent wrong behavior on a documented feature (filtered search quietly returns the wrong result set), which is why I'm rating it blocking rather than advisory.

Fix is one case, mirroring the body-side `promotableKind` (`cmd/service/flags.go:243`):

```go
func schemaTypeKind(t string) valueKind {
	switch t {
	case "integer", "number":
		return kindInt
	case "boolean":
		return kindBool
	case "array":
		return kindStringSlice
	}
	return kindString
}
```

Please add a test that passes a repeated `--tag` and asserts both values reach the query string, so this can't regress.

### P1 — Bare `marketplace skill download <id>` returns `HTTP_302` instead of the artifact

`skill.download` (`marketplace.json:2109`) defaults to a `302` redirect to the artifact and only returns JSON when `format=json` is passed. The operation does **not** set `x-octo-binary-response`, so the client does not opt into redirect handling. In `internal/client/client.go:253`, a 3xx without `binaryResp` is treated as an error (`unexpected redirect to …`). So a user who runs `marketplace skill download <id>` without `--response-format json` gets a confusing `HTTP_302` error.

The documented install flow in `skills.md` always uses `--response-format json`, and that path is tested and works, so this is not data loss — but the bare command is a discoverability footgun (the natural first thing to type fails cryptically). Two acceptable fixes: set `x-octo-binary-response: true` on `skill.download` so the redirect Location is surfaced as an envelope (matching `file.download`'s handling), or make the flag required / default it to `json`. If you choose to leave it, please make the SKILL docs state that `--response-format json` is mandatory for this command.

### P2 — Required request bodies / properties are not enforced client-side

The loader discards operation-level `requestBody.required` and property-level `required` only appends "(required)" to help text (`internal/registry/loader.go:366`, `cmd/service/flags.go:208`); an empty invocation resolves to a nil body (`cmd/service/run.go`). So `skill create` (requires `parse_task_id`), `skill publish` (requires `skill_upload_id`), `skill-upload create` (requires `file_name`/`file_size`), etc. reach the backend with missing mandatory fields and fail with a round-trip validation error instead of a fast local one. This is pre-existing engine behavior, not introduced here, and the backend does reject — so it's advisory. Worth a follow-up to mark required promoted fields with `MarkFlagRequired`, consistent with how query/header required flags are already handled.

### P2 — MCP install assumes `quick_start.slug` is always present

`mcp.md:41` unconditionally uses `quick_start.slug` as the `mcpServers` key. Per the spec the schema also exposes `server_name`, and `slug` may be absent on older listings. Recommend a documented fallback (e.g. `slug` → `server_name`) so a legacy record doesn't produce an empty config key. Docs-only change.

### P2 — Underscore→dash rewrite is a (benign here) behavior shift for leaf/group names

`strings.ReplaceAll(_, -)` now applies to all intermediate and leaf command segments (`cmd/service/service.go:84,97`), not just marketplace. I verified that no existing (non-marketplace) operationId contains an underscore in any segment, so there is no rename of any current command — but it's worth a note that any future spec relying on an underscore in a command name would now be silently hyphenated. No action needed for this PR.

## Coverage notes

- Verified: command-tree shape and normalized-path uniqueness (no collisions/shadowing from the `start=0/1` branch or the dash rewrite), space-header + base-URL contract, enum/tags schema, the two P1s reproduced against the generated commands, `go build` / `go vet` / `go test ./...` green, `git diff --check` clean.
- Not verified (out of local reach): live backend acceptance of the generated request shapes; the full `$ref` integrity of the 4234-line generated spec beyond the operations cited; the real publish/download/MCP flows against a running marketplace (the PR description states these were run manually by the author).
- Note: the PR is currently `CONFLICTING` with `main` — the conflict is in the `CLAUDE.md` / `README.md` skills list only (a parallel `octo-html` change touched the same lines), not in any Go code. Please rebase before merge.

## Verdict

Two blocking issues: array query filters are silently dropped (wrong search results on a documented feature) and bare `skill download` fails with a cryptic redirect error. Both are small, well-scoped fixes. Requesting changes; happy to re-review promptly once the array-flag fix (with a repeated-value test) and the download path are addressed.

kense-lab:
Block 已修复，当前 head `f36a7a3`：

- ParamInfo 保留 array items，string array query 注册为 StringSlice
- 双 `--tag cli --tag devops` 测试确认生成两个 query value
- Skill 下载文档明确 `--response-format json` 必须使用，避免 302 误解
- `go test ./...`、`git diff --check` 通过

@Jerry-Xin @yujiawei 请基于当前 head 复审。

kense-lab:
@lml2468 @mochashanyao 当前 head `f36a7a3` 尚无有效 review；两个 Block 已修复且全量测试通过，请检查当前 diff。

yujiawei (approved):
# Code Review — PR #91 (octo-cli)

**Head SHA reviewed:** `f36a7a3e045ff8cf8ba7cfcfee3b99f955a5b270` · merge-base `640e71a`

Re-review after the two prior P1 blockers were reported fixed. `go build ./...`, `go vet ./...`, and the full `go test ./... -count=1` suite all pass locally; `git diff --check` is clean. This review draws on two independent automated passes plus manual verification of the load-bearing findings against the checked-out head.

## 1. Spec compliance

**Spec: ✅**

The change delivers exactly the stated scope and nothing out of scope:

- New embedded `marketplace` OpenAPI spec (26 operations across Skill + MCP resource families), wired into the metadata-driven engine.
- `attachOperation` extended to preserve the operationId's first segment as a resource group when it differs from the service name, so `skill.list` and `mcp.list` become addressable under `marketplace` instead of colliding as `marketplace list`.
- Underscore→dash normalization on intermediate and leaf segments (`skill_upload.create` → `marketplace skill-upload create`).
- Agent-facing docs under `skills/octo-marketplace/` (SKILL.md router + skills.md + mcp.md), README/CLAUDE.md/loader doc updates, and embed tests.

No missing, extra, or divergent behavior versus the summary.

## 2. Code quality

**Quality: Approved**

### Both prior P1 blockers are fixed and verified end-to-end

**P1 (array query filters silently dropped) — RESOLVED.** The fix threads through the whole pipeline correctly:
- `internal/registry/loader.go` — `ParamInfo` gains an `Items *SchemaInfo`, and `buildDetail` resolves the schema `items` (nil-safe: only set when `sch["items"]` is a map).
- `cmd/service/flags.go` — new `queryParamKind(p)` returns `kindStringSlice` for `type: array` with string items, registering a `StringSliceVar`.
- `cmd/service/run.go:107-109` — `buildQuery` emits `q.Add(name, v)` per value, producing repeated query params.

Verified by dry-run: `marketplace mcp list --tag cli --tag devops` now sends `?tag=cli&tag=devops` (both values reach the wire). `TestMarketplaceMCPSearchFiltersRequest` asserts `--tag cli --tag devops` → `tag: {cli, devops}` and covers `transport/visibility/source/created_by_type` as arrays. I enumerated every array query param in the spec — all 13 are `items.type: string`, so all are covered.

**P1 (bare `skill download <id>` returns HTTP_302) — RESOLVED via documentation.** `skills.md:34-36` now states `--response-format json` is mandatory and explains the client intentionally does not follow the artifact redirect. This was one of the acceptable fixes called out in the prior review; the documented install flow uses the flag, is tested, and works.

### Other verified points

- **`attachOperation` — no regression.** For every existing spec, the first operationId segment equals the service name (`start=1`, prior behavior preserved) and no intermediate/leaf segment contains an underscore (the `ReplaceAll` is a verified no-op). Only `marketplace` triggers `start=0`. I generated the full tree from the built binary: `marketplace {skill,mcp,skill-category,mcp-category,skill-upload,skill-icon-upload,mcp-icon-upload,skill-parse-task}` with correct leaves, and confirmed all 26 normalized command paths are unique (no collisions/shadowing).
- **Install-flow safety** (SHA-256 match, `..`/absolute/symlink/device rejection, atomic move, never-execute-archive-scripts) is specified in `skills.md` for the agent runtime to honor. No extraction code ships in this PR, so this is the correct layer; flagging that it remains doc-enforced, not compiler-enforced.

### Non-blocking findings (P2)

- **P2 — Non-string array query params would silently fall back to scalar.** `queryParamKind` (`cmd/service/flags.go:249`) only maps `array` → `kindStringSlice` when `Items.Type == "string"`. An `integer`/`boolean` array, or an array whose `items` is a `$ref`, would fall through to `kindString`. The marketplace spec has no such shape today (all 13 array params are inline string arrays), so this is fail-closed with no live impact — but it's a latent gap in the generic engine worth a follow-up if a future spec needs it.
- **P2 — Underscore→dash normalization is lossy and unvalidated.** `strings.ReplaceAll(_, -)` at `service.go:83,97` means two distinct operationIds like `foo_bar.list` and `foo-bar.list` would converge to the same command. No collision exists in any current spec (verified across all embedded specs), but there is no guard against a future one.
- **P2 — `skill download` redirect footgun is doc-enforced only.** If a caller omits `--response-format json`, the redirect path in `internal/client/client.go:253` surfaces the signed `Location` in the error message (`unexpected redirect to "..."`), which is in tension with the docs' "never log the short-lived URL" rule. Note this code is pre-existing (not introduced by this PR) and only triggers on the documented-misuse path. A stronger fix would default/require the flag in the spec and redact the URL in the redirect error, but that is not a merge blocker here.
- **P2 — Required promoted body fields are not enforced client-side.** Pre-existing engine behavior: `skill create` / `skill publish` / `skill-upload create` reach the backend with missing mandatory fields and fail on a round-trip instead of a fast local error. Advisory; consider `MarkFlagRequired` on required promoted fields in a follow-up.

## 3. Overall verdict

**APPROVE**

Spec ✅ AND Quality Approved. Both previously blocking P1 issues are fixed and verified end-to-end (array filters send repeated params; download flow documented). The remaining items are P2 advisories about latent gaps in the generic engine with no impact on the shipped marketplace surface.

## 4. Suggestions

- Extend `queryParamKind` to cover `integer`/`boolean`/`$ref` array items (map to slice kinds) so the next multi-value spec doesn't hit the same class of bug the array-string fix just closed.
- Add a spec-load-time uniqueness check on normalized command paths so an accidental `_` vs `-` operationId collision fails loud at startup rather than silently shadowing a command.
- Consider making `skill.download` default to `format=json` (or redact the `Location` in the redirect error) so the natural bare invocation doesn't emit a signed URL.

## 5. Additional notes

- Test coverage is strong and meaningful: request construction (method/path/base-url/space-header per op), visibility enum drift vs backend, string-array tags schema, the foreign-domain tree shape, and skill embedding are all pinned. The new `TestMarketplaceMCPSearchFiltersRequest` specifically guards the array-filter regression.

kense-lab:
@Jerry-Xin 你在 `65f8e4d` 提出的 Block 已于当前 head `f36a7a3` 修复：ParamInfo 保留 array items、string array 使用 StringSlice、双 `--tag` 请求测试验证两个值均到达 URL。请基于当前 head 复审并更新 review 状态。

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #91 (octo-cli)
**PR:** Mininglamp-OSS/octo-cli#91 — feat(marketplace): add OpenAPI-driven skill and MCP commands
**Head SHA:** `f36a7a3e045ff8cf8ba7cfcfee3b99f955a5b270`
**Scope:** 7 changed files (+4629/−29), 2 new skill docs, 1 new spec (4234 lines)

---

## Summary

This PR adds the **marketplace** service to the octo-cli OpenAPI spec registry, exposing 26 operations across two catalog domains (Skill and MCP) via the `octo-cli marketplace` command tree. Key engineering changes:

1. **Foreign operationId routing** (`service.go`): `attachOperation` now preserves the first segment of operationIds whose prefix differs from the service name (e.g., `skill.list` → `marketplace skill list`).
2. **Array query parameter support** (`loader.go` + `flags.go`): `ParamInfo` gains an `Items` field; `queryParamKind` maps `array`+`string-items` to cobra `StringSliceVar`, enabling `--tag cli --tag devops`.
3. **Underscore→dash normalization** (`service.go`): Intermediate command segments convert `_` to `-` for CLI-friendly names (e.g., `mcp_category` → `mcp-category`).
4. **Comprehensive test coverage** (4 test files, 372 new lines) covering registry shape, request building, command tree, tag handling, visibility enums, download flow, and foreign-operationId routing.
5. **Agent-facing skill docs** (`SKILL.md`, `skills.md`, `mcp.md`) with progressive-disclosure, safety rules, and SHA-256 verification workflows.

---

## Verification

| Area | Status | Evidence |
|---|---|---|
| Foreign operationId routing | ✅ Correct | `service.go:78-80` — `start=0` when `segs[0] != d.Service`; grep confirms only marketplace has foreign operationIds |
| Backward compatibility (existing specs) | ✅ Safe | All 9 existing specs have operationIds matching service name; `start=1` path unchanged |
| Underscore→dash conversion | ✅ Safe | 7 marketplace operationIds have underscores; no existing specs affected; `findChild` deduplicates intermediate groups |
| Array query param wiring | ✅ Correct | `loader.go:359-361` resolves `schema.items` via `resolveSchema`; `flags.go:256-258` maps to `kindStringSlice`; `run.go:107-109` uses `q.Add` for repeated params |
| Pagination declarations | ✅ Correct | Cursor pagination on `skill.list` and `skill.mine.list` only; `mcp.list` correctly uses offset pagination without `x-octo-pagination` |
| Spec embed | ✅ Correct | `skills.go` `//go:embed */*.md` pattern picks up `octo-marketplace/*.md`; `TestOctoMarketplaceReferencesEmbedded` verifies |
| Test coverage | ✅ Comprehensive | 26 operations verified in registry shape test; request/response tests cover path params, query params, headers, body; command tree test validates domain structure |

---

## Findings

No P0/P1 issues; two P2 items and one nit below.

### P2 — Global underscore→dash conversion in `attachOperation`

**File:** `cmd/service/service.go:82`

The `strings.ReplaceAll(segs[i], "_", "-")` conversion is now applied globally to every spec's operationId segments, not just marketplace. Grep confirms no existing specs (bot, docs, event, file, group, html, matter, message, thread) have underscores in operationIds, so this is backward-compatible today. Consider adding a loader-level assertion or registry validation that rejects operationIds with unexpected underscore patterns, so a future spec author gets a clear error at build time rather than a silent command-name change at runtime.

**Diff-scope:** New behavior introduced by this PR. Affects all specs going forward.
**Severity:** P2 — no current breakage, but future-spec footgun.

### P2 — `mcp.list` offset pagination not documented as different from skill cursor pagination

**File:** `internal/registry/specs/marketplace.json:326` / `skills/octo-marketplace/mcp.md`

`mcp.list` uses offset pagination (`page`/`page_size` query params, `OffsetList` response) and correctly omits `x-octo-pagination`, so `--page-all` is not registered. The agent-facing docs (`mcp.md`) correctly avoid mentioning `--page-all`, but an agent familiar with the skill list docs (which use `--page-all`) might try it and get a cobra unknown-flag error. Consider a one-line note in `mcp.md`: "MCP listings use offset pagination; `--page-all` is not available."

**Diff-scope:** New content introduced by this PR.
**Severity:** P2 — doc gap, not a code defect.

### Nit — Array query params with non-string items silently fall back to string

**File:** `cmd/service/flags.go:256`

When `type: "array"` but `Items` is nil or `Items.Type` is not `"string"`, the function falls through to `kindString`. This means array-of-integer or array-of-object query params would silently become single-string flags. This is safe today (all marketplace array params have `items.type: "string"`), but consider logging a warning or returning an error kind for unsupported array item types to surface spec authoring mistakes early.

**Diff-scope:** New code path in this PR.
**Severity:** Nit — no current impact.

---

## Data-Flow Tracing

| Consumed datum | Source | Traced path | Reaches consumer? |
|---|---|---|---|
| `ParamInfo.Items` | `loader.go:359` — `sch["items"].(map[string]any)` inside `buildDetail` | `resolveSchema(doc, items)` → `pi.Items = &resolved` → `queryParamKind(p)` checks `p.Items.Type == "string"` | ✅ Yes — flows to `StringSliceVar` binding at `flags.go:92-93`, then `q.Add` at `run.go:108` |
| `d.Service` for foreign check | `loader.go:429` — `stringOf(info["x-octo-service"])` | `Registry.ListOperations` → `OperationInfo.Service` → `attachOperation` compares `segs[0] != d.Service` | ✅ Yes — marketplace spec declares `"marketplace"`; `skill.list` first segment `"skill"` ≠ `"marketplace"` → `start=0` |
| Underscore→dash in leaf name | `service.go:97` — `strings.ReplaceAll(segs[len(segs)-1], "_", "-")` | Passed to `buildOperationCmd` as `verb` parameter → becomes cobra `Use` string | ✅ Yes — `skill_upload.parse` → leaf `parse` under intermediate `skill-upload` |
| Cursor pagination config | `marketplace.json:1729` — `x-octo-pagination` on `skill.list` | `loader.go` parses into `OperationDetail.Pagination` → `service.go:153-157` registers `--page-all` and `--page-limit` flags | ✅ Yes — `skill list --page-all` works; `mcp list` has no pagination extension so no `--page-all` |
| Enum values for visibility | `marketplace.json:3955` — `["public","private","space"]` on `CreateRequest.visibility` | `loader.go` parses `sch["enum"]` → `ParamInfo.Enum` → `flags.go:74-76` appends enum hint to flag description | ✅ Yes — `--visibility` flag shows "(one of: public, private, space)" |

---

## Blind-Spot Checklist

| Check | Result | Evidence |
|---|---|---|
| **C1 — Dual-path parity** | N/A | No paired add/remove or subscribe/unsubscribe operations in this PR |
| **C2 — Control-flow ordering / nesting** | Clear | `attachOperation` is called once per operation in `registerServiceCommands`; `findChild` deduplicates intermediate groups so `skill.list` and `skill.create` share the `skill` node without collision |
| **C3 — Authorization boundary** | N/A | No new endpoints or auth middleware; spec declares `x-octo-risk` and `security` per operation, consumed by existing auth flow |
| **C4 — Auth lifecycle / container-member cascade** | N/A | No auth changes |
| **C5 — Build/runtime parity** | Clear | `go:embed` pattern verified; `TestOctoMarketplaceReferencesEmbedded` asserts `skills.md` and `mcp.md` are embedded; spec is embedded via `//go:embed specs/*.json` in loader |
| **C6 — Governance/doc self-consistency** | Clear | `SKILL.md` routing table correctly points to `skills.md` and `mcp.md`; safety rules (confirmation before mutation, no secret printing) are consistent across all three docs |

---

## Cross-Round Blocker Recheck

---

## Extra Observations

- The `skill.mine.list` operationId correctly resolves to `marketplace skill mine list` (three levels: skill → mine → list). The `findChild` mechanism reuses the `skill` intermediate group created by `skill.list`, `skill.create`, etc.
- The `mcp.mine.list` operationId similarly becomes `marketplace mcp mine list`, sharing the `mcp` node with `mcp.list`, `mcp.create`, etc.
- All 26 operations declare `x-octo-space-header: true` (via service-level declaration) and `x-octo-base-url: "OCTO_API_BASE_URL"`, consistent with the existing spec contract.
- The spec includes `x-octo-risk: "write"` on mutating operations and `"read"` on read-only operations, correctly consumed by the existing risk-aware confirmation flow.

---

[Octo-Q] verdict: APPROVE — No P0/P1 blockers found. The PR cleanly extends the OpenAPI-driven CLI architecture with well-tested marketplace support. Two P2 suggestions (registry validation for underscore patterns, offset-pagination doc note) and one nit (array item type fallback) are non-blocking.

lml2468 (approved):
## Review — head `f36a7a3e045f`

Reviewed from a local checkout at this head: ran `go build ./...`, `go vet ./...` (incl. `-tags integration`), and `go test ./cmd/service ./internal/registry ./skills` — all green. Verified the two 🔴 raised at the prior head end-to-end.

### Gate 1 — Spec compliance ✅
Change stays within the stated intent (OpenAPI-driven Skill/MCP domain + CLI generator support for a multi-domain service). The `service.go`/`loader.go` multi-domain change is a **no-op for every existing spec** (each existing `x-octo-service` equals its operationId first segment), so it introduces no regression to `docs`/`matter`/etc.

### Gate 2 — Code quality ✅ (one P2 note)

**🔴 #1 — array filters collapsed to scalar (repeated `--tag` lost): FIXED + regression-locked ✅**
- `loader.go` adds `ParamInfo.Items`; `flags.go`'s new `queryParamKind()` maps `type:array` + `items.type:string` → `kindStringSlice`; `run.go` emits via `q.Add()` (repeat) instead of `q.Set()` (overwrite).
- `TestMarketplaceMCPSearchFiltersRequest` asserts `--tag cli --tag devops` → `tag:[cli,devops]`. Confirmed passing.

**P2 (MEDIUM) — bare `skill.download` returns 302 → surfaced as an error, and the presigned URL is echoed into the error text**
- `skill.download` does **not** set `x-octo-binary-response` (unlike `file.download`, which does). A bare `marketplace skill download <id>` (without `--response-format json`) hits a 302; the client's non-binary path returns:
  `HTTP_302: unexpected redirect to "https://…/artifact.zip?sig=<token>"`
- Two consequences: (1) a confusing "unexpected redirect" error on an otherwise valid command; (2) the signed redirect URL is printed into the error message — which conflicts with this skill's own rule "Never print presigned URLs, tokens".
- **Why non-blocking:** the documented primary path — `skill download <id> --response-format json` — returns 200 with `{download_url, file_sha256}` and is covered by `TestMarketplaceDownloadRequest` (passing). The core agent flow is unaffected; the leaked URL is the caller's own short-lived URL in their own terminal.
- **Suggested fix (pre-merge, ~1 line):** set `x-octo-binary-response: true` on `skill.download` (align with `file.download`) so the 302 `Location` is wrapped into a JSON envelope instead of raising/echoing, plus a small test. Alternatively, document that the bare form is unsupported and require `--response-format json`.

**Other checks (all clean):** 26 operations, command tree has no collisions after `_`→`-` normalization; `TestMarketplaceRegistryShape` enumerates all 26 (method/path/base-url/space-header); destructive `skill delete` / `mcp delete` are `ExactArgs(1)` (no positional-arg misuse — the repo's historical defect class does not recur); `marketplace.json` is valid (20 paths / 73 schemas, all `$ref`s resolve). `mcp.list` uses offset pagination (no `x-octo-pagination`) consistent with the backend and documented as manual paging — non-blocking.

### Coverage / blind spots
- **Verified:** build, full changed-pkg tests, real-binary command tree/flags/args/schema loading, both 🔴 claims, backward-compat of the generator change.
- **Not verified (needs CI / human):** DB/network-backed integration (upload/download/parse-poll/delete against the live backend); server-side behavior described in the spec is out of this PR's scope.

### Verdict: **APPROVE** (with one P2 to clean up before merge)
Both gates pass in substance: the more serious 🔴 (silent filter-value loss) is confirmed fixed and regression-locked, and the feature is mature and end-to-end verified. The bare `skill.download` 302 is a real, PR-introduced P2 rough edge (not data loss; the documented path is safe and tested) — strongly recommended to add `x-octo-binary-response: true` before merge, but it does not block.

Jerry-Xin (approved):
APPROVE. The prior blocking issue is fixed and regression-locked; one non-blocking P2 remains.

### ✅ Prior 🔴 resolved — repeatable array query filters
- `internal/registry/loader.go`: `ParamInfo` now carries `Items`, and query-parameter `items` are parsed.
- `cmd/service/flags.go`: new `queryParamKind()` maps `type:array` + `items.type:string` to `kindStringSlice` (→ `StringSliceVar`).
- `cmd/service/run.go`: `kindStringSlice` now uses `q.Add()` (append) instead of `q.Set()` (overwrite).
- Regression lock: `TestMarketplaceMCPSearchFiltersRequest` asserts `--tag cli --tag devops` → `tag: [cli, devops]` (both values preserved). Verified fixed at this head.

### 🟡 Non-blocking (P2) — bare `skill.download` 302 handling
- `skill.download` (`GET /market/api/v1/skills/{skill_id}/download`) does not set `x-octo-binary-response` (unlike `file.download`). A bare invocation returning a 302 hits the "unexpected redirect" error path, and the signed redirect URL gets echoed into the error message — which conflicts with the skill's own guidance not to print presigned URLs.
- Severity P2 (not merge-blocking): the documented path `skill download <id> --response-format json` returns 200 + `{download_url, file_sha256}` and is covered by `TestMarketplaceDownloadRequest`; core agent flows are unaffected, and the leaked URL is short-lived and only visible to the caller in their own terminal.
- Suggested fix (fast-follow): add `x-octo-binary-response: true` to `skill.download` (align with `file.download`) + a test, so the 302 Location is wrapped in a JSON envelope instead of surfacing as an error.

### ✅ Highlights
- Marketplace operations grouped cleanly under distinct Skill and MCP command families; multi-domain operation IDs preserved.
- 26 operations with no command-tree collisions; destructive `delete` ops use `ExactArgs(1)`.
- Spec is valid (paths + schemas, refs resolve); strong test coverage for request construction, visibility rules, Space headers, and embedded skill references.
- `go test ./...` and `git diff --check` pass.

PR #105, opened 2026-07-24
fix(html): add docs-html gateway prefix

Ranwanglc:
## Summary

Routes every `octo-cli html` operation through the dedicated Docs HTML gateway prefix:

- changes all registered HTML API paths from `/v1/*` to `/docs-html/v1/*`
- updates the bundled `octo-html` skill to document the same prefix
- adds a registry contract test covering every registered HTML operation

The `docs` domain and published viewer URLs such as `/d/<slug>/v/<n>` are unchanged.

## Linked Spec

- Global engineering standard: `octo-spec@1.2.0`
- Runtime contract: the gateway and Docs HTML backend must serve `/docs-html/v1/*`
- No dedicated issue or product spec is linked to this PR

### 1. What changes on the load-bearing path?

Before this PR, the CLI registry resolved HTML operations under the generic `/v1/*` path. Because command dispatch uses the registry path verbatim, every `octo-cli html` request was sent through that old route.

After this PR, all 20 registered HTML operations resolve under `/docs-html/v1/*`. The service identifier and `OCTO_API_BASE_URL` handling remain unchanged; only the request path prefix changes. Published document URLs are not rewritten.

### 2. What could break?

- If the gateway or Docs HTML backend does not serve `/docs-html/v1/*` before this client change is deployed, HTML commands will return 404.
- A missed registry path would leave one operation on the old route.
- Accidentally changing viewer URLs such as `/d/<slug>/v/<n>` would break published document links.
- Other CLI domains must not inherit the Docs HTML prefix.

No authentication behavior, request payload, response schema, environment variable, or persisted data format changes.

### 3. How do we know it works?

- The registry contract test iterates every HTML operation and requires the `/docs-html/v1/` prefix.
- Existing operation-count coverage ensures the HTML operation list cannot disappear and make the prefix test pass vacuously.
- The built CLI resolves `html list --dry-run` to a URL under `/docs-html/v1/docs`.
- Repository-wide review confirmed all HTML registry paths and skill documentation use the new prefix, with no bare old paths or double-prefixes.
- CI build, Go tests, npm tests, history, and vulnerability checks pass.

## How verified

- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] CI-pinned `golangci-lint` on the touched Go package
- [x] Registry contract test covers all 20 HTML operations
- [x] Built CLI dry-run resolves the new gateway path
- [x] `git diff --check`
- [x] Independent code review: approved

## Deployment prerequisites / required environment variables

[...]

PR #106, opened 2026-07-27
feat: add read-only summary commands

wanyiwang06:
## Summary

- add `summary list`, `summary get`, and `summary result` commands
- use the owner-scoped bot Summary API with Bearer authentication and no client-supplied Space header
- add the `octo-summary` skill and update shared skill guidance and CLI documentation
- cover command registration, authentication headers, request paths, queries, and pagination with tests

## Dependency

Depends on Mininglamp-OSS/octo-smart-summary#172. Merge and deploy the Summary backend API before releasing these CLI commands.

## Validation

- `go test ./...`
- `go vet ./...`
- `python3 /home/mlamp/.codex/skills/.system/skill-creator/scripts/quick_validate.py skills/octo-summary`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **`octo-cli summary` read-only domain** — `summary list|get|result` lets a
+  personal Agent discover and cite existing summaries visible to its human
+  owner. The embedded spec targets the gateway's `/summary/api/v1/bot/*`
+  mount, suppresses client space headers, and ships with the `octo-summary`
+  Skill. Listing filters title or topic; result reads include citation metadata
+  while the backend omits surrounding-message context for bot requests.
 - **`octo-cli message search` family** (6 subcommands) — full-text message and
   file search: `message search` (messages), `search all` (messages + files),
   `search files`, `search media` (images/videos, in-channel only, no keyword),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   mount, suppresses client space headers, and ships with the `octo-summary`
   Skill. Listing filters title or topic; result reads include citation metadata
   while the backend omits surrounding-message context for bot requests.
+  **Currently withheld** behind `x-octo-disabled` (skill `disabled: true`): the
+  commands and skill are not listed by the CLI until the Summary backend
+  (octo-smart-summary#172) is merged and deployed; `octo-cli schema summary.*`
+  still introspects. Flip both flags in a one-line follow-up once the backend is live.
 - **`octo-cli message search` family** (6 subcommands) — full-text message and
   file search: `message search` (messages), `search all` (messages + files),
   `search files`, `search media` (images/videos, in-channel only, no keyword),
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -76,7 +76,7 @@ octo-cli version
 
 `octo-cli auth login` stores a bot token (read from a hidden prompt, `--with-token` stdin, or `--token-file` — never argv) under a profile keyed by `--bot-id`/`--profile`. `status`/`list` show metadata only (tokens always masked); `logout` removes a profile.
 
-Bot-type capability and per-command flags are in `docs/octo-cli-design.md`. Agent-facing usage lives under `skills/` (`octo-shared`, `octo-matter` (withheld — see above), `octo-messaging`, `octo-files`, `octo-docs`, `octo-html`, `octo-marketplace`) — keep those in sync when command shapes change.
+Bot-type capability and per-command flags are in `docs/octo-cli-design.md`. Agent-facing usage lives under `skills/` (`octo-shared`, `octo-matter` (withheld — see above), `octo-summary` (withheld — backend pending), `octo-messaging`, `octo-files`, `octo-docs`, `octo-html`, `octo-marketplace`) — keep those in sync when command shapes change.
 
 ## Environment
 
--- a/README.md
+++ b/README.md
@@ -289,7 +289,8 @@ Machine-readable usage docs for AI Agents live under [`skills/`](./skills/):
   share codes & per-uid grants, media assets, inline comments, agent element
   read/replace.
 - [`octo-summary`](./skills/octo-summary/SKILL.md) — discover, read, and cite
-  existing summaries visible to the personal Agent's human owner.
+  existing summaries visible to the personal Agent's human owner. **Temporarily
+  withheld** (backend API pending — octo-smart-summary#172; not currently loadable).
 
 These docs are also **embedded in the binary**, so a released `octo-cli` ships them:
 
--- a/README.md
+++ b/README.md
@@ -288,6 +288,8 @@ Machine-readable usage docs for AI Agents live under [`skills/`](./skills/):
   **separate backend** from `octo-docs`): publish immutable versions, drafts,
   share codes & per-uid grants, media assets, inline comments, agent element
   read/replace.
+- [`octo-summary`](./skills/octo-summary/SKILL.md) — discover, read, and cite
+  existing summaries visible to the personal Agent's human owner.
 
 These docs are also **embedded in the binary**, so a released `octo-cli` ships them:
 
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -446,6 +446,9 @@ func TestDisabledServiceWithheldFromCommandTree(t *testing.T) {
 	if topLevelCmd(root, "matter") != nil {
 		t.Error("disabled service `matter` must not appear in the command tree")
 	}
+	if topLevelCmd(root, "summary") != nil {
+		t.Error("disabled service `summary` must not appear in the command tree")
+	}
 	if topLevelCmd(root, "message") == nil {
 		t.Error("enabled service `message` must still appear")
 	}
@@ -472,11 +475,17 @@ func TestDisabledServiceHiddenFromGlobalSchemaList(t *testing.T) {
 		if s == "matter" {
 			t.Error("global schema --list must not list disabled service `matter`")
 		}
+		if s == "summary" {
+			t.Error("global schema --list must not list disabled service `summary`")
+		}
 	}
 	for _, op := range data["operations"].([]any) {
 		m, _ := op.(map[string]any)
 		if m["service"] == "matter" {
 			t.Errorf("global schema --list leaked a matter op: %v", m["id"])
+		}
+		if m["service"] == "summary" {
+			t.Errorf("global schema --list leaked a summary op: %v", m["id"])
 		}
 	}
 }
@@ -500,5 +509,10 @@ func TestDisabledServiceStillIntrospectable(t *testing.T) {
 	// Explicit operation lookup still resolves.
 	if _, _, err := execRoot(t, f, "schema", "matter.list"); err != nil {
 		t.Errorf("explicit `schema matter.list` must still resolve: %v", err)
+	}
+
+	// summary is likewise disabled but must stay introspectable.
+	if _, _, err := execRoot(t, f, "schema", "summary.list"); err != nil {
+		t.Errorf("explicit `schema summary.list` must still resolve: %v", err)
 	}
 }
--- a/cmd/service/summary_test.go
+++ b/cmd/service/summary_test.go
@@ -1,0 +1,98 @@
+package service
+
+import (
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/Mininglamp-OSS/octo-cli/internal/registry"
+)
+
+func TestSummaryRegistryShape(t *testing.T) {
+	reg := registry.MustNew()
+	wants := map[string]string{
+		"summary.list":   "/summary/api/v1/bot/summaries",
+		"summary.get":    "/summary/api/v1/bot/summaries/{id}",
+		"summary.result": "/summary/api/v1/bot/summaries/{id}/result",
+	}
+	for id, path := range wants {
+		op, ok := reg.GetOperation(id)
+		if !ok {
+			t.Fatalf("missing %s", id)
+		}
+		if op.Method != http.MethodGet || op.Path != path || op.SpaceHeader || op.Risk != "read" || op.BaseURLEnv != "OCTO_API_BASE_URL" {
+			t.Errorf("%s: got method=%s path=%s space=%v risk=%s base=%s", id, op.Method, op.Path, op.SpaceHeader, op.Risk, op.BaseURLEnv)
+		}
+	}
+	if op, _ := reg.GetOperation("summary.list"); op.Pagination != nil {
+		t.Error("summary.list uses a backend {total,items} page, so generic --page-all must stay disabled")
+	}
+}
+
+func TestSummaryTreeShape(t *testing.T) {
+	root, _, _ := rootWithService(t, func(http.ResponseWriter, *http.Request) {})
+	summary := findCmd(root, "summary")
+	if summary == nil {
+		t.Fatal("missing summary service command")
+	}
+	for _, leaf := range []string{"list", "get", "result"} {
+		if !contains(childNames(summary), leaf) {
+			t.Errorf("summary: missing %q; got %v", leaf, childNames(summary))
+		}
+	}
+	if list := findCmd(summary, "list"); list == nil || list.Flags().Lookup("page-all") != nil {
+		t.Error("summary list must exist without --page-all")
+	}
+}
+
+func TestSummaryListSendsBearerWithoutSpaceHeader(t *testing.T) {
+	var gotPath, gotQuery, gotAuth, gotSpace string
+	root, _, _ := rootWithServiceSpaced(t, "spoofed-space", func(w http.ResponseWriter, r *http.Request) {
+		gotPath, gotQuery = r.URL.Path, r.URL.RawQuery
+		gotAuth, gotSpace = r.Header.Get("Authorization"), r.Header.Get("X-Space-Id")
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"code":0,"message":"ok","data":{"total":0,"items":[]}}`))
+	})
+	root.SetArgs([]string{"summary", "list", "--keyword", "review", "--page-size", "10"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if gotPath != "/summary/api/v1/bot/summaries" || !strings.Contains(gotQuery, "keyword=review") || !strings.Contains(gotQuery, "page_size=10") {
+		t.Errorf("request = %s?%s", gotPath, gotQuery)
+	}
+	if gotAuth == "" || gotSpace != "" {
+		t.Errorf("auth=%q space=%q", gotAuth, gotSpace)
+	}
+}
+
+func TestSummaryResultPath(t *testing.T) {
+	var gotPath string
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotPath = r.URL.Path
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"code":0,"data":{"content":"done"}}`))
+	})
+	root.SetArgs([]string{"summary", "result", "42"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if gotPath != "/summary/api/v1/bot/summaries/42/result" {
+		t.Fatalf("path=%s", gotPath)
+	}
+}
+
+func TestSummaryGetPath(t *testing.T) {
+	var gotPath string
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotPath = r.URL.Path
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"code":0,"data":{"task_id":42}}`))
+	})
+	root.SetArgs([]string{"summary", "get", "42"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if gotPath != "/summary/api/v1/bot/summaries/42" {
+		t.Fatalf("path=%s", gotPath)
+	}
+}
--- a/cmd/service/summary_test.go
+++ b/cmd/service/summary_test.go
@@ -40,8 +40,11 @@ func TestSummaryTreeShape(t *testing.T) {
 			t.Errorf("summary: missing %q; got %v", leaf, childNames(summary))
 		}
 	}
-	if list := findCmd(summary, "list"); list == nil || list.Flags().Lookup("page-all") != nil {
-		t.Error("summary list must exist without --page-all")
+	list := findCmd(summary, "list")
+	if list == nil {
+		t.Fatal("summary list command must exist")
+	} else if list.Flags().Lookup("page-all") != nil {
+		t.Error("summary list must not expose --page-all (backend returns a {total, items} page, no cursor)")
 	}
 }
 
--- a/cmd/skills_test.go
+++ b/cmd/skills_test.go
@@ -77,8 +77,14 @@ func TestLoadSkillsExcludesDisabled(t *testing.T) {
 	if hasSkill(entries, "octo-matter") {
 		t.Error("disabled skill `octo-matter` must not be listed")
 	}
+	if hasSkill(entries, "octo-summary") {
+		t.Error("disabled skill `octo-summary` must not be listed")
+	}
 	if _, err := skills.FS.ReadFile("octo-matter/SKILL.md"); err != nil {
 		t.Errorf("octo-matter/SKILL.md must stay embedded (only the listing filters it): %v", err)
+	}
+	if _, err := skills.FS.ReadFile("octo-summary/SKILL.md"); err != nil {
+		t.Errorf("octo-summary/SKILL.md must stay embedded (only the listing filters it): %v", err)
 	}
 }
 
--- a/internal/registry/loader_test.go
+++ b/internal/registry/loader_test.go
@@ -11,7 +11,7 @@ func TestNewLoadsAllServices(t *testing.T) {
 		t.Fatalf("New: %v", err)
 	}
 	got := r.ListServices()
-	want := []string{"bot", "docs", "event", "file", "group", "html", "marketplace", "matter", "message", "thread"}
+	want := []string{"bot", "docs", "event", "file", "group", "html", "marketplace", "matter", "message", "summary", "thread"}
 	if len(got) != len(want) {
 		t.Fatalf("ListServices: got %d services, want %d (%v)", len(got), len(want), got)
 	}
@@ -45,6 +45,7 @@ func TestAllDomainOperationCounts(t *testing.T) {
 		"docs":        31,
 		"html":        20,
 		"marketplace": 25,
+		"summary":     3,
 	}
 	totalWant := 0
 	for svc, want := range expected {
--- a/internal/registry/loader_test.go
+++ b/internal/registry/loader_test.go
@@ -173,6 +173,7 @@ func TestServiceSpaceHeaderContract(t *testing.T) {
 		{"message", "message.send", true},
 		{"matter", "matter.create", true},
 		{"marketplace", "skill.get", true},
+		{"summary", "summary.list", false},
 		{"docs", "docs.create", false},
 		{"bot", "bot.register", false},
 		{"thread", "thread.create", false},
--- a/internal/registry/specs/summary.json
+++ b/internal/registry/specs/summary.json
@@ -9,7 +9,7 @@
   "x-octo-base-url": "OCTO_API_BASE_URL",
   "x-octo-space-header": false,
   "x-octo-disabled": true,
-  "x-octo-status-values": ["pending(0)", "waiting_confirm(1)", "processing(2)", "completed(3)", "failed(4)", "cancelled(5)"],
+  "x-octo-status-values": ["waiting(0)", "waiting_confirm(1)", "processing(2)", "completed(3)", "failed(4)", "cancelled(5)"],
   "paths": {
     "/summary/api/v1/bot/summaries": {
       "get": {
@@ -19,9 +19,9 @@
         "x-octo-risk": "read",
         "parameters": [
           { "name": "page", "in": "query", "schema": { "type": "integer", "default": 1, "minimum": 1 }, "description": "1-based page number." },
-          { "name": "page_size", "in": "query", "x-octo-flag": "page-size", "schema": { "type": "integer", "default": 20, "minimum": 1, "maximum": 100 }, "description": "Items per page (1-100)." },
+          { "name": "page_size", "in": "query", "x-octo-flag": "page-size", "schema": { "type": "integer", "default": 20, "minimum": 1, "maximum": 100 }, "description": "Items per page (1-100); values >100 are silently reset to 20 server-side." },
           { "name": "keyword", "in": "query", "schema": { "type": "string" }, "description": "Substring matched against summary title OR topic." },
-          { "name": "status", "in": "query", "schema": { "type": "integer", "enum": [0, 1, 2, 3, 4, 5] }, "description": "Task status filter: 0=pending, 1=waiting_confirm, 2=processing, 3=completed, 4=failed, 5=cancelled." },
+          { "name": "status", "in": "query", "schema": { "type": "integer", "enum": [0, 1, 2, 3, 4, 5] }, "description": "Task status filter. 0=waiting (caller-specific aggregate: status 0, OR a pending participant/schedule invitation, OR an unsubmitted personal result on a multi-participant task — so status-0 rows may not have status field 0), 1=waiting_confirm, 2=processing, 3=completed, 4=failed, 5=cancelled." },
           { "name": "trigger_type", "in": "query", "x-octo-flag": "trigger-type", "schema": { "type": "integer", "enum": [1, 2, 3] }, "description": "How the summary was triggered: 1=manual, 2=scheduled, 3=agent." },
           { "name": "origin_channel_id", "in": "query", "x-octo-flag": "origin-channel-id", "schema": { "type": "string" }, "description": "Filter to summaries whose origin channel matches this id." },
           { "name": "created_after", "in": "query", "x-octo-flag": "created-after", "schema": { "type": "string", "format": "date-time" }, "description": "RFC3339 timestamp, e.g. 2026-07-01T00:00:00Z. Malformed values are silently ignored by the server (filter not applied)." },
@@ -31,26 +31,44 @@
         ],
         "responses": {
           "200": {
-            "description": "Summary page",
+            "description": "Summary page (wrapped in the backend {code, message, data} envelope; via octo-cli the jq path is .data.data.items[]).",
             "content": {
               "application/json": {
                 "schema": {
                   "type": "object",
+                  "description": "Backend envelope. octo-cli nests this under its own `data`, so the effective jq path is `.data.data.*`.",
                   "properties": {
-                    "total": { "type": "integer", "description": "Total matching summaries (not just this page)." },
-                    "items": {
-                      "type": "array",
-                      "items": {
-                        "type": "object",
-                        "properties": {
-                          "task_id": { "type": "integer" },
-                          "task_no": { "type": "string" },
-                          "title": { "type": "string" },
-                          "status": { "type": "integer", "description": "0=pending,1=waiting_confirm,2=processing,3=completed,4=failed,5=cancelled" },
-                          "trigger_type": { "type": "integer", "description": "1=manual,2=scheduled,3=agent" },
-                          "origin_channel_id": { "type": "string" },
-                          "created_at": { "type": "string", "description": "RFC3339 timestamp" },
-                          "updated_at": { "type": "string", "description": "RFC3339 timestamp" }
+                    "code": { "type": "integer", "description": "0 on success." },
+                    "message": { "type": "string" },
+                    "data": {
+                      "type": "object",
+                      "properties": {
+                        "total": { "type": "integer", "description": "Total matching summaries (not just this page)." },
+                        "items": {
+                          "type": "array",
+                          "items": {
+                            "type": "object",
+                            "properties": {
+                              "task_id": { "type": "integer" },
+                              "task_no": { "type": "string" },
+                              "title": { "type": "string" },
+                              "topic": { "type": "string", "description": "Free-text topic; keyword also matches this." },
+                              "status": { "type": "integer", "description": "0=waiting,1=waiting_confirm,2=processing,3=completed,4=failed,5=cancelled" },
+                              "trigger_type": { "type": "integer", "description": "1=manual,2=scheduled,3=agent" },
+                              "summary_mode": { "type": "integer" },
+                              "origin_channel_id": { "type": "string" },
+                              "sources": { "type": "array", "description": "Source channels/scopes the summary was built from." },
+                              "participants": { "type": "array", "description": "Participants of a multi-person summary." },
+                              "creator_name": { "type": "string" },
+                              "total_msg_count": { "type": "integer" },
+                              "created_at": { "type": "string", "description": "RFC3339 timestamp" },
+                              "completed_at": { "type": "string", "description": "RFC3339 timestamp; empty until completed." },
+                              "activity_at": { "type": "string", "description": "RFC3339 timestamp of last activity." },
+                              "is_unread": { "type": "boolean" },
+                              "needs_attention": { "type": "boolean" },
+                              "schedule_id": { "type": "integer", "description": "Present for scheduled summaries." }
+                            }
+                          }
                         }
                       }
                     }
@@ -72,25 +90,33 @@
         ],
         "responses": {
           "200": {
-            "description": "Summary detail",
+            "description": "Summary detail (wrapped in the backend {code, message, data} envelope; via octo-cli the jq path is .data.data.*).",
             "content": {
               "application/json": {
                 "schema": {
                   "type": "object",
                   "properties": {
-                    "task_id": { "type": "integer" },
-                    "task_no": { "type": "string" },
-                    "title": { "type": "string" },
-                    "status": { "type": "integer", "description": "0=pending,1=waiting_confirm,2=processing,3=completed,4=failed,5=cancelled" },
-                    "trigger_type": { "type": "integer", "description": "1=manual,2=scheduled,3=agent" },
-                    "origin_channel_id": { "type": "string" },
-                    "created_at": { "type": "string", "description": "RFC3339 timestamp" },
-                    "updated_at": { "type": "string", "description": "RFC3339 timestamp" },
-                    "result": {
+                    "code": { "type": "integer" },
+                    "message": { "type": "string" },
+                    "data": {
                       "type": "object",
-                      "description": "Current visible result; null until the task completes.",
                       "properties": {
-                        "content": { "type": "string", "description": "Summary body (markdown)." }
+                        "task_id": { "type": "integer" },
+                        "task_no": { "type": "string" },
+                        "title": { "type": "string" },
+                        "topic": { "type": "string" },
+                        "status": { "type": "integer", "description": "0=waiting,1=waiting_confirm,2=processing,3=completed,4=failed,5=cancelled" },
+                        "trigger_type": { "type": "integer", "description": "1=manual,2=scheduled,3=agent" },
+                        "origin_channel_id": { "type": "string" },
+                        "created_at": { "type": "string", "description": "RFC3339 timestamp" },
+                        "updated_at": { "type": "string", "description": "RFC3339 timestamp" },
+                        "result": {
+                          "type": ["object", "null"],
+                          "description": "Current visible result; null until the task completes.",
+                          "properties": {
+                            "content": { "type": "string", "description": "Summary body (markdown)." }
+                          }
+                        }
                       }
                     }
                   }
@@ -104,33 +130,46 @@
     "/summary/api/v1/bot/summaries/{id}/result": {
       "get": {
         "operationId": "summary.result",
-        "summary": "Get the current summary body and citation metadata. Citation context_before/context_after is omitted on bot routes.",
-        "description": "Citation context_before/context_after is omitted on bot routes.",
+        "summary": "Get the current summary body and citation metadata. Citation context_before/context_after is omitted on bot routes; for a by-person summary the caller does not own, plain `citations` is returned empty while `team_citations` still carries the index→author mapping.",
+        "description": "Citation context_before/context_after is omitted on bot routes. For SummaryMode=by-person, plain `citations` is empty unless the caller owns the producing personal result (`team_citations` still populated).",
         "x-octo-risk": "read",
         "parameters": [
           { "name": "id", "in": "path", "required": true, "schema": { "type": "integer" }, "description": "Summary task_id." }
         ],
         "responses": {
           "200": {
-            "description": "Current summary result",
+            "description": "Current summary result (wrapped in the backend {code, message, data} envelope; via octo-cli the jq path is .data.data.*).",
             "content": {
               "application/json": {
                 "schema": {
                   "type": "object",
                   "properties": {
-                    "task_id": { "type": "integer" },
-                    "content": { "type": "string", "description": "Summary body (markdown), with [n] citation markers." },
-                    "citations": {
-                      "type": "array",
-                      "description": "Grounding messages for [n] markers. context_before/context_after are omitted on bot routes.",
-                      "items": {
-                        "type": "object",
-                        "properties": {
-                          "index": { "type": "integer", "description": "The [n] marker this entry grounds." },
-                          "sender": { "type": "string" },
-                          "sent_at": { "type": "string", "description": "RFC3339 timestamp" },
-                          "source": { "type": "string", "description": "Origin channel/source label." },
-                          "message_seq": { "type": "integer", "description": "Message sequence in the source channel." }
+                    "code": { "type": "integer" },
+                    "message": { "type": "string" },
+                    "data": {
+                      "type": "object",
+                      "properties": {
+                        "content": { "type": "string", "description": "Summary body (markdown), with [n] citation markers." },
+                        "citations": {
+                          "type": "array",
+                          "description": "Grounding messages for [n] markers. context_before/context_after are omitted on bot routes; empty for a by-person summary the caller does not own.",
+                          "items": {
+                            "type": "object",
+                            "properties": {
+                              "index": { "type": "integer", "description": "The [n] marker this entry grounds." },
+                              "sender": { "type": "string" },
+                              "sent_at": { "type": "string", "description": "RFC3339 timestamp" },
+                              "source": { "type": "string", "description": "Origin channel/source label." },
+                              "message_seq": { "type": "integer", "description": "Message sequence in the source channel." },
+                              "content": { "type": "string", "description": "Quoted message text (not stripped on bot routes; use this to attribute claims)." },
+                              "channel_id": { "type": "string" },
+                              "channel_type": { "type": "integer" }
+                            }
+                          }
+                        },
+                        "team_citations": {
+                          "type": "array",
+                          "description": "Index→author mapping for by-person / team summaries; always returned."
                         }
                       }
                     }
--- a/internal/registry/specs/summary.json
+++ b/internal/registry/specs/summary.json
@@ -1,0 +1,57 @@
+{
+  "openapi": "3.1.0",
+  "info": {
+    "title": "Octo Summary Bot Read API",
+    "version": "1.0.0",
+    "description": "Read-only access to summaries visible to the bot owner's identity. The backend resolves owner and space from the bot token."
+  },
+  "x-octo-service": "summary",
+  "x-octo-base-url": "OCTO_API_BASE_URL",
+  "x-octo-space-header": false,
+  "paths": {
+    "/summary/api/v1/bot/summaries": {
+      "get": {
+        "operationId": "summary.list",
+        "summary": "List summaries visible to the bot owner",
+        "description": "Returns a page in {total, items}. keyword matches title or topic. Space and effective reader are resolved server-side from the bot token.",
+        "x-octo-risk": "read",
+        "parameters": [
+          { "name": "page", "in": "query", "schema": { "type": "integer", "default": 1, "minimum": 1 } },
+          { "name": "page_size", "in": "query", "x-octo-flag": "page-size", "schema": { "type": "integer", "default": 20, "minimum": 1, "maximum": 100 } },
+          { "name": "keyword", "in": "query", "schema": { "type": "string" }, "description": "title or topic substring" },
+          { "name": "status", "in": "query", "schema": { "type": "integer" } },
+          { "name": "trigger_type", "in": "query", "x-octo-flag": "trigger-type", "schema": { "type": "integer" } },
+          { "name": "origin_channel_id", "in": "query", "x-octo-flag": "origin-channel-id", "schema": { "type": "string" } },
+          { "name": "created_after", "in": "query", "x-octo-flag": "created-after", "schema": { "type": "string", "format": "date-time" } },
+          { "name": "created_before", "in": "query", "x-octo-flag": "created-before", "schema": { "type": "string", "format": "date-time" } },
+          { "name": "sort_by", "in": "query", "x-octo-flag": "sort-by", "schema": { "type": "string", "enum": ["created_at", "updated_at"], "default": "created_at" } },
+          { "name": "sort_order", "in": "query", "x-octo-flag": "sort-order", "schema": { "type": "string", "enum": ["asc", "desc"], "default": "desc" } }
+        ],
+        "responses": { "200": { "description": "Summary page" } }
+      }
+    },
+    "/summary/api/v1/bot/summaries/{id}": {
+      "get": {
+        "operationId": "summary.get",
+        "summary": "Get summary metadata and current visible result",
+        "x-octo-risk": "read",
+        "parameters": [
+          { "name": "id", "in": "path", "required": true, "schema": { "type": "integer" } }
+        ],
+        "responses": { "200": { "description": "Summary detail" } }
+      }
+    },
+    "/summary/api/v1/bot/summaries/{id}/result": {
+      "get": {
+        "operationId": "summary.result",
+        "summary": "Get the current summary body and citation metadata",
+        "description": "Citation context_before/context_after is omitted on bot routes.",
+        "x-octo-risk": "read",
+        "parameters": [
+          { "name": "id", "in": "path", "required": true, "schema": { "type": "integer" } }
+        ],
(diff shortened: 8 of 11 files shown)


mochashanyao (dismissed):
[Octo-Q · automated review]

---

Jerry-Xin (dismissed):
Project relevance gate passed; this PR extends the `octo-cli` command registry, docs, tests, and embedded skills for a relevant read-only Summary domain.

✅ Highlights
- The Summary spec follows the existing metadata-driven service pattern and correctly declares the service, gateway base URL, and `x-octo-space-header:false` at `internal/registry/specs/summary.json:8`.
- Space-header suppression is wired correctly end to end: the operation carries `SpaceHeaderSet && !SpaceHeader` into `SuppressSpaceHeader` at `cmd/service/run.go:55`, and the transport only sets `X-Space-Id` when `!suppressSpaceHeader` at `internal/client/client.go:220` (and mirrored in dry-run at `client.go:377`). This means the client will not forward a client-supplied space for bot Summary routes.
- The new command tests are real integration coverage, not mock-fitted contracts: `TestSummaryListSendsBearerWithoutSpaceHeader` runs with a spaced credential and asserts the request carries `Authorization` but no `X-Space-Id`, plus correct path and query (`keyword`, `page_size`) at `cmd/service/summary_test.go:48`. Path/registry-shape tests cover `summary.list|get|result` methods, paths, risk=read, base URL, and disabled generic `--page-all`.
- Registry counts are kept honest: `loader_test.go` updates the service list and adds `summary: 3` to the per-domain operation-count assertions, so a missing/extra operation would fail the build.
- The skill guidance clearly documents owner-scoped `bf_*` auth, no `--space`, manual page-based pagination, citation handling, and intentionally omitted surrounding context, and `skills_test.go` asserts the skill is embedded with the key markers.

💬 Non-blocking
- No blocking code issues found.
- `internal/registry/specs/summary.json` sets `x-octo-base-url: OCTO_API_BASE_URL`, but `serviceForBaseURL` currently returns an empty string, so per-operation base-URL routing relies solely on the shared `OCTO_API_BASE_URL` env plus the `/summary/api/v1/bot/*` path prefix. This matches the other bot-scoped services and is not a defect for this PR; just noting the `BaseURLEnv` field is effectively informational at the routing layer today.
- Per the PR description, the Summary backend (dependency PR in the summary backend repo) must be merged and deployed before these CLI commands are released. This is a release-coordination requirement, not a CLI blocker.

Validation:
- `go test ./cmd/service ./internal/registry ./skills`
- `go test ./...`
- `go vet ./...`

yujiawei (changes requested):
# Code Review — PR #106 (octo-cli)

Reviewed at head SHA `7e0df2e19fbaba759886cd716f0a90a88d6f9443`, base `main` (merge-base `29ec0bc`).

Verification performed locally, not just by reading the diff:

- `go build ./...`, `go vet ./...`, `go test ./...` — **all pass**.
- Built the binary and exercised the real command surface with `--dry-run` (`OCTO_SPACE_ID` set) to confirm wire-level behaviour.
- Cross-checked the three request paths and all ten query parameters against the backend dependency (`octo-smart-summary#172`, head `6f5be7d`) and against the gateway route table in `octo-deployment` (`docker/nginx/conf.d/octo.conf.template:247-273` and `helm/octo/templates/configmap-nginx.yaml:168-171`).

---

## 1. Spec compliance

**Spec: ✅**

Everything the PR description claims is actually in the diff, and nothing beyond it:

- `summary list|get|result` register correctly (`cmd/service/summary_test.go:30-49`, verified live: `octo-cli summary list --help` renders all three).
- **Gateway prefix is correct.** `/summary/api/v1/bot/*` (`internal/registry/specs/summary.json:12,33,44`) is stripped by `rewrite ^/summary/(.*) /$1 break;` at the nginx layer, landing on the backend's `GET /api/v1/bot/summaries[/:id[/result]]` (`octo-smart-summary internal/api/router/router.go`, bot group). This is the exact failure mode #105 had to fix for `html`, and here it checks out in both the docker and helm templates.
- **Space-header suppression is real, not just unit-tested.** With `OCTO_SPACE_ID=sp_123` loaded, `summary list --dry-run` emits only `Authorization: Bearer …`, while `marketplace skill list --dry-run` on the same credential emits `X-Space-Id: sp_123`. The mechanism (`x-octo-space-header: false` → `SpaceHeaderSet && !SpaceHeader` → `SuppressSpaceHeader`, `cmd/service/run.go:55`) is the generic one, no carve-out. It also matches the backend, which resolves owner and space from verify-bot and ignores the client header (`internal/middleware/space.go:28-42`).
- All ten declared query parameters exist as real backend filters (`octo-smart-summary internal/api/handler/task.go:492-568`); `keyword` really does match title **or** topic, as the CHANGELOG says.
- `--page-all` is correctly absent (no `x-octo-pagination`), matching the backend's `{total, items}` page shape, and the test pins it.
- No write operations, no over-built flags, no unrequested surface.

## 2. Code quality

### P1-1 — `--created-after` / `--created-before` are silently dropped on non-RFC3339 input, and nothing tells the caller RFC3339 is required

`internal/registry/specs/summary.json:25-26`

```json
{ "name": "created_after",  "in": "query", "x-octo-flag": "created-after",  "schema": { "type": "string", "format": "date-time" } },
{ "name": "created_before", "in": "query", "x-octo-flag": "created-before", "schema": { "type": "string", "format": "date-time" } },
```

Three things compose into a wrong-answer bug:

1. The registry loader never reads `format` — `ParamInfo` (`internal/registry/loader.go:142-158`) has no `Format` field, so `format: date-time` is dropped on load and never reaches `--help` or `octo-cli schema`.
2. Neither parameter has a `description`, so the generated flag help is blank. Verified on the built binary:
   ```
   --created-after string
   --created-before string
   ```
3. The backend parses the value and **silently ignores the filter when parsing fails** (`octo-smart-summary internal/api/handler/task.go:548-553`):
   ```go
   if s := c.Query("created_after"); s != "" {
       if t, err := time.Parse(time.RFC3339, s); err == nil {
           whereSQL += " AND created_at >= ?"
           args = append(args, t)
       }
   }
   ```
   No `else`, no error.

So the natural agent input goes through and returns *unfiltered* results with `ok: true`:

```
$ octo-cli summary list --created-after 2026-07-01 --dry-run
… /summary/api/v1/bot/summaries?created_after=2026-07-01
```

An agent that asked for "summaries created after July 1" gets every summary it can see and has no signal that the filter was discarded — then cites them as date-scoped. For a skill whose entire purpose is grounded citation, that is a correctness problem, not a cosmetic one.

**Fix:** add a description stating the required format on both parameters (e.g. `"RFC3339 timestamp, e.g. 2026-07-01T00:00:00Z; malformed values are ignored by the server"`). Surfacing `format` through `ParamInfo` would fix the whole class, but a description is the minimum.

### P1-2 — All three operations ship with no response schema, which the architecture doc lists as a requirement

`internal/registry/specs/summary.json:30,41,53`

```json
"responses": { "200": { "description": "Summary page" } }
```

`docs/architecture-design.md:293` states:

> **Response schemas required** — Agent needs field names for --jq and decision-making.

`summary.json` is the only spec in the registry with **0 of 3** operations declaring a 2xx content schema (for comparison: `marketplace` 26/26, `docs` 19/31). Verified against the built binary:

```
$ octo-cli schema summary.result
… "parameters":[{"in":"path","name":"id","required":true,"type":"integer"}], … (no response_schema)
```

This lands hardest on this specific feature. `skills/octo-summary/SKILL.md:38` instructs the agent to "attribute claims to the returned citation entry using its index, sender, sent time, source, and message sequence" — none of those field names are discoverable from the CLI, so an agent must guess field names for `--jq` or fetch and eyeball raw JSON first. `total`/`items`, `task_id`, and the citation object shape are all already known here (they are in the backend handler); declaring them is cheap.

### P1-3 — Enabled on merge while the backend dependency is still unmerged, though the repo has a purpose-built gate for exactly this

`internal/registry/specs/summary.json:1-10` (no `x-octo-disabled`), `skills/octo-summary/SKILL.md:1-4` (no `disabled:`)

`octo-smart-summary#172` is still **open**. The PR body says "Merge and deploy the Summary backend API before releasing these CLI commands", so the risk is acknowledged — but it is currently guarded only by someone remembering it at tag time. Releases are `workflow_dispatch` on an arbitrary tag (`.github/workflows/release-publish.yml`), so any release cut for an unrelated reason inside that window ships a `summary` domain that 404s, plus a bundled skill that actively tells agents to use it (`octo-cli skills` lists it via the `*/SKILL.md` glob in `cmd/skills.go:29`).

The repo already solves this by construction for `matter`, whose "backend API is not yet stable":

- `internal/registry/specs/matter.json:9` → `"x-octo-disabled": true`
- `skills/octo-matter/SKILL.md:4` → `disabled: true`

Disabled domains stay loaded, so `octo-cli schema summary.list` still introspects, but the commands and the skill are withheld from discovery.

**Fix:** either hold this PR until the backend merges and deploys, or land it with `x-octo-disabled: true` + skill `disabled: true` and flip both in a one-line follow-up once the backend is live. Note that the CHANGELOG entry under `## [Unreleased]` currently reads as a shipped feature either way.

### P2-1 — `status` and `trigger_type` are bare integers with no enum and no description

`internal/registry/specs/summary.json:22-23`

```json
{ "name": "status", "in": "query", "schema": { "type": "integer" } },
{ "name": "trigger_type", "in": "query", "x-octo-flag": "trigger-type", "schema": { "type": "integer" } },
```

`summary list --help` renders `--status int` and `--trigger-type int` with empty help text. The backend's values are a fixed closed set (`octo-smart-summary internal/model/model.go:60-65`: `Pending=0, WaitingConfirm=1, Processing=2, Completed=3, Failed=4, Cancelled=5`), and `skills/octo-summary/SKILL.md:19` uses the magic literal `--status 3` with no explanation of what 3 means. `matter.json` declares enums for its status filter and `x-octo-status-values` at the spec level; both are available here. Declaring `enum: [0,1,2,3,4,5]` plus a description would make the flag self-documenting — `formatEnum` already renders it into `--help` (`cmd/service/flags.go:73-75`).

More broadly, `summary.json` documents 1 of its 10 query parameters, the lowest ratio in the registry (`docs` and `marketplace` are at 100%). For an agent-facing CLI these descriptions *are* the API.

### P2-2 — Operation-level `description` text is never surfaced, so the two most important behavioural caveats are invisible

`internal/registry/specs/summary.json:16,48`

The spec carefully documents "Returns a page in {total, items}" and "Citation context_before/context_after is omitted on bot routes" as operation `description` fields. But `buildDetail` only reads `op["summary"]` — there is no `Description` field on `OperationInfo`/`OperationDetail` (`internal/registry/loader.go:129-136, 325-334`), so neither string reaches `--help` or `octo-cli schema` (confirmed in the `schema summary.result` output above). Today the only place an agent can learn about the missing citation context is the SKILL.md prose. Worth either promoting these into the `summary` field or adding `Description` to the loader.

### P2-3 — `CLAUDE.md` domain count and skill list not updated

`CLAUDE.md:26` still reads `## Command Structure (9 domains, 104 operations)`; enabled counts are now **10 domains / 118 operations** (measured via `EnabledServices()` / `EnabledOperations()`). `CLAUDE.md:79` enumerates the bundled skills and omits `octo-summary`, even though this PR did update the equivalent lists in `README.md:291-292` and `skills/octo-shared/SKILL.md:219`. Note the operation count was already stale before this PR (the `message search` family did not update it either), so only the domain count and the skill list are newly inaccurate.

### P2-4 — `TestServiceSpaceHeaderContract` not extended to the new service

`internal/registry/loader_test.go:166-197`

That table's own doc comment says it "pins the space-header declaration of **every** service spec so an accidental flip is caught", and it is a hardcoded case list — adding a service does not automatically extend it. `summary` is missing, so the repo-wide drift guard now has a hole. `TestSummaryRegistryShape` does assert the same property, so behaviour is covered; this is about keeping the contract table honest. One line: `{"summary", "summary.list", false}`.

### P2-5 — Test failure message conflates two different failures

`cmd/service/summary_test.go:43-45`

```go
if list := findCmd(summary, "list"); list == nil || list.Flags().Lookup("page-all") != nil {
    t.Error("summary list must exist without --page-all")
}
```

The short-circuit is safe, but "missing command" and "unexpected `--page-all` flag" produce the identical message. Splitting them costs two lines and saves a debugging round-trip.

---

## 3. Overall verdict

**CHANGES_REQUESTED** (Spec ✅ AND Quality Approved is required for approval; quality has P1 findings).

The wiring is genuinely good — the gateway prefix, the space-header suppression, the Bearer path, and every query parameter check out against both the backend and the deployment route table, and I verified the suppression end-to-end rather than trusting the unit test. What's missing is the agent-facing contract: this domain exists so an agent can find and *cite* a summary, and right now it ships with no response schemas, 1 of 10 parameters documented, and a date filter that fails silently.

## 4. Suggested changes

1. Add descriptions to `created_after` / `created_before` naming RFC3339 explicitly (P1-1). Consider surfacing `format` through `ParamInfo` so the whole class is fixed.
2. Declare 200 response schemas for all three operations — at minimum `{total, items[]}` for `list`, `task_id` + result fields for `get`, and `content` + the citation object shape for `result` (P1-2).
3. Decide the rollout gate explicitly: hold for the backend, or land with `x-octo-disabled: true` + skill `disabled: true` and a follow-up flip (P1-3).
4. Add `enum` + descriptions for `status` / `trigger_type`, and descriptions for the remaining bare parameters (P2-1).
5. Add `{"summary", "summary.list", false}` to `TestServiceSpaceHeaderContract` (P2-4).
6. Update `CLAUDE.md:26` domain count and the `CLAUDE.md:79` skill list (P2-3).

## 5. Additional observations (not blocking, outside this PR's scope)

- `RegisterServiceCommands` iterates `reg.ListServices()` (`cmd/service/service.go:31`), not `EnabledServices()` — yet `loader.go:97-100` documents `EnabledServices` as "the single source of truth… command registration and global discovery both use it so the filter can't be forgotten at one site". Only `matter` is disabled today and it is separately withheld, so nothing is broken, but the invariant the comment asserts is not the one the code implements. This matters for P1-3: if you take the `x-octo-disabled` route, verify the commands are actually withheld from the tree, don't assume it.
- Path arguments are substituted as opaque strings (`url.PathEscape`, `cmd/service/run.go:30`), so `summary get abc` reaches the network instead of failing locally despite `"type": "integer"`. Correctly escaped and no traversal risk (`summary get "abc/../../etc"` → `abc%2F..%2F..%2Fetc`), and it matches the existing `event.ack` precedent, so I'd leave it — noting it only because the declared integer type is decorative here. The loader likewise drops `minimum`/`maximum`, so `--page-size 101` is transmitted and silently reset to 20 server-side.
- Backend `summary_task.id` is an auto-increment sequence, not a Snowflake id (see the `ROW_NUMBER() OVER (PARTITION BY … CONCAT('t', id) …)` folding comment in `task.go`), so the `integer` typing carries no 64-bit precision risk. Checked because the declared type invited the question.
- `summary get`'s spec summary promises "metadata and current visible result"; the backend delivers, and `GetResult` additionally falls back to the owner's completed personal result for `ModeByPerson` tasks (`task.go:1114-1122`). The skill's framing of `get` vs `result` is accurate.

wanyiwang06:
@yujiawei 已按你的 review 全部修完，请复审当前 head `4f2b3d4`：

**P1**
- **P1-1** created_after/created_before 加了 RFC3339 description（并注明 malformed 值被服务端静默忽略）。
- **P1-2** 三个 op 都补了 2xx response schema：list `{total, items[]}`、get 详情、result `{content, citations[{index,sender,sent_at,source,message_seq}]}` —— `octo-cli schema summary.{list,get,result}` 均已能输出 response_schema。
- **P1-3** 加 `x-octo-disabled: true` + skill `disabled: true`（后端 octo-smart-summary#172 未合前）：命令与 skill 从 discovery 撤下、`schema` 仍可 introspect（与 matter 一致）；后端上线后一行翻开即可。

**P2**
- P2-1 status/trigger_type 加 enum + description（status 0-5、trigger 1-3）+ 顶层 `x-octo-status-values`。
- P2-2 `{total,items}` 与「context omitted on bot routes」两条 caveat 提进 operation summary（能到 --help/schema）。
- P2-3 CLAUDE.md skill 列表补 octo-summary（withheld；因 summary 已 disabled，域数维持 9）。
- P2-4 TestServiceSpaceHeaderContract 固定 `{summary, summary.list, false}`。
- P2-5 拆分 summary-list「命令缺失 vs 有 --page-all」的合并断言。

**自验**：`go build/vet/test ./...` 全绿；binary 实测 `schema` 三个 op 都 surface response schema + 参数 enum/description；disabled 域命令已撤下但 schema introspection 正常。

Jerry-Xin (dismissed):
Re-review at head 4f2b3d4e8312 (previous APPROVE was on 7e0df2e). Verified the direct incremental delta only (compare is ahead:1, behind:0 — clean single commit, no rebase drift).

## Delta (5 files)
- internal/registry/specs/summary.json (+107/-18): param descriptions + response schemas + enums, and `x-octo-disabled: true`.
- skills/octo-summary/SKILL.md: frontmatter `disabled: true`.
- cmd/service/summary_test.go, internal/registry/loader_test.go, plus the repo agent-guide doc: test/doc follow-ups.

## Peer CR (created-after / created-before) — resolved in practice
The prior CR flagged that the time-range params had no format validation, silently ignored bad input, and lacked help text. This commit:
- Adds explicit RFC3339 help text with examples on both params, and honestly documents the server-side silent-ignore behavior (filter not applied on malformed input).
- Gates the entire summary domain behind `x-octo-disabled: true` (spec) + `disabled: true` (skill) because the backend is not yet merged. Verified the gate is effective, not a no-op: the truthy helper accepts JSON boolean `true`, and the disabled service is withheld from the command tree and the schema-list while remaining introspectable — matching the established pattern already used by the matter service, with passing coverage.

Because the commands are withheld from discovery until the backend lands, the silent-ignore concern is moot in the shipped surface; when re-enabled, users get clear RFC3339 guidance. These params are server-side pass-through with no client handler, so client-side time.Parse validation is not warranted here.

## Verification
- go build / vet / test green locally (cmd/..., internal/registry/...).
- loader_test pins `{summary, summary.list, false}` (space-header suppression contract intact).
- summary_test split the conflated list assertion and hardened the --page-all check.

✅ Highlights
- Response schemas now make field names discoverable for --jq / schema introspection.
- Status/trigger_type enums + descriptions added; disabled-domain approach mirrors existing convention.

No blocking items. Approving.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1111 (octo-web)

## Summary

This PR adds a permanent "?" agent CLI onboarding help button to the Docs list header (`packages/docs/src/pages/OnboardingHelp.tsx`), addressing discoverability lost when the docs list narrowed to owned/recently-viewed entries. The button opens a modal dialog with a copyable self-bind prompt — a set of CLI instructions the user sends to their bot, which then reads its own token and configures `octo-cli` independently. No secrets are user-supplied; the external link uses `rel="noopener noreferrer"`. The PR is additive (7 files, 334 insertions, 0 deletions) with a thorough test suite covering persistent visibility, dialog interaction, clipboard wiring, reference link, and prompt safety across both locales.

## Verification

- ✅ **i18n contract** — new `onboarding.*` keys verified present in both `en-US.json` and `zh-CN.json` with identical structure (8 keys each). Access path `t('docs.onboarding.prompt')` maps correctly through the existing `useI18n` hook re-exported from `@octo/base` (`packages/docs/src/octoweb/index.ts:457`).
- ✅ **Security posture** — external `<a>` tag uses `target="_blank" rel="noopener noreferrer"` (`packages/docs/src/pages/OnboardingHelp.tsx:92-93`), preventing tabnabbing. No secrets are exposed in the prompt text; the bot reads its own token from config.
- ✅ **Prompt safety (test-locked)** — the test suite at `packages/docs/src/pages/DocsHome.test.tsx:2245-2256` asserts balanced quotes, no `OCTO_BOT_TOKEN=` shell assignments, and no `="<` placeholder-in-quotes patterns in both locales. These guards prevent a bot from executing a malformed shell export.
- ✅ **Clipboard degradation** — `navigator.clipboard?.writeText()` with optional chaining and catch block (`packages/docs/src/pages/OnboardingHelp.tsx:30-35`) degrades gracefully: clipboard denial still leaves the prompt visible and selectable.
- ✅ **CSS variable resolution** — all `--octo-*` semantic tokens consumed by the new styles (`--octo-bg`, `--octo-fg`, `--octo-muted`, `--octo-accent`, `--octo-border`, `--octo-hover`, `--octo-modal-automated review`) are defined at `:root` in `packages/docs/src/editor/styles.css:3-10`.
- ✅ **Config fallback** — `CLI_REFERENCE_URL` uses `envOr()` with `import.meta.env?.VITE_DOCS_CLI_REFERENCE_URL` and a real default (`https://github.com/Mininglamp-OSS/octo-cli`), not a placeholder (`packages/docs/src/config.ts:125-127`).

Static analysis only at head `cb8c4c67a794ef71dfc963eed6877c543eb523f6`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 and one nit below.

### P2 — Transparent overlay lacks backdrop dimming (`packages/docs/src/editor/styles.css:3383`)

`.octo-docs-help-overlay` uses `background: transparent`, so the dialog floats over page content with no visual separation. Users may not immediately realize the background is inert — the only dismiss affordance beyond the close button is clicking the overlay itself, but without dimming the overlay boundary is invisible. This diverges from the modal overlay convention elsewhere in the codebase (e.g., the editor share dialog and the matter picker both use a semi-opaque backdrop). A `background: rgba(0, 0, 0, 0.3)` or similar would anchor the visual affordance.

### Nit — Unused `--octo-help-*` CSS custom property tier (`packages/docs/src/editor/styles.css:3363-3455`)

Every `--octo-help-*` variable referenced in the new styles (e.g., `--octo-help-control-size`, `--octo-help-space-2`, `--octo-help-z-overlay`) is consumed but never defined in the docs package. The three-tier pattern `var(--octo-help-*, var(--wk-*, <hardcoded>))` works because the fallback always resolves — but the first tier is dead CSS. Either define `--octo-help-*` at `:root` for future themability, or collapse to the simpler `var(--wk-*, <hardcoded>)` two-tier form.

## Human-verify

1. The self-bind prompt references `npm install -g @mininglamp-oss/octo-cli` and a `curl ... | sh` install script at `https://raw.githubusercontent.com/Mininglamp-OSS/octo-cli/main/install.sh`. Worth confirming that the `octo-cli` npm package and the install script URL are actually published and accessible to end users. This is **not** a merge blocker for this PR — flagging for product confirmation.

- Dialog dismiss: overlay click (`onClick={() => setOpen(false)}`) and close button both work; `e.stopPropagation()` on the dialog prevents accidental dismiss from inner clicks.
- i18n reactivity: `useI18n()` (not the one-shot `t()`) ensures the button/dialog copy updates on runtime locale switch.
- `aria-modal="true"` and `aria-label` on the dialog provide screen-reader accessibility.
- No route/deep-link coupling — the dialog is self-contained with no URL state, so it cannot interfere with document target resolution.
- The `copied` state timer (`window.setTimeout` at line 32) is a 2-second reset. Not cleaned up on unmount, but `setCopied(false)` on an unmounted component is a no-op in React 18+ — harmless.
- The `OnboardingHelp` import in `DocsHome.tsx:32` and its placement in the list header at `:869` is list-state independent — always rendered regardless of docs list contents, matching the PR's stated goal.

## Verdict: APPROVED

Clean, well-structured additive PR. The onboarding help component correctly addresses the discoverability gap without introducing security, data-flow, or accessibility regressions. The P2 (transparent overlay) is a cosmetic improvement worth considering but not blocking.

yujiawei (changes requested):
# Code Review — PR #106 (octo-cli)

Re-review at head SHA `4f2b3d4e8312c2496f45477d1b520cdebfe4aca1`, base `main` (merge-base `29ec0bc`). Round 2 — the previous round's findings were raised against `7e0df2e`.

Verification performed locally, not just by reading the diff:

- `go build ./...`, `go vet ./...`, `go test ./...` — **all pass**.
- Built the binary and confirmed the new gate empirically: `summary` is absent from `octo-cli --help`, absent from `schema --list` (`services` = 9, `operations` = 115), `octo-summary` is absent from `octo-cli skills`, and `octo-cli schema summary.{list,get,result}` still introspects. Withholding happens in `withholdDisabledServices` (`cmd/root.go:65-80`), not in `RegisterServiceCommands` — so the concern raised last round about `cmd/service/service.go:31` iterating `ListServices()` does not apply.
- Cross-checked every newly declared response field name against the backend dependency (`octo-smart-summary#172`, head `6f5be7d`, still open): `ListSummaries` (`internal/api/handler/task.go:481-789`), `GetSummary` (`:878-1097`), `GetResult` (`:1099-1164`), `Citation` (`internal/model/model.go:217-228`), and the status/trigger constants (`:60-77`).
- Reproduced the actual CLI output shape by pointing the binary at a local server returning the backend's real envelope (see P1-1).

---

## 1. Spec compliance

**Spec: ✅**

All eight items from the previous round are present, and I verified each on the built binary rather than in the diff:

| Prior finding | Status | Evidence |
|---|---|---|
| P1-1 RFC3339 documented | ✅ | `summary.json:27-28`; `schema summary.list` renders both descriptions incl. the "malformed values are silently ignored" caveat |
| P1-2 2xx response schemas | ✅ present (content has defects — see P1-1/P1-2 below) | `summary.json:32-70, 73-105, 113-146`; `schema summary.result` emits `response_schema` |
| P1-3 rollout gate | ✅ | `summary.json:11` + `skills/octo-summary/SKILL.md:3`; verified withheld from command tree, `schema --list`, and `skills` |
| P2-1 status/trigger enums | ✅ and **values are correct** | `summary.json:24-25` + `:12`; matches `model.go:60-65` (0-5) and `:71-77` (1=manual, 2=scheduled, 3=agent) |
| P2-2 caveats into `summary` | ✅ | `summary.json:17, 107` now surface via `schema` |
| P2-3 CLAUDE.md | ✅ skill list; domain count correctly left at 9 (summary is disabled, and `EnabledServices()` is indeed 9) |
| P2-4 space-header contract | ✅ | `internal/registry/loader_test.go:176` |
| P2-5 test message split | ✅ | `cmd/service/summary_test.go:43-48` |

No over-build, no unrequested surface, still read-only. One documentation nit: the PR body still describes the three commands as added without mentioning that they now ship withheld behind `x-octo-disabled` — worth a line so the release notes reader isn't surprised.

## 2. Code quality

The response schemas landed, but their *content* does not match what the backend puts on the wire. Since the whole point of adding them was to let an agent write `--jq` without guessing, a schema that names the wrong paths is the failure mode the request was meant to prevent.

### P1-1 — Declared response schemas omit the backend's `{code, message, data}` envelope, so every documented field sits one level deeper than the schema says

`internal/registry/specs/summary.json:32-70` (and `:73-105`, `:113-146`)

The Summary backend wraps every success in its own envelope (`octo-smart-summary internal/api/handler/task.go:216-233`):

```go
func ok(c *gin.Context, data interface{}) {
    c.JSON(http.StatusOK, apiResponse{Code: 0, Message: "ok", Data: data})
}
```

`octo-cli` does not unwrap that. `output.WriteSuccess` places the raw upstream body under the envelope's `data` unchanged, and its only special case is the `{data: [...], pagination: {...}}` shape, which this is not (`internal/output/envelope.go:37-56, 98-121`). There is no `x-octo-unwrap`-style extension in the registry, and `ResponseSchema` is only ever serialized for `octo-cli schema` output — it is never used to validate or reshape a response (`internal/registry/loader.go:206, 385`).

Reproduced against a local server returning the backend's exact body:

```console
$ curl -s $BASE/summary/api/v1/bot/summaries
{"code": 0, "message": "ok", "data": {"total": 1, "items": [...], "attention_count": 0, ...}}

$ octo-cli api GET /summary/api/v1/bot/summaries
{
  "data": {
    "code": 0,
    "data": { "items": [ { "task_id": 7, ... } ], "total": 1, ... },
    "message": "ok"
  },
  "identity": { ... },
  "ok": true
}
```

So the real path is `.data.data.items[]`, while the declared schema documents `total` / `items` as if they were the body's top level. Every existing skill in this repo teaches the single-level convention — `--jq '.data[].id'` (`skills/octo-shared/SKILL.md:171`), `--jq '.data.url'` (`skills/octo-files/SKILL.md:45`), `jq '.data.sheetList'` (`skills/octo-docs/sheet.md:144`) — because every other backend here returns its payload bare. `skills/octo-summary/SKILL.md:23` says only "Results are page-based `{total, items}`" and ships no `--jq` example, so an agent generalizing from `octo-shared` will write `.data.items[]` and get `null` with `ok: true`.

Note the repo already has a precedent for declaring the envelope: `marketplace.json` response schemas `$ref` a `response.Data-…` component whose shape is `{"data": {...}}`, i.e. the raw body including its wrapper.

**Fix:** either declare the envelope in the schema (`{code, message, data: {total, items[]}}`) so `octo-cli schema` describes the real body like `marketplace` does, or keep the payload-only schema and state the extra hop explicitly — plus one worked `--jq '.data.data.items[]'` example in `SKILL.md`. Declaring it matches the existing convention and needs no prose.

### P1-2 — Two declared fields are never returned by the backend

`internal/registry/specs/summary.json:53` and `:121`

- **`items[].updated_at`** (list). `ListSummaries` builds each item explicitly (`octo-smart-summary task.go:730-758`) with `created_at`, `completed_at`, `result_edited_at`, `activity_at` — there is no `updated_at` key. Repo-wide, the only `"updated_at"` response key in that handler file is at `task.go:969`, which is `GetSummary`, not the list. So `summary.get`'s `updated_at` (`summary.json:88`) is correct and the list's is not. (`sort_by=updated_at` still works — that sorts on the column; it just isn't projected into the item.)
- **`task_id`** (result). `GetResult` returns `content, citations, team_citations, total_msg_count, total_token_used, model_version, version, operation_type, operation_note, parent_result_id, generated_at, result_is_edited, result_edited_at` (`task.go:1144-1163`). No `task_id`. The `ModeByPerson` bot fallback path (`writeBotPersonalResult`, `task.go:1165-1190`) also omits it.

A phantom field is worse than an absent one: an agent that keys a decision off `updated_at` or `task_id` gets `null` and no error. Both are one-line deletions.

### P2-1 — Citation schema omits the field an agent needs to quote, and omits the case where `citations` comes back empty for a legitimate caller

`internal/registry/specs/summary.json:123-140`

The declared citation object is `{index, sender, sent_at, source, message_seq}`. The wire object is `model.Citation` (`octo-smart-summary internal/model/model.go:217-228`), which also carries **`content`** (the quoted message text), `channel_id`, and `channel_type`. `content` is not stripped on bot routes — `citationsForRequest` clears only `context_before` / `context_after` (`task.go:1192-1204`). `skills/octo-summary/SKILL.md:39` tells the agent to "attribute claims to the returned citation entry", which in practice means quoting `content`; neither the schema nor the skill mentions it exists.

Two related omissions in the same block:

- **`team_citations`** is always returned (`task.go:1146`) and is undeclared.
- For `SummaryMode == ModeByPerson`, plain `citations` is replaced with `[]` unless the caller *owns* the producing personal result (`callerPlainCitationsVisible`, `task.go:196-214`, applied at `:1138-1142` and `:934-940`). A bot owner reading a colleague's by-person summary therefore gets `citations: []` — indistinguishable from "this summary has no citations" — while `team_citations` still carries the index→author mapping. The op summary documents only the weaker `context_before`/`context_after` omission (`summary.json:107`), so the stronger strip is invisible.

Worth one sentence in the `summary.result` description plus declaring `content` and `team_citations`.

### P2-2 — List-item schema omits most of what the skill tells the agent to select on

`internal/registry/specs/summary.json:44-54`

Declared: 8 fields. Returned: ~25 (`task.go:730-758`), including `topic`, `summary_mode`, `sources`, `participants`, `creator_name`, `total_msg_count`, `completed_at`, `activity_at`, `is_unread`, `needs_attention`, `schedule_id`. `SKILL.md:25` instructs the agent to "choose a candidate using its title, topic, sources, participants, time range, and `task_id`" — of those, `topic`, `sources`, `participants`, and the time range are absent from the schema. `topic` matters most: `keyword` matches title **OR** topic (`task.go:541-544`), so an agent cannot see *why* a result matched. The three `*_count` aggregates at the page level are also undeclared.

### P2-3 — `status 0` is documented as `pending`, but the backend treats it as an aggregate "waiting" state

`internal/registry/specs/summary.json:24`

The description reads `0=pending`. The backend special-cases that value (`task.go:517-527`): `status = 0` **OR** the caller has a pending participant invitation **OR** a pending schedule invitation **OR** an unsubmitted personal result on a multi-participant task. Its own comment says "Waiting is a caller-specific aggregate state". So `--status 0` returns rows whose `status` field is not 0, which reads as a bug from the caller's side. The other five values are exact.

### P2-4 — `CHANGELOG.md` advertises the domain as shipped, though it is withheld

`CHANGELOG.md:10-16`

The `## [Unreleased] / ### Added` entry describes `summary list|get|result` as available and says the domain "ships with the `octo-summary` Skill". With `x-octo-disabled: true` none of the three commands appear in the tree and the skill is not listed by `octo-cli skills`. Whoever cuts the next tag will publish release notes for a feature the binary withholds. Add the withheld-until-backend qualifier here.

### P2-5 — `withheld` marker applied inconsistently across the four skill lists

The repo advertises the skill inventory in four places. This PR added the marker to one of them:

- `CLAUDE.md:79` → `octo-summary` (withheld — backend pending) ✅
- `README.md:291-292` → listed as a normal available skill ❌ (contrast `README.md:279-280`, which spells out "**Temporarily withheld** … not listed by `octo-cli skills`" for `octo-matter`)
- `skills/octo-shared/SKILL.md:219` → listed as a normal available skill ❌ (contrast `:214` for `octo-matter`)
- `octo-cli skills` output → correctly excluded ✅

Since `octo-shared` is the skill an agent loads *first*, the stale line there is the one most likely to send an agent after a command that does not exist.

### P2-6 — Nothing pins the withheld state for `summary`

`cmd/cmd_test.go:428-500` locks all three consumption surfaces for `matter` (tree / global schema list / still-introspectable), and `cmd/skills_test.go:70-79` pins that `octo-matter` drops out of `loadSkills`. Neither is a table, so `summary` is not covered. `cmd/service/summary_test.go:30-48` asserts the subtree *exists*, which is still true at `RegisterServiceCommands` level but is no longer what a user sees — nothing would fail if someone dropped `x-octo-disabled` or the skill's `disabled:` line by accident. Given the flag is load-bearing until the backend ships, one added case in each test is cheap insurance.

## 3. Overall verdict

**CHANGES_REQUESTED** (Spec ✅ AND Quality Approved is required to approve; quality has P1 findings.)

The rollout gate is the right call and it demonstrably works — I verified withholding across all three surfaces rather than trusting the flag. The parameter documentation is now genuinely good, and every enum value and label checks out against the backend constants. What still blocks is the artifact that was the point of the last round: the response schemas describe a body one envelope level off from what the CLI actually prints, and two of their fields don't exist on the wire. Both are small edits, and both are much easier to fix now than after the domain flips on.

## 4. Suggested changes

1. Declare the backend's `{code, message, data}` envelope in all three response schemas (matching the `marketplace.json` precedent), or document the `.data.data` hop and add one worked `--jq` example to `SKILL.md` (P1-1).
2. Delete `items[].updated_at` from `summary.list` and `task_id` from `summary.result` (P1-2).
3. Add `content` (and ideally `team_citations`) to the citation schema, and one sentence noting that plain `citations` is emptied for callers who do not own the producing personal result on by-person tasks (P2-1).
4. Extend the list-item schema with at least `topic`, `sources`, `participants`, and the time range, so the skill's selection guidance is actually discoverable (P2-2).
5. Reword `status: 0` as the aggregate "waiting" state (P2-3).
6. Qualify the `CHANGELOG.md` entry as withheld, and add the withheld marker to `README.md:291` and `skills/octo-shared/SKILL.md:219` (P2-4, P2-5).
7. Add a `summary` case to the disabled-withholding tests (P2-6).

## 5. Additional observations (not blocking)

- The backend dependency is still open at the same head as last round (`octo-smart-summary#172`, `6f5be7d`), so the gate is doing real work rather than guarding a race that already resolved.
- `result` is declared as `"type": "object"` while its own description says it is null until the task completes (`summary.json:89-94`). Strictly this should be `["object", "null"]` under OpenAPI 3.1, but it has no runtime consequence here: the registry never validates responses (`loader.go:206, 385`), so nothing can reject or crash on a `null`. Cosmetic only — noting it so the next reader does not re-derive it.
- Similarly, `citations` cannot arrive as JSON `null`: `GetCitations` returns `[]Citation{}` for an empty/invalid column (`model.go:276-285`) and `citationsForRequest` re-allocates via `make(...)` for every bot request (`task.go:1195`). The strict `"type": "array"` declaration is safe.
- The `get` vs `result` shape asymmetry (body nested under `result.content` in one, at the root in the other) is *faithful* to the backend — `GetSummary` nests, `GetResult` does not. It looks suspicious but is correct; the real problem in that area is the envelope level (P1-1).
- `page_size > 100` is still transmitted and silently reset to 20 server-side (`task.go:496-499`), because the loader drops `minimum`/`maximum`. Same pre-existing class as the dropped `format`; out of scope here.

wanyiwang06:
@yujiawei 二轮意见已全部处理，请复审 `43a71ab`（谢谢逐条对着后端核，指得很准）：

**P1（阻断）**
- **P1-1 信封**：三个 op 的 response schema 现在都声明后端 `{code, message, data}` 信封（对齐 marketplace），`octo-cli schema` 描述的就是真实 body——实际 jq 路径 `.data.data.*`。已实测 `schema summary.{list,result}` 顶层是 `code/data/message`。
- **P1-2 幻影字段**：删掉 `items[].updated_at`（list）和 `task_id`（result）——后端确实不返回。

**P2**
- P2-1 citation 加 `content`（可引用原文，bot 路由不 strip）+ `channel_id/channel_type` + `team_citations`；并注明 by-person 非属主时 plain `citations` 为空。
- P2-2 list-item 补 `topic/sources/participants/summary_mode/creator_name/total_msg_count/completed_at/activity_at/is_unread/needs_attention/schedule_id`。
- P2-3 `status 0` 改述为 caller-specific 聚合「waiting」态。
- P2-4 CHANGELOG 标注 withheld-until-backend。
- P2-5 README + octo-shared/SKILL.md 补 withheld 标记（之前只在 CLAUDE.md）。
- P2-6 三个 disabled-withholding 测试 + loadSkills 测试都加了 summary/octo-summary 用例。
- result schema 用 `["object","null"]`（你提的 cosmetic）。

**自验**：`go build/vet/test ./...` 全绿；binary 实测 schema 显示 `{code,message,data}` 信封 + `data.{total,items}`/`{content,citations,team_citations}`，无 updated_at/task_id 幻影字段。

Jerry-Xin (approved):
## Re-review APPROVE — head `43a71abed48f`

Re-reviewed the single new commit on top of the previously-approved `4f2b3d4e8312`. `compare` is `ahead:1, behind:0` (no rebase) — a clean spec/docs/test-only delta across 6 files. `go build ./...`, `go vet ./...`, and `go test ./cmd ./internal/registry/... ./internal/output/...` all green locally at this head.

### True delta (files/lines)
- `internal/registry/specs/summary.json` (+83/-48) — response-schema rework + enum/param-doc fixes
- `CHANGELOG.md`, `README.md`, `skills/octo-shared/SKILL.md` — withheld-until-backend qualifier added
- `cmd/cmd_test.go` (+14), `cmd/skills_test.go` (+6) — `summary` added to the disabled-service test matrix (withheld from tree + `schema --list`, still introspectable, skill excluded) — mirrors the existing `matter` pattern

### Resolves the prior CHANGES_REQUESTED (byte-verified at head)
- **Response envelope** — all three success schemas are now wrapped in `{code, message, data}`, with an explicit note that octo-cli nests this again, so the effective path is `.data.data.*`. Descriptions call this out on every 200.
- **Phantom fields removed** — `items[].updated_at` (list) and `task_id` (result) are gone; both were never on the wire.
- **Citation completeness** — citation objects now declare `content` (the quotable text, not stripped on bot routes), `channel_id`, `channel_type`; `team_citations` is declared; the by-person empty-`citations` caveat is documented on `summary.result`.
- **List fields** — `topic`, `summary_mode`, `sources`, `participants`, `creator_name`, `total_msg_count`, `completed_at`, `activity_at`, `is_unread`, `needs_attention`, `schedule_id` added.
- **`status 0`** — relabeled `pending → waiting` with the caller-specific aggregate semantics spelled out; `page_size >100 → silently reset to 20` documented; `result` made nullable (`["object","null"]`).
- **Withheld notice** — CHANGELOG/README/octo-shared now state the domain ships withheld behind `x-octo-disabled` (skill `disabled: true`) pending the backend, still introspectable via `schema summary.*`.

### On the envelope point specifically
Independently: the response schema here is **descriptive introspection metadata only** — `ResponseSchema` is assigned once in the loader and never read again; runtime output handling unwraps envelopes generically by probing keys and never consults the declared schema shape. So a wrapper mismatch could not break parsing; it was a doc-accuracy issue. That said, declaring the envelope was the right call: these schemas exist to guide agent `--jq`, the CLI double-nests to `.data.data.*` (genuinely surprising), and `marketplace.json` already sets the precedent of declaring the wrapper. The explicit jq-path note makes this more accurate than the bare-payload style used elsewhere. Correct and well-scoped.

### Domain still withheld
`x-octo-disabled: true` is unchanged and skill `disabled: true` — the three commands stay out of the command tree and `schema --list`, so there is no live-command or 404 risk regardless of backend readiness.

### 🔴 Blocking
None.

### 🟡 Non-blocking
- `check-sprint` is red (sprint governance, non-code) — resolve per team process before merge; unrelated to this diff.

### ✅ Highlights
Every prior finding addressed with byte-level accuracy, docs match the wire, disabled-service test coverage extended to match the established `matter` pattern, and the rollout stays gated. Ready to merge once governance CI is satisfied.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #131 (octo-docs-backend)

Reviewed at head `734fb907`. Scope: full-text document search API with permission down-push, Kafka index producer (Redis Stream → Kafka migration), OpenSearch client, ACL epoch fix in `acceptInvite.ts`, and collab server indexing hook.

## Summary

This PR adds a full-text document search API backed by OpenSearch with a clean permission down-push architecture: MySQL computes the caller's visible doc_id set at query time (`listVisibleDocIdSet`), then pushes it as a `terms` filter into the OpenSearch query. No per-hit MySQL re-check — the visible set is the authoritative gate. The index producer migrates from Redis Stream to Kafka (`kafkajs` lazy singleton), and the `acceptInvite.ts` ACL path is fixed to publish the epoch refresh signal after transaction commit, matching the other ACL mutation paths.

The architecture is sound and the code is well-tested: route-level tests, unit tests for the OS query builder, and an e2e integration script with a soft-delete leak regression test. No P0/P1 issues found; two P2 items and two nits below.

## Verification

- ✅ **Permission down-push correct** — `listVisibleDocIdSet` (`src/db/repos/docMetaRepo.ts:396`) computes visibility as owner OR doc_member OR (space-member AND share_scope=anyone), filtered by `status=1`. The e2e script (`scripts/e2e-p4-verify.mts`) confirms soft-deleted docs are excluded from the visible set even when OS holds a stale `status=1` copy.
- ✅ **SQL injection surface clean** — all user-supplied values (`uid`, `spaceId`, `ownedBots`) flow through parameterized queries with correct bind ordering.
- ✅ **Fail-closed on search errors** — `searchDocsHandler` (`src/api/routes/docs.ts:413`) catches OS errors and returns 503; if `config.search.enabled` is false, returns 503 without connecting.
- ✅ **Kafka fire-and-forget safe** — `enqueueDocIndex` (`src/search/docIndexQueue.ts:25`) swallows producer failures, returns false, and never disturbs the authoritative collab document store.
- ✅ **TLS default secure** — `opensearchTlsRejectUnauthorized` defaults to `true`; the escape hatch is scoped to https nodes only (`src/search/osClient.ts:42`).
- ✅ **Epoch publish post-commit** — `refreshAndPublish` in `acceptInvite.ts` is captured in-transaction but fired after commit, preventing phantom reindex signals on rollback.
- ✅ **Kafka singleton safe** — `getKafkaProducer()` (`src/db/kafka.ts:25`) uses lazy init with shared in-flight connect; failure clears cached promise for retry.
- ✅ **Shutdown cleanup** — `closeKafkaProducer()` wired into graceful shutdown at `src/index.ts:19`.
- ✅ **Test coverage adequate** — route-level tests cover permission down-push (member/non-member), empty-set short-circuit, pagination clamping, docType filter, and OS error → 503. Unit tests cover OS query construction. e2e script covers real MySQL + OS integration including ik Chinese tokenization.

## Findings

No P0/P1 issues. Two P2 items and two nits below.

### P2 — MySQL error path in search handler inconsistent with OS error path (`src/api/routes/docs.ts:399`)

`listVisibleDocIdSet` is not wrapped in try/catch. A MySQL failure (connection pool exhausted, timeout, transient error) propagates as an unhandled async rejection to the Express error handler, typically returning 500. The OpenSearch path below already catches its own errors and returns a clean 503 (`search unavailable`). Wrapping the MySQL call similarly would give a consistent, non-information-leaking response on either backend failure.

### P2 — Stale JSDoc on `listVisibleDocIdSet` (`src/db/repos/docMetaRepo.ts:396`)

The JSDoc states "Deliberately does NOT enumerate anyone_in_space (space-share) docs", but the code below DOES enumerate them when `isSpaceMember === true` (via `OR m.share_scope = ANYONE`). The route handler's own comment at `src/api/routes/docs.ts:354` correctly describes this enumeration. Update the JSDoc to match the code — e.g., "Enumerates the FULL visible set including space-share docs for confirmed members".

### Nit — Duplicated `isIndexableDocument` logic (`src/collab/server.ts:55`)

`isIndexableDocument` is an exact duplicate of `isSearchIndexedDoc` in `src/search/docIndexQueue.ts` (both parse the document name and check `kind === 'document' || kind === 'whiteboard'`). Since `collab/server.ts` already imports `enqueueDocIndex` from `docIndexQueue.ts`, importing `isSearchIndexedDoc` instead of duplicating keeps one source of truth for the indexable-type gate.

### Nit — `closeKafkaProducer` missing error guard (`src/db/kafka.ts:42`)

`closeKafkaProducer` awaits `p.disconnect()` without try/catch. If disconnect throws (broker unreachable during shutdown), the error propagates to the signal handler and can prevent the subsequent `closePool()` from running. A try/catch with a `console.warn` would match the best-effort style used elsewhere.

## Human-verify

1. **OpenSearch index mapping** — the `ik` analyzer, `multi_match` fields (`title^2`, `body`), and `doc_id` keyword type are assumed to be configured in the index mapping deployed by ops. Not verifiable from this repo alone. **Not** a merge blocker.
2. **Kafka topic provisioning** — `octo.docindex.v1` must exist with appropriate partition count and retention before the producer goes live. The `KAFKA-MIGRATION-NOTES.md` documents this as out-of-scope. **Not** a merge blocker.
3. **`octo-doc-indexer` consumer compatibility** — the Kafka message shape `{documentName, kind, ts}` must match what the consumer expects. The `ts` field is documented as diagnostic-only (not for ordering). **Not** a merge blocker for this producer-side PR.

- Route registration: `POST /docs/search` is mounted before `/:docId` param routes, preventing path-parameter shadowing.
- `normalizeTypeFilter` validates `doc_type` against a known enum set before passing to the OS query.
- `pageSize` is clamped to `config.search.pageSizeMax` (default 50), preventing unbounded OS queries.
- `track_total_hits: true` ensures accurate pagination totals.
- `_source` filtering limits OS response to needed fields (`doc_id`, `title`, `doc_type`, `updated_at`, `space_id`).
- Gray-release flags `SEARCH_ENABLED` and `SEARCH_INDEX_ENABLED` both default to `false`.

## Verdict: APPROVED

Solid implementation of the permission down-push pattern with correct security posture (MySQL as source of truth, fail-closed on errors, no per-hit re-check). The Kafka migration is clean and well-documented. The two P2 items (MySQL error handling consistency, stale JSDoc) are non-blocking and straightforward to address.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. Permission down-push architecture is correct; MySQL as source of truth for visibility, OS for text matching. Soft-delete leak regression test confirms the fix. Two P2 nits (MySQL error path consistency, stale JSDoc) are non-blocking.

yujiawei (approved):
# Code Review — PR #106 (octo-cli)

Re-review at head SHA `43a71abed48f42d511ad58e7d5867e1c164eb7ed`, base `main` (merge-base `29ec0bc`). Round 3 — the previous rounds were raised against `7e0df2e` and `4f2b3d4`.

Verification performed locally, not just by reading the diff:

- `go build ./...`, `go vet ./...`, `go test ./... -count=1` — **all pass** (12/12 packages).
- Built the binary and reproduced the actual wire shape against a local server returning the backend's exact envelope, to confirm the newly declared `{code, message, data}` schema is accurate (see below).
- Re-confirmed the withholding gate empirically on this head: `summary` absent from `octo-cli --help`, `schema --list` reports `services=9 / operations=115` with no `summary`, `octo-cli skills` does not list `octo-summary`, and `octo-cli schema summary.{list,get,result}` still introspects.
- Cross-checked every declared response field against the backend dependency (`octo-smart-summary#172`, head `6f5be7d`, **still open**): `ListSummaries` (`internal/api/handler/task.go:730-759`), `GetSummary` (`:878-1097`), `GetResult` (`:1099-1163`), `writeBotPersonalResult` (`:1165-1190`), `model.Citation` / `model.TeamCitation` (`internal/model/model.go:217-252`), and `SummaryTask.ID` (`:126`).

---

## 1. Spec compliance

**Spec: ✅**

Six of the seven changes requested last round landed, and I verified each against the built binary and the backend rather than in the diff:

| Prior finding | Status | Evidence |
|---|---|---|
| P1-1 declare the `{code, message, data}` envelope | ✅ **and empirically correct** | `summary.json:37-45, 96-102, 144-150`; reproduced below |
| P1-2 drop phantom `items[].updated_at` / `result.task_id` | ✅ | neither key present; `summary.get`'s `updated_at` (`:112`) correctly retained — the backend does return it at `task.go:969` |
| P2-1 citation `content` + `team_citations` + by-person caveat | ✅ | `summary.json:164-166, 170-173, 133-134`; matches `model.Citation` field-for-field |
| P2-2 richer list-item schema | ⚠️ mostly — **`time_range_start` / `time_range_end` still missing** (P2-2 below) | `summary.json:51-70` |
| P2-3 reword `status: 0` as the aggregate state | ✅ and accurate | `summary.json:24` matches the special-case at `task.go:517-528` |
| P2-4 / P2-5 withheld markers in CHANGELOG / README / octo-shared | ✅ all four lists now consistent | `CHANGELOG.md:16-19`, `README.md:291-293`, `skills/octo-shared/SKILL.md:219`, `CLAUDE.md:79` |
| P2-6 pin the withheld state | ✅ | `cmd/cmd_test.go:449-451, 478-480, 513-517`, `cmd/skills_test.go:80-82, 87-89` |

The envelope fix is correct, and I confirmed it end-to-end rather than trusting the declaration. `splitPagination` only flattens a `{data: [...], pagination: {...}}` body (`internal/output/envelope.go:98-115`), which this is not, so the whole backend body lands under the CLI's own `data`:

```console
$ octo-cli api GET /summary/api/v1/bot/summaries
{ "data": { "code": 0, "data": { "items": [ { "task_id": 7, ... } ], "total": 1, ... }, "message": "ok" }, "ok": true }

$ octo-cli api GET /summary/api/v1/bot/summaries --jq '.data.data.items[].task_id'
7
```

Still read-only, still three operations, no new flags or commands beyond what the previous round asked for. No over-build.

## 2. Code quality

**Quality: Approved**

Everything that blocked the last two rounds is resolved and verified. What remains is schema completeness and typing precision — all P2, none of it reachable by a user today because the domain is withheld behind a gate I re-verified on this head. The natural place to sweep these is the same one-line follow-up that flips `x-octo-disabled` off; that is the last moment they are free.

### P2-1 — `["object", "null"]` is silently dropped by the loader, so `result` now has *no* declared type at all

`internal/registry/specs/summary.json:114`

`SchemaInfo.Type` is a plain `string` (`internal/registry/loader.go:166`) and is populated through `stringOf` (`:585-588`), which type-asserts to `string` and yields `""` for a JSON array. With `omitempty`, the key vanishes:

```console
$ octo-cli schema summary.get   # .data.response_schema…data.properties.result
{
  "description": "Current visible result; null until the task completes.",
  "properties": { "content": { "description": "Summary body (markdown).", "type": "string" } }
}
```

At the previous head this field read `"type": "object"`. The OpenAPI 3.1 union is strictly more correct as source text but strictly *less* informative as projected output, so this is a small regression in what an agent can introspect. **This one is my error to own** — I suggested the union last round and labelled it cosmetic without checking that the loader can represent it. Either revert to `"type": "object"` and state the null case in the `description` (which is already there), or extend `SchemaInfo` to carry union types. The description alone conveys it; the empty `type` does not.

### P2-2 — list-item schema still omits the time range the skill tells the agent to select on

`internal/registry/specs/summary.json:51-70` vs `octo-smart-summary task.go:741-742`

`skills/octo-summary/SKILL.md:25` says to "choose a candidate using its title, topic, sources, participants, **time range**, and `task_id`". `topic`, `sources` and `participants` were added this round; `time_range_start` / `time_range_end` were not. `created_at` / `completed_at` / `activity_at` are not substitutes — they describe when the *task* ran, not the conversation window the summary covers, which is the field an agent actually needs to decide "does this summary cover the period I was asked about?". Two lines.

### P2-3 — two nullable list fields are declared as non-null scalars, and their descriptions say "empty" where the wire says `null`

`internal/registry/specs/summary.json:65` and `:69`

- `completed_at` → `"type": "string"`, described as "empty until completed". The backend builds it as `*string` and leaves it nil when there is no result (`task.go:692-697, 749`), so the wire value is `null`, not `""`.
- `schedule_id` → `"type": "integer"`, described as "Present for scheduled summaries". The key is *always* present; it is `null` for unscheduled tasks (`task.go:720-723, 738`).

Reproduced against a server returning the backend's exact body:

```console
$ octo-cli api GET /summary/api/v1/bot/summaries --jq '.data.data.items[0] | {completed_at, schedule_id}'
{ "completed_at": null, "schedule_id": null }
```

An agent writing `select(.completed_at != "")` — the natural reading of "empty until completed" — filters nothing. Reword both to name `null` explicitly. (Given P2-1, prefer a description over a union type here.)

### P2-4 — three array fields are declared with no `items`, including the one the by-person caveat points the agent at

`internal/registry/specs/summary.json:60-61` (`sources`, `participants`) and `:170-173` (`team_citations`)

All three are bare `{"type": "array"}`. `team_citations` is the sharpest case: `summary.result`'s own summary text (`:133`) tells the caller that for a by-person summary they do not own, plain `citations` comes back empty and `team_citations` "still carries the index→author mapping" — but the schema gives no way to discover that the mapping is `{index, user_id, user_name}` (`model.TeamCitation`, `internal/model/model.go:246-252`). The whole point of declaring response schemas was to remove that guess. `sources` items are `{source_type, source_id, source_name}` (`task.go:681-687`) and `participants` items are `{user_id, user_name, status[, confirmed_at]}` (`:661-670`); `SchemaInfo.Items` already exists in the loader (`loader.go:169`) and nests correctly, as the `citations` block on the same operation demonstrates.

### P2-5 — `summary.get`'s schema is now markedly thinner than `summary.list`'s, and omits the citations that operation does return

`internal/registry/specs/summary.json:98-121`

`summary.get` declares 10 top-level fields; the backend returns roughly 25, including `summary_mode`, `time_range_start`/`_end`, `sources`, `participants`, `error_message`, `result_id`, `result_is_edited`, `permissions`, `personal_result`, and `members`. More pointedly, its `result` object declares only `content`, while the backend puts `citations`, `team_citations`, `total_msg_count`, `version` and `generated_at` in there too — under the *same* by-person privacy gate documented on `summary.result` (`task.go:930-936` mirrors `:1136-1142`). `SKILL.md:34` says "Use `get` for metadata plus the current visible result", so an agent that follows `get` cannot discover from the schema that citations are available there at all. Either enrich `get`'s schema to match the effort now spent on `list`/`result`, or make `SKILL.md` explicit that citations are only discoverable via `result`.

### P2-6 — `summary.result` omits the two fields that tell an agent the body was human-edited, and the personal-result fallback shape

`internal/registry/specs/summary.json:151-174` vs `task.go:1155-1162` and `:1165-1189`

`GetResult` always returns `result_is_edited` and `result_edited_at`. For a skill whose stated purpose is grounded citation, "a human edited this body after it was generated" is material: the `[n]` markers in an edited `content` are no longer guaranteed to line up with the `citations` array. Neither the schema nor `SKILL.md` mentions it.

Separately, on the bot route a `ModeByPerson` task with no team result falls back to `writeBotPersonalResult`, which returns a **different** shape — `result_source: "personal"` (`task.go:1178`), `personal_result_id`, `personal_version_id`, `team_citations: []` — and this is a path bot callers will hit routinely. The declared `{content, citations, team_citations}` is a valid subset of both branches, so nothing breaks, but `result_source` is the discriminator an agent would want and it is undeclared.

## 3. Overall verdict

**APPROVED** (Spec ✅ and no P0/P1 remaining.)

Three rounds in, every blocking finding is closed and independently re-verified: the gateway prefix, the space-header suppression, all ten query parameters, every enum value, the withholding gate across all four consumption surfaces, and now the response envelope reproduced against a real server. The remaining items are schema-completeness polish on a domain that is not reachable from the shipped binary, so continuing to iterate on them here trades real ping-pong cost for no user-visible gain.

Concrete ask instead of another round: fold P2-1 through P2-6 into the same follow-up PR that removes `x-octo-disabled` from `summary.json` and `disabled: true` from the skill. That flip is the moment the schema starts being load-bearing, and it is a natural checkpoint. I would treat P2-4 (`team_citations` items) and P2-2 (time range) as the two worth not skipping, since both are named in the skill's own instructions to the agent.

## 4. Suggested changes

1. Revert `result` to `"type": "object"` and let the existing description carry the null case, or teach `SchemaInfo` union types (P2-1).
2. Add `time_range_start` / `time_range_end` to the list-item schema (P2-2).
3. Reword `completed_at` and `schedule_id` to say `null`, not "empty" / "present only when…" (P2-3).
4. Give `sources`, `participants`, and `team_citations` an `items` schema (P2-4).
5. Bring `summary.get` up to parity, notably its `result.citations` / `result.team_citations` (P2-5).
6. Declare `result_is_edited` / `result_edited_at`, and note the `result_source: "personal"` fallback branch (P2-6).

## 5. Additional observations (not blocking, and not introduced by this PR)

- **`schema --list` error and domain paths leak disabled services.** `runSchemaList` builds `payload["services"]` and the unknown-service hint from `reg.ListServices()` (`cmd/schema.go:64, 72`), and only the global-discovery branch swaps in `EnabledServices()` (`:85`). So `octo-cli schema --list nosuch` replies `known services: [bot docs event file group html marketplace matter message summary thread]`, and `schema --list docs` echoes the same unfiltered list. `loader.go:91-92` explicitly says `ListServices` is "deliberately NOT filtered — use EnabledServices / EnabledOperations for a caller-facing view", so these two call sites contradict that contract. **This predates this PR** — `matter` has leaked the same way since it was disabled — so it is not this author's to fix, but this PR does double the exposure. Worth a separate one-line change.
- **`task_id` carries no 64-bit precision risk.** `SummaryTask.ID` is `int64` with `autoIncrement` (`model.go:126`), not a Snowflake, so the declared bare `integer` cannot realistically exceed 2^53 and no `format: int64` or string encoding is warranted. Checked because the declared type invites the question.
- **The `get` vs `result` nesting asymmetry is faithful**, not a bug: `GetSummary` nests the body under `result`, `GetResult` returns it at the root. The schemas now mirror that correctly.
- **`keyword` is interpolated into `title LIKE ? OR topic LIKE ?` with `%` wrapping and no wildcard escaping** (`task.go:540-542`), so a keyword containing `%` or `_` acts as a wildcard. Parameterised, so no injection — purely a surprising-match edge. Backend-side and out of scope here.
- The backend dependency is still open at the same head as both prior rounds (`octo-smart-summary#172`, `6f5be7d`), so the gate continues to do real work rather than guarding an already-resolved race.
- `page_size > 100` is still transmitted and silently reset to 20 server-side (`task.go:496-500`) because the loader drops `minimum`/`maximum`; the parameter description now says so explicitly, which is the right mitigation short of fixing the loader. Same pre-existing class as the dropped `format`.

PR #107, opened 2026-07-27
fix(release): pin GoReleaser to requested tag

liuooo:
## What

Promoting a commit from `v0.9.0-next.3` to `v0.9.0` leaves both tags on the
same commit. GoReleaser selected the prerelease tag, so the stable release
workflow produced prerelease-named archives and embedded the prerelease
version. npm packaging then correctly refused to publish the mismatched files.

## How

- Pass the requested workflow tag through `GORELEASER_CURRENT_TAG`.
- Keep the existing single-build release graph and CI evidence gate unchanged.

## Testing

- `actionlint .github/workflows/release-publish.yml`
- `git diff --check`
- Reproduced locally with both tags on commit `29ec0bc`; GoReleaser generated
  all six `octo-cli_0.9.0_*` archives with `version=0.9.0`.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -90,6 +90,10 @@ jobs:
           version: "~> v2"
           args: release --clean --skip=publish,announce
         env:
+          # A stable tag may promote the exact commit used by a prerelease tag.
+          # GoReleaser otherwise chooses one of the colocated tags itself and
+          # can embed/archive the prerelease version during a stable release.
+          GORELEASER_CURRENT_TAG: ${{ inputs.tag }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       # Run-scoped artifact consumed by both release-upload and publish-npm in


mochashanyao (approved):
[Octo-Q · automated review]

---

## Review: octo-cli PR#107 — fix(release): pin GoReleaser to requested tag
**Head SHA:** `aeafd03066f7ad4a077e56879097b9f376205e0d`
**Diff scope:** 1 file, +4 / −0 (`.github/workflows/release-publish.yml`)

---

### 1. Verification Summary

✅ **Fix is correct and minimal.** Setting `GORELEASER_CURRENT_TAG: ${{ inputs.tag }}` in the goreleaser step env pins the version derivation to the user-requested tag, which directly prevents the described bug (GoReleaser auto-selecting a colocated prerelease tag when both `v0.9.0-next.3` and `v0.9.0` point to the same commit).

---

### 2. Data Flow Trace

| Consumer | Upstream Source | Flow Verified |
|---|---|---|
| `GORELEASER_CURRENT_TAG` (env) | `workflow_dispatch` input `inputs.tag` | ✅ Validated as `^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?$` at step "Validate tag input" (line ~68) before reaching goreleaser step |
| `{{.Version}}` in `.goreleaser.yaml` ldflags + archive `name_template` | `GORELEASER_CURRENT_TAG` env → GoReleaser internal tag resolution | ✅ GoReleaser v2 uses this env var to override git-derived tag; `{{.Version}}` strips the `v` prefix automatically |
| `dist/*.tar.gz` filenames (e.g., `octo-cli_0.9.0_linux_amd64.tar.gz`) | `{{.Version}}` from above | ✅ Correct — consumed by `release-upload` and `publish-npm` downstream jobs |

**Data flows cleanly from validated input → env → GoReleaser → artifacts. No short-circuit, no empty-value risk.**

---

### 3. Findings

**None.** No P0/P1/P2 issues identified.

---

### 4. Diff-scope Analysis (R2)

- **New:** The 4 added lines (3 comment + 1 env var). No existing behavior modified.
- **Not amplified:** The checkout step already fetches `refs/tags/${{ inputs.tag }}`; this env var is additive and only constrains GoReleaser's internal tag resolution.

---

### 5. Blind-spot Checklist (R5)

**C1 — Dual-path parity:** CLEAR. Only one `goreleaser-action` invocation exists in the repo (this step in `release-publish.yml`). No other workflow calls goreleaser, so no parity gap.

**C2 — Control-flow ordering / nesting:** CLEAR. The env var is scoped to the single goreleaser step. No other step or job inherits it.

**C3 — Authorization boundary:** N/A. CI workflow; no new endpoint or tool exposure.

**C4 — Authorization lifecycle:** N/A.

**C5 — Build/runtime parity:** CLEAR. The `--skip=publish,announce` flags mean goreleaser only builds locally; artifacts are uploaded separately by `release-upload`. The env var correctly affects only the build step's version embedding.

**C6 — Governance/docs:** N/A. No policy or doc changes.

---

### 6. Additional Notes

The inline comment explaining *why* the env var is needed is well-placed and will help future maintainers understand the colocated-tag pitfall without needing to dig through commit history.

---

### 7. Cross-round Blocker Review (R6)

N/A — first review round.

---

**[Octo-Q] verdict: APPROVE** — Correct, minimal fix with clear data-flow traceability and no side effects.

Jerry-Xin (approved):
Summary: This PR is in scope for `octo-cli` and correctly addresses the release workflow’s colocated stable/prerelease tag ambiguity without changing the existing build, release upload, or npm publish graph.

✅ Highlights
- 🔵 Suggestion / Positive: Setting `GORELEASER_CURRENT_TAG` at `.github/workflows/release-publish.yml:96` aligns GoReleaser’s `.Version` with the validated workflow input, which matches the archive naming expected by `.goreleaser.yaml` and the npm packaging workflow.
- 🔵 Suggestion / Positive: The existing pre-checkout tag validation at `.github/workflows/release-publish.yml:64` keeps the new env var constrained to v-prefixed semver input before it reaches GoReleaser.
- 🔵 Suggestion / Positive: The downstream npm path already resolves package version from the same requested tag and verifies checksums before packaging, so this change preserves the single-build release artifact model.

No blocking or non-blocking issues found. I did not run the workflow locally; review was based on the checked-out workflow and related release packaging context.

yujiawei (approved):
# Code Review — PR #107 (octo-cli)

Reviewed at head SHA `aeafd03066f7ad4a077e56879097b9f376205e0d`, merge-base `29ec0bc`. Diff scope: 1 file, +4 / −0 (`.github/workflows/release-publish.yml`).

## 1. Spec compliance

**Spec: ✅**

The stated goal is "make the stable release workflow build the tag the operator asked for, without changing the release graph or the CI-evidence gate".

- **Missing work:** none. The single root cause is addressed at the single place where it manifests.
- **Extra work:** none. No new inputs, jobs, permissions, or `needs` edges; the job graph (`publish → build → release-upload → publish-npm → cleanup`) and the `validate_run_id` evidence gate are byte-for-byte unchanged.
- **Deviation:** none. The value pinned is `inputs.tag`, i.e. exactly the tag already validated pre-checkout (`.github/workflows/release-publish.yml:64-72`) and already used for `actions/checkout` (`:77`), `gh release upload` (`:126`) and `publish-npm` (`:139`). The build step is now consistent with every other consumer of the same input.

## 2. Code quality

**Quality: Approved**

### Root cause and fix — verified, not assumed

GoReleaser resolves the current tag in a fixed precedence order, and `GORELEASER_CURRENT_TAG` is checked **first**, ahead of any git-derived guess (`internal/pipe/git/git.go`, `getTag()`); `ctx.Version` is then simply `strings.TrimPrefix(currentTag, "v")`. The fallback that produced the bug is `git tag --points-at HEAD` sorted by the default `git.tag_sort: -version:refname`. Reproduced against the actual commit:

```
$ git tag --points-at 29ec0bc --sort=-version:refname
v0.9.0-next.3
v0.9.0
```

The prerelease tag sorts first, so `.Version` became `0.9.0-next.3` — which explains both the misnamed archives and the embedded version via the `ldflags` `-X ...cmd.Version={{.Version}}` in `.goreleaser.yaml:18`. Setting the env var removes the guess entirely. Confirmed correct.

Two follow-on behaviours I checked rather than took on trust:

- **The dirty/ref validation still passes.** GoReleaser's `validate()` runs `git describe --exact-match --tags --match <CurrentTag>`. Since `build` checks out `refs/tags/${{ inputs.tag }}` with `fetch-depth: 0` (`:74-78`), HEAD carries that exact tag, so the pin does not defeat the guard — and it would still fail fast (`ErrWrongRef`) if a tag were ever pinned that does not point at the checked-out commit. Good property to keep.
- **Changelog / previous-tag resolution is unaffected in practice.** `GORELEASER_PREVIOUS_TAG` is left unset, so `getPreviousTag()` may resolve to the colocated prerelease tag and yield an empty changelog — but the step runs with `--skip=publish,announce` and release notes are produced by the reusable `publish` job (`:45-53`), so GoReleaser's changelog output is never consumed. No action needed; noting it so it isn't mistaken for a latent bug later.

Also checked: no injection surface (the input is interpolated into an `env:` value, not into a `run:` script, and is regex-validated before checkout); `actionlint` and the workflow-sanity checks pass; no other GoReleaser invocation exists in the repo (`npm-publish.yml` only consumes the `goreleaser-dist` artifact), so there is no second site needing the same pin.

### P2 — mismatched archives can still reach the public Release before anything notices

Not introduced by this PR, and not a merge blocker, but this PR is the natural place to record it: the *only* thing that cross-checks "what GoReleaser built" against "what tag we are releasing" is `npm/scripts/prepare-packages.js:108-109`, which fails on a missing `octo-cli_<version>_<os>_<arch>.tar.gz`. That check lives in `publish-npm`, which `needs: [publish, build, release-upload]` (`:135`) — i.e. it runs *after* `release-upload` has already attached `dist/*.tar.gz` to the public GitHub Release (`:128`). That ordering is exactly why the reported incident left a stable Release holding prerelease-named assets while npm refused.

Inside `build`, `if-no-files-found: error` (`:108`) only proves *some* files exist, not that they carry the requested version. A three-line assertion at the end of `build` would move detection ahead of the irreversible step:

```yaml
      - name: Assert built version matches requested tag
        env:
          TAG: ${{ inputs.tag }}
        run: |
          set -euo pipefail
          version="${TAG#v}"
          built=$(jq -r .version dist/metadata.json)
          if [ "$built" != "$version" ]; then
            echo "::error::GoReleaser built '$built' but the requested tag is '$TAG'."
            exit 1
          fi
```

(Equivalently: `test -f "dist/octo-cli_${version}_linux_amd64.tar.gz"`.) Cheap, and it makes any future recurrence — a different resolution path, a config change, a GoReleaser v2 minor bump under the floating `version: "~> v2"` — fail before publishing anything.

### P2 — the same ambiguity remains for runs that don't go through this workflow

The env pin fixes CI, which is what matters; but a maintainer running `goreleaser release` locally on a commit with colocated tags still hits the original auto-detection. If you want the repo itself to be unambiguous, `.goreleaser.yaml` supports `git.prerelease_suffix` (fed to `versionsort.suffix`, so prerelease tags sort *below* the stable one) and `git.ignore_tags`. Worth considering as a complement — **not** a replacement: the env pin is strictly better for the workflow because it encodes operator intent, and deliberately releasing a prerelease tag must keep working.

### Nit / process

- The added comment (`:93-95`) explains *why* rather than *what*, and matches the existing commentary style in this file. Good.
- `check-sprint` is currently red on this PR ("No linked issue / Sprint not set" family of errors from the reusable check, current sprint `Sprint W30`). Unrelated to the diff, but it will need clearing before merge if that check is required.
- Operational check, already fine: the live `v0.9.0` Release currently carries the correct `octo-cli_0.9.0_*` assets plus `checksums.txt`, so no leftover prerelease-named assets need cleanup.

## 3. Verdict

**APPROVED** — minimal, correctly targeted, verified against GoReleaser's actual tag-resolution order and reproduced against the real colocated tags. Both P2 items are hardening suggestions and can land separately.

## 4. Suggested follow-ups

1. Add the version assertion to the `build` job so a version mismatch fails *before* `release-upload` mutates the public Release.
2. Decide whether to also make tag selection unambiguous at the `.goreleaser.yaml` level for non-workflow runs.
3. Consider a short release runbook note on tag promotion (there is no release doc in the repo today), since "promote the prerelease commit to a stable tag" is now a supported path with a non-obvious history.

## 5. Coverage note (what this review did and did not cover)

Covered: the full `release-publish.yml` job graph, `.goreleaser.yaml`, `npm-publish.yml`'s consumption of the artifact, `prepare-packages.js` version handling, GoReleaser's upstream `getTag`/`validate`/`getPreviousTag` implementation, and the live tag/Release state.

Not covered (would require running the workflow): an end-to-end stable-release dry run on a commit with colocated tags — the PR reports this was reproduced locally, and I did not independently execute it. Also not covered: the org-level reusable workflows (`Mininglamp-OSS/.github@v1`) used by `publish`, `check-sprint`, and `workflow-sanity`, which are outside this repo.

PR #115, opened 2026-07-30
feat(loop): expose complete Fleet business commands

### LINKED ISSUE #116: feat(loop): expose the complete Fleet public API
## Summary

Complete the octo-cli Loop integration so all supported Fleet business resources are available through the stable public `/v1` API.

## Scope

- Add workspace, label, project, comment, attachment, skill, runtime, and autopilot commands.
- Keep task, execution, expert, expert template, and expert team commands aligned with the same public contract.
- Support explicit workspace selection for human credentials.
- Preserve workspace-bound Loop credential behavior.
- Keep the CLI registry synchronized with the Fleet OpenAPI contract.

## Acceptance criteria

- The Loop registry exposes all documented Fleet public operations.
- Commands route through the configured Fleet endpoint under `/v1`.
- Public vocabulary does not expose legacy issue, agent, or squad names.
- Unit and command integration tests pass.

liuooo:
## What

Expose the complete Fleet public Loop API through octo-cli, expanding the existing task and expert integration to the remaining business resources.

## How

- Extend the embedded Loop OpenAPI registry from 61 to 136 operations.
- Add workspace, label, project, comment, attachment, skill, runtime, and autopilot command groups.
- Add explicit `--workspace-id` selection for human credentials while preserving workspace-bound Loop credentials.
- Keep Fleet routing on `/v1/*` and avoid operation ID collisions with Marketplace skill commands.

## Tests

- `go test ./...`
- Command-tree coverage for every new resource group.
- Fleet endpoint, workspace header, multipart, and registry contract coverage.

## Checklist

- [x] Tests added or updated
- [x] Documentation and embedded API contract updated
- [x] Conventional Commits used

Closes #116

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,14 +34,6 @@ Key properties:
 - **Agent-first output.** A stable JSON envelope with identity, data,
   pagination, and rate-limit metadata; a small fixed error taxonomy.
 
-### Reusable Go client
-
-The public `github.com/Mininglamp-OSS/octo-cli/client` package is independent
-from Cobra and CLI profile storage. A single client instance can route both
-Octo and Fleet/Loop requests through service-specific endpoints and tokens.
-Loop DTOs and business compatibility aliases deliberately stay outside this
-transport package.
-
 ## Domains
 
 | Domain    | Ops | Purpose                                                        |
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@ deterministic taxonomy. There is no interactive I/O.
 
 ## Architecture
 
-octo-cli is **metadata-driven**. The entire command tree — 104 operations
-across 9 domains — is auto-registered at startup from OpenAPI 3.x specs
+octo-cli is **metadata-driven**. The command tree is auto-registered at startup
+from OpenAPI 3.x specs
 embedded into the binary. Adding or changing an endpoint means editing a
 spec, not the code.
 
@@ -34,6 +34,14 @@ Key properties:
 - **Agent-first output.** A stable JSON envelope with identity, data,
   pagination, and rate-limit metadata; a small fixed error taxonomy.
 
+### Reusable Go client
+
+The public `github.com/Mininglamp-OSS/octo-cli/client` package is independent
+from Cobra and CLI profile storage. A single client instance can route both
+Octo and Fleet/Loop requests through service-specific endpoints and tokens.
+Loop DTOs and business compatibility aliases deliberately stay outside this
+transport package.
+
 ## Domains
 
 | Domain    | Ops | Purpose                                                        |
@@ -47,6 +55,7 @@ Key properties:
 | `message` | 10  | Messaging — send, edit, sync, read-receipt; search (search/all/files/media/around/groups, in-channel or cross-channel) |
 | `file`    | 4   | Files — upload, download, credentials, presigned URLs          |
 | `event`   | 2   | Event polling — list, ack                                      |
+| `loop`    | 61  | Fleet control plane — tasks, executions, experts, expert templates, and expert teams |
 
 ## Installation
 
@@ -133,6 +142,10 @@ octo-cli docs content get doc-123          # returns the body + base version tok
 octo-cli docs import doc-123 --file ./notes.md      # replaces a doc from .md/.markdown/.docx
 octo-cli docs export doc-123 --export-format pdf -o ./notes.pdf
 octo-cli docs members set doc-123 --data '{"uid":"u-1","role":"writer"}'
+
+# Fleet/Loop can use the same gateway or an independent local endpoint.
+OCTO_LOOP_API_BASE_URL="http://127.0.0.1:8081" \
+  OCTO_BOT_TOKEN="octo_loop_your_credential" octo-cli loop task list
 octo-cli docs comments add doc-123 --data '{"body":"looks good"}'
 
 # Spreadsheets — read the live cells + base version, then batch-edit under If-Match.
@@ -198,7 +211,8 @@ All backend services are accessed through a single API base URL.
 
 | Var                 | Purpose                                                  |
 |---------------------|----------------------------------------------------------|
-| `OCTO_BOT_TOKEN`    | Bot token (`app_*`, `bf_*`, or `uk_*`). Required.       |
+| `OCTO_BOT_TOKEN`    | Active Octo or Loop bearer credential. Required.         |
+| `OCTO_LOOP_API_BASE_URL` | Optional Fleet/Loop endpoint; defaults to `OCTO_API_BASE_URL`. |
 | `OCTO_API_BASE_URL`  | Unified API base URL for all services. Required.          |
 | `OCTO_BOT_ID`       | Select/assert the bot credential by robot id (see `--bot-id`). |
 | `OCTO_CONFIG_DIR`   | Override the config/credential directory (default `~/.octo-cli`). |
--- a/cmd/auth.go
+++ b/cmd/auth.go
@@ -26,7 +26,7 @@ import (
 func newAuthCmd(f *cmdutil.Factory) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "auth",
-		Short: "Manage stored bot credentials (profiles)",
+		Short: "Manage stored Octo credentials (profiles)",
 	}
 	cmd.AddCommand(
 		newAuthLoginCmd(f),
@@ -44,8 +44,8 @@ func newAuthLoginCmd(f *cmdutil.Factory) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "login",
-		Short: "Store a bot token (read from a hidden prompt or stdin, never argv)",
-		Long: "Store a bot token under a profile. Identify the bot with --bot-id " +
+		Short: "Store a credential (read from a hidden prompt or stdin, never argv)",
+		Long: "Store an Octo or Loop bearer credential under a profile. Identify a bot with --bot-id " +
 			"(its robot id, recommended) and/or --profile (a friendly name). The " +
 			"token is read from a hidden prompt on a terminal, or from stdin with " +
 			"--with-token, or from --token-file — never from the command line.",
@@ -76,7 +76,7 @@ func newAuthLoginCmd(f *cmdutil.Factory) *cobra.Command {
 			}
 			if token == "" {
 				return failErr(f, output.ErrValidation(
-					"empty token", "provide a non-empty app_*, bf_*, or uk_* bot token"))
+					"empty token", "provide a non-empty Octo or Loop bearer credential"))
 			}
 
 			store, err := f.AuthStore()
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -72,14 +72,15 @@ func newConfigShowCmd(f *cmdutil.Factory) *cobra.Command {
 			}
 
 			payload := map[string]any{
-				"api_base_url":     cfg.APIBaseURL,
-				"space_id":         nullableString(space),
-				"format":           cfg.Format,
-				"bot_token":        maskToken(token),
-				"bot_token_source": nullableString(source),
-				"bot_kind":         botKind(token),
-				"profile":          nullableString(profileName),
-				"robot_id":         nullableString(robotID),
+				"api_base_url":      cfg.APIBaseURL,
+				"loop_api_base_url": nullableString(cfg.LoopAPIBaseURL),
+				"space_id":          nullableString(space),
+				"format":            cfg.Format,
+				"bot_token":         maskToken(token),
+				"bot_token_source":  nullableString(source),
+				"bot_kind":          botKind(token),
+				"profile":           nullableString(profileName),
+				"robot_id":          nullableString(robotID),
 			}
 			raw, err := json.Marshal(payload)
 			if err != nil {
--- a/cmd/config_test.go
+++ b/cmd/config_test.go
@@ -12,7 +12,8 @@ func TestMaskToken(t *testing.T) {
 		{"app_abcdefgh12345678", "app_ab***5678"},
 		{"bf_something", "bf_so***hing"},
 		{"app_tiny", "app_***"}, // body too short to reveal
-		{"short", "***"},        // unknown prefix
+		{"octo_loop_abcdefghijk", "octo_loop_ab***hijk"},
+		{"short", "***"}, // unknown prefix
 		{"", nil},
 		{"unknown_format_token", "***"}, // unknown prefix: reveal nothing
 	}
--- a/cmd/service/service.go
+++ b/cmd/service/service.go
@@ -210,10 +210,12 @@ func extractPathParams(path string) []string {
 	}
 }
 
-// serviceForBaseURL maps the spec's x-octo-base-url env-var name to the
-// config-level service key used by client.Request.Service. With the unified
-// gateway model all services route to the same URL, so this always returns
-// empty (default service). Retained for interface compatibility.
-func serviceForBaseURL(_ string) string {
+// serviceForBaseURL maps a spec endpoint to the config-level routing key. Loop
+// may use a standalone Fleet endpoint during local integration; all existing
+// Octo services continue to use the unified default URL.
+func serviceForBaseURL(baseURLEnv string) string {
+	if baseURLEnv == "OCTO_LOOP_API_BASE_URL" {
+		return "loop"
+	}
 	return ""
 }
--- a/cmd/service/service_test.go
+++ b/cmd/service/service_test.go
@@ -91,6 +91,13 @@ func TestRegisterServiceCommands_TreeShape(t *testing.T) {
 		"message": {"edit", "read-receipt", "send", "sync"},
 		"event":   {"ack", "list"},
 	}
+
+	loop := findCmd(root, "loop")
+	for _, resource := range []string{"task", "execution", "expert", "expert-template", "expert-team"} {
+		if findCmd(loop, resource) == nil {
+			t.Errorf("loop: missing resource %q; got %v", resource, childNames(loop))
+		}
+	}
 	for svc, wantCmds := range want {
 		svcCmd := findCmd(root, svc)
 		if svcCmd == nil {
@@ -113,6 +120,46 @@ func TestRegisterServiceCommands_TreeShape(t *testing.T) {
 	timeline := findCmd(matter, "timeline")
 	if timeline == nil || !contains(childNames(timeline), "add") || !contains(childNames(timeline), "list") || !contains(childNames(timeline), "delete") {
 		t.Errorf("matter timeline should nest add/list/delete; got %v", childNames(timeline))
+	}
+}
+
+func TestLoopCommandUsesFleetEndpointAndPublicPath(t *testing.T) {
+	var gotPath, gotAuth, gotSpace string
+	fleet := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		gotPath = r.URL.Path
+		gotAuth = r.Header.Get("Authorization")
+		gotSpace = r.Header.Get("X-Space-Id")
+		_, _ = w.Write([]byte(`{"data":{"task_id":"task-1"}}`))
+	}))
+	defer fleet.Close()
+
+	tf := cmdutil.NewTestFactory()
+	cfg := &config.Config{
+		APIBaseURL:     "http://octo.invalid",
+		LoopAPIBaseURL: fleet.URL,
+		BotToken:       "octo_loop_test",
+		Format:         "json",
+	}
+	cred := &credential.BotCredential{Token: "octo_loop_test", SpaceID: "space-from-legacy-profile"}
+	tf.SetConfig(cfg)
+	tf.SetCredential(cred)
+	tf.SetClient(client.New(cfg, cred, client.Options{NoRetry: true}))
+	tf.RegistryFunc = registry.MustNew
+
+	root := &cobra.Command{Use: "octo-cli", SilenceUsage: true, SilenceErrors: true}
+	RegisterServiceCommands(root, tf.Factory)
+	root.SetArgs([]string{"loop", "task", "get", "task-1"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("loop task get: %v", err)
+	}
+	if gotPath != "/api/v1/tasks/task-1" {
+		t.Fatalf("path = %q", gotPath)
+	}
+	if gotAuth != "Bearer octo_loop_test" {
+		t.Fatalf("Authorization = %q", gotAuth)
+	}
+	if gotSpace != "" {
+		t.Fatalf("Loop public API must not receive X-Space-Id, got %q", gotSpace)
 	}
 }
 
--- a/cmd/service/service_test.go
+++ b/cmd/service/service_test.go
@@ -93,7 +93,11 @@ func TestRegisterServiceCommands_TreeShape(t *testing.T) {
 	}
 
 	loop := findCmd(root, "loop")
-	for _, resource := range []string{"task", "execution", "expert", "expert-template", "expert-team"} {
+	for _, resource := range []string{
+		"attachment", "autopilot", "comment", "execution", "expert",
+		"expert-template", "expert-team", "label", "project", "runtime",
+		"skill", "skill-file", "task", "workspace",
+	} {
 		if findCmd(loop, resource) == nil {
 			t.Errorf("loop: missing resource %q; got %v", resource, childNames(loop))
 		}
@@ -152,7 +156,7 @@ func TestLoopCommandUsesFleetEndpointAndPublicPath(t *testing.T) {
 	if err := root.Execute(); err != nil {
 		t.Fatalf("loop task get: %v", err)
 	}
-	if gotPath != "/api/v1/tasks/task-1" {
+	if gotPath != "/v1/tasks/task-1" {
 		t.Fatalf("path = %q", gotPath)
 	}
 	if gotAuth != "Bearer octo_loop_test" {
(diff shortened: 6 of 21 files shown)


mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #115 (octo-cli)

## Summary

This PR exposes the complete Fleet (Loop) business API surface through octo-cli: a 7,359-line OpenAPI 3.1 spec (`internal/registry/specs/loop.json`) registers 136 operations across 96 paths (tasks, executions, experts, expert teams, autopilots, runtimes, skills, labels, projects, workspaces, comments, attachments). The code changes are focused and minimal: a new `OCTO_LOOP_API_BASE_URL` env var for optional endpoint separation, a `loop_credential` token prefix, generic `$ref` resolution for shared components (parameters, request bodies, responses), and string/label updates from "bot token" to "Octo or Loop bearer credential". The spec correctly declares `x-octo-space-header: false` so the Loop public API never receives `X-Space-Id`. Test coverage is thorough — the integration test `TestLoopCommandUsesFleetEndpointAndPublicPath` verifies URL routing, auth header, and space header suppression end-to-end.

## Verification

Static analysis only at head `b59632808ea6218ba31883115d24c16ddf39e87c`; build and tests not executed in this environment.

- ✅ **URL routing chain** — spec `x-octo-base-url: "OCTO_LOOP_API_BASE_URL"` → loader `BaseURLEnv` → `serviceForBaseURL()` returns `"loop"` → `Config.ServiceURL("loop")` returns `LoopAPIBaseURL` when set, falls through to `APIBaseURL` when empty. All string literals match across `cmd/service/service.go:218`, `internal/config/config.go:67`, and the spec.
- ✅ **Space header suppression** — spec `x-octo-space-header: false` → loader sets `SpaceHeader=false, SpaceHeaderSet=true` → `run.go:55` computes `SuppressSpaceHeader: true`. Test `TestLoopCommandUsesFleetEndpointAndPublicPath` asserts `X-Space-Id` is absent on the wire.
- ✅ **Token masking** — `MaskToken` adds `octo_loop_` prefix handling. `extractJSONSchema` handles nil body safely (nil map access returns zero value in Go). All `$ref` resolution failures produce nil that downstream code handles.
- ✅ **Component $ref resolution** — new `followComponentRef` generic helper correctly handles parameters, requestBodies, and responses sections. All three call sites check for nil returns before use: parameters `continue`, requestBody flows to nil-safe `extractJSONSchema`, responses checked `if r == nil { continue }`.
- ✅ **Security** — no operation-level security overrides found (all 136 ops inherit global `bearer_auth` / `octo_session`). No hardcoded credentials in spec, docs, or code. `octo_session` apiKey scheme (`Token` header) correctly declared alongside `bearer_auth`.
- ✅ **Test coverage** — 4 new test functions cover: command tree shape, URL routing + auth + space suppression, workspace header flag, component ref resolution, extended business contract, and the new credential prefix.

## Findings

No P0/P1 issues. One P2 documentation discrepancy.

### P2 — README operation count stale (`README.md:50`)

The service table claims `loop` has "61" operations, but `TestAllDomainOperationCounts` asserts 136 and the spec contains 136 operations across 96 paths. The figure "61" does not correspond to any derivable count (paths=96, operations=136, tags=13, resource commands=14). Update to 136 to match the loader output.

- AUTH/VISIBILITY: all changes are wording/copy updates ("bot token" → "Octo or Loop bearer credential") and new additive code. No auth guard logic changed. `TokenKind` adds a new format classifier without affecting authorization.
- BYTE-CAP: `MaskToken` byte-level masking is consistent with existing prefix behavior (head=2, tail=4, minMiddle=3). The `octo_loop_` prefix is handled identically to `app_`/`bf_`/`uk_`.
- INPUT-SHAPE: no input validation logic changed. `Config.Load()` adds one `os.Getenv` call. Empty `LoopAPIBaseURL` correctly falls through to `APIBaseURL`.
- PERSISTENCE/CACHE: no persisted state modified. New config field is read-only from env.
- The 5-commit history includes test plans and reports that are useful documentation artifacts but add ~1,100 lines of non-runtime content.

## Verdict: APPROVED

Solid, well-structured PR. The URL routing chain, space header suppression, credential handling, and component ref resolution are all correctly implemented and tested. The only issue is a stale operation count in the README table.

Jerry-Xin (changes requested):
Solid, well-scoped addition of the Loop/Fleet command surface (136 operations) driven entirely from the embedded `loop.json` OpenAPI spec, with sensible credential and routing plumbing. However, CI `build & test` is RED on a real lint failure, so this needs one change before it can merge.

## 🔴 Blocking

- **golangci-lint (gocyclo) fails — hard CI block.** `internal/registry/loader.go:325` `buildDetail` now has cyclomatic complexity 16 (> 15). The three new `$ref` resolution branches (parameters `$ref`, `requestBody` `$ref`, response `$ref`) pushed it over the threshold. Reproduced locally at the CI-pinned version (golangci-lint v2.12.2):
  ```
  internal/registry/loader.go:325:1: cyclomatic complexity 16 of func `buildDetail` is high (> 15) (gocyclo)
  ```
  Suggested fix: extract the parameter loop (and/or the requestBody/response resolution) into small helpers such as `buildParameters(doc, op)` / `buildRequestBody(doc, op)` so `buildDetail` drops back under 15. Purely mechanical, keeps behavior identical.

## 💬 Non-blocking

- **No `x-octo-risk` annotations on destructive Loop operations.** `loop.json` contains 18 `DELETE` + 5 `PUT` mutating ops (e.g. `task.delete`, `workspace.delete`, `project.delete`, `expert_team.delete`, `autopilot.delete`, `loop.skill.delete`, `expert.skill.replace`, `autopilot.signing_secret.set`) but carries **zero** `x-octo-risk` markers, while every other embedded spec annotates risk (matter=14, marketplace=26, docs=32, message=10, group=9, …). The `Risk` field is surfaced in `octo-cli schema` help output (`cmd/service/service.go:185`), so the omission means comparable destructive Loop commands lose the risk hint their Octo counterparts show. Not a security hole (the field is display-only; there is no hard confirmation guard in the CLI today), but worth adding for consistency and operator safety.

## ✅ Highlights

- **Routing is safe with no deploy-order hazard.** `Config.ServiceURL` returns `LoopAPIBaseURL` only when `OCTO_LOOP_API_BASE_URL` is explicitly set, otherwise falls back to the unified `APIBaseURL`. Loop commands therefore target the same gateway by default and the standalone Fleet endpoint is strictly opt-in — no risk of shipping commands that call an undeployed service.
- **Credential handling is correct.** New `octo_loop_` prefix is classified in `TokenKind` and properly masked in `MaskToken` (no head/tail leak for unknown kinds). Doc/message wording updates are consistent across `auth`, `config`, `factory`, and `env_provider`.
- **Additions are wired and tested.** 136 loop operations load through the registry; `$ref` resolution for parameters/requestBody/responses is exercised by `TestLoopPublicContractResolvesSharedComponents` and `TestLoopExtendedBusinessContract`, and the service-count test asserts `loop: 136`. Local `go build ./...` passes and `internal/registry`, `internal/config`, `internal/credential` tests are green.
- No hardcoded IDs, tokens, or endpoints in the new code; base URLs remain env-driven.

Once the gocyclo failure is resolved and CI `build & test` goes green, this is good to go.

Jerry-Xin (approved):
Clean, well-scoped addition of the Loop/Fleet command surface (136 operations) driven entirely from the embedded `loop.json` OpenAPI spec, with correct credential and routing plumbing. CI blocker resolved — approving.

## ✅ Verified

- **Lint green.** Verified locally at the CI-pinned golangci-lint v2.12.2: `0 issues` across `./...`. The earlier gocyclo failure on `buildDetail` (complexity 16 > 15) is resolved — the `$ref` resolution logic was extracted into `operationParameters` / `operationRequestBody` helpers, keeping behavior identical while dropping complexity under threshold.
- **Build + tests pass.** `go build ./...` clean; full `go test ./...` green, including the new `TestLoopPublicContractResolvesSharedComponents`, `TestLoopExtendedBusinessContract`, and the `loop: 136` operation-count assertion.

## ✅ Highlights

- **Routing is safe with no deploy-order hazard.** `Config.ServiceURL` returns `LoopAPIBaseURL` only when `OCTO_LOOP_API_BASE_URL` is explicitly set, otherwise falls back to the unified `APIBaseURL`. Loop commands target the same gateway by default, and the standalone Fleet endpoint is strictly opt-in — no risk of shipping commands that call an undeployed service.
- **Credential handling is correct.** New `octo_loop_` prefix is classified in `TokenKind` and properly masked in `MaskToken` (no head/tail leak for unknown kinds). Doc/message wording updates are consistent across `auth`, `config`, `factory`, and `env_provider`.
- **Additions are wired and covered.** 136 loop operations load through the registry; `$ref` resolution for parameters/requestBody/responses is exercised by tests. No hardcoded IDs, tokens, or endpoints in the new code; base URLs remain env-driven.

## 💬 Non-blocking (follow-up, not required for merge)

- **No `x-octo-risk` annotations on destructive Loop operations.** `loop.json` contains 18 `DELETE` + 5 `PUT` mutating ops (e.g. `task.delete`, `workspace.delete`, `project.delete`, `expert_team.delete`, `autopilot.delete`, `loop.skill.delete`, `expert.skill.replace`, `autopilot.signing_secret.set`) but carries **zero** `x-octo-risk` markers, while every other embedded spec annotates risk (docs=32, marketplace=26, matter=14, message=10, group=9, …). The `Risk` field is surfaced in `octo-cli schema` help output (`cmd/service/service.go:185`), so comparable destructive Loop commands lose the risk hint their Octo counterparts show. Not a security issue (the field is display-only; the CLI has no hard confirmation guard today), but worth adding for consistency and operator safety in a follow-up.

Nice work — LGTM.

yujiawei (changes requested):
# Code Review — PR #115 (octo-cli)

Reviewed at head `15d15bc`, diffed against merge-base `479ccbf`. (I began on `b596328`; `15d15bc` — `refactor(registry): simplify operation detail parsing` — landed mid-review. It extracts `operationParameters` / `operationRequestBody` out of `buildDetail` with no behaviour change: the only semantic delta is `d.Parameters` becoming an empty non-nil slice instead of `nil` when an operation declares no parameters, which is invisible in output because the field is `json:"parameters,omitempty"` (`internal/registry/loader.go:204`). Every finding below was re-verified against `15d15bc`; line numbers refer to that commit.) `go build ./...` and `go test ./...` both pass locally (go1.26.5). I also built the binary and exercised the new command tree with `--dry-run` / `--verbose` to check real wire behaviour, and cross-compared `loop.json` against the eleven pre-existing embedded specs for convention drift.

Note for context: `internal/registry/specs/loop.json` does not exist on `main`, so relative to the base branch this PR introduces the **entire** 136-operation Loop surface, not an increment on 61.

**Verdict: CHANGES_REQUESTED.** Two blocking issues (tenancy selector, advertised-but-unimplemented auth scheme) plus one confirmed secret-exposure path. The engine changes themselves are clean and the routing/space-header work is correct and well tested — the problems are concentrated in the embedded spec's completeness and in the docs that ship with it.

---

## 1. Contract / scope compliance

What the PR body claims, and what I could verify:

| Claim | Status |
|---|---|
| Registry extended to 136 operations | ✅ 136 operations across 96 paths |
| workspace / label / project / comment / attachment / skill / runtime / autopilot groups added | ✅ all 14 resource groups register |
| Fleet routing stays on `/v1/*` | ✅ every path is `/v1/*` |
| No operation-ID collisions with Marketplace skill commands | ✅ verified 0 duplicate `operationId` across all 12 specs; `loop.skill.*` correctly namespaces under `loop skill` |
| Explicit `--workspace-id` for human credentials | ⚠️ **partial** — 69 of 136 operations only (see 2.1) |
| "Documentation and embedded API contract updated" (checklist) | ❌ several stale/contradictory docs (see 3.1–3.3) |
| "Command-tree coverage for every new resource group" | ⚠️ the new assertion checks that the 14 resource *nodes* exist but not their leaf verbs, unlike the pre-existing `want` map in the same test which enumerates verbs per service (`cmd/service/service_test.go:95`) |

`X-Space-Id` suppression works as designed — with `OCTO_SPACE_ID` set, `loop task get <id> --dry-run` emits only `Authorization`, no space header. Endpoint override works too: with `OCTO_LOOP_API_BASE_URL` set, the URL becomes `http://fleet.example/v1/tasks` while other domains stay on `OCTO_API_BASE_URL`. Both are correct and covered by the new tests.

---

## 2. Blocking findings

### 2.1 [P1] Workspace selector is missing on the primary resource families

`WorkspaceIDHeader` (`loop.json:6582`, surfaced as `--workspace-id`) is attached to 69 of 136 operations. The 67 that lack it include the **entire** `task`, `expert`, `expert_team`, `execution`, and `workspace` families — i.e. the resources this CLI exists to drive.

The two operations are semantically identical in scope, yet only one is selectable:

```jsonc
// loop.json:105-118  — task.list
"description": "Returns one workspace-scoped page of tasks.",
"operationId": "task.list",
"parameters": [
  { "$ref": "#/components/parameters/Page" },
  { "$ref": "#/components/parameters/PageSize" }
]                                    // <-- no WorkspaceIDHeader

// loop.json:2755-2770 — label.list
"description": "List labels in the authenticated workspace.",
"operationId": "label.list",
"parameters": [
  { "$ref": "#/components/parameters/Page" },
  { "$ref": "#/components/parameters/PageSize" },
  { "$ref": "#/components/parameters/WorkspaceIDHeader" }
]
```

Observable result:

```
$ octo-cli loop label list --help
      --workspace-id string   Explicit workspace selector for human credentials; ...

$ octo-cli loop task list --help
      --page int        One-based page number. (default 1)
      --page-size int   Number of resources per page. (default 20)
                        # no --workspace-id
```

`loop.json:149` (`task.create`) has `"description": "Creates a task in the authenticated workspace."` and also no header. So with a human credential spanning more than one workspace there is no way to say *which* workspace a task is created in — the write lands wherever the server defaults, silently. For a change whose stated purpose includes "explicit `--workspace-id` selection for human credentials", that is the case that matters most.

The same asymmetry produces a genuinely confusing pair for one concept:

```
loop task reaction remove <task-id>     --emoji <e>            # typed body, no --workspace-id
loop comment reaction remove <id>       --data '{...}'  --workspace-id <w>   # no --emoji
```

**Ask:** either attach `WorkspaceIDHeader` to every workspace-scoped operation, or state in the spec/skill docs why the `task`/`expert` families resolve their workspace differently. If the split mirrors `octo-fleet/openapi/public-v1.yaml` exactly, please say so in the PR description — that turns this from a bug into a documented upstream constraint.

### 2.2 [P1] `octo_session` is declared as an auth scheme but the transport cannot send it

```jsonc
// loop.json:6485-6489
"octo_session": {
  "type": "apiKey",
  "in": "header",
  "name": "Token",
  "description": "Octo session token accepted directly by Fleet."
}
```

It is also listed in the root `security` block (`loop.json:77`), and `skills/octo-loop/SKILL.md:14-17` instructs users to *"Set `OCTO_BOT_TOKEN` to a credential Fleet accepts directly, such as an Octo Session credential"*.

But the client only ever emits a bearer header, unconditionally:

```go
// internal/client/client.go:217-219
if c.cred != nil && c.cred.Token != "" {
    httpReq.Header.Set("Authorization", "Bearer "+c.cred.Token)
}
```

`grep -rn '"Token"' --include=*.go .` returns nothing; no code reads `components.securitySchemes` at all. So a session credential put in `OCTO_BOT_TOKEN` goes out as `Authorization: Bearer <session>`, never as `Token: <session>`. Either Fleet accepts sessions as bearer — in which case the `octo_session` scheme is misleading and should be dropped from the spec — or it does not, in which case the flow the skill doc recommends is broken. There is no test for it either way. In a change flagged security-sensitive, an advertised authentication path with no implementation and no coverage shouldn't land unresolved.

**Ask:** pick one — implement the `Token` header behind an explicit selector, or remove `octo_session` from the spec and correct `SKILL.md`.

### 2.3 [P1] Secret-bearing request bodies are logged in cleartext

This spec introduces the first operations in the CLI whose request body is itself a secret: `autopilot.signing_secret.set` (`loop.json:6361`) and `expert.environment.update` (`loop.json:2003`, `custom_env` map). No other embedded spec has a secret-valued body field.

Both reach code that prints the body verbatim:

```go
// internal/client/client.go:230-232
if c.options.Verbose && len(body) > 0 {
    c.verbosef("request body: %s", truncate(string(body), 1024))
}
// internal/client/client.go:390 (renderDryRun)
out["body"] = bodyField
```

Reproduced — note the secret was supplied the *safe* way, via `--data @-`:

```
$ echo '{"signing_secret":"SUPER_SECRET_VALUE_123"}' | octo-cli --verbose \
    loop autopilot signing-secret set <ap-id> <trigger-id> --data @-
[octo] PUT http://.../signing_secret
[octo] request body: {"signing_secret":"SUPER_SECRET_VALUE_123"}      <-- stderr

$ ... --dry-run ...  ->  "body": { "signing_secret": "SUPER_SECRET_VALUE_123" }   <-- stdout envelope
```

`Authorization` is masked in both paths, so the redaction hook already exists — the body just isn't covered. This is worth fixing here rather than later, because the repository already treats this threat model as settled: `cmd/auth.go:47-51` deliberately refuses to take a token from argv ("never from the command line"). Adding endpoints that hand a secret to `--data` on argv, and then echo it to stderr under `--verbose`, walks that back. Agents run with `--verbose` in CI, where stderr is persisted.

**Ask:** add a spec-driven marker (e.g. `x-octo-sensitive-body`, or per-property `x-octo-secret`) and redact those fields in both `verbosef` and `renderDryRun`. Also add a `--data @-` hint to the two operations' descriptions.

---

## 3. Non-blocking findings

### 3.1 [P2] `loop.json` is the only spec with no `x-octo-risk` on any operation

All 136 Loop operations omit it. Every other spec annotates 100% of its operations:

```
bot.json    {'write': 4, 'read': 2}          matter.json  {'write': 10, 'read': 3, 'high-risk-write': 1}
docs.json   {'write': 18, 'read': 13}        message.json {'write': 3, 'read': 7}
...                                          loop.json    {None: 136}
```

`x-octo-risk` is documented in `docs/architecture-design.md:286` and is what `buildLongDesc` renders as the `[risk: …]` tag in `--help` (`cmd/service/service.go:185-187`). So roughly twenty destructive operations now ship with no safety classification:

```
$ octo-cli loop runtime archive-and-delete --help
Archive experts and delete runtime
operationId: runtime.archive_and_delete
POST /v1/runtimes/{runtime_id}/archive_agents_and_delete      # no [risk: ...] label
```

Same for `workspace.delete`, `project.delete`, `task.delete`, `loop.skill.delete`. It is metadata only today (there is no runtime confirmation gate — `TestMatterDelete_ExecutesDirectly` confirms that), but it is the signal agents and `octo-cli schema` consumers read before a destructive call. `runtime.archive_and_delete` (`loop.json:4769`) is the obvious `high-risk-write` candidate.

### 3.2 [P2] Every Loop response schema is opaque, so `octo-cli schema` returns no field names

```jsonc
// loop.json:6943
"ResourceEnvelope": { "type": "object", "required": ["data"],
  "properties": { "data": { "type": "object", "additionalProperties": true } } }
```

`CollectionEnvelope.data.items` is the same. All 136 operations resolve to it. Compare:

```
$ octo-cli schema task.get   -> response_schema: { data: { type: object } }
$ octo-cli schema matter.get -> response_schema: { matter: { id, title, description,
                                  created_at, deadline, remind_at, seq_no, ... },
                                  assignees: [...], channels: [...] }
```

`docs/architecture-design.md:293` states this as a requirement: *"**Response schemas required** — Agent needs field names for `--jq` and decision-making."* With >50% of the CLI surface now opaque, `--jq` filtering over Loop responses is guesswork. Related: 27 write operations use `FlexibleRequest` (`loop.json:7328`, `additionalProperties: true`, zero properties), so no body flags are promoted and `--help` offers nothing:

```
$ octo-cli loop workspace create --help
      --data string   JSON request body (string, @file, or @- for stdin). ...
```

I'd rather not block on this since it's a large mechanical change, but it should be a tracked follow-up before this becomes the reference for how new domains are added — retro-fitting 136 operations later rarely happens.

### 3.3 [P2] Docs shipped with the PR are stale or self-contradictory

- `README.md:50` — `| loop | 61 | Fleet control plane — tasks, executions, experts, expert templates, and expert teams |`. The count is 136, and the description omits 8 of the 14 groups (workspace, label, project, comment, attachment, skill, runtime, autopilot).
- `README.md:190-192` — the token-kind table still lists only `app_*` / `bf_*` / `uk_*`; `octo_loop_*` was added to `TokenKind`/`MaskToken` but not here.
- `README.md:136-141` — the new Loop example is inserted **inside** the Documents example block, between `docs members set` and `docs comments add`, breaking that grouping.
- `skills/octo-loop/SKILL.md` documents only 5 of the 14 groups, and line 40 asserts *"Repository operations, daemon device enrollment, task claim/heartbeat, and local runtime paths are intentionally not part of this CLI surface"* — while `runtime.local_skill_import.create`, `runtime.local_skill_request.create`, `runtime.model_request.create`, and `runtime.update_request.create` **are** exposed. Please reconcile the claim with the surface.
- `docs/loop-integration-test-report.md` is evidence for a different revision: it says `Branch: feat/loop-integration-test`, `commit: a18d5ba`, and reports the Fleet contract comparison as `61/61 operations match` across 5 groups. Head is `b596328` with 136 operations across 14 groups, so the contract-parity evidence covers **less than half** of what ships. Please re-run the comparison against head, or mark the report as scoped to the earlier commit.

### 3.4 [P2] `firstSuccessSchema` still drops `$ref` for 2xx codes outside 200/201/204

The PR adds ref-following to the ordered branch:

```go
// internal/registry/loader.go:503-515
order := []string{"200", "201", "204"}
for _, code := range order {
    if r, ok := resps[code].(map[string]any); ok {
        if ref := stringOf(r["$ref"]); ref != "" { r = followResponseRef(doc, ref) }   // added
        ...
```

but the generic fallback immediately below it was not updated:

```go
// internal/registry/loader.go:516-526
for code, r := range resps {
    if !strings.HasPrefix(code, "2") { continue }
    if rm, ok := r.(map[string]any); ok {
        if s := extractJSONSchema(rm); s != nil { ... }     // no $ref follow
    }
}
```

`loop.json` is the first spec with `202`-only responses, so this now fires:

```
$ octo-cli schema task.rerun        -q '.data | has("response_schema")'  -> false
$ octo-cli schema task.quick_create -q '.data | has("response_schema")'  -> false
$ octo-cli schema task.get          -q '.data | has("response_schema")'  -> true
```

One-line fix: hoist the same `$ref` follow into the fallback loop.

### 3.5 [P2] Path-item-level `parameters` are silently ignored, so `schema` hides positional args

`operationParameters` reads only `op["parameters"]` (`internal/registry/loader.go:369`). `loop.json` is the first spec to declare parameters at the path-item level — 34 of its 96 paths do; all other 11 specs use 0. Every one of those is `in: path`, so positional-arg *handling* is unaffected (`extractPathParams` derives them from the path string), but discoverability is:

```
$ octo-cli schema task.get   -> no "parameters" key at all
$ octo-cli schema matter.get -> parameters: [{name: id, in: path, required: true}, ...]
```

An agent introspecting `task.get` cannot learn that it takes a required `task_id`. Either merge path-item parameters in the loader, or move them to operation level in the spec for consistency with the other eleven.

### 3.6 [P2] `identity.type` is hard-coded `"bot"`

```go
// internal/cmdutil/factory.go:274
id := map[string]any{"type": "bot"}
```

Every envelope reports `"type": "bot"`, including for `octo_loop_*` credentials. That contradicts the contract this PR newly asserts in `internal/credential/token.go:3-5` ("never determine a Loop principal class") and gives log consumers a wrong actor type when the principal is a human. A neutral value (or omitting the field when the kind is not a bot prefix) would be more honest.

### 3.7 [P2] Unresolvable `$ref` fails open

`followComponentRef` returning `nil` makes the parameter loop `continue` (`internal/registry/loader.go:376-380`) — a typo'd `$ref` silently removes a flag, e.g. `--workspace-id`, with no diagnostic. The `requestBody` branch (`operationRequestBody`, `:405-415`) doesn't nil-check at all; it happens to be safe only because reads on a nil map return zero values. Both would be better as loud failures at registry load. Also latent: a parameter whose `schema` is itself a `$ref` is read without resolution (`:389`), yielding an empty type — not triggered by this spec, but worth guarding while you're in here.

---

## 4. Nits

- `x-octo-pagination` is absent from all 136 operations, so `--page-all` — advertised as a global flag in `README.md:256` — works nowhere under `loop`. That is *correct*, since `OffsetPagination` exposes `total`/`page`/`page_size` and the engine's `parsePage` needs `has_more`/`next_cursor`. Worth one README sentence so it doesn't read as an oversight.
- Template-generated descriptions read oddly: `workspace.list` is `"List workspaces in the authenticated workspace."`, `workspace.create` is `"Create workspace in the authenticated workspace."` (all six `workspace.*` entries). Not user-visible today (only `summary` reaches `--help`), but it does suggest the spec was generated and not proofread — which is consistent with 3.1/3.2.
- `docs/architecture-design.md:305` caps operationId sub-resource depth at 2; `attachment.content.get` and `autopilot.trigger_config.update` produce depth-3 paths (`loop attachment content get`). Fine in practice, but the doc rule now has exceptions.

---

## 5. Things I could not verify — please have a human confirm

1. **Spec parity with upstream.** `x-octo-source` points at `octo-fleet/openapi/public-v1.yaml`, which I cannot read from here. I could not confirm that the 136 operation IDs, paths, methods, and — critically — the 69/136 `X-Workspace-ID` split match upstream. If the split is upstream's, 2.1 becomes documentation; if it is a transcription gap, it is a bug.
2. **Whether Fleet accepts an Octo session token as `Authorization: Bearer`.** This decides which half of 2.2 applies.
3. **Authenticated behaviour.** The PR's own report lists cross-workspace isolation, the human/bot/Loop permission matrix, and revoked-credential handling as *not yet executed*. Given `--workspace-id` is a tenancy boundary supplied by an untrusted-to-the-CLI flag, a real cross-workspace isolation test against a test Fleet endpoint is the check I would most want before this is relied on.

Scope of my own pass, for the record: I read the full non-spec diff, sampled `loop.json` around the security-relevant operations rather than auditing all 7,359 lines field-by-field, ran the unit suite, and drove the binary via `--dry-run`/`--verbose`. I did not run against a live Fleet endpoint and did not review the committed HTML report's rendering.

PR #118, opened 2026-07-30
feat(docs): add full-text search command

### LINKED ISSUE #117: feat(docs): expose full-text document search in the CLI
Track the CLI integration for permission-scoped online-document full-text search, including cursor pagination and documentation.

wwwzzzcccc:
## Summary

Expose the existing permission-scoped online-document full-text search API as `octo-cli docs search`.

## Changes

- add `docs search` with keyword, repeatable document-type, cursor, and page-size flags
- extend metadata-driven pagination with configurable item/cursor response paths and POST body cursor progression
- stop automatic pagination on repeated cursors
- document the new command in README, project instructions, and the embedded docs skill
- add registry, request-shape, cursor progression, and repeated-cursor tests

## Linked Spec

- Fixes #117
- Backend contract: `POST /v1/bot/docs/search`

## How verified

- `go test -race -shuffle=on -count=1 ./...`
- `go vet ./...`
- `go build ./cmd/octo-cli`
- `golangci-lint run` (v2.12.2): 0 issues
- `go mod tidy` with no `go.mod` / `go.sum` drift
- `gofmt -l .`: clean
- `git diff --check`: clean
- local Kafka + OpenSearch + backend + indexer E2E
- deployed test environment E2E: keyword search, type filters, manual cursor, and `--page-all`

### What does this change actually do?

Before this change, agents had to use the generic API command to call document search and manually manage keyset cursors. After it, the metadata-driven command tree exposes `docs search`; a normal request preserves the backend `{total, items, nextCursor}` object, while `--page-all` follows body cursors and emits the merged item array.

### What could break?

The shared pagination parser is the main blast radius. Existing pagination contracts keep their historical defaults, custom paths are opt-in through `x-octo-pagination`, and tests cover both legacy behavior and the new document-search shape. A repeated-cursor guard prevents malformed backends from looping until the page cap.

### How do you know it works?

Unit and race tests cover command registration, request bodies, type filters, cursor placement, custom paths, and stop conditions. Real E2E tests verified the Bot-authenticated backend route, permission filtering, OpenSearch results, manual cursor progression, and all-page aggregation.

## Checklist

- [x] Code, comments, and commit message are in English
- [x] README command reference and examples updated
- [x] CLAUDE.md command list updated
- [x] New feature includes tests


Jerry-Xin (approved):
The PR is in scope for `octo-cli` and cleanly extends the existing metadata-driven command registry to expose `docs search`.

💬 Non-blocking

🟡 Warning: `--keyword` is documented as required via the schema, but promoted body flags are not locally enforced by the generic flag generator. `cmd/service/flags.go:208` only appends “(required)” to the help text, and `cmd/service/flags.go:211-226` registers the flag without `MarkFlagRequired`, so `octo-cli docs search --page-all` can still reach the backend with no `q`. This appears to be an existing generator limitation, not a new architectural problem, but this command makes it more visible because keyword is the primary input. See `internal/registry/specs/docs.json:98`.

✅ Highlights

🔵 Suggestion-strength positive: The pagination extension is implemented in the shared path rather than special-casing `docs.search`, and existing defaults for `{data, pagination}` are preserved in `cmd/service/run.go:346-360`.

🔵 Good coverage: The tests cover command registration, body flag aliases, POST body cursor progression, default response preservation, custom response paths, and repeated-cursor termination.

Verification: I ran `go test ./...`; all tests passed.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #118 (octo-cli)

## Summary

This PR adds a `docs search` full-text search command (POST `/v1/bot/docs/search`) and generalizes the pagination engine to support non-standard response shapes. The key production change is in `cmd/service/run.go`: `parsePage` now reads items, cursor, and has-more from configurable JSON paths declared via `x-octo-pagination` in the spec, instead of the hardcoded `{data, pagination:{has_more, next_cursor}}` envelope. A repeated-cursor guard (`seenCursors` map) prevents infinite pagination loops. The `PaginationInfo` struct gains an `ItemsField` field. Tests cover flag registration, request body mapping, single-page envelope preservation, multi-page cursor progression, page-all termination, and repeated-cursor detection. Documentation (SKILL.md, CLAUDE.md, README.md) is updated consistently.

Overall: well-scoped, well-tested change with clean backward compatibility across all existing paginated operations.

## Verification

Local checks at head `ca59c939`:
- `go test ./...` — pass (all 12 packages)
- `go build ./...` — clean

- ✅ **Backward compatibility** — All existing paginated specs (matter.list, message.search, marketplace list/publish) declare `cursorField` and `hasMoreField` in the old `pagination.*` shape with empty `itemsField`. The new `parsePage` at `cmd/service/run.go:353-361` falls through to the hardcoded defaults (`"data"`, `"pagination.next_cursor"`, `"pagination.has_more"`). Existing `TestPagination_PageLimitStops` and `TestPagination_PageAllMergesAllPages` still pass, confirming runtime equivalence.
- ✅ **Cursor routing** — For `docs.search`, `cursor` is a body property (not in OpenAPI `parameters[]`), so `cursorIsQueryParam` correctly returns `false`. The `withCursorBody` path at `:292-300` creates a new map each iteration — no mutation of the previous page's request.
- ✅ **Repeated-cursor guard** — `seenCursors` map at `:256` breaks the loop when the server returns the same `nextCursor` twice. `TestDocsSearch_PageAllStopsOnRepeatedCursor` verifies exactly 2 calls before stop.
- ✅ **valueAtPath safety** — `:389-400` handles: empty path (returns nil), non-map intermediate (returns nil), missing key (returns nil). Safe for all existing and new specs.
- ✅ **hasMore inference** — When `hasMoreField` is empty (docs.search omits it), `parsePage` at `:383` infers `hasMore = cursor != ""`. Correct for the `{total, items, nextCursor}` response shape where absence of `nextCursor` means last page.
- ✅ **Numeric precision** — `UseNumber()` decoder at `:340` preserves number precision. All existing cursor fields are strings and all hasMore fields are booleans, so this switch does not affect them.

## Findings

No P0/P1 issues. No P2 issues. The implementation is clean, the backward compatibility is verified for all existing paginated operations, and the test coverage is thorough — including flag shape, body mapping, single-page envelope preservation, multi-page cursor progression, and the repeated-cursor safety net.

## Data Flow Trace

- `docs.json` x-octo-pagination → `loader.go:393` buildDetail parses `itemsField` into `PaginationInfo.ItemsField`
- `PaginationInfo` → `run.go:333` parsePage reads `pag.ItemsField`, `pag.CursorField`, `pag.HasMoreField`
- `valueAtPath(page, "items")` extracts items array from response → `json.Marshal` roundtrip → `[]json.RawMessage`
- `valueAtPath(page, "nextCursor")` extracts cursor string → `seenCursors` check → `withCursorBody` places into next request body
- `hasMoreField == ""` → `hasMore = cursor != ""` → loop break condition in `runPaginated:270`
- All data flows verified: consumed values trace back to their upstream sources and reach the consumption points correctly.

## Blind Spot Checklist

- **C1 (dual-path parity)** — N/A. No paired add/remove or subscribe/unsubscribe paths touched. The single new endpoint is self-contained.
- **C2 (control-flow ordering)** — Clear. `parsePage` is called once per page in `runPaginated`. No nested or recursive call sites. `cursorIsQueryParam` is called once per page iteration.
- **C3 (authorization boundary)** — N/A. The search endpoint uses the same auth as other docs operations. No new auth paths.
- **C4 (authorization lifecycle)** — N/A. No auth changes.
- **C5 (build/note ≠ runtime)** — Clear. Tests verify runtime behavior (body mapping, cursor progression, repeated cursor detection), not just build-time correctness.
- **C6 (governance/docs self-consistency)** — Clear. SKILL.md, CLAUDE.md, and README.md all consistently report 105 operations / 32 docs ops.

## Cross-round Blocker Re-check

N/A — first review pass.

- `valueAtPath` dot-separated path traversal handles all edge cases (missing keys, non-map intermediates, empty paths) without panicking.
- `withCursorBody` shallow-clones the body map — values are primitives or immutable slices, so no aliasing issues across pages.
- `parsePage` with `UseNumber()` does not affect cursor (string) or hasMore (bool) extraction for any existing spec.
- The `seenCursors` map is bounded by `--page-limit` (default 10) — no unbounded memory growth.
- The `docs.json` spec correctly places `/v1/bot/docs/search` before `/v1/bot/docs/{docId}` — no route ambiguity.
- The `required: ["q"]` body schema maps to `--keyword` via `x-octo-flag` — cobra enforces the required flag.
- `--doc-type` is optional; the enum `["doc", "sheet", "board"]` is declared in the spec for server-side validation.
- SKILL.md accurately documents cursor-based pagination behavior, page-all, and page-limit.
- CLAUDE.md and README.md operation counts updated consistently (104→105, docs 31→32).

## Verdict: APPROVED

Clean, well-tested implementation. The pagination generalization is backward-compatible with all existing specs, the new search endpoint is correctly wired from spec through loader to the pagination engine, and the repeated-cursor guard is a useful safety addition. No correctness, security, or compatibility concerns.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Clean backward compatibility verified for all existing paginated operations. Tests thoroughly cover new functionality.

yujiawei (approved):
# Code Review — PR #118 (octo-cli)

Reviewed at head `ca59c939a6fad0aaca66b9e816f940f1ade3ae99` against merge-base `479ccbfacda999540674e92d44530fc8f3a50878`.
Verified locally: `go build ./...`, `go test ./...` — all green. CI `build & test (1.24.x)` and `npm test` pass on this head.

I also ran two independent second-opinion passes over the diff and reconciled their findings below, including the places where I disagree with them.

---

## 1. Spec compliance

**Spec: ✅**

Against the linked issue #117 ("expose full-text document search in the CLI, including cursor pagination and documentation"):

- **Missing:** none. `docs search` is registered (`internal/registry/specs/docs.json:98`), cursor pagination works in both manual (`--cursor`) and automatic (`--page-all`) modes, and docs are updated in three places (`README.md`, `CLAUDE.md`, `skills/octo-docs/SKILL.md`).
- **Extra:** the generalization of `parsePage` (configurable `itemsField` / `cursorField` / `hasMoreField`) is not literally in the issue text, but it is *required* — the new endpoint returns `{total, items, nextCursor}`, which does not match the historical `{data, pagination:{has_more,next_cursor}}` envelope the shared parser hard-coded. Adding the paths to the existing `x-octo-pagination` extension instead of adding endpoint-specific code is the right call for a metadata-driven CLI. Not scope creep.
- **Deviation:** none. Operation counts are internally consistent — the README domain table sums to 105 (`docs` 32 + `html` 20 + `matter` 14 + `group` 9 + `thread` 8 + `message` 10 + `bot` 6 + `file` 4 + `event` 2), matching the `104 → 105` edits in `README.md:15` and `CLAUDE.md:26`, and `internal/registry/loader_test.go:44` pins `docs: 32`.
- Regression coverage for the legacy envelope already exists and still passes on this head — `TestPagination_PageAllMergesAllPages` (query cursor, `matter.list`) and `TestPagination_BodyCursorEndpoint` (body cursor, `message.search`). That is the single most important thing about this diff and it is covered.

---

## 2. Code quality

**Quality: Approved (with nits)**

No P0/P1. Every finding below needs a malformed backend response, a hypothetical future spec, or is documentation/formatting — none blocks merge. Ordered by how much I'd want them addressed.

### P2-1 — `parsePage` turns three former hard errors into silent success

`cmd/service/run.go:362`, `:375`, `:343`. The rewrite from a typed `json.Unmarshal` into `[]json.RawMessage` to a generic map walk changed three failure modes from *loud* to *silent*. Verified by running the old and new implementations side by side:

```go
// run.go:362
rawItems, ok := valueAtPath(page, itemsField).([]any)
if !ok {
    rawItems = nil
}
```

| Response | Before | After |
|---|---|---|
| `{"data":{},"pagination":{"has_more":true,"next_cursor":"c2"}}` | `PAGINATION_PARSE` error | `items=0`, `hasMore=true` → keeps paging, emits `[]` with `ok:true` |
| `{"data":[],"pagination":{"next_cursor":123}}` | `cannot unmarshal number into ... type string` | `cursor="123"` (this coercion is *intended* — see P2-2) |
| `{"data":[],"pagination":{"next_cursor":{"id":"c2"}}}` | `cannot unmarshal object into ... type string` | `cursor="map[id:c2]"` — a Go debug string is sent to the backend as the cursor |
| `{"data":[...],"pagination":{...}} garbage` | `invalid character 'g' after top-level value` | accepted (`json.Decoder.Decode` stops after the first value) |

This applies to all eight pre-existing paginated operations, not just `docs search`. No shipping backend produces these shapes, which is why this is P2 and not a blocker — but it works against this file's own stated design principle. `runPaginated` immediately above carries a comment calling its `OutputPath` guard *"a fail-loud guard against a future spec that ... would otherwise silently corrupt --output"*. The most likely real-world trigger here is exactly that: a future spec with a typo'd `itemsField`, which now yields an empty result set with `ok:true` instead of an error.

Suggestion: return `PAGINATION_PARSE` when `itemsField` resolves to a non-nil non-array (distinguish "absent" from "wrong type"), and reject cursor values that are neither string nor `json.Number`. Keeping `json.Number → string` is correct and deliberate, since `docs comments list` / `docs versions list` use integer id cursors.

### P2-2 — an omitted `hasMoreField` means two different things

`cmd/service/run.go:354-359`:

```go
hasMoreField = pag.HasMoreField
// Older specs omitted itemsField but did declare the historical cursor
// and has-more paths. Keep data as the default in that case.
if pag.HasMoreField == "" && pag.CursorField == "" {
    hasMoreField = "pagination.has_more"
}
```

Two problems:

1. **The comment describes different code than what is written.** It talks about `itemsField` and "keep `data` as the default"; the condition is about `hasMoreField`/`cursorField`. Whatever the intent was, this comment will mislead the next reader.
2. **The semantics of omitting `hasMoreField` depend on whether `cursorField` is set**, which is surprising for a declarative spec extension:
   - `cursorField` set, `hasMoreField` omitted → "non-empty cursor means more pages", and a response's explicit `has_more: false` is **ignored**. Verified: `{"data":[{"a":1}],"pagination":{"has_more":false,"next_cursor":"c9"}}` with `PaginationInfo{CursorField:"pagination.next_cursor"}` returns `hasMore=true`.
   - both omitted → falls back to `pagination.has_more`. So a spec declaring only `itemsField` silently gets the legacy has-more path: verified `{"items":[{"id":"1"}],"pagination":{"next_cursor":"c2"}}` with `PaginationInfo{ItemsField:"items"}` returns `hasMore=false` and stops after one page, despite a perfectly good default-path cursor.

Note the `HasMoreField=="" && CursorField==""` branch is unreachable today — all eight existing paginated ops declare both `cursorField` and `hasMoreField` explicitly, which is also why this diff carries no live regression. It is purely a trap for the next spec author. Suggestion: make the "cursor presence implies more pages" mode an explicit opt-in (e.g. `hasMoreField: "-"` or a `cursorImpliesMore: true` boolean) rather than inferring it from an absent field, and fix the comment.

### P2-3 — the repeated-cursor guard stops the loop but still emits the duplicate page

`cmd/service/run.go:256`, `:269-277`. Both second-opinion passes flagged this independently, and it holds up:

```go
merged = append(merged, data...)   // :269 — appended first
if !hasMore || nextCursor == "" { break }
if _, seen := seenCursors[nextCursor]; seen { break }   // :273 — checked after
seenCursors[nextCursor] = struct{}{}
```

Also, `seenCursors` is not seeded with the caller's initial `--cursor` (`:256`). So starting at `--cursor c1` against a backend that echoes `nextCursor: "c1"` costs one redundant request whose items are appended before the break.

To be fair about the size of this: it is a strict improvement over the previous behavior (which burned all 10 pages), and it only produces duplicate *items* when a backend returns identical data for a genuinely different cursor. I do not agree with the framing that backends "commonly" do this. But the guard's own test doesn't pin the part that matters — `cmd/service/docs_test.go:306-320` asserts `calls != 2` only, never that the merged output contains one item rather than two. Suggestion: seed `seenCursors` with the initial cursor and check membership *before* issuing the request, and extend that test to assert the merged item count.

### P2-4 — `--page-all` output no longer preserves the backend's JSON key order

`cmd/service/run.go:362-372`. Items are now decoded into `map[string]any` and re-encoded with `json.Marshal`, so keys come out alphabetically sorted instead of in the order the backend sent them. Verified on input `{"data":[{"zeta":"<em>hi</em> & co","alpha":1,"big":12345678901234567890,"f":1.10}]}`:

```
before: [{"zeta":"<em>hi</em> & co","alpha":1,"big":12345678901234567890,"f":1.10}]
after:  [{"alpha":1,"big":12345678901234567890,"zeta":"<em>hi</em> & co"}]
```

Worth stating precisely what *doesn't* change, because one of the second-opinion passes rated this P1 on the grounds that `--page-all` now "corrupts" content via HTML escaping. That part is wrong, and I checked rather than assumed: `encoding/json` runs `compact(..., escapeHTML=true)` over the bytes a `json.RawMessage` produces, so `<`, `>` and `&` were already escaped identically before this diff. `decoder.UseNumber()` also correctly preserves numeric literals — `12345678901234567890` and `1.10` survive byte-for-byte. So the only real delta is key order, which is cosmetic for a JSON-envelope CLI consumed via `jq`. P2, and worth a line in the PR description since it affects all eight existing `--page-all` operations, not just the new one.

Related (advisory only): decoding the whole response into generic `map[string]any` / `[]any` allocates considerably more than the old `[]json.RawMessage`. Not worth optimizing at current page sizes; noting it so it's a conscious trade-off.

### P2-5 — the new spec operation is a single unformatted line

`internal/registry/specs/docs.json:98`. The whole `/v1/bot/docs/search` object (~2.3 KB) is on one line, sharing that line with the existing `"/v1/bot/docs/{docId}"` key, in a file that is otherwise pretty-printed at 2-space indent. CI has `gofmt`, `go vet` and `golangci-lint` but no JSON formatter, so nothing catches it — and `CONTRIBUTING.md:86` explicitly says to use the existing specs as a template. The content is valid and semantically correct (I confirmed re-indenting is a pure no-op semantically), but as written this region is effectively unreviewable in a diff and will be painful in any future merge conflict. Please re-indent to match the rest of the file.

### P2-6 — `SKILL.md` pagination note leaves `--page-all` ambiguous for two commands

`skills/octo-docs/SKILL.md:83-89`. The rewrite dropped the old prose that said the docs list endpoints don't offer `--page-all`. Now the `docs list` bullet says "`--page-all` is not offered" and the `docs search` bullet says to use it, but the `docs comments list` / `docs versions list` bullet — describing an almost identical `{items, nextCursor}` shape — says nothing either way. Those two commands genuinely have no `--page-all` flag (confirmed via `--help`), so an agent reading this will try it and hit an unknown-flag error. Either add "; `--page-all` is not offered" to that bullet, or — now that the engine supports this response shape — declare `x-octo-pagination` on those two operations in a follow-up.

### P2-7 — one missing regression test on the axis this refactor actually changed

`cmd/service/service_test.go:417`. `TestParsePage_CustomFieldPaths` covers only the fully-valid custom-path case. The behavior most at risk from this rewrite is the *legacy* one: that a declared `hasMoreField` still wins over cursor presence. No test covers `has_more: false` together with a non-empty `next_cursor` — `TestPagination_PageAllMergesAllPages` ends with both `has_more:false` and `next_cursor:""`, so it can't tell which condition stopped the loop. A two-line table case would pin it.

---

## 3. Overall verdict

**APPROVE**

Spec ✅ and no P0/P1 findings. The load-bearing part of this diff is the shared `parsePage` rewrite, and the pre-existing `--page-all` tests for both the query-cursor and body-cursor paths still pass unchanged, which is the evidence that matters most. The new command's flag surface is clean (caller-facing `--keyword` / `--doc-type` / `--page-size` over wire names `q` / `docType` / `pageSize`), the cursor correctly stays in the POST body and never leaks into the query string, and `withCursorBody` preserves `q` and `docType` across pages since `resolveBody` always yields a `map[string]any`.

Everything above is a nit or a follow-up. The two I'd most like fixed before merge, since both are cheap: **P2-5** (re-indent the spec — it's a formatting no-op) and **P2-2**'s wrong comment (one line).

## 4. Suggested follow-ups

1. Re-indent `internal/registry/specs/docs.json` to 2-space, and consider adding a JSON-format check to CI so this can't recur.
2. Fix the `hasMoreField` comment, and make cursor-implies-more an explicit spec opt-in instead of an inferred one.
3. Make `parsePage` fail loud on a wrong-typed items path and a non-string/non-number cursor.
4. Seed `seenCursors` with the initial `--cursor` and check before the request; assert merged item count in the repeated-cursor test.
5. Add the `has_more:false` + non-empty `next_cursor` regression case.
6. Note the `--page-all` key-order change in the PR description — it affects all eight existing paginated operations.

## 5. Additional observations (outside the diff)

- `check-sprint / check-sprint` is failing on this PR at the "Validate Sprint field on Project Board" step. It's project-board metadata, not a code defect, and it isn't a required check — but it needs clearing separately.
- Body-schema `required` (here `["q"]`) is only rendered as a `(required)` suffix in `--help` (`cmd/service/flags.go:208-210`); it is never passed to `MarkFlagRequired`, so `docs search` with no `--keyword` reaches the backend. Pre-existing engine behavior across every operation, not introduced here — flagging in case that gap is news.
- `--doc-type`'s enum lives on `items` rather than on the property, so the engine's "(one of: …)" help rendering (`cmd/service/flags.go:205-207`) doesn't fire. The description text spells the values out manually, so the UX is fine; noting it in case array-of-enum help is worth handling generically.
- Unverified by me, and taken on the PR description's word: that the backend actually returns `{total, items, nextCursor}` with exactly this casing, that `nextCursor` is absent (rather than `null` or `""`) on the last page, that the `doc`/`sheet`/`board` enum matches the backend and the HTML type is genuinely excluded server-side, and that permission scoping works. Also unverified: any server-side cap on `pageSize`. The claimed local and deployed E2E runs cover these; I reviewed the diff and ran the suite, not the live API.

PR #113, opened 2026-07-29
feat(summary): add bot summary create command (withheld)

wanyiwang06:
## Summary

- add `octo-cli summary create` with required `--idempotency-key` and JSON request body validation
- define `summary create|list|get|result` for schema introspection while keeping the domain withheld until the backend is deployed
- embed the disabled `octo-summary` Skill with its eventual write-safety workflow
- keep owner, bot identity, and Space server-resolved; the CLI never sends a client Space header for Summary routes
- add registry and command tests for request path, headers, payload, risk, and skill discovery

## Dependencies

- includes the read-only Summary CLI foundation from #106
- requires the deployed backend from Mininglamp-OSS/octo-smart-summary#181

The Summary domain intentionally remains behind `x-octo-disabled` and the Skill behind `disabled: true` in this PR. A separate follow-up will enable both only after the backend is merged, deployed, and `BOT_SUMMARY_CREATE_ENABLED` is on.

## Safety

- creation requires explicit sources and time range
- retries must reuse the same stable idempotency key
- the Skill requires user confirmation before the write
- source authorization and effective identity remain enforced server-side

## Verification

```text
go test ./...
```

End-to-end tested locally with OpenClaw, `octo-cli --profile summary-test`, and the traditional Summary worker workflow.

Closes #114

PR #119, opened 2026-07-30
feat(docs): add full-text search command

### LINKED ISSUE #117: feat(docs): expose full-text document search in the CLI

wwwzzzcccc:
## Summary

Expose the existing permission-scoped online-document full-text search API as `octo-cli docs search`.

## Changes

- add `docs search` with keyword, repeatable document-type, cursor, and page-size flags
- extend metadata-driven pagination with configurable item/cursor response paths and POST body cursor progression
- seed repeated-cursor detection from an explicitly supplied first-page cursor
- document the command in README, project instructions, and the embedded docs skill
- add registry, request-shape, cursor progression, and loop-prevention tests

## Linked Spec

- Fixes #117
- Backend contract: `POST /v1/bot/docs/search`

## How verified

- `go test -race -shuffle=on -count=1 ./...`
- `go vet ./...`
- `go build ./cmd/octo-cli`
- `golangci-lint run` (v2.12.2): 0 issues
- `go mod tidy` with no module-file drift
- `gofmt -l .` and `git diff --check`: clean
- local Kafka + OpenSearch + backend + indexer E2E
- deployed test-environment E2E for keyword, type filters, cursor, and `--page-all`
- local final commit `bb419cf` reviewed and APPROVED before PR creation

### What does this change actually do?

Before this change, agents had to use the generic API command and manually manage search cursors. After it, `docs search` preserves the backend `{total, items, nextCursor}` object for a single page, while `--page-all` follows body cursors and emits one merged item array.

### What could break?

The shared pagination parser is the main blast radius. Existing contracts retain historical defaults; custom paths are opt-in. Repeated-cursor detection is seeded from the initial request and covers both self-repeating cursors and later loops before duplicate requests are sent.

### How do you know it works?

Tests cover registration, request body flags, body cursor placement, custom response paths, manual continuation cursors, normal progression, and loop stop conditions. Real E2E tests verified Bot authentication, permission filtering, OpenSearch results, cursor progression, and all-page aggregation.

## Checklist

- [x] Code, comments, and commit message are in English
- [x] README command reference and examples updated
- [x] CLAUDE.md command list updated
- [x] New feature includes tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **`octo-cli docs search`** — permission-scoped full-text search across online
+  documents, sheets, and boards via `POST /v1/bot/docs/search`. Supports
+  repeatable `--doc-type` filters, manual cursor continuation, and `--page-all`.
+  The metadata-driven pagination engine now supports custom item/cursor paths,
+  explicit cursor-based has-more inference, POST body cursors, and opt-in
+  repeated-cursor detection with a `PAGINATION_LOOP` error.
 - **`octo-cli summary` domain** — `summary create|list|get|result` lets a
   personal Agent create owner-only summaries from explicit authorized sources,
   then discover and cite summaries visible to its human owner. The embedded
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -23,7 +23,7 @@
 - **Search subjects** (`message search` family): a `bf_` token searches as the bot, or as a real person with `--on-behalf-of <uid>` (OBO — requires an active grant); a `uk_` token searches as the real person it belongs to. An `app_` token cannot search — the CLI rejects it locally (`validation`, in `internal/client/search_route.go`) before any request, distinct from a server-side `FORBIDDEN`.
 - `OCTO_SPACE_ID` (or `--space`) supplies space context for platform-scoped bots. Space-scoped bots resolve their space server-side.
 
-## Command Structure (9 domains, 104 operations)
+## Command Structure (9 domains, 105 operations)
 
 Service commands are auto-registered. The hand-written leaves are `schema`, `version`, `api` (generic passthrough), `config`, `auth`, and the cobra-generated `completion`.
 
@@ -48,7 +48,7 @@ octo-cli thread    create | list | get | members
 octo-cli file      upload | download | credentials | presigned
 octo-cli bot       register | set-commands | user-info | space-members | typing | heartbeat
 octo-cli event     list | ack
-octo-cli docs      create | list | get | rename | delete | forward-grant
+octo-cli docs      create | list | search | get | rename | delete | forward-grant
                content  get|edit
                sheet    get|edit
                scene    get|edit|export
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ deterministic taxonomy. There is no interactive I/O.
 
 ## Architecture
 
-octo-cli is **metadata-driven**. The entire command tree — 104 operations
+octo-cli is **metadata-driven**. The entire command tree — 105 operations
 across 9 domains — is auto-registered at startup from OpenAPI 3.x specs
 embedded into the binary. Adding or changing an endpoint means editing a
 spec, not the code.
@@ -38,7 +38,7 @@ Key properties:
 
 | Domain    | Ops | Purpose                                                        |
 |-----------|-----|----------------------------------------------------------------|
-| `docs`    | 31  | Documents, spreadsheets & whiteboards — lifecycle, body content, sheet cells (paged read), board scenes, members, comments, versions, attachments |
+| `docs`    | 32  | Documents, spreadsheets & whiteboards — lifecycle, full-text search, body content, sheet cells (paged read), board scenes, members, comments, versions, attachments |
 | `html`    | 20  | Interactive HTML documents (octo-doc, **separate backend** from `docs`) — publish immutable versions, drafts, per-doc share codes & per-uid grants, media assets, inline comments, agent element read/replace |
 | `matter`  | 14  | Todos/tasks — **temporarily withheld** while the backend API stabilizes |
 | `summary` | 4   | Personal-bot summaries — create owner-only summaries from explicit sources, then discover/read/cite. **Temporarily withheld** while the create backend (Mininglamp-OSS/octo-smart-summary#181) is merged, deployed, and enabled |
@@ -126,9 +126,10 @@ octo-cli thread create group-abc --name "design review"
 octo-cli file upload --file ./report.pdf
 octo-cli file download abc123 --jq '.data.url'
 
-# Docs — create/list, then read and incrementally edit the live body.
+# Docs — create/list/search, then read and incrementally edit the live body.
 octo-cli docs create --title "Design notes"
 octo-cli docs list --sort updatedAt:desc
+octo-cli docs search --keyword "quarterly plan" --doc-type doc --page-all
 octo-cli docs get doc-123
 octo-cli docs content get doc-123          # returns the body + base version token
 octo-cli docs import doc-123 --file ./notes.md      # replaces a doc from .md/.markdown/.docx
--- a/cmd/service/docs_test.go
+++ b/cmd/service/docs_test.go
@@ -2,12 +2,14 @@ package service
 
 import (
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
 	"net/url"
 	"strings"
 	"testing"
 
+	"github.com/Mininglamp-OSS/octo-cli/internal/output"
 	"github.com/Mininglamp-OSS/octo-cli/internal/registry"
 )
 
@@ -28,7 +30,7 @@ func TestDocs_TreeShape(t *testing.T) {
 	if docs == nil {
 		t.Fatal("missing docs service command")
 	}
-	for _, leaf := range []string{"create", "list", "get", "rename", "delete", "forward-grant"} {
+	for _, leaf := range []string{"create", "list", "search", "get", "rename", "delete", "forward-grant"} {
 		if !contains(childNames(docs), leaf) {
 			t.Errorf("docs: missing leaf %q; got %v", leaf, childNames(docs))
 		}
@@ -69,6 +71,7 @@ func TestDocs_RegistryShape(t *testing.T) {
 	cases := map[string]want{
 		"docs.create":              {"POST", "/v1/bot/docs"},
 		"docs.list":                {"GET", "/v1/bot/docs"},
+		"docs.search":              {"POST", "/v1/bot/docs/search"},
 		"docs.get":                 {"GET", "/v1/bot/docs/{docId}"},
 		"docs.rename":              {"PATCH", "/v1/bot/docs/{docId}"},
 		"docs.delete":              {"DELETE", "/v1/bot/docs/{docId}"},
@@ -205,6 +208,184 @@ func TestDocs_NoPagination(t *testing.T) {
 	}
 	if list.Flags().Lookup("page-all") != nil {
 		t.Error("docs list must not have --page-all")
+	}
+}
+
+func TestDocsSearch_FlagsAndRequestBody(t *testing.T) {
+	var gotMethod, gotPath string
+	var gotBody map[string]any
+	root, tf, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotMethod, gotPath = r.Method, r.URL.Path
+		if err := json.NewDecoder(r.Body).Decode(&gotBody); err != nil {
+			t.Errorf("decode body: %v", err)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"total":1,"items":[{"docId":"d1","title":"Plan","docType":"doc","updatedAt":123,"spaceId":"s1","highlight":"roadmap"}]}`))
+	})
+
+	search := findCmd(findCmd(root, "docs"), "search")
+	if search == nil {
+		t.Fatal("docs search command missing")
+	}
+	for _, name := range []string{"keyword", "doc-type", "cursor", "page-size", "page-all", "page-limit"} {
+		if search.Flags().Lookup(name) == nil {
+			t.Errorf("docs search missing --%s", name)
+		}
+	}
+	if search.Flags().Lookup("q") != nil || search.Flags().Lookup("docType") != nil {
+		t.Error("docs search must expose caller-facing flag names, not wire names")
+	}
+
+	root.SetArgs([]string{"docs", "search", "--keyword", "roadmap", "--doc-type", "doc", "--doc-type", "board", "--cursor", "c1", "--page-size", "25"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if gotMethod != http.MethodPost || gotPath != "/v1/bot/docs/search" {
+		t.Errorf("request = %s %s", gotMethod, gotPath)
+	}
+	if gotBody["q"] != "roadmap" || gotBody["cursor"] != "c1" || gotBody["pageSize"] != float64(25) {
+		t.Errorf("body = %#v", gotBody)
+	}
+	types, ok := gotBody["docType"].([]any)
+	if !ok || len(types) != 2 || types[0] != "doc" || types[1] != "board" {
+		t.Errorf("docType = %#v", gotBody["docType"])
+	}
+
+	var env struct {
+		OK   bool `json:"ok"`
+		Data struct {
+			Total int `json:"total"`
+			Items []struct {
+				DocID string `json:"docId"`
+			} `json:"items"`
+		} `json:"data"`
+	}
+	if err := json.Unmarshal(tf.Out.Bytes(), &env); err != nil {
+		t.Fatalf("parse envelope: %v -- out=%s", err, tf.Out.String())
+	}
+	if !env.OK || env.Data.Total != 1 || len(env.Data.Items) != 1 || env.Data.Items[0].DocID != "d1" {
+		t.Errorf("single-page response did not preserve backend object: %+v", env)
+	}
+}
+
+func TestDocsSearch_PageAllUsesBodyCursorAndStopsWithoutNextCursor(t *testing.T) {
+	pages := []string{
+		`{"total":3,"items":[{"docId":"d1"},{"docId":"d2"}],"nextCursor":"next-2"}`,
+		`{"total":3,"items":[{"docId":"d3"}]}`,
+	}
+	var cursors []string
+	root, tf, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		var body map[string]any
+		_ = json.NewDecoder(r.Body).Decode(&body)
+		cursor, _ := body["cursor"].(string)
+		cursors = append(cursors, cursor)
+		if queryCursor := r.URL.Query().Get("cursor"); queryCursor != "" {
+			t.Errorf("cursor leaked into query: %q", queryCursor)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(pages[len(cursors)-1]))
+	})
+	root.SetArgs([]string{"docs", "search", "--keyword", "plan", "--page-all"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if len(cursors) != 2 || cursors[0] != "" || cursors[1] != "next-2" {
+		t.Fatalf("body cursor progression = %v", cursors)
+	}
+	var env struct {
+		Data []struct {
+			DocID string `json:"docId"`
+		} `json:"data"`
+	}
+	if err := json.Unmarshal(tf.Out.Bytes(), &env); err != nil {
+		t.Fatalf("parse envelope: %v", err)
+	}
+	if len(env.Data) != 3 {
+		t.Errorf("merged data = %+v", env.Data)
+	}
+}
+
+func TestDocsSearch_PageAllRejectsRepeatedCursor(t *testing.T) {
+	calls := 0
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		calls++
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"total":99,"items":[{"docId":"d"}],"nextCursor":"same"}`))
+	})
+	root.SetArgs([]string{"docs", "search", "--keyword", "plan", "--page-all", "--page-limit", "10"})
+	err := root.Execute()
+	var exitErr *output.ExitError
+	if !errors.As(err, &exitErr) || exitErr.Code != "PAGINATION_LOOP" {
+		t.Fatalf("error = %#v, want PAGINATION_LOOP", err)
+	}
+	if calls != 2 {
+		t.Errorf("repeated cursor should fail after 2 requests, got %d", calls)
+	}
+}
+
+func TestDocsSearch_FinalPageStopsWithoutCursor(t *testing.T) {
+	calls := 0
+	root, tf, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		calls++
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"total":1,"items":[{"docId":"d1"}],"nextCursor":""}`))
+	})
+	root.SetArgs([]string{"docs", "search", "--keyword", "plan", "--page-all"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if calls != 1 {
+		t.Fatalf("final page should stop after 1 request, got %d", calls)
+	}
+	var env struct {
+		Data []struct {
+			DocID string `json:"docId"`
+		} `json:"data"`
+	}
+	if err := json.Unmarshal(tf.Out.Bytes(), &env); err != nil {
+		t.Fatalf("parse envelope: %v", err)
+	}
+	if len(env.Data) != 1 || env.Data[0].DocID != "d1" {
+		t.Errorf("merged data = %+v", env.Data)
+	}
+}
+
+func TestDocsSearch_PageLimitPreemptsLoopDetection(t *testing.T) {
+	calls := 0
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		calls++
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"total":99,"items":[{"docId":"d"}],"nextCursor":"same"}`))
+	})
+	root.SetArgs([]string{"docs", "search", "--keyword", "plan", "--page-all", "--page-limit", "1"})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("page-limit boundary must not report a loop: %v", err)
+	}
+	if calls != 1 {
+		t.Fatalf("page-limit should stop after 1 request, got %d", calls)
+	}
+}
+
+func TestDocsSearch_PageAllSeedsManualCursor(t *testing.T) {
+	calls := 0
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		calls++
+		var body map[string]any
+		_ = json.NewDecoder(r.Body).Decode(&body)
+		if body["cursor"] != "same" {
+			t.Errorf("cursor = %#v, want same", body["cursor"])
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"total":1,"items":[{"docId":"d1"}],"nextCursor":"same"}`))
+	})
+	root.SetArgs([]string{"docs", "search", "--keyword", "plan", "--cursor", "same", "--page-all"})
+	err := root.Execute()
+	var exitErr *output.ExitError
+	if !errors.As(err, &exitErr) || exitErr.Code != "PAGINATION_LOOP" {
+		t.Fatalf("error = %#v, want PAGINATION_LOOP", err)
+	}
+	if calls != 1 {
+		t.Fatalf("manual repeated cursor should fail after 1 request, got %d", calls)
 	}
 }
 
--- a/cmd/service/run.go
+++ b/cmd/service/run.go
@@ -352,13 +352,7 @@ func runPaginated(ctx context.Context, f *cmdutil.Factory, rt *operationRuntime,
 	// leaving only the last page's bytes. No spec op declares both today, so this
 	// never fires in practice; it is a fail-loud guard against a future spec that
 	// wires the two together and would otherwise silently corrupt --output.
-	if firstReq.OutputPath != "" {
-		err := output.ErrWithHint(
-			"internal",
-			"PAGINATION_OUTPUT_CONFLICT",
-			"operation is paginated and also writes a binary body to --output; these are mutually exclusive",
-			"operation-spec bug: an op with x-octo-pagination must not also declare an inline 2xx binary body",
-		)
+	if err := validatePaginationRequest(firstReq); err != nil {
 		_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
 		return err
 	}
@@ -369,50 +363,34 @@ func runPaginated(ctx context.Context, f *cmdutil.Factory, rt *operationRuntime,
 		return err
 	}
 	pag := rt.detail.Pagination
-	cursorParam := pag.CursorParam
-	if cursorParam == "" {
-		cursorParam = "cursor"
-	}
-	limit := 10
-	if rt.pageLimit != nil && *rt.pageLimit > 0 {
-		limit = *rt.pageLimit
-	}
-
+	cursorParam, limit := paginationControls(rt, pag)
 	merged := make([]json.RawMessage, 0, 64)
 	req := *firstReq
+	seenCursors := paginationSeenCursors(&req, rt, pag, cursorParam)
 	for page := 0; page < limit; page++ {
 		body, err := cli.Do(ctx, &req)
 		if err != nil {
 			_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
 			return err
 		}
-		data, nextCursor, hasMore, perr := parsePage(body)
+		data, nextCursor, hasMore, perr := parsePage(body, pag)
 		if perr != nil {
 			_ = f.EmitError(perr) //nolint:errcheck // best-effort emit before returning err
 			return perr
 		}
 		merged = append(merged, data...)
-		if !hasMore || nextCursor == "" {
+		if !hasMore || nextCursor == "" || page+1 >= limit {
 			break
 		}
-		// Prepare next request. Copy the whole struct so no field is silently
-		// dropped (RawBody/ContentType/BinaryResponse/OutputPath). The cursor
-		// goes wherever the spec declares it: a query parameter (GET endpoints
-		// like matter.list) or a JSON body field (POST endpoints like the
-		// message.search family). Either way we clone the container before
-		// setting so the previous page's request is never mutated.
-		next := req
-		if cursorIsQueryParam(rt, cursorParam) {
-			nextQ := url.Values{}
-			for k, vs := range req.Query {
-				nextQ[k] = append([]string(nil), vs...)
-			}
-			nextQ.Set(cursorParam, nextCursor)
-			next.Query = nextQ
-		} else {
-			next.Body = withCursorBody(req.Body, cursorParam, nextCursor)
+		if cursorWasSeen(seenCursors, nextCursor) {
+			loopErr := paginationLoopError(nextCursor)
+			_ = f.EmitError(loopErr) //nolint:errcheck // best-effort emit before returning err
+			return loopErr
 		}
-		req = next
+		if seenCursors != nil {
+			seenCursors[nextCursor] = struct{}{}
+		}
+		req = requestWithCursor(&req, rt, cursorParam, nextCursor)
 	}
 
 	out, err := json.Marshal(merged)
@@ -420,6 +398,72 @@ func runPaginated(ctx context.Context, f *cmdutil.Factory, rt *operationRuntime,
 		return err
 	}
 	return f.EmitSuccess(out)
+}
+
+func paginationControls(rt *operationRuntime, pag *registry.PaginationInfo) (cursorParam string, limit int) {
+	cursorParam = pag.CursorParam
+	if cursorParam == "" {
+		cursorParam = "cursor"
+	}
+	limit = 10
+	if rt.pageLimit != nil && *rt.pageLimit > 0 {
+		limit = *rt.pageLimit
+	}
+	return cursorParam, limit
+}
+
+func paginationSeenCursors(req *client.Request, rt *operationRuntime, pag *registry.PaginationInfo, cursorParam string) map[string]struct{} {
+	if !pag.RejectCursorRepeats {
+		return nil
+	}
+	return initialSeenCursors(req, rt, cursorParam)
+}
+
+func validatePaginationRequest(req *client.Request) *output.ExitError {
+	if req.OutputPath == "" {
+		return nil
+	}
+	return output.ErrWithHint(
+		"internal",
+		"PAGINATION_OUTPUT_CONFLICT",
+		"operation is paginated and also writes a binary body to --output; these are mutually exclusive",
+		"operation-spec bug: an op with x-octo-pagination must not also declare an inline 2xx binary body",
+	)
+}
+
+func cursorWasSeen(seen map[string]struct{}, cursor string) bool {
+	if seen == nil || cursor == "" {
+		return false
+	}
+	_, ok := seen[cursor]
+	return ok
+}
+
+func paginationLoopError(cursor string) *output.ExitError {
+	return output.ErrWithHint(
+		"internal",
+		"PAGINATION_LOOP",
+		fmt.Sprintf("backend repeated pagination cursor %q", cursor),
+		"retry without --page-all or report the backend cursor loop",
+	)
+}
+
+// requestWithCursor copies the whole request so no field is silently dropped.
+// It clones the query/body container before setting the cursor, leaving the
+// previous page's request unchanged.
+func requestWithCursor(req *client.Request, rt *operationRuntime, cursorParam, cursor string) client.Request {
+	next := *req
+	if cursorIsQueryParam(rt, cursorParam) {
+		nextQ := url.Values{}
+		for key, values := range req.Query {
+			nextQ[key] = append([]string(nil), values...)
+		}
+		nextQ.Set(cursorParam, cursor)
+		next.Query = nextQ
+	} else {
+		next.Body = withCursorBody(req.Body, cursorParam, cursor)
+	}
+	return next
 }
 
 // cursorIsQueryParam reports whether this operation declares its pagination
@@ -434,6 +478,29 @@ func cursorIsQueryParam(rt *operationRuntime, cursorParam string) bool {
 		}
 	}
 	return false
+}
+
+// initialSeenCursors seeds the repeated-cursor guard with the cursor already
+// carried by the first request. This prevents --cursor C --page-all from
+// requesting C twice when a backend incorrectly returns C as its own next cursor.
+func initialSeenCursors(req *client.Request, rt *operationRuntime, cursorParam string) map[string]struct{} {
+	seen := map[string]struct{}{}
+	if cursor := requestCursor(req, rt, cursorParam); cursor != "" {
+		seen[cursor] = struct{}{}
+	}
+	return seen
+}
+
+func requestCursor(req *client.Request, rt *operationRuntime, cursorParam string) string {
+	if cursorIsQueryParam(rt, cursorParam) {
+		return req.Query.Get(cursorParam)
+	}
+	if body, ok := req.Body.(map[string]any); ok {
+		if cursor, ok := body[cursorParam].(string); ok {
+			return cursor
+		}
+	}
+	return ""
 }
 
 // withCursorBody returns a shallow clone of the JSON body map with the cursor
@@ -451,21 +518,115 @@ func withCursorBody(body any, cursorParam, nextCursor string) map[string]any {
 	return next
 }
 
-// parsePage extracts {data:[], pagination:{has_more, next_cursor}} from a
-// backend response. Tolerant: missing fields → empty data, no more pages.
-func parsePage(body []byte) (items []json.RawMessage, cursor string, hasMore bool, exitErr *output.ExitError) {
+// parsePage extracts pagination fields using x-octo-pagination paths. Defaults
+// preserve the historical {data, pagination:{has_more,next_cursor}} contract.
+// Specs without a has-more field must explicitly opt into cursor inference.
+func parsePage(body []byte, pag *registry.PaginationInfo) (items []json.RawMessage, cursor string, hasMore bool, exitErr *output.ExitError) {
 	if len(body) == 0 {
 		return nil, "", false, nil
 	}
-	var page struct {
-		Data       []json.RawMessage `json:"data"`
-		Pagination struct {
-			HasMore    bool   `json:"has_more"`
-			NextCursor string `json:"next_cursor"`
-		} `json:"pagination"`
+	fields := paginationFieldsFor(pag)
+	items, err := parsePageItems(body, fields.items)
+	if err != nil {
+		return nil, "", false, paginationParseError(err)
 	}
-	if err := json.Unmarshal(body, &page); err != nil {
-		return nil, "", false, output.ErrWithHint("internal", "PAGINATION_PARSE", err.Error(), "response did not match expected pagination shape")
+	cursor, err = parsePageCursor(body, fields.cursor)
+	if err != nil {
+		return nil, "", false, paginationParseError(err)
 	}
-	return page.Data, page.Pagination.NextCursor, page.Pagination.HasMore, nil
+	if fields.inferHasMore {
+		return items, cursor, cursor != "", nil
+	}
+	hasMore, err = parsePageHasMore(body, fields.hasMore)
+	if err != nil {
+		return nil, "", false, paginationParseError(err)
+	}
+	return items, cursor, hasMore, nil
+}
+
+type paginationFields struct {
+	items        string
+	cursor       string
+	hasMore      string
+	inferHasMore bool
+}
+
+func paginationFieldsFor(pag *registry.PaginationInfo) paginationFields {
+	fields := paginationFields{items: "data", cursor: "pagination.next_cursor", hasMore: "pagination.has_more"}
+	if pag == nil {
+		return fields
+	}
+	if pag.ItemsField != "" {
+		fields.items = pag.ItemsField
+	}
+	if pag.CursorField != "" {
+		fields.cursor = pag.CursorField
+	}
+	if pag.HasMoreField != "" {
+		fields.hasMore = pag.HasMoreField
+	}
+	fields.inferHasMore = pag.InferHasMore
+	return fields
+}
+
+func parsePageItems(body []byte, path string) ([]json.RawMessage, error) {
+	raw, found, err := rawValueAtPath(body, path)
+	if err != nil || !found {
+		return nil, err
+	}
+	var items []json.RawMessage
+	if err := json.Unmarshal(raw, &items); err != nil {
+		return nil, fmt.Errorf("field %q must be an array: %w", path, err)
+	}
+	return items, nil
+}
+
+func parsePageCursor(body []byte, path string) (string, error) {
+	raw, found, err := rawValueAtPath(body, path)
+	if err != nil || !found || string(raw) == "null" {
+		return "", err
+	}
+	var cursor string
+	if err := json.Unmarshal(raw, &cursor); err != nil {
+		return "", fmt.Errorf("field %q must be a string or null: %w", path, err)
+	}
+	return cursor, nil
+}
+
+func parsePageHasMore(body []byte, path string) (bool, error) {
+	raw, found, err := rawValueAtPath(body, path)
+	if err != nil || !found || string(raw) == "null" {
+		return false, err
+	}
+	var hasMore bool
+	if err := json.Unmarshal(raw, &hasMore); err != nil {
+		return false, fmt.Errorf("field %q must be a boolean or null: %w", path, err)
+	}
+	return hasMore, nil
+}
+
+func paginationParseError(err error) *output.ExitError {
+	return output.ErrWithHint("internal", "PAGINATION_PARSE", err.Error(), "response did not match expected pagination shape")
+}
+
+// rawValueAtPath resolves dot-separated object keys while retaining the exact
+// JSON bytes of the selected value. Literal dots in response keys are not
+// supported by x-octo-pagination paths.
+func rawValueAtPath(body []byte, path string) (json.RawMessage, bool, error) {
+	if path == "" {
+		return nil, false, nil
+	}
+	current := json.RawMessage(body)
+	for _, part := range strings.Split(path, ".") {
+		var object map[string]json.RawMessage
+		if err := json.Unmarshal(current, &object); err != nil {
+			return nil, false, fmt.Errorf("field path %q traverses a non-object: %w", path, err)
+		}
+		next, ok := object[part]
+		if !ok {
+			return nil, false, nil
+		}
+		current = next
+	}
+	return current, true, nil
 }
--- a/cmd/service/service_test.go
+++ b/cmd/service/service_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"mime"
 	"net/http"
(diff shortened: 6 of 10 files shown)


Jerry-Xin (dismissed):
**APPROVE** — clean, well-tested addition of `docs search` (full-text) plus a careful generalization of the `--page-all` engine. Supersedes #118 (now CLOSED); reviewed on its own merits at head `bb419cfe`.

Verified locally at the live head:
- `go build ./...` clean; `go test ./cmd/service/... ./internal/registry/...` pass (targeted `TestDocsSearch_*`, `TestParsePage_CustomFieldPaths`, `TestGetOperationDocsSearch_Pagination`, op-count all green).
- `golangci-lint run ./...` at the CI-pinned v2.12.2: **0 issues**.
- CI: `build & test (1.24.x)` green, osv-scan green. (`check-sprint` red is the board gate, non-blocking.)

**Backward-compat (byte-verified at head):** every pre-existing paginated op — `matter.list`, `matter.timeline.list`, `message.search`/`.all`/`.files`/`.media`, `skill.list`, `skill.mine.list` — still declares `cursorField: pagination.next_cursor` + `hasMoreField: pagination.has_more` and **no `itemsField`**. In `parsePage`, an empty `itemsField` defaults to `data` and the set cursor/has-more paths are used as before, so the historical `{data, pagination:{has_more,next_cursor}}` contract is fully preserved. The `HasMoreField == "" && CursorField == ""` fallback correctly does not fire for these (both are set). Confirmed by `TestParsePage_CustomFieldPaths` and the untouched `matter.list` pagination test.

**Correctness highlights:**
- `docs.search` pagination is cursor-in-body (POST), and the page loop routes the cursor to the body (not the query) via `withCursorBody`; `TestDocsSearch_PageAllUsesBodyCursorAndStopsWithoutNextCursor` asserts the cursor never leaks into the query string. With `hasMoreField` absent, `hasMore = cursor != ""` — matches the documented `{total, items, nextCursor}` contract.
- `seenCursors` loop guard (seeded from the first request's cursor) prevents an infinite/duplicate loop when a backend echoes the same `nextCursor`; both the backend-echo and manually-seeded-`--cursor` cases are covered by tests.
- `UseNumber()` avoids float coercion of large integer IDs when re-marshaling items.
- The earlier "loud-fail → silent-success" worry does not manifest as a regression: invalid JSON still returns a `PAGINATION_PARSE` ExitError, and tolerance for missing/mismatched fields (empty result, no error) is identical to the prior struct-based decode.

**Security/wiring:** `x-octo-risk: "read"` is present and correct. `--keyword` is a JSON body value marshaled via `encoding/json` (generic passthrough — no SQL/shell/string concatenation), so there is no client-side injection surface; query semantics are the backend's concern. Flags are caller-facing (`--keyword`, `--doc-type`) rather than wire names (`q`, `docType`), asserted by tests. Docs/README/SKILL op counts bumped consistently (104→105, docs 31→32).

🟡 Non-blocking:
- `--keyword` backs the schema-`required` `q`, but `registerBodyFlags` only appends "(required)" to the flag description — it does **not** call `cmd.MarkFlagRequired` for body-required fields (unlike header/query params). So omitting `--keyword` isn't blocked client-side; it relies on the server to reject the missing `q`. This is a pre-existing generator limitation across the CLI, not specific to this PR; the spec correctly declares `"required":["q"]`. Worth addressing generically in the generator later.
- Soft deploy-order: this ships a new `POST /v1/bot/docs/search` client op. If the backend endpoint isn't deployed yet, the command returns a server error at runtime — graceful, and consistent with the metadata-driven "ship spec independently" model, but flagging in case rollout ordering matters.

Nice work on the pagination generalization and the thorough table-driven tests.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. Clean, well-tested feature addition with good backward compatibility. One formatting nit (minified spec entry).

yujiawei (changes requested):
# Code Review — PR #119 (octo-cli)

Reviewed at head `bb419cf`, against merge base `479ccbf`. Verified locally: `go build ./...`, `go vet ./...`, `go test ./...` all pass; `internal/registry/specs/docs.json` parses.

**Verdict: CHANGES_REQUESTED** — one blocking issue (P1) in the shared pagination loop, which affects existing commands outside this feature's scope. The `docs search` command itself looks well-built.

---

## 1. Spec compliance

Linked issue #117 asks for CLI integration of permission-scoped full-text document search, including cursor pagination and documentation.

**Spec: ✅**

- **Missing:** none. `docs search` is registered (`internal/registry/specs/docs.json:98`), cursor pagination works via `--cursor` / `--page-all` / `--page-limit`, and docs are updated in `README.md`, `CLAUDE.md`, and `skills/octo-docs/SKILL.md`.
- **Over-build:** one item to call out, though I'm not counting it as a spec failure. The repeated-cursor loop guard (`cmd/service/run.go:257`, `273-277`) is not required by the new endpoint's contract, yet it changes runtime behavior for **every** pre-existing paginated operation (`matter.list`, the `message.search` family, the marketplace list ops). The PR's E2E evidence covers `docs search` only. See P1 below.
- **Deviation:** none. The operation counts are internally consistent — `README.md` 104→105 matches the domain-table sum, `docs` 31→32 matches `internal/registry/loader_test.go:49`.

The generalization of `parsePage` to spec-declared item/cursor paths *is* necessary work for the `{total, items, nextCursor}` response shape, and I read it as in scope.

---

## 2. Code quality

### P1 — The repeated-cursor guard silently truncates `--page-all` and neuters `--page-limit`

`cmd/service/run.go:273-277`

```go
if _, seen := seenCursors[nextCursor]; seen {
    break
}
seenCursors[nextCursor] = struct{}{}
```

The guard compares each returned cursor against **every** cursor seen so far and exits the loop with no error, no warning, and `ok: true`. That assumes an opaque cursor is unique per page. Plenty of backends don't work that way — a scroll handle / session token that stays constant while you post it back for successive batches is a common pagination style, and so is an alternating token.

This is demonstrable today, not hypothetical. `cmd/service/service_test.go:401-415` serves a constant `next_cursor:"c"`:

```go
page := `{"data":[{"id":"x"}],"pagination":{"has_more":true,"next_cursor":"c"}}`
...
root.SetArgs([]string{"matter", "list", "--page-all", "--page-limit", "2"})
```

Raising that to `--page-limit 5` and asserting 5 calls fails:

```
service_test.go:413: expected 5 calls bounded by --page-limit, got 2
```

So for any backend with a stable cursor token, `--page-limit` no longer bounds anything — the guard does, at 2 pages. Before this PR the same invocation walked all 5. The caller gets a partial result set in an envelope that is byte-for-byte indistinguishable from a complete one.

Second effect: the repeated page's items are merged *before* the repeat is detected, so the output also contains duplicates. With a one-item page I measured `calls=2 mergedItems=2`.

Third effect, on test coverage: `TestPagination_PageLimitStops` now passes for the wrong reason. Both mechanisms stop at 2 for that fixture, so the test can no longer distinguish them — a future regression in `--page-limit` would go undetected.

Worth noting that before this PR an infinite-loop backend was already bounded by `--page-limit` (default 10), so the guard's upside is saving up to 9 wasted requests, while its downside is silently incomplete data on unrelated commands.

Suggested fix, in rough order of preference:

1. Narrow the guard to what its own docstring at `run.go:318-320` describes — break only when the returned cursor equals the cursor **just sent** (self-repeat), not any historical cursor. That covers the stated motivation (`--cursor C --page-all` re-requesting `C`) without penalizing stable-token backends across the board.
2. Additionally require the page to have returned zero items before treating a repeat as a loop.
3. Whichever guard you keep, don't exit silently — surface a diagnostic (stderr under `--verbose`, or a `PAGINATION_LOOP` error) so a caller can tell truncation from completion.
4. Update `service_test.go:401` to use advancing cursors (`c1`, `c2`, …) so `--page-limit` stays pinned independently, and add a case asserting the guard's own behavior.

### P2 — `--page-all` now reorders the keys of every returned item

`cmd/service/run.go:388-397`

```go
for _, item := range rawItems {
    raw, err := json.Marshal(item)
    ...
}
```

Items used to pass through as `json.RawMessage`, preserving the backend's byte layout. They now round-trip through `map[string]any`, and `json.Marshal` emits map keys in sorted order. Confirmed:

```
OLD: [{"z":1,"a":"<b>hi</b>","m":12345678901234567890}]
NEW: [{"a":"<b>hi</b>","m":12345678901234567890,"z":1}]
```

This applies to every pre-existing paginated command, not just `docs search`. Semantically harmless for JSON consumers, but it is an unannounced output change — please note it in `CHANGELOG.md`. Credit where due: `decoder.UseNumber()` (`run.go:365`) is the right call and keeps large integers from being mangled into `float64`, as the output above shows.

### P2 — `parsePage` no longer fails loudly on an unexpected response shape

`cmd/service/run.go:359-397`

Previously, unmarshalling into the fixed struct rejected anything that wasn't `{data, pagination}` and returned a `PAGINATION_PARSE` error. Now `decoder.Decode(&page)` accepts any valid JSON, and `valueAtPath` returns `nil` for a path that doesn't resolve:

```go
rawItems, ok := valueAtPath(page, itemsField).([]any)
if !ok {
    rawItems = nil
}
```

A bare JSON array, an `itemsField` typo in a spec, or an `items` that is an object rather than an array all now produce a successful `ok: true, data: []`. A spec bug becomes "no results," which is a hard failure mode to debug. Consider distinguishing "path absent" (tolerate — legitimately empty last page) from "path present but not an array" (error).

### P2 — Default-path resolution couples two unrelated spec fields

`cmd/service/run.go:373-382`

```go
hasMoreField = pag.HasMoreField
// Older specs omitted itemsField but did declare the historical cursor
// and has-more paths. Keep data as the default in that case.
if pag.HasMoreField == "" && pag.CursorField == "" {
    hasMoreField = "pagination.has_more"
}
```

Net effect: declaring `cursorField` alone silently flips has-more detection from "read `pagination.has_more`" to "infer from a non-empty cursor." That's correct for the specs in the tree today (all existing ones declare both fields, so behavior is unchanged — I checked all of `marketplace.json`, `message.json`, `matter.json`), but the implicit dependency between `cursorField` and `hasMoreField` is a trap for the next spec author. An explicit opt-in (a dedicated `inferHasMore: true`, or a sentinel) would read better. A table-driven test over the four `(cursorField, hasMoreField)` combinations would also lock the intent down; `TestParsePage_CustomFieldPaths` only covers the both-declared case.

### P2 — Flag naming diverges from sibling `docs` commands

From `--help` on the built binary:

```
docs list    --pageSize int     items per page (1-100, default 20)
docs search  --page-size int    items per page
             --doc-type strings document type filter
docs create  --docType string
```

`docs search` uses kebab-case (`--doc-type`, `--page-size`) via `x-octo-flag`; its siblings expose the raw camelCase wire names. Kebab-case is the better style, but the split inside one command group is a papercut for scripted callers. Either add camelCase aliases here, or file a follow-up for a consistent repo-wide rename.

### P2 — The new operation is one 1729-character minified line

`internal/registry/specs/docs.json:98`

The rest of the file is 2-space pretty-printed; the only other long lines in it are long `description` strings sitting on properly indented lines. As written, any future edit to `docs.search` shows up as a single whole-line diff, which is effectively unreviewable. Nothing in `lefthook.yml` or the `Makefile` checks spec formatting (both cover Go only), so this won't self-correct. Please reformat to match the file.

### P2 — Envelope pagination metadata was not generalized alongside `parsePage`

`internal/output/envelope.go:100-110`

`splitPagination` is still hardcoded to `{data, pagination}`, so a single-page `docs search` gets no `_pagination` block and callers must reach into `data.nextCursor`. This is documented in the spec description and in `skills/octo-docs/SKILL.md`, so I take it as deliberate — but it leaves the metadata-driven pagination story half-finished. Worth a follow-up issue.

### Nits

- `cmd/service/run.go:401` — `cursor = fmt.Sprint(value)` will happily stringify a non-scalar cursor (`map[...]`) and put it on the wire. A type check plus `PAGINATION_PARSE` would fail more usefully.
- `cmd/service/run.go:408-423` — `valueAtPath` splits on `.`, so a response key containing a literal dot is unreachable. Fine for every spec in the tree; worth a one-line comment.
- `docs search --keyword x --doc-type html` is accepted locally and sends `{"docType":["html"]}`, even though the spec description says HTML is intentionally not exposed. I checked before flagging this as a defect: enums are **never** validated client-side in this CLI (`docs create --docType bogus` and `docs list --sort bogus` both pass straight through, and `cmd/service/validate.go` only covers whiteboard fractional indexes). So this is pre-existing framework behavior, not something this PR introduced. Mentioning it only so the "intentionally not exposed" wording isn't read as a local guarantee.
- Same category: the spec marks `q` required, but body-level `required` only appends "(required)" to the help text — `MarkFlagRequired` is not called (`cmd/service/flags.go:171-176`), so `docs search --dry-run` with no `--keyword` builds a request with no body at all. Framework-wide and pre-existing; listed for completeness.

### What I checked and found correct

- `withCursorBody` (`run.go:344-353`) clones per page, so the previous page's body is never mutated — and `resolveBody` does return `map[string]any`, so the type assertions in `withCursorBody` / `requestCursor` hold for this operation. `docs search --keyword plan --page-all` correctly carries `q` forward onto page 2.
- Cursor placement: body for this POST endpoint, never leaked into the query string — asserted in `TestDocsSearch_PageAllUsesBodyCursorAndStopsWithoutNextCursor`.
- All three terminal conditions work: absent `nextCursor`, `""`, and `null` all stop the loop.
- No regression in resolved field paths for existing paginated ops — every one of them declares both `cursorField: "pagination.next_cursor"` and `hasMoreField: "pagination.has_more"` and no `itemsField`, so resolution is byte-identical to the old hardcoded behavior.
- The new tests are substantive rather than tautological; they assert wire shape, flag surface, cursor progression, and merged output.
- `x-octo-risk: "read"` is the right classification for a search POST.

---

## 3. Blind spots / what I could not verify

- The real backend's cursor semantics for `POST /v1/bot/docs/search` — whether `nextCursor` advances per page or is a stable handle, and whether it is returned on the final page. This is exactly what determines the blast radius of the P1.
- Whether the backends behind `matter.list`, the `message.search` family, and the marketplace list ops emit advancing cursors. The PR's E2E evidence covers `docs search` only, and those three families are the ones newly exposed to the guard.
- Permission scoping, OpenSearch relevance, and indexer behavior — not reachable from the CLI repo.
- `golangci-lint` is not installed in my environment, so I relied on the PR's claim of 0 issues.
- I did not audit the separate `html` domain for a competing search surface.

## 4. Summary of requested changes

Blocking: narrow or instrument the repeated-cursor guard (P1) and restore an independent `--page-limit` assertion in `service_test.go:401`.

Non-blocking but please address in this PR if convenient: reformat the `docs.json` entry (P2), and add a `CHANGELOG.md` note for the item key-ordering change (P2). The rest can be follow-ups.

Jerry-Xin (dismissed):
The PR is in scope for `octo-cli`: it adds a metadata-driven `docs search` command and extends the existing pagination engine for that command’s backend contract.

💬 Non-blocking

🟡 Warning: `docs.search` declares `q` as required in `internal/registry/specs/docs.json:117`, but promoted body fields are only labeled as required in help text and are not locally validated in `cmd/service/flags.go:208`. This matches the current generic `--data` escape-hatch design, so it is not a blocker, but `octo-cli docs search` without `--keyword` will reach the backend instead of failing early with a local validation error.

✅ Highlights

🔵 Suggestion-level positive note: The pagination extension keeps legacy `{data, pagination}` defaults intact in `cmd/service/run.go:427`, while making custom item/cursor paths opt-in for `docs.search`.

🔵 Suggestion-level positive note: Repeat-cursor detection is scoped behind `rejectCursorRepeats` and seeded from an initial cursor, which avoids changing behavior for existing paginated commands while protecting the new search loop in `cmd/service/run.go:242` and `cmd/service/run.go:254`.

Verification performed: `go test ./cmd/service ./internal/registry`, `go test ./...`, and `go vet ./...` all passed locally.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #119 (octo-cli)

## Summary

This PR adds `docs search` — a full-text search command backed by `POST /v1/bot/docs/search` — and extends the shared pagination engine with configurable item/cursor response paths, POST body cursor progression, cursor-inference mode, and repeated-cursor loop detection. The refactoring decomposes the monolithic `parsePage` and `runPaginated` into focused helpers (`paginationControls`, `paginationSeenCursors`, `requestWithCursor`, `rawValueAtPath`, etc.) while keeping existing operations on their historical defaults via opt-in spec fields.

The change is well-scoped, well-tested, and maintains backward compatibility. Two P2 behavioral tightenings are noted below.

## Verification

Static analysis only at head `efe942b4c384216cf6ad3b769dc8e1a11888e3f5`; build and tests not executed in this environment.

- ✅ **Backward compatibility** — `paginationFieldsFor` defaults (`data`, `pagination.next_cursor`, `pagination.has_more`) match the old hardcoded struct tags. Existing specs (matter, message, marketplace) set `cursorField`/`hasMoreField` explicitly, overriding the same values. New fields (`ItemsField`, `InferHasMore`, `RejectCursorRepeats`) are opt-in with zero-value defaults that preserve old behavior.
- ✅ **Cursor loop detection** — `paginationSeenCursors` returns nil for ops without `RejectCursorRepeats`, so `cursorWasSeen` is always false — existing operations are unaffected. For docs.search, the seed correctly captures the initial `--cursor` value, and the map is updated after each page.
- ✅ **Test coverage** — Registration (`TestDocs_TreeShape`, `TestDocs_RegistryShape`, `TestGetOperationDocsSearch_Pagination`), request body shape (`TestDocsSearch_FlagsAndRequestBody`), body cursor progression (`TestDocsSearch_PageAllUsesBodyCursorAndStopsWithoutNextCursor`), loop detection (`TestDocsSearch_PageAllRejectsRepeatedCursor`, `TestDocsSearch_PageAllSeedsManualCursor`), custom field paths (`TestParsePage_CustomFieldPaths`), inference opt-in (`TestParsePage_InferHasMoreRequiresOptIn`), type validation (`TestParsePage_RejectsInvalidFieldTypes`), and legacy stable-cursor backward compatibility (`TestPagination_LegacyStableCursorStillUsesPageLimit`).
- ✅ **Docs consistency** — CLAUDE.md (105 ops, search in command tree), README.md (105 ops, 32 docs ops, search example), SKILL.md (search usage + pagination note) all aligned.
- ✅ **Spec contract** — `docs.json` declares `POST /v1/bot/docs/search` with `required: ["q"]`, `docType` array enum `[doc, sheet, board]`, cursor/pageSize, and response `{total, items[], nextCursor}`. Pagination extension sets `itemsField: "items"`, `cursorField: "nextCursor"`, `inferHasMore: true`, `rejectCursorRepeats: true`.

## Findings

No P0/P1 issues. Two P2 behavioral tightenings below.

### P2 — `parsePageItems` rejects null items where old code accepted silently (`cmd/service/run.go:457`)

The old `parsePage` used a Go struct (`Data []json.RawMessage`), which unmarshals JSON `null` to a nil slice with no error. The new `parsePageItems` explicitly errors: `field "data" must be an array, got null`. This tightening applies to all existing paginated operations (matter.list, message.search.*, marketplace.*), not just docs.search.

Practical risk is low — response schemas declare items as `"type": "array"` (not nullable), and backends should always return arrays. Consider either (a) treating `null` as empty for backward parity, or (b) documenting this tightening in the commit message.

### P2 — `rawValueAtPath` errors on null intermediate objects (`cmd/service/run.go:491`)

When a path like `"pagination.next_cursor"` traverses through `"pagination": null`, `json.Unmarshal` into `map[string]json.RawMessage` fails. The old struct-based parser would yield zero values (`HasMore=false`, `NextCursor=""`). Same scope and risk profile as the null-items case above — low practical risk for conforming backends, but a subtle behavioral change for all paginated operations.

- `requestWithCursor` deep-clones query values and creates a new body map — no mutation of the previous page's request.
- `withCursorBody` shallow-clones the body map; cursor is set at the top level only, so nested references are safely shared.
- `initialSeenCursors` correctly seeds from query or body cursor, and skips empty cursors.
- `cursorWasSeen` short-circuits on nil map and empty cursor — no false positives for ops without loop detection.
- `truthy` handles both boolean and string `"true"` for the new spec fields.
- `TestPagination_PageLimitStops` was updated to use unique cursors (avoiding false loop detection), and a separate `TestPagination_LegacyStableCursorStillUsesPageLimit` proves stable cursors still work for ops without `rejectCursorRepeats`.
- `rawValueAtPath` correctly handles empty path (returns `nil, false, nil`), but is never called with empty path from production code since `paginationFieldsFor` always sets non-empty defaults.

## Verdict: APPROVED

Well-structured feature addition with thorough test coverage and clean backward compatibility through opt-in spec fields. The two P2 items are behavioral tightenings with low practical risk — not merge blockers.

yujiawei (changes requested):
# Code Review — PR #119 (octo-cli)

Re-review at head `efe942b`, against merge base `479ccbf`. Verified locally at this head: `go build ./...`, `go vet ./...`, and `go test ./...` all pass; `internal/registry/specs/docs.json` parses; `octo-cli docs search --help` and `octo-cli schema docs.search` inspected on the built binary.

**Verdict: CHANGES_REQUESTED** — one blocking issue (P1): a three-line addition to `parsePageItems` narrows the response contract for all eight pre-existing paginated operations. Everything else is non-blocking.

Credit first: the previous round's blocking issue is genuinely fixed, and fixed well. The repeated-cursor guard is now **opt-in per spec** (`rejectCursorRepeats`), it **fails loud** (`PAGINATION_LOOP`) instead of silently truncating, `inferHasMore` replaced the implicit `cursorField`/`hasMoreField` coupling, items stay `json.RawMessage` so byte layout and key order are preserved, and `TestPagination_PageLimitStops` now uses advancing cursors with a dedicated `TestPagination_LegacyStableCursorStillUsesPageLimit` pinning legacy behavior independently. The minified spec entry is reformatted. That is a complete response to every item raised.

---

## 1. Spec compliance

Linked issue #117 asks for CLI integration of permission-scoped full-text document search, including cursor pagination and documentation.

**Spec: ✅**

- **Missing:** none. `docs search` is registered (`internal/registry/specs/docs.json:98-110`), cursor pagination works via `--cursor` / `--page-all` / `--page-limit`, and docs are updated in `README.md`, `CLAUDE.md`, and `skills/octo-docs/SKILL.md`.
- **Over-build:** resolved. Last round I flagged that the loop guard changed runtime behavior for every pre-existing paginated operation. It is now gated on `rejectCursorRepeats`, which only `docs.search` declares — `paginationSeenCursors` (`cmd/service/run.go:288-293`) returns `nil` for every other op, and `cursorWasSeen` short-circuits on `nil`. I byte-checked all eight pre-existing paginated ops (`matter.json:42,266`, `message.json:145,184,223,262`, `marketplace.json:1729,2587`): every one declares `cursorField: pagination.next_cursor` + `hasMoreField: pagination.has_more`, no `itemsField`, no `inferHasMore`, no `rejectCursorRepeats`. `paginationFieldsFor` resolves those to exactly the old hardcoded paths.
- **Deviation:** none. Operation counts are internally consistent: `README.md:15` and `CLAUDE.md:26` 104→105, the `README.md` domain table sums to 105, and `docs` 31→32 matches `internal/registry/loader_test.go:45`.

---

## 2. Code quality

### P1 — `data: null` is now a hard failure on every pre-existing paginated command

`cmd/service/run.go:450-452`

```go
if string(raw) == "null" {
    return nil, fmt.Errorf("field %q must be an array, got null", path)
}
```

The old decoder unmarshalled into a fixed struct with a `Data []json.RawMessage` field. Go's `encoding/json` unmarshals a JSON `null` into a slice as a no-op — nil slice, no error — so `{"data":null,"pagination":{...}}` was a *valid empty page*. This PR turns it into `PAGINATION_PARSE`.

Measured at both heads. Old decoder shape:

```
OLD data:null -> err=<nil> items=[] hasMore=false cursor=""
```

Current head, `parsePage([]byte(`{"data":null,"pagination":{"has_more":false,"next_cursor":null}}`), nil)`:

```
PAGINATION_PARSE: field "data" must be an array, got null
```

End-to-end on an untouched pre-existing op, `matter list --page-all` against a stub returning that body:

```
calls=1 err=PAGINATION_PARSE: field "data" must be an array, got null out=
```

Note `out=` — empty. The error path at `run.go:251-253` returns before `f.EmitSuccess`, so **every page already accumulated in `merged` is discarded**. If page 1 returns 50 items with `has_more:true` and page 2 returns `data:null`, the caller gets a nonzero exit and zero rows where it previously got 50 rows and `ok:true`.

Why this matters rather than being theoretical: `null` for an empty list is what a Go handler emits by default for an uninitialized slice, and the CLI's own contract tolerated it until this commit — so no backend was ever obliged to avoid it. This affects `matter list`, the four `message search` ops, and the two marketplace list ops, none of which this PR intends to touch. The PR description states "Existing contracts retain historical defaults"; for this one case that is no longer true.

There is a related asymmetry worth fixing in the same edit: an **absent** items path is tolerated (returns empty, `run.go:447-449`), and a path under a `null` parent is tolerated (`rawValueAtPath` unmarshals `null` into a nil map, so the key lookup misses), but an **explicitly null** items path is fatal. Those three cases should not diverge.

Suggested fix — delete the three lines. `json.Unmarshal([]byte("null"), &items)` already yields a nil slice with no error, which restores the historical behavior exactly, and the wrong-type cases in `TestParsePage_RejectsInvalidFieldTypes` (`{"data":{}}`) still error through `run.go:454-456`. I applied that deletion locally and re-ran `go test ./...`: fully green, so no existing test pins the current behavior. If you prefer to keep rejecting `null`, it needs to be gated behind an opt-in spec field the way `inferHasMore` and `rejectCursorRepeats` are, so pre-existing ops are unaffected — plus a test either way, since nothing covers `null` today.

### P2 — The loop guard preempts the loop's own terminators

`cmd/service/run.go:254-262`

```go
if cursorWasSeen(seenCursors, nextCursor) {
    ...
    return loopErr
}
merged = append(merged, data...)
if !hasMore || nextCursor == "" {
    break
}
```

The repeat check runs *before* both the merge and the `!hasMore` terminator. So a page that says "no more results" while echoing a cursor already used produces a hard `PAGINATION_LOOP` and discards every page fetched so far — even though no further request would have been issued. That is a false positive by construction: the repeat is unobservable when the walk was already ending.

Latent today, not live: it needs `rejectCursorRepeats` together with an explicit `hasMoreField`, and `docs.search` uses `inferHasMore` instead (where `hasMore == cursor != ""`, so a non-empty repeated cursor always implies `hasMore`). No spec in the tree hits it. Still worth closing while the code is fresh — move the repeat check below the `!hasMore || nextCursor == ""` break, and, for the page-limit boundary, only treat a repeat as a loop when another request would actually follow. A table-driven test over (`repeat`, `hasMore`, `pageLimit remaining`) would pin the intent.

Separately, on a *genuine* loop the same code discards all previously merged pages. Failing loud was the right call, but consider putting the count of pages/items dropped into the error hint so a caller can tell how much it lost.

### P2 — No `CHANGELOG.md` entry

`CHANGELOG.md:8` — the `[Unreleased] / Added` section, and the file's own header ("All notable changes to `octo-cli` are documented in this file").

Every comparable prior change has an entry — the `message search` family and the `summary` domain both do. This PR adds a caller-facing command family *and* extends the `x-octo-pagination` contract with three new fields (`itemsField`, `inferHasMore`, `rejectCursorRepeats`) plus a new `PAGINATION_LOOP` error code that agent callers may need to handle. `CHANGELOG.md` is untouched (`git diff --name-only 479ccbf efe942b` has no changelog line).

### P2 — Flag naming still diverges inside the `docs` group

From `--help` on the binary built at this head:

```
docs list    --pageSize int      items per page (1-100, default 20)
docs search  --page-size int     items per page
             --doc-type strings  document type filter
docs create  --docType string
```

Carried over from the previous round and still non-blocking. Kebab-case is the better style; the split inside one command group is a papercut for scripted callers. Either add camelCase aliases here or file a follow-up for a repo-wide rename.

### P2 — `docs search` gets `--page-all` while its identical-shaped siblings do not

`skills/octo-docs/SKILL.md:78-89` now documents that `docs search` returns `{total, items, nextCursor}` and supports `--page-all`, while `docs comments list` and `docs versions list` return `{items, nextCursor}` and still require manual `--cursor` walking. The machinery this PR builds (`itemsField` + `inferHasMore`) is exactly what those two need. Not a defect in this diff — the doc is honest about the split — but it is now a one-line-per-op follow-up rather than an engine change. Worth an issue.

### P2 — Envelope pagination metadata still not generalized

`internal/output/envelope.go:100-120` — `splitPagination` remains hardcoded to `{data, pagination}` with an array-valued `data`, so a single-page `docs search` gets no `_pagination` block and callers must reach into `data.nextCursor`. Documented in the spec description and in `SKILL.md`, and asserted by `TestDocsSearch_FlagsAndRequestBody`, so I read it as deliberate. Follow-up.

### Nits

- `internal/registry/specs/docs.json:178-183` — `"type": ["integer", "null"]` on `updatedAt` is dropped by `stringOf` (`internal/registry/loader.go`), so `octo-cli schema docs.search` renders `"updatedAt": {}` with no type at all. Pre-existing framework behavior, not introduced here — `summary.json:114` already has the same shape and `octo-cli schema summary.get` loses `result`'s type identically. Mentioning it because agents introspect via `schema`, and a plain `"type": "integer"` would render correctly today.
- `docs/octo-cli-search-command-tree.md:14` still reads `docs (31)` and omits `search` from the direct-leaf list. It reads as a point-in-time design doc for the message-search work rather than living reference material, so probably fine to leave — but if it is meant to stay current, it needs the same +1 the README got.
- `cmd/service/run.go:491-508` — `rawValueAtPath` splits on `.`, so a response key containing a literal dot is unreachable. Correctly called out in the doc comment now.
- `internal/registry/loader.go:191` — `LimitParam` is parsed and stored but never read anywhere outside tests. Pre-existing dead field; `docs.search` reasonably declines to set it.

### What I checked and found correct

- **Backward compatibility of the field-path resolution**, byte-verified per op as described in section 1. `paginationFieldsFor` (`run.go:427-443`) defaults to `items:"data"`, `cursor:"pagination.next_cursor"`, `hasMore:"pagination.has_more"` — identical to the old hardcoded struct.
- **The guard is genuinely opt-in.** `TestPagination_LegacyStableCursorStillUsesPageLimit` walks all 5 pages against a fixed `next_cursor:"stable"` on `matter.list`, which is precisely the regression from last round, now pinned.
- **`--page-limit` is independently pinned again.** `TestPagination_PageLimitStops` uses advancing `c1..c5`, so it no longer passes for the wrong reason.
- **`inferHasMore` is a real opt-in.** `TestParsePage_InferHasMoreRequiresOptIn` asserts both directions on the same body.
- **Loud failure on wrong types** (`TestParsePage_RejectsInvalidFieldTypes`) covers items-as-object, cursor-as-object, has-more-as-string. I additionally confirmed a bare top-level JSON array and a numeric `next_cursor` both produce `PAGINATION_PARSE`.
- **Key order and large integers are preserved.** Items stay `json.RawMessage` end to end — no `map[string]any` round-trip, so no sorted-key reordering and no float64 mangling. This is strictly better than the previous head.
- **Cursor plumbing.** `requestWithCursor` / `withCursorBody` (`run.go:327-341`, `383-395`) clone per page, so the previous page's body is never mutated and `q` / `pageSize` carry forward. `cursorIsQueryParam` correctly routes `docs.search` to the body — `TestDocsSearch_PageAllUsesBodyCursorAndStopsWithoutNextCursor` asserts the cursor never leaks into the query string. Verified `--page-size 25` survives onto page 2.
- **Terminal conditions.** Absent, `""`, and `null` `nextCursor` all stop the walk.
- **Guard behavior has its own tests now** — `TestDocsSearch_PageAllRejectsRepeatedCursor` (fails after 2 requests) and `TestDocsSearch_PageAllSeedsManualCursor` (fails after 1, seeded from `--cursor`).
- **Spec entry is properly pretty-printed** at 2-space indent, matching the rest of `docs.json`.
- `x-octo-risk: "read"` is right for a search POST. Flag surface (`--keyword`/`--doc-type`/`--cursor`/`--page-size`) matches the wire names via `x-octo-flag`, and `TestDocsSearch_FlagsAndRequestBody` asserts the wire names are *not* exposed.

---

## 3. Blind spots — what I could not verify

- **Whether any Octo backend actually emits `"data": null`** for an empty page. This is the reachability question for the P1 and it is not answerable from this repo. My position is that the CLI should not narrow a contract it previously accepted, independent of current backend behavior.
- The real cursor semantics of `POST /v1/bot/docs/search` — whether `nextCursor` is absent or echoed on the final page. The PR reports E2E verification; I could not reproduce it.
- Permission scoping, OpenSearch relevance, and indexer behavior — not reachable from the CLI repo.
- `golangci-lint` is not installed in my environment; I relied on the PR's claim of 0 issues at v2.12.2.
- I did not audit the separate `html` domain for a competing search surface.
- The `README.md` 9-domain table excludes `marketplace` (25) and `summary` (3), so the documented "105 operations" is not the 133 that `loader_test.go` pins. That framing predates this PR and its +1 increments are self-consistent, so I am not treating it as a finding here.

## 4. Summary of requested changes

**Blocking:** restore tolerance for `data: null` (delete `cmd/service/run.go:450-452`, or gate it behind an opt-in spec field), and add a test covering the null case either way.

**Non-blocking, nice to have in this PR:** add the `CHANGELOG.md` entry, and reorder the repeat check below the `!hasMore` terminator. The rest are fine as follow-ups.

Jerry-Xin (dismissed):
# Code Review — PR #119 (octo-cli)

Re-review at head `f9b0b47a`, base `main`. Verified locally at this head: `gofmt`, `go vet`, `go mod tidy` (no drift), `golangci-lint v2.12.2` (CI-pinned, 0 issues), and `go test -race ./...` across all 12 packages — all green.

**Verdict: APPROVE.**

## Credit — the backward-compat regression flagged last round is real, and it is now fixed

Full credit to @yujiawei and @mochashanyao for catching this on the prior head (`efe942b`), and to the author for the surgical fix. I under-called it in my previous APPROVE at `efe942b`; their P1 was correct.

**The regression (at `efe942b`):** `parsePageItems` contained
```go
if string(raw) == "null" {
    return nil, fmt.Errorf("field %q must be an array, got null", path)
}
```
The pre-#119 decoder unmarshalled `data` into a `[]json.RawMessage`, so a backend response of `{"data":null,"pagination":{...}}` (a valid empty page — e.g. an empty result set) decoded to a nil slice with **no error**. The strict `null`-check turned that into a hard `PAGINATION_PARSE` failure for all eight pre-existing paginated operations (matter.list, matter.timeline.list, message.search×4, skill.list, skill.mine.list) whose specs use the default `data` items path.

**The fix at `f9b0b47`** (`cmd/service/run.go`): the explicit `null` guard is removed. `null` items now fall through to `json.Unmarshal(raw, &items)`, which is a no-op for a JSON `null` → nil slice, no error — restoring the legacy contract. I byte-verified the default path (no `itemsField`; `cursorField=pagination.next_cursor`; `hasMoreField=pagination.has_more`) end-to-end:

- `{"data":null,"pagination":{"has_more":false,"next_cursor":null}}` → empty page, no error (cursor guard `string(raw)=="null"` returns `""`; has-more guard returns `false`).
- `{"data":null}` (no pagination) → empty page, no error (`rawValueAtPath` returns `found=false` for the missing `pagination.*` paths).
- `{"data":null,"pagination":null}` (null intermediate) → empty page, no error. Unmarshalling `null` into a `map[string]json.RawMessage` yields a nil map, so the subsequent key lookup returns `found=false` rather than erroring — addressing the null-intermediate concern.

The regression test `TestParsePage_LegacyNullItemsAreEmpty` pins exactly this: `data:null` on the default path returns nil items, empty cursor, `hasMore=false`, no error. Confirmed passing. Importantly, `TestParsePage_RejectsInvalidFieldTypes` still passes — genuinely malformed shapes (object for items, string for has-more) are still rejected. Only `null` is tolerated, which is the correct narrowing.

The second delta hunk in `runPaginated` reorders the loop so the page append + break condition (`!hasMore || nextCursor == "" || page+1 >= limit`) run before the repeated-cursor check, so a `--page-limit` boundary no longer misfires `PAGINATION_LOOP`. Covered by the new `TestDocsSearch_FinalPageStopsWithoutCursor` and `TestDocsSearch_PageLimitPreemptsLoopDetection`.

## Good parts from the prior head still hold (byte-verified at `f9b0b47`)

- Opt-ins are still scoped to `docs.search` only: it is the sole op declaring `itemsField: items`, `inferHasMore: true`, `rejectCursorRepeats: true`. All eight legacy ops still declare only `cursorField: pagination.next_cursor` + `hasMoreField: pagination.has_more` — no opt-ins leaked onto them.
- `rejectCursorRepeats` fails loud with `PAGINATION_LOOP` (not silent truncation); `inferHasMore` is explicit opt-in.
- `docs.search` is `x-octo-risk: read`; the `summary` domain remains `x-octo-disabled: true`.
- No hardcoded secrets/IDs/endpoints in the delta. `docs.search` wired (`docs.json` operationId `docs.search`) with 6 `TestDocsSearch*` cases; CHANGELOG updated.

## Non-blocking

🟡 The PR currently reports `CONFLICTING` against `main` — needs a rebase before merge. Not a review blocker, but flagging so it does not stall the merge.

yujiawei (dismissed):
# Code Review — PR #119 (octo-cli)

Third-round review at head `f9b0b47`, merge base `479ccbf`.

**Verdict: APPROVED** — both blocking issues from the previous round are fixed, verified by execution, and covered by new tests. No P0/P1 remains. One merge-readiness item and a set of P2/follow-ups below.

Verified locally at this head:

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `go test -race -shuffle=on -count=1 ./...` | all 12 packages pass |
| `golangci-lint run ./...` (v2.12.2, CI-pinned) | **0 issues** |
| `gofmt -l .` / `git diff --check` | clean |
| merged tree (`f9b0b47` + `origin/main` `de8b9ac`, CHANGELOG resolved) | builds, full suite passes |

`golangci-lint` at the CI-pinned version was unavailable to me last round; running it now closes that blind spot with a direct result rather than a relayed claim.

---

## 1. Spec compliance

Linked issue #117 asks for CLI integration of permission-scoped full-text document search, including cursor pagination and documentation.

**Spec: ✅**

- **Missing:** none. `docs search` is registered (`internal/registry/specs/docs.json:98-110`), cursor pagination works via `--cursor` / `--page-all` / `--page-limit`, and docs are updated in `README.md`, `CLAUDE.md`, and `skills/octo-docs/SKILL.md`. `CHANGELOG.md:11-16` now carries an entry — that was the one requested item still outstanding.
- **Over-build:** none. Resolved in the previous round and still true: the repeated-cursor guard is gated on `rejectCursorRepeats`, which only `docs.search` declares.
- **Deviation:** none. Operation counts stay internally consistent (`README.md:15` and `CLAUDE.md:26` at 105; `docs` 31→32 matches `internal/registry/loader_test.go`).

## 2. Code quality

**Quality: Approved**

### Both previous blocking items are genuinely fixed

**`data: null` tolerance restored** — `cmd/service/run.go:445-456`. The three-line rejection is gone, so `json.Unmarshal([]byte("null"), &items)` yields a nil slice with no error, exactly as the old struct-based decoder did.

I re-ran the end-to-end probe that failed last round, against an untouched pre-existing op:

```
# matter list --page-all; page 1 = 2 items + has_more:true, page 2 = {"data":null,...}
PROBE calls=2 err=<nil> out={"data":[{"id":"a"},{"id":"b"}], ..., "ok": true}
```

Both items from page 1 survive and the exit is clean. Previously this produced `PAGINATION_PARSE` with an empty result and every merged page discarded. `data:null` on page 1 alone also returns `{"data":[],"ok":true}`. `TestParsePage_LegacyNullItemsAreEmpty` (`cmd/service/service_test.go:470-479`) pins it, and the wrong-type cases in `TestParsePage_RejectsInvalidFieldTypes` still fail loud.

**Loop guard no longer preempts the loop's own terminators** — `cmd/service/run.go:254-262`:

```go
merged = append(merged, data...)
if !hasMore || nextCursor == "" || page+1 >= limit {
    break
}
if cursorWasSeen(seenCursors, nextCursor) {
```

The merge now happens before any exit, and the repeat check sits below the terminators, so a repeat is only an error when another request would actually follow. The added `page+1 >= limit` closes the page-limit boundary I asked about.

I checked the obvious way this reorder could go wrong — a genuine loop slipping past the guard — and it does not:

```
constant cursor "same":     calls=2  err=PAGINATION_LOOP  (detected)
alternating cursor A/B/A:   calls=3  err=PAGINATION_LOOP  (detected)
--page-limit 1 + repeat:    calls=1  err=<nil>            (boundary, correctly not a loop)
```

Every new break terminates the walk, so no bypass can result in a duplicate or unbounded request. Legacy equivalence also holds at the limit boundary — `matter list --page-all --page-limit 3` against advancing cursors still makes 3 requests and merges all 3 pages, so moving `merged = append` above the break did not drop the last page. `TestDocsSearch_PageLimitPreemptsLoopDetection` and `TestDocsSearch_FinalPageStopsWithoutCursor` pin the new behavior, and `TestPagination_PageLimitStops` / `TestPagination_LegacyStableCursorStillUsesPageLimit` continue to pin the legacy path independently.

For legacy ops the reorder is a no-op by construction: `paginationSeenCursors` (`cmd/service/run.go:288-293`) returns `nil` when `rejectCursorRepeats` is unset, and `cursorWasSeen` short-circuits on `nil`, so the guard was already unreachable for all eight pre-existing paginated operations.

### P2 — Merge readiness: the branch conflicts with `main`, and CI has not run at this head

This is the only thing I would act on before merging, and it is mechanical rather than a code defect.

`main` has advanced to `de8b9ac` (#113), which also touches `cmd/service/run.go`, `internal/registry/loader.go`, `README.md`, `CLAUDE.md`, and `CHANGELOG.md`. The PR is `CONFLICTING`; the only textual conflict is `CHANGELOG.md`, where both sides insert at the top of `[Unreleased] / Added`.

The knock-on effect is worth stating explicitly, because it is easy to misread the check list as healthy. At head `f9b0b47` only the `pull_request_target` workflows ran (Check Sprint, PR Labeler). The `pull_request`-triggered ones — **CI (build & test), OSV-Scanner, History Check** — produced no check run at all, because GitHub cannot construct the merge ref for a conflicting PR. They did run on all three earlier heads (`ca59c93`, `bb419cf`, `efe942b`). So there is currently **no CI evidence at the head being reviewed**:

```
f9b0b47   Check Sprint   failure     (board gate, non-blocking)
f9b0b47   PR Labeler     success
          CI / OSV-Scanner / History Check — never created
```

I am not blocking on this for two reasons. First, I ran the equivalent checks myself at this head *and* on the merged tree (table at the top) — all green, including `golangci-lint` at the CI-pinned v2.12.2, so the substance is verified rather than assumed. Second, the rebase is unavoidable and self-correcting: resolving the conflict creates a new head, which triggers the missing workflows and — since `main` has `dismiss_stale_reviews: true` — dismisses this approval, so a fresh round happens regardless. Please rebase and confirm CI is green on the new head before merging.

A note for that next round, since this PR is already three rounds deep: the rebase is a `CHANGELOG.md` conflict resolution, so the re-review should be a diff-only check of the rebase rather than a re-litigation of the whole change. The engineering questions here are settled.

### P2 — A genuine `PAGINATION_LOOP` still discards every page already fetched

`cmd/service/run.go:258-262`, `315-322`. Failing loud on a real loop is the right call and I am not asking to change it. But the caller currently gets a non-zero exit and an empty `data`, with no indication of how much was dropped:

```
docs search --page-all against a constant nextCursor:
  calls=2  err=PAGINATION_LOOP: backend repeated pagination cursor "same"  out=""
```

Two pages of results were fetched and thrown away. Carrying the dropped page/item count in the hint would let a caller decide whether to retry narrower. Carried over from last round; still a nice-to-have, not a blocker.

### P2 — `parsePageHasMore` accepts only a strict JSON boolean

`cmd/service/run.go` — a `hasMoreField` returning `"true"` or `1` produces `PAGINATION_PARSE` rather than coercing. I looked at whether this should be loosened and concluded it should not: strict typing is the deliberate contract here, `TestParsePage_RejectsInvalidFieldTypes` pins it, and silent coercion of a pagination terminator is exactly the kind of tolerance that turns a backend bug into a truncated result set. Recording it as a known boundary for whoever adds the next `hasMoreField` spec, not as a change request. `docs.search` uses `inferHasMore` and never reaches this path.

### P2 — Follow-ups, unchanged from last round

- **Flag naming diverges inside the `docs` group.** `docs search --page-size` / `--doc-type` (kebab) vs `docs list --pageSize` and `docs create --docType` (camel), confirmed on the binary built at this head. Kebab is the better style; the split is a papercut for scripted callers. Repo-wide rename or camelCase aliases, either as a follow-up.
- **`docs comments list` and `docs versions list` return the same `{items, nextCursor}` shape but still require manual cursor walking.** The `itemsField` + `inferHasMore` machinery this PR builds is exactly what they need, so this is now a one-line-per-op change rather than an engine change. Worth an issue.
- **Envelope pagination metadata is not generalized.** `internal/output/envelope.go` `splitPagination` remains hardcoded to `{data, pagination}`, so a single-page `docs search` gets no `_pagination` block. Documented in the spec description and `SKILL.md`, and asserted by `TestDocsSearch_FlagsAndRequestBody`, so I read it as deliberate.

### Nits

- `docs/octo-cli-search-command-tree.md:14` still reads `docs (31)` and omits `search`. It reads as a point-in-time design doc rather than living reference material, so probably fine to leave — but if it is meant to stay current it needs the same +1 the README got.
- `internal/registry/specs/docs.json:178-183` — `"type": ["integer", "null"]` on `updatedAt` is dropped by `stringOf`, so `octo-cli schema docs.search` renders `"updatedAt": {}` with no type. Pre-existing framework behavior (`summary.json` has the same shape), mentioned only because agents introspect via `schema`.

### Re-confirmed correct at this head

- Backward compatibility of field-path resolution: all eight pre-existing paginated ops declare `cursorField: pagination.next_cursor` + `hasMoreField: pagination.has_more`, no `itemsField`, no `inferHasMore`, no `rejectCursorRepeats`; `paginationFieldsFor` resolves those to exactly the old hardcoded paths.
- Cursor plumbing: `requestWithCursor` / `withCursorBody` clone per page, so the previous page's body is never mutated and `q` / `pageSize` carry forward. `cursorIsQueryParam` routes `docs.search` to the body, and the cursor never leaks into the query string.
- Items stay `json.RawMessage` end to end — key order preserved, no float64 mangling of large integer IDs.
- Terminal conditions: absent, `""`, and `null` `nextCursor` all stop the walk.
- `x-octo-risk: "read"` (`internal/registry/specs/docs.json:103`) is correct for a search POST. `--keyword` is a JSON body value marshaled via `encoding/json` — no injection surface.

## 3. Blind spots

- The live cursor semantics of `POST /v1/bot/docs/search` — specifically whether `nextCursor` is absent or echoed on the final page. The PR reports E2E verification; I could not reproduce it against a real backend. This matters because with `inferHasMore` a non-empty echoed cursor on a terminal page is indistinguishable from "there is more", and would surface as `PAGINATION_LOOP`.
- Permission scoping, OpenSearch relevance, and indexer behavior — not reachable from this repo.
- CI's own result at this head, for the reason described above; substituted with local equivalents.
- I did not audit the separate `html` domain for a competing search surface.

## 4. Summary

Nothing blocking. Rebase onto `main` (CHANGELOG conflict) and confirm CI is green on the resulting head; that push will dismiss this approval, so plan on one more short round scoped to the rebase diff. The P2 items are all fine as follow-ups.

lml2468 (dismissed):
## Review — head `f9b0b47a4bcc`

Reviewed from a local checkout at this head. `go build` / `go vet` clean, **golangci-lint v2.12.2 (CI-pinned) → 0 issues**, full `go test ./...` green in a clean env.

This head fixes the backward-compat regression yujiawei/mochashanyao flagged on the prior head — credit to them, and I under-called it myself last round (I said "full suite green, no regression" but no test exercised `data:null` until now).

### The regression, now fixed ✅
- **Was:** `parsePageItems` hard-errored on `data: null` (`field "data" must be an array, got null`). Any of the 8 existing paginated ops whose backend returns an empty result set as `data:null` would get `PAGINATION_PARSE` — a real break vs the pre-refactor tolerant decode (which treated `null` as an empty slice).
- **Fix (delta `efe942b..f9b0b47`):** removed the explicit `null`→error guard, so `json.Unmarshal("null", &items)` is a no-op (nil slice, no error) — matching pre-refactor behavior. Verified end-to-end:
  - `{"data":null,"pagination":{…}}` → empty page, no error ✅
  - intermediate `pagination:null` → empty page, no error (null → nil map → not-found, graceful) ✅ (mochashanyao's 🟡 is a non-issue in practice)
  - `TestParsePage_LegacyNullItemsAreEmpty` locks it; `TestParsePage_RejectsInvalidFieldTypes` still passes (genuinely malformed shapes still loud-fail — the tightening is kept, it just no longer misfires on null).
- `runPaginated` also moved append+break ahead of the repeated-cursor check (+`page+1>=limit`), avoiding a spurious `PAGINATION_LOOP` on the final page.

### Prior-round items still good
- Repeated-cursor guard is **opt-in** (`rejectCursorRepeats`, only docs.search) → zero effect on the 8 legacy paginated ops; a repeat now errors loudly (`PAGINATION_LOOP`) rather than silently truncating.
- `inferHasMore` explicit opt-in; strict typed parsePage; items retained as `json.RawMessage` bytes (large-int precision + key order preserved).
- docs.search: `x-octo-risk: read`, HTML type excluded; 8 legacy specs untouched (fully backward-compatible); summary domain still withheld (#113 constraint intact).

### Coverage / blind spots
- **Verified locally:** build, vet, golangci-lint (CI version), full `go test ./...` (clean env), the null-data fix and intermediate-null path, backward-compat of the 8 legacy paginated ops.
- **Merge prerequisites (not a code-side block, but resolve before merging):**
  1. PR is **`CONFLICTING` vs main — needs a rebase** before it can merge.
  2. GitHub's `build & test (1.24.x)` has **not yet reported on this head** (only label/check-sprint/code-review show); I verified build/lint/tests locally with the CI-pinned toolchain — wait for GitHub CI to go green too before merging.
- The `cmd`/`cmdutil` auth/factory tests fail only under a stray `OCTO_BOT_ID` env; clean env passes; untouched by this PR — environmental.

### Non-blocking (🟡)
- `--keyword` (required `q`) shows "(required)" in help but has no `MarkFlagRequired` — a pre-existing generator limitation (the `--data` escape hatch), not introduced here.

### Verdict: **APPROVE**
The `data:null` backward-compat regression is fixed and regression-locked; the pagination refactor's good parts (opt-in loop guard, loud loop error, strict-but-null-tolerant parsing) hold; full suite + CI-version lint green locally. Code side approved. Gate the actual merge on (1) rebase to resolve the conflict and (2) GitHub `build & test` going green on the rebased head.

lml2468 (approved):
## Re-review — new head `864a12f768df` (rebased)

The prior merge prerequisite (CONFLICTING vs main) is resolved: the branch was rebased onto current main (`de8b9ac`), and GitHub now reports **`mergeable: MERGEABLE`**. Re-verified fresh (not assuming a pure rebase, since main's own evolution of `run.go`/`loader.go` came in with it).

### Approved behaviors preserved after rebase ✅
- `parsePageItems` stays **null-tolerant** (no `data:null`→error guard) — the backward-compat fix is intact, not re-regressed.
- Repeated-cursor guard remains **opt-in** (`rejectCursorRepeats`); repeat → loud `PAGINATION_LOOP`; `inferHasMore` explicit opt-in.
- `docs.search` pagination opt-ins intact (`itemsField:items`, `cursorField:nextCursor`, `inferHasMore`, `rejectCursorRepeats`); `x-octo-risk: read`.
- summary domain still withheld (`x-octo-disabled: true`) — the #113 constraint is untouched by this rebase.

### Verified on the rebased head
- `go build` / `go vet` clean; **golangci-lint v2.12.2 (CI-pinned) → 0 issues**; full `go test ./...` green in a clean env.
- 8 legacy paginated ops still backward-compatible; `TestParsePage_LegacyNullItemsAreEmpty` + `RejectsInvalidFieldTypes` pass.

### Non-blocking (🟡)
- `--keyword` (required `q`) has no `MarkFlagRequired` — pre-existing generator limitation, not introduced here.

### Merge prerequisite now satisfied / remaining
- ✅ Conflict resolved (MERGEABLE after rebase).
- ⏳ Confirm GitHub `build & test (1.24.x)` goes green on this rebased head before merging (I verified locally with the CI-pinned toolchain).

### Verdict: **APPROVE** (re-anchored to `864a12f768df`)
Rebase cleanly resolved the conflict and preserved every approved behavior (null-tolerant parse, opt-in loop guard, loud loop error, withheld summary); build/lint/full-suite green locally. Code side approved; gate merge on GitHub CI going green on this head.

Jerry-Xin (approved):
The PR is in scope and ready to merge; it cleanly adds `docs search` through the existing metadata-driven command system and keeps shared pagination behavior backward-compatible.

💬 Non-blocking
- 🔵 Suggestion: `updatedAt` in `internal/registry/specs/docs.json:178` uses an OpenAPI union type array, but the registry loader currently reads schema type with `stringOf` at `internal/registry/loader.go:548`, so `octo-cli schema docs.search` may omit that field’s type. Existing specs already have similar nullable shapes, so this is not a blocker, but schema fidelity would improve if union types were represented deliberately.

✅ Highlights
- Pagination extensions are opt-in and preserve legacy defaults for existing `{data, pagination}` endpoints.
- Cursor loop detection is scoped to the new search contract via `rejectCursorRepeats`, avoiding behavior changes for older stable-cursor endpoints.
- Tests cover registration, body flag mapping, body cursor progression, custom response paths, cursor inference, page-limit behavior, and loop prevention.

Verification run:
- `go test ./cmd/service ./internal/registry`
- `go test ./...`

yujiawei (approved):
# Code Review — PR #119 (octo-cli)

Reviewed at head `864a12f`, diffed against merge-base `de8b9ac` (branch is up to date with `main`).

Verification performed locally at this SHA: `go build ./...`, `go vet ./...`, `go test -count=1 ./...` — all green. I also built the binary and exercised the new command's flag surface via `schema docs.search` / `--dry-run`, and ran a differential harness comparing the old and new `parsePage` implementations over a 16-case corpus of legacy response shapes (details under P2-1).

---

## 1. Spec compliance

**Spec: ✅**

Against the linked tracker (#117 — "CLI integration for permission-scoped online-document full-text search, including cursor pagination and documentation"):

- **Missing:** none. `docs search` exists (`internal/registry/specs/docs.json:96-112`), cursor pagination works via `--cursor` / `--page-all` / `--page-limit`, and docs are updated in all four places this repo expects (`README.md:132`, `CLAUDE.md:26,36`, `CHANGELOG.md:9-14`, `skills/octo-docs/SKILL.md:59-61,84-86,105`).
- **Out of scope:** none blocking. The change to the *shared* pagination engine is not scope creep — the backend returns `{total, items, nextCursor}`, which cannot be walked by the legacy `{data, pagination:{has_more,next_cursor}}` parser, so generalizing it is a precondition for the "cursor pagination" half of the issue. One small piece of speculative generality: `hasMoreField` (`internal/registry/loader.go:200`, `cmd/service/run.go:565-567`) is now configurable but no embedded spec uses a custom value — it is only reachable from a unit test. Symmetric with `itemsField`/`cursorField`, so I am not blocking on it, just noting it as a knob with no production consumer yet.
- **Divergence:** none. Operation counts are updated consistently (104→105, docs 31→32) and enforced by `internal/registry/loader_test.go:45`, so the docs can't silently drift.

One note on the edited test: `TestPagination_PageLimitStops` (`cmd/service/service_test.go:402`) was changed from a stable cursor to distinct per-page cursors. That is not a coverage loss — the original scenario is preserved verbatim by the new `TestPagination_LegacyStableCursorStillUsesPageLimit` (`cmd/service/service_test.go:418`), which pins that a legacy op with `rejectCursorRepeats` unset still walks to `--page-limit` on a repeating cursor. Good instinct to add that guard explicitly.

---

## 2. Code quality

**Quality: Approved**

No P0 or P1 findings. The design holds up under adversarial reading; a few notes below.

### What I checked and found correct

- **Loop guard vs `--page-limit` ordering** (`cmd/service/run.go:382-393`). The new `page+1 >= limit` term in the break is not just an optimization — it is load-bearing for correctness of the guard. Because the repeated-cursor check runs *after* that break, the guard can only fire when another request would actually be issued. Traced the boundary cases: `--cursor same --page-all --page-limit 1` correctly makes one request and exits clean instead of raising a spurious `PAGINATION_LOOP`; `--cursor same --page-all` (default limit) correctly raises `PAGINATION_LOOP` after exactly one request. A given cursor is never sent twice on any path.
- **No cross-file collision with the message-search router.** `routeSearchPath` (`internal/client/search_route.go:61-65`) only rewrites paths under `/v1/bot/messages/` whose suffix begins with `_search`, so `/v1/bot/docs/search` is untouched — it is not accidentally rerouted to the `/v1/user/*` mount, and it is not rejected for `app_` tokens. Worth stating explicitly because "search" in the operation name is exactly the kind of thing that trips a prefix matcher.
- **Typed-nil trap avoided.** `validatePaginationRequest` / `paginationLoopError` / `parsePage` all return `*output.ExitError` and are compared against `nil` while still concretely typed (`run.go:355`, `385`, `377`), so no `error`-interface-wrapping-a-nil-pointer bug.
- **Required-field enforcement is real.** `docs search` with no `--keyword` fails locally: `VALIDATION_ERROR … request body is missing required field(s): q`. Verified against the built binary.
- **Body is preserved across pages.** `withCursorBody` (`run.go:510-519`) clones the whole map, so `q` / `docType` / `pageSize` survive to page 2+, and the cursor stays in the body rather than leaking into the query string (asserted at `cmd/service/docs_test.go:282-284`).

### P2 — non-blocking

**P2-1 · `parsePage` silently drops the old parser's case-insensitivity.** `cmd/service/run.go:615-632`

`rawValueAtPath` does exact map-key lookup, whereas the previous `json.Unmarshal` into a tagged struct matched field names **case-insensitively**. I built a differential harness over 16 legacy response shapes (nulls, non-object roots, wrong-typed `data`/`next_cursor`/`has_more`, `"pagination": [1,2]`, extra keys) and old/new agreed on **every** case except one:

```
body = {"DATA":[{"id":1}],"Pagination":{"Has_More":true,"Next_Cursor":"C"}}
  old: items=1 cursor="C" hasMore=true
  new: items=0 cursor=""  hasMore=false
```

So the rewrite is faithful except that a backend returning non-lowercase pagination keys would now yield a *silently empty* `--page-all` result instead of paging. Every backend behind the nine existing paginated ops declares lowercase `snake_case`, so I do not think this is reachable today — but the failure mode is silent rather than loud, and the PR body's "existing contracts retain historical defaults" claim is very slightly stronger than what the code delivers. Suggest documenting the narrowing in the `rawValueAtPath` doc comment (it already documents the literal-dot limitation; case sensitivity deserves the same line).

**P2-2 · `docs search` has no way to scope results to a space.** `internal/registry/specs/docs.json:10`

`docs.json` sets `x-octo-space-header: false` document-wide, so `--space` / `OCTO_SPACE_ID` is a no-op for the new command — confirmed with `--dry-run`, which emits only the `Authorization` header. Meanwhile each result item carries `spaceId` (`docs.json:196-198`), which implies results can span spaces for a platform-scoped bot. This is consistent with the rest of the `docs` domain (including `docs list`), so it is not a regression, but for a full-text search surface it is worth confirming that cross-space fan-out is the intended contract rather than an inherited default.

**P2-3 · `updatedAt` union type is invisible to `schema docs.search`.** `internal/registry/specs/docs.json:178-183`, `internal/registry/loader.go:548`

The response declares `"type": ["integer", "null"]`, but the loader reads `type` with `stringOf`, which yields `""` for an array. Result: `octo-cli schema docs.search` renders `"updatedAt": {}` with no type at all — the one field an agent would most want typed. Sibling declarations in the same file use a scalar (`docs.json:86`, `docs.json:229`). `"type": "integer"` conveys the same thing to a consumer, since a JSON `null` is representable regardless. (There is one prior union in `summary.json:208`, so this is a pre-existing loader gap rather than something this PR introduced — but this PR is where it becomes agent-visible on a new command.)

**P2-4 · `--doc-type` enum is documentation-only.** `cmd/service/run.go:243-251`, `cmd/service/flags.go:204-207`

`validateBodySchema` checks `required` and `minItems` but never `enum`, so `--doc-type bogus` is forwarded to the backend verbatim (verified via `--dry-run`). Also, because the enum sits on `items` rather than the property, `formatEnum` does not decorate the help text — you worked around that by hand-writing "(doc, sheet, board)" into the description, which reads fine. This matches how every other enum in the CLI behaves, so it is a pre-existing convention, not a defect in this PR. Flagging only because the spec description says the `html` type is "intentionally not exposed", and a caller can in fact still pass it.

**P2-5 · Root object is re-parsed 2–3× per page.** `cmd/service/run.go:528-543`

`parsePage` calls `rawValueAtPath` once per field, and each call re-unmarshals the full page body into `map[string]json.RawMessage`. With `highlight` snippets in search results the bodies are not tiny. Bounded by `--page-limit` (default 10), so the practical cost is negligible and I would not restructure for it — but resolving the root map once and walking it for all three paths would be strictly simpler *and* cheaper.

**P2-6 · Minor inconsistency in nil-handling of `pag`.** `cmd/service/run.go:404`, `416` vs `556-558`

`paginationFieldsFor` defensively handles `pag == nil`, while `paginationControls` and `paginationSeenCursors` dereference it unguarded. Both are safe in practice (`runPaginated` is only reachable when `d.Pagination != nil`, per `cmd/service/service.go:150`), and the unguarded form matches the pre-PR code — but since the three helpers were extracted together in this PR, picking one convention would read better.

### Test-coverage gaps

- No test asserts that non-cursor body fields (`q`, `docType`, `pageSize`) actually survive to page 2+. `TestDocsSearch_PageAllUsesBodyCursorAndStopsWithoutNextCursor` and the pre-existing `TestPagination_BodyCursorEndpoint` both only inspect `body["cursor"]`. Since `--page-all` on a search endpoint dropping `q` would mean an unfiltered backend query, this is the one assertion I would most like added — a two-line change to the existing handler.
- `parsePage` is now unit-tested directly and `docs.search` is covered end-to-end, but there is still no operation-level pagination test for `matter.timeline.list`, three of the four `message.search` ops, or either marketplace op. Not this PR's debt to pay, but this PR is what makes those paths share newly-rewritten code.

---

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. This is careful work on a genuinely risky surface — the shared pagination parser feeds nine operations, and the rewrite turned out to be behaviourally identical to its predecessor on every legacy shape I could construct save one benign case-sensitivity narrowing. Opting `inferHasMore` and `rejectCursorRepeats` in per-spec rather than changing defaults is the right call, and adding `TestPagination_LegacyStableCursorStillUsesPageLimit` to pin the old behaviour before generalizing shows the blast radius was actually thought about rather than asserted.

## 4. Suggested follow-ups

1. Add the "body fields survive paging" assertion described above (highest value of anything here).
2. Change `updatedAt` to `"type": "integer"` so `schema docs.search` reports a type (P2-3).
3. Add one line to the `rawValueAtPath` doc comment noting that path lookup is case-sensitive, unlike the struct-tag matching it replaced (P2-1).
4. Confirm the cross-space search contract is intentional (P2-2).

## 5. Additional notes

- The single-page response is an object (`{total, items, nextCursor}`) while `--page-all` emits a bare array — the shape changes with the flag, which will surprise anyone writing one `--jq` expression for both. This is pre-existing `--page-all` engine behaviour rather than something introduced here, and you documented it in both the spec description and `skills/octo-docs/SKILL.md:84-86`, so no action needed; noting it only because agents are the primary consumer and `total` is unavailable in the `--page-all` path.
- I could not independently verify the backend wire contract. `POST /v1/bot/docs/search` and the field names `q` / `docType` / `pageSize` / `total` / `items` / `nextCursor` do not appear in any repository I can read from here, so their correctness rests entirely on the E2E runs described in the PR body. If any name differs, the command fails at runtime with nothing in this repo's test suite catching it. Worth a second pair of eyes from whoever owns the backend handler before release.

PR #125, opened 2026-08-10
feat(html): support canonical HTML document creation

Ranwanglc:
## Summary
- model canonical HTML creation as `idempotency_key` with no document ref
- model republish as an existing `slug`/document ref with no idempotency key
- expose backend publish fields at CLI `data.slug` and `data.doc_id`
- add `html draft create` and use `<doc-ref>` for all later operations

## Linked Spec
- Depends on Mininglamp-OSS/octo-docs-backend#166 and Mininglamp-OSS/octo-docs-html#33.
- New documents return `data.slug == data.doc_id`; legacy documents retain their old slug.
- Same-name publishing is not an identity or update mechanism.

## How verified
- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `go build ./...`
- `golangci-lint run` — 0 issues
- `make ci`
- independent cross-repository contract review: no blocking P0/P1 findings

### What changed on the load-bearing path?
The generated CLI now validates two mutually exclusive publish intents: canonical create (`html` + `idempotency_key`) and update (`html` + existing document ref). Successful HTML envelopes are unwrapped so callers persist `data.slug` directly. Canonical draft creation is exposed as a separate operation.

### What could break?
Old automation that used an unknown slug to create a document is rejected; it must use canonical create. Existing legacy slugs remain valid document refs. The CLI must be rolled out only after both server PRs support this contract.

### How was it confirmed?
Registry, generated-command, conditional-validation, response-unwrapping, path-help, draft-create, skill, full, race, lint, vet, and build tests pass.

## Required environment variables
None added. Existing `OCTO_API_BASE_URL` and bot credential configuration remain required.

Closes #126

PR #131, opened 2026-08-17
docs(html): state the no-JavaScript rule for published documents

Ranwanglc:
## Summary

Documents published through `octo-cli html publish` / `html draft` must not contain JavaScript — the octo-doc backend rejects them with `400 html_contains_javascript`. The skill never mentioned this, so an agent generating a document had no reason not to reach for `<script>` or an `onclick` handler, and only discovered the contract from a failed publish.

Worth stating even independently of the gate: these documents render inside an iframe with `sandbox="allow-same-origin"` and **no** `allow-scripts`. Script never ran. A generated page that "needs JS" was silently inert for the reader, which is a worse failure than a 400 — nobody notices.

So this documents the rule at the point of use rather than leaving it to be learned by rejection.

## Related Issue

None — follows the corresponding server-side change in `octo-docs-html`.

## Changes

`skills/octo-html/SKILL.md`, bumped `0.1.0` → `0.2.0`:

- **§1 Document lifecycle** — a no-JavaScript rule ahead of the `publish` examples: the exact constructs the backend refuses (`<script>` including inside `<svg>`, `on*` attributes, `javascript:`/`vbscript:`/scriptable `data:` URLs in the usual sinks, non-empty `srcdoc`), the CSS equivalents that do work (`:hover`, `:target`, `:checked` + sibling selectors, `<details>`/`<summary>`, transitions), and what remains allowed (CSS, inline `style`, `<iframe>`, `<noscript>`, `<meta http-equiv=refresh>`) so the rule is not read as broader than it is.
- **Errors** — `400 html_contains_javascript` documented alongside the existing cases, including that `details.violations` lists every offending construct with `kind`/`tag`/`attr`/`line`, so a document is fixed in one pass instead of one violation per round-trip.
- Both places tell a generating agent to regenerate without script rather than retry unchanged or try to encode around the gate.

Documentation only — no command, flag, or output change.

## Testing

```
go build ./...          # ok
go test ./... -count=1  # all packages ok (includes skills registry tests)
gofmt -l .              # clean
```

## Checklist

- [x] Conventional Commit
- [x] Skill version bumped
- [x] Documentation matches the server behaviour it describes
- [x] No behaviour change to the CLI itself

### DIFF
--- a/skills/octo-html/SKILL.md
+++ b/skills/octo-html/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: octo-html
-version: 0.1.0
+version: 0.2.0
 description: HTML docs domain (octo-doc) — create and govern self-contained interactive HTML documents, immutable versions, drafts, sharing, media, comments, and agent element edits. This is a DIFFERENT backend from the `octo-docs` (CRDT/Yjs) domain. Load after octo-shared.
 metadata:
   requires:
@@ -52,6 +52,30 @@ deployed before this CLI is released.
   `{ok:false,error:{type,code,message,hint,detail}}` envelope.
 
 ## 1. Create and publish
+
+**Documents are declarative: no JavaScript.** The backend rejects any publish or
+draft whose HTML carries script, with `400` and the stable code
+`html_contains_javascript`. This is not advisory — there is no flag to opt out.
+When generating a document, never emit:
+
+- `<script>` elements (including inside `<svg>`),
+- `on*` event-handler attributes (`onclick`, `onload`, `onerror`, …),
+- `javascript:` / `vbscript:` / scriptable `data:` URLs in `href`, `src`,
+  `xlink:href`, `action`, `formaction`, `object[data]`, or a non-empty `srcdoc`.
+
+Express interaction with CSS instead — `:hover`, `:target`, `:checked` +
+sibling selectors, `<details>`/`<summary>`, transitions and animations all work
+and cover most of what a document needs. Do not assume script is merely inert:
+the host application embeds documents in a sandboxed iframe, but the document's
+own version URL serves the stored HTML as a top-level page, so script that got
+stored would execute for every reader who opens that link. CSS, inline `style`,
+`<iframe>`, `<noscript>` and `<meta http-equiv=refresh>` are all still allowed.
+
+On rejection the error `details.violations` lists every offending construct with
+its `kind`, `tag`, `attr` and 1-based `line`. Fix all of them in one pass — the
+list is complete (capped at 50, with `details.truncated` set when it overflows).
+Do not retry the same document unchanged, and do not try to work around the gate
+by encoding or splitting the script.
 
 ```bash
 # Canonical create: no --slug. The CLI generates the idempotency key.
@@ -169,6 +193,10 @@ an element's tag or nearby heading unless necessary.
 - `401 / 403` — missing or insufficient capability.
 - `404` — document reference (canonical doc_id or legacy slug), comment, or aid
   not found.
+- `400 html_contains_javascript` — the published or draft HTML carries
+  JavaScript (see the no-JavaScript rule in §1). `details.violations` names every
+  offending construct with its `kind` / `tag` / `attr` / `line`; regenerate the
+  document without script rather than retrying it unchanged.
 - Element replacement rejects multiple top-level elements, scripts/styles,
   inline event handlers, and `javascript:` URLs.
 


Jerry-Xin (dismissed):
This relevant, documentation-only change accurately clarifies the backend’s no-JavaScript contract.

✅ Highlights

- Clear prohibited/allowed constructs at [skills/octo-html/SKILL.md:39](skills/octo-html/SKILL.md:39).
- Actionable error guidance at [skills/octo-html/SKILL.md:197](skills/octo-html/SKILL.md:197).
- Appropriate version bump.
- `git diff --check` and `go test ./... -count=1` pass.
- No blocking or non-blocking issues found.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #131 (octo-cli)

## Summary

Docs-only change to `skills/octo-html/SKILL.md` (+29/−1, version 0.1.0 → 0.2.0) that states the no-JavaScript rule for octo-doc documents at the point of use: the exact constructs the octo-docs-html backend rejects with `400 html_contains_javascript` (`<script>` incl. inside `<svg>`, `on*` handlers, `javascript:`/`vbscript:`/scriptable `data:` URLs in the usual sinks, non-empty `srcdoc`), CSS-based interaction equivalents, what remains allowed, and an Errors-section entry describing `details.violations`. No command, flag, or CLI behaviour change. Review finds no P0/P1: the doc is internally consistent and nothing in any public repo contradicts the documented gate; remaining items are one P2 wording/verifiability point and two nits.

## Verification

Static analysis only at head `256de5e23b7e8611e1c01d17b653b0eea1535c6d` (merge-base `e627723`); build and tests not executed in this environment. Cross-repo context pulled read-only from public clones of `octo-server`, `octo-web`, and `octo-cli`.

- ✅ **Diff scope** — exactly one file changed (`skills/octo-html/SKILL.md`), three hunks (frontmatter version line 3; §1 rule block lines 40–63; Errors entry lines 200–203).
- ✅ **Self-consistency (C6)** — new §1 text agrees with the existing Errors section, the `{data}/{error}` envelope prose, and the §6 element-replace fragment rule; the Errors entry's "see §1" cross-reference resolves.
- ✅ **No CLI-side contradiction** — the CLI is spec-driven (`internal/registry/specs/html.json`) and surfaces backend envelopes unchanged; there is no client-side HTML validation and no opt-out/bypass flag anywhere in the spec or command tree, matching the doc's "no flag to opt out".
- ✅ **Version bump** — 0.1.0 → 0.2.0 in frontmatter, consistent with a doc-contract change; nothing in-repo pins the old version.
- ⚠️ **Backend contract not independently verifiable** — the gate's source of truth is the `octo-docs-html` service, which is not public (no `html_contains_javascript` in `octo-server` main or any of 78 remote branches; no public `octo-doc`/`octo-docs-html` repo). See Human-verify.

## Findings

No P0/P1 issues; one P2 and two nits below, all non-blocking.

### P2 — Rendering claim is unverifiable and reads flatter than the evidence (`skills/octo-html/SKILL.md:52`)

"Script would not run anyway: documents render inside an iframe with `sandbox="allow-same-origin"` and no `allow-scripts`" cannot be verified from any public code. The only public embedding surface, `octo-web` `DocTab.tsx:77` (OCT-138 Stage A skeleton, not yet wired to a production panel), sets `sandbox="allow-scripts allow-same-origin allow-forms allow-popups"` with a vitest assertion locking `allow-scripts` in, and its design comment says the doc page's JS is needed to fetch the capability cookie same-origin. That evidence is consistent with a two-layer design (service viewer shell runs JS; authored content sits in an inner script-less iframe), but as written the sentence states the outer embedding has no scripts. Diff-scope: new text introduced by this PR. Severity: P2, not P1 — even if the claim were wrong, the publish/draft gate is the enforcement point and agent behaviour does not change harmfully. Fix: qualify the sentence to name the octo-docs-html viewer's content iframe, or rest the justification on the gate alone.

### Nit — "complete list" vs "capped at 50" self-contradiction (`skills/octo-html/SKILL.md:59`)

"Fix all of them in one pass — the list is complete (capped at 50, with `details.truncated` set when it overflows)": when `truncated` is set the list is by definition not complete, so one pass cannot fix everything. Suggested wording: "complete up to 50 items; when `details.truncated` is set, fix the reported batch and re-publish to surface the remainder."

### Nit — allow-list scope vs element-replace fragment rule (`skills/octo-html/SKILL.md:56`)

"CSS, inline `style`, `<iframe>`, `<noscript>` and `<meta http-equiv=refresh>` are all still allowed" is scoped to publish/draft by its §1 context, but §6 element-replace fragments additionally reject `<style>`. An agent editing via element replace could misread the allow-list. Add a parenthetical scoping the sentence to publish/draft and pointing at §6's stricter fragment rule.

## Data-flow tracing (claim → upstream source → does it hold?)

- `400 html_contains_javascript` on publish/draft → upstream: octo-docs-html backend (private). Not defined in `octo-cli` spec `html.json` (documents 200 envelopes only); absent from `octo-server` main + all 78 remote branches. PR body states it "follows the corresponding server-side change in octo-docs-html". No counter-evidence found; independently unverifiable → Human-verify item 1.
- `details.violations` shape (`kind`/`tag`/`attr`/1-based `line`, cap 50, `truncated`) → same private upstream; unverifiable → Human-verify item 1.
- iframe `sandbox="allow-same-origin"` without `allow-scripts` → upstream should be the octo-docs-html viewer renderer (private). Public check: `octo-web` `DocTab.tsx:77` outer iframe carries `allow-scripts` (test-locked); consistent with two-layer design, does not directly corroborate the flat claim → P2 above.
- allow-list (`<iframe>`, `<noscript>`, `<meta http-equiv=refresh>`, inline `style`) → private sanitizer; nothing in public repos contradicts; the listed URL sinks (`src` covers iframe `src`; non-empty `srcdoc` banned) are internally coherent.
- "no flag to opt out" → verified on the CLI side: no such flag in `html.json` or the command tree.

## Blind-spot checklist (C1–C6)

- C1 dual-path parity: clear — doc covers both gate ends it claims (publish and draft); element-replace is a separately documented gate (§6, pre-existing, untouched).
- C2 control-flow ordering / nested reuse: clear — docs-only; the new rule text is cross-referenced once (Errors section) and the reference resolves.
- C3 authorization boundary: N/A — no capability/permission surface changed; the text explicitly closes an opt-out rather than opening one.
- C4 authorization lifecycle / cascade: N/A — no auth-state text changed.
- C5 build pass ≠ runtime path: hit, downgraded — PR reports `go build`/`go test` clean (plausible; not re-run here), but for a docs PR the "runtime" is backend behaviour, which is private; not treated as mitigated by the author's checklist — see Human-verify. No counter-evidence found, so not a P1.
- C6 governance/doc self-consistency: hit → produced the two nits above (truncated wording, allow-list scope); no conflict with other skill sections or repo docs.

## Cross-round blocker re-check (R6)

N/A — first review round for this PR on this issue (no prior comments/blockers).

## Human-verify

1. All backend-specific claims — exact error code string, `details.violations` field names and 1-based `line`, the 50-item cap and `truncated` flag, the allow-list, "no opt-out flag", and the content-iframe sandbox — live in the private `octo-docs-html` service and could not be independently verified from public code. The author's checklist asserts the doc matches server behaviour. **Not a merge blocker for this PR**; worth a backend-side confirmation by the final reviewer.
2. If `draft promote` does not re-run the gate on promotion, a draft saved before the gate existed could promote JS-carrying HTML; confirm the gate covers promote as well as save. **Not a blocker for this docs PR.**

## Verdict: APPROVED

Docs-only change that is internally consistent, correctly version-bumped, and contradicted by nothing in the public repos; the gate's enforcement claims are attributable to the private octo-docs-html backend and flagged for backend-side confirmation rather than blocking. The P2 (qualify the iframe rendering claim) and the two nits are non-blocking suggestions.

[Octo-Q] verdict: APPROVE — 仅 P2/nit（iframe 渲染表述需限定措辞 + 两处措辞 nit），无 P0/P1；后端契约在私有 octo-docs-html 仓，公开代码无反证，建议终审侧向后端确认。

yujiawei (dismissed):
# Code Review — PR #131 (octo-cli)

Reviewed at head SHA `256de5e2`, merge-base `e6277239`. Docs-only diff: `skills/octo-html/SKILL.md` (+29/−1), skill version `0.1.0` → `0.2.0`.

**Verdict: APPROVE.** No P0/P1. The change closes a real agent footgun — a hard backend gate that was previously discoverable only by a failed publish — and the actionable core of the new text (don't emit script, fix every violation in one pass, don't retry unchanged, don't try to smuggle script past the gate) is correct. Five non-blocking nits below; none of them need to hold up the merge.

## 1. Scope check

- The diff does exactly what the title and description say: it documents an existing server-side contract. No command, flag, spec, or output change is bundled in — confirmed, the diff touches one markdown file and nothing under `cmd/`, `internal/`, or `internal/registry/specs/`.
- Skill frontmatter version bumped as claimed (`SKILL.md:3`). Minor-level bump is right for additive content.
- Verified locally at head: `go build ./...` clean, `gofmt -l .` clean, `go test ./skills/... -count=1` ok. (I did not run the full `./...` suite; nothing in this diff is reachable from Go code beyond the `//go:embed */*.md` glob in `skills/skills.go:13`, which is glob-based and needs no registration change.)
- Nothing over-built, nothing obviously missing relative to the stated goal.

## 2. Findings

### P2 — the `details.violations` path is one level up from where a CLI caller will find it

`skills/octo-html/SKILL.md:58` and `:201` tell the agent that on rejection "the error `details.violations` lists every offending construct". That is literally true of the *backend* error object, but the CLI does not hand the backend body through at that shape:

- `internal/output/errors.go:266-278` — `newBackendError` stores the **whole** response body in `ExitError.Detail`.
- `internal/output/envelope.go:116-117` — `WriteError` emits it as `error.detail`.

So for a matters-shaped backend envelope the reachable path in the CLI's stderr output is `.error.detail.error.details.violations`, not `.details.violations`. I confirmed this empirically with a throwaway test against `ParseBackendError` + `WriteError` (deleted afterwards):

```json
{"ok":false,"error":{"code":"html_contains_javascript","detail":{"error":{"code":"html_contains_javascript","details":{"violations":[{"kind":"element","tag":"script","line":12}],"truncated":false}}},...}}
```

Why this is only a nit and not a blocker: `skills/octo-shared/SKILL.md:116` already documents the composition ("matters (structured): `{error:{code, message, details}}` → passes through into `detail` unchanged"), and `octo-html` declares `requires: skills: ["octo-shared"]`, so a reader that followed the load order can derive the nesting. The whole envelope is also pretty-printed to stderr, so the list is visible regardless of depth.

Suggestion: spell the path once, the way `skills/octo-matter/SKILL.md:134` does for its own error case. Note that the matter line is itself off by one (`error.detail.details` vs. the actual `error.detail.error.details`), so if you touch this, fixing both together is the cheaper move.

Related, pre-existing and out of this diff's scope: `SKILL.md:197` says "The CLI surfaces the backend envelope unchanged", which is what makes a bare `details.` two lines later read as a top-level path. The sentence was already there; worth a separate pass.

### P2 — "the list is complete (capped at 50 …)" says both things at once

`SKILL.md:59-60`: "Fix all of them in one pass — the list is complete (capped at 50, with `details.truncated` set when it overflows)."

An agent optimising for one round-trip will take "the list is complete" as the operative clause and skip the `truncated` check, which is exactly the case where a second pass is required. Suggested rewrite: "Fix all of them in one pass. The list covers every violation up to 50; if `details.truncated` is set there are more, so expect another round after the first fix."

### P2 — the `srcdoc` bullet is grammatically attached to the URL-scheme rule

`SKILL.md:47-48`:

```
- `javascript:` / `vbscript:` / scriptable `data:` URLs in `href`, `src`,
  `xlink:href`, `action`, `formaction`, `object[data]`, or a non-empty `srcdoc`.
```

Read strictly, this says a `srcdoc` is refused when it *contains* one of those URL schemes. The PR description says something stronger — any non-empty `srcdoc` is refused, regardless of content. If the stronger rule is what the backend implements, `srcdoc` should be its own bullet; as written, an agent could reasonably conclude that a script-free `srcdoc` is fine. This matters more than usual because `:55` lists `<iframe>` as allowed, so `<iframe srcdoc="…">` is a natural thing to reach for.

### P2 — `<meta http-equiv=refresh>` is listed as allowed, but probably won't fire in the described viewer

`SKILL.md:52-56` establishes that documents render in an iframe with `sandbox="allow-same-origin"` and no `allow-scripts`, then lists `<meta http-equiv=refresh>` among constructs that "are all still allowed".

Per the HTML Standard's *parse a sandboxing directive*, omitting `allow-scripts` also sets the **sandboxed automatic features browsing context flag** (verified in the spec text), and declarative refresh is gated on that flag in the browsers I'm aware of — I have not driven the actual viewer to confirm, so treat this as strong prior rather than settled fact. If it holds, meta refresh is accepted by the gate but silently inert at render time, which is precisely the failure mode `:52-54` warns about. A short parenthetical ("accepted by the gate; the sandbox may still suppress it") keeps the allow-list honest. `<noscript>` and nested `<iframe>` in the same sentence are fine — `<noscript>` genuinely renders when scripting is disabled, and a nested frame inherits the sandbox flags.

### P2 — the same rule is absent from the schema surface the skill points agents at

`internal/registry/specs/html.json:534` already states the constraint inline for `html.element.replace` ("must be exactly one top-level element and free of scripts / inline event handlers / `javascript:` URLs"), but the `html.publish` (`:43`) and `html.draft.save` (`:172`) descriptions say nothing about it, and neither operation declares a `400` response. Since this skill's own "Schema lookup" section sends agents to `octo-cli schema html.publish`, an agent that introspects instead of reading the skill still learns nothing. Mirroring one sentence into those two descriptions would make the two agent-facing surfaces agree. Follow-up, not part of this PR.

## 3. Coverage — what I could not verify

The backend that enforces this gate is not in this repository: `html_contains_javascript`, `violations`, and `truncated` appear nowhere in the tree outside the added text, and `internal/registry/specs/html.json` declares no error responses for the HTML operations. So the exact code string, the `kind`/`tag`/`attr`/`line` shape, the cap of 50, the `truncated` flag, the precise sink list, and the sandbox attribute of the viewer are all taken on the author's word — they are consistent and specific, but this review cannot confirm them against source. If the server-side change is available, linking it in the description would let the next reader check the doc against the implementation.

Two coverage gaps in the text itself, both minor: the rule is stated for `publish` and `draft` but not for `element replace` (covered separately at `:204-205`, though it is unclear whether that path returns the same code), and `asset add` is not mentioned at all — worth a sentence if an uploaded SVG asset is scanned too.

yujiawei:
Follow-up to my review above — one point I flagged as "unverifiable" turns out to have public counter-evidence, so I'm upgrading it from a coverage note to a concrete suggestion. Verdict is unchanged (still approve; this does not block).

**`skills/octo-html/SKILL.md:52-54`** — "Script would not run anyway: documents render inside an iframe with `sandbox="allow-same-origin"` and no `allow-scripts`".

The only public embedding surface for these documents contradicts that as written. In `octo-web`, `packages/dmworkbase/src/Components/DocTab/DocTab.tsx` is explicitly the octo-docs-html document tab (its `src` is the doc service render URL, e.g. `/d/<slug>/v/1`), and line 77 sets:

```
sandbox="allow-scripts allow-same-origin allow-forms allow-popups"
```

with a test asserting `allow-scripts` stays in (`__tests__/DocTab.test.tsx:59`). The component's own design note gives the reason: the doc page's JavaScript performs a same-origin fetch to obtain the HttpOnly capability cookie. So the page served at `/d/<slug>/v/<n>` does run script.

That is compatible with a two-layer design — viewer shell runs its own JS, authored content sits in an inner script-less frame — but the sentence as written describes the outer embedding, and it is the second of the two justifications the section offers ("script would not run anyway"). If the inner content frame is the one without `allow-scripts`, naming it would make the claim checkable; if it isn't, the sentence should go and the paragraph should rest on the publish/draft gate alone, which is the actual enforcement point and stands on its own.

Two knock-on notes:

- My earlier nit about `<meta http-equiv=refresh>` was reasoned *from* this sandbox sentence (no `allow-scripts` also sets the sandboxed automatic features flag). If the sandbox description changes, re-check that nit against the real flags rather than carrying it over.
- Separately, the allow-list at `:55-56` ("CSS, inline `style`, `<iframe>`, `<noscript>` and `<meta http-equiv=refresh>` are all still allowed") is scoped to publish/draft by its §1 context, but §6 `element replace` additionally rejects `<style>` in a fragment. A four-word parenthetical scoping the sentence to publish/draft would stop an agent editing via `element replace` from generalising it.

Jerry-Xin (approved):
Re-review after `main` was merged into the branch (head 35ebe38).

**Delta isolation.** The new head adds only the merge commit `35ebe38` ("Merge branch 'main' into docs/html-no-javascript"), which pulls in df54c37 `fix(mail): align Agent commands with owner authorization (#133)` — already reviewed and merged to main separately via #133, not owned by this PR. Byte-verified: `skills/octo-html/SKILL.md` is byte-identical between the previously approved head and this one (empty diff for that path). This PR's own surface against the merge base remains documentation-only: the version bump, the §1 no-JavaScript rule, and the `400 html_contains_javascript` Errors entry.

**Verdict: APPROVE — no blocking findings.** The PR-owned content is unchanged from what was reviewed and approved at the previous head, so the prior analysis stands:

- The §1 framing is correct and is exactly what justifies the hard `400` gate: the document's own version URL serves the stored HTML as a top-level page, so script that got stored would execute for every reader who opens that link (`skills/octo-html/SKILL.md:66-70`).
- The prohibited/allowed construct lists and the Errors entry cross-reference each other correctly; the remediation guidance (fix all violations in one pass; do not retry unchanged; do not encode around the gate) is the right agent behaviour.
- Docs-only: no command, flag, or output change in this repo; the gate itself lives in the octo-docs-html backend, and nothing in this repo contradicts the stated behaviour.

**Disposition of outstanding items — all carried as non-blocking.** This push changed no doc text, so every note from the previous round remains open:

1. 🔵 (Jerry-Xin, prior round) `details.violations` location: the CLI embeds the whole backend body under `error.detail` (`internal/output/errors.go` ParseBackendError keeps the raw body as Detail; `internal/output/envelope.go` WriteError emits it), so readers should resolve `details.violations` inside the `error.detail` payload rather than expecting a top-level `details` field. The skill already documents the `{ok:false,error:{type,code,message,hint,detail}}` envelope at `skills/octo-html/SKILL.md:52`; a one-line note or small JSON example tying `details.violations` to `error.detail` (lines 74, 76, 197) would remove the hunt.
2. 🔵 (Jerry-Xin, prior round) The PR description still calls script "silently inert" while the final doc correctly states stored script executes via the version URL. Worth aligning the description; cosmetic.
3. 🟡 (mochashanyao) Error-path precision — same root as item 1; carried unchanged.
4. 🟡 (mochashanyao) CSS guidance vs element-replacement gating — §1 allows CSS for publish while the Errors entry says element replacement refuses scripts/styles. These are different operations with different gates, but stating explicitly that `<style>` is allowed at publish while element replacement rejects it would remove the ambiguity. Carried unchanged.
5. 🟡 (yujiwei) CLI-accessible error-path clarity — same root as item 1; carried.
6. 🟡 (yujiwei) Doc wording precision — carried unchanged.

None of these change agent behaviour harmfully: the violations list ships verbatim inside the CLI's error envelope and is fully recoverable, and the fix-in-one-pass guidance is correct regardless of the exact path.

✅ Highlights (unchanged from the prior round)
- Precise enumeration of prohibited constructs and the CSS equivalents that do work.
- Actionable remediation: fix all violations in one pass, do not retry unchanged or encode around the gate.
- Appropriate version bump; documentation-only surface.

yujiawei (approved):
# Code Review — PR #131 (octo-cli)

Reviewed at head SHA `461d5969`, merge-base `0134bcb` (#125). Diff: one file,
`skills/octo-html/SKILL.md`, +29/−1, skill frontmatter `0.1.0` → `0.2.0`.

This is round 2. The three reviews on `256de5e2` were dismissed when the branch
was rebased onto #125, so nothing from that round carries forward automatically
and this verdict is issued against the current head only.

**What actually changed since `256de5e2`:** the rebase absorbed #125's rewrite of
the surrounding skill, and one sentence of this PR's own added text was reversed
— from *"Script would not run anyway: documents render inside an iframe with
`sandbox="allow-same-origin"` and no `allow-scripts`, so a page that needs JS is
silently inert"* to *"Do not assume script is merely inert … the document's own
version URL serves the stored HTML as a top-level page, so script that got stored
would execute for every reader who opens that link"* (`skills/octo-html/SKILL.md:68-72`).
That is a real correction, not a rewording, and it changes what the rule rests
on. Two wording nits raised in round 1 were not addressed; they are repeated
below since the dismissal cleared them.

---

## 1. Scope compliance

**Spec: ✅**

- **Nothing missing.** The PR promises to state the no-JavaScript rule at the
  point of use and to document the error; both landed — the §1 block
  (`skills/octo-html/SKILL.md:56-78`) and the Errors entry (`:196-199`).
- **Nothing over-built.** "Documentation only — no command, flag, or output
  change" holds. The diff touches no file under `cmd/`, `internal/`, or
  `internal/registry/specs/`.
- **No deviation.** The version bump is real (`:3`) and minor-level is right for
  additive content. Note that nothing in the repo consumes it: `loadSkills`
  parses only `description:` and `disabled:` from frontmatter
  (`cmd/skills.go:53-80`), so the field is convention, not a pinned contract —
  which also means the bump cannot break anything.

Verified locally at head: `go build ./...` clean, `gofmt -l .` clean,
`go test ./... -count=1` green across all 12 packages with tests — the PR's
testing claims reproduce. The "no flag to opt out" claim checks out on the CLI side — the
`html.publish` request body in `internal/registry/specs/html.json` declares only
`slug`, `html`, `idempotency_key`, `version`, `meta`, `mount_type`, `group_no`,
`thread_id`, and there is no client-side HTML validation anywhere on this path.

---

## 2. Code quality

**Quality: Approved** — no P0/P1. Five P2s, all non-blocking.

### P2 — `details.violations` is not the path a CLI caller can reach

`skills/octo-html/SKILL.md:74` and `:197` tell the agent that on rejection
"`details.violations` lists every offending construct". That is true of the
*backend* object, but the CLI does not surface it at that depth. A
matters-shaped backend envelope hits Layer 1 of `parseBackendError`, which passes
the **whole response body** as the detail payload:

```go
// internal/output/errors.go:191-198
if len(body) > 0 && json.Unmarshal(body, &mEnv) == nil && mEnv.Error.Code != "" {
    ...
    return newBackendError(mEnv.Error.Code, mEnv.Error.Message, fallbackType, mEnv.Error.Hint, body)
}
```

and `WriteError` emits it verbatim under `detail`
(`internal/output/envelope.go:117`: `errObj["detail"] = ee.Detail`). So the
reachable path on stderr is `.error.detail.error.details.violations`, not
`details.violations`. An agent that writes
`-q '.error.details.violations'` gets nothing and concludes the list is absent.

Not a blocker: `skills/octo-shared/SKILL.md` already documents that a structured
backend envelope passes through into `detail` unchanged, `octo-html` declares
`requires: skills: ["octo-shared"]`, and the whole envelope is pretty-printed
anyway. But spelling the path once here would remove the inference step for the
exact reader this paragraph is written for.

### P2 — "the list is complete (capped at 50 …)" asserts both halves of a contradiction

`skills/octo-html/SKILL.md:75-76`:

> Fix all of them in one pass — the list is complete (capped at 50, with
> `details.truncated` set when it overflows).

When `truncated` is set the list is by definition *not* complete, and one pass
cannot fix everything. The sentence's headline clause ("the list is complete")
is the one an agent optimising for a single round-trip will act on, which is
precisely the case where a second pass is required. Suggested wording: *"The
list covers every violation up to 50; if `details.truncated` is set there are
more, so expect another round after the first fix."*

Raised in round 1 by two reviewers; unchanged at this head.

### P2 — the `srcdoc` clause is grammatically attached to the URL-scheme rule

`skills/octo-html/SKILL.md:63-64`:

```
- `javascript:` / `vbscript:` / scriptable `data:` URLs in `href`, `src`,
  `xlink:href`, `action`, `formaction`, `object[data]`, or a non-empty `srcdoc`.
```

Read strictly, this says a `srcdoc` is refused when it *contains* one of those
schemes. The PR description states something stronger — any non-empty `srcdoc`
is refused regardless of content. If the stronger rule is what the backend
implements, `srcdoc` belongs in its own bullet; as written, an agent can
reasonably conclude a script-free `srcdoc` is fine. This is load-bearing because
`:72` lists `<iframe>` as allowed, so `<iframe srcdoc="…">` is the natural thing
to reach for. Also raised in round 1; unchanged.

### P2 — the reversed rationale is now a security claim, and the PR description and commit message still assert its opposite

The shipped text (`:68-71`) says script *would* execute for readers who open the
version URL. The PR body ("these documents render inside an iframe with
`sandbox="allow-same-origin"` and **no** `allow-scripts`. Script never ran.") and
the commit message body ("so script would not have run even before the gate
existed -- a page written that way is silently inert") still say the opposite,
and the commit message is what lands in `main`'s history permanently. Please
amend both to match the text you shipped.

Two consequences worth a sentence in the doc, if the new claim is the accurate
one:

- The gate stops being "belt and braces over an inert surface" and becomes the
  only thing between a generated document and script running same-origin for
  every reader of a share link. That is worth saying plainly — it is a stronger
  reason to obey the rule than "it wouldn't work anyway".
- Under that model the allow-list on the same line deserves a caveat.
  `<meta http-equiv=refresh>` on a top-level-served page is a live navigation
  primitive aimed at every reader, and `<iframe>` embeds a third-party document
  that runs its own script in its own origin. Both were harmless under the old
  "sandboxed, inert" framing; neither is under the new one. Listing them flatly
  as "still allowed" reads as an endorsement.

This is the one place I would most like a citation. The file already sets the
precedent two sections up (`:41-43` names octo-docs-backend#166 and
octo-docs-html#33 as the rollout dependency); linking the server PR that
introduced this gate would let the next reader check the doc against the
implementation instead of against the author's recollection.

### P2 — the same rule is missing from the schema surface this skill points agents at

`internal/registry/specs/html.json:1058` already states the constraint inline for
`html.element.replace` ("must be exactly one top-level element and free of
scripts / inline event handlers / `javascript:` URLs"), but the `html.publish`
and `html.draft.save` descriptions say nothing about it, and neither operation
declares a `400` response at all (the spec documents only `200` for these
paths). This skill's own "Schema lookup" section (`:203-210`) sends agents to
`octo-cli schema html.publish`, so an agent that introspects rather than reads
the skill still learns nothing about the hardest constraint on the endpoint.
Mirroring one sentence into those two descriptions would make the two
agent-facing surfaces agree. Follow-up, not part of this PR.

---

## 3. Merge readiness (not a review finding)

`check-sprint` is failing on this PR: *"No linked issue found. Please add a
'Closes #<issue>' reference to the PR description."* The PR body says
"## Related Issue — None". That check is independent of this review and will
keep the PR from merging until a linked issue on the board is added.

`build & test (1.24.x)`, `npm test`, `osv-scan` and the history/label checks all
pass.

---

## 4. Coverage — what this review could not verify

The service that enforces the gate is not in this repository. `html_contains_javascript`,
`violations` and `truncated` appear nowhere in the tree outside the added text,
and the HTML operations in `internal/registry/specs/html.json` declare no error
responses. A search for the enforcing service outside this repo did not turn it
up either. So the exact code string, the `kind` / `tag` / `attr` / 1-based `line`
shape, the 50-item cap, the `truncated` flag, the precise sink list, and the
top-level-serving claim are all taken on the author's word. They are internally
consistent and nothing in this repo contradicts them — but this review confirms
consistency, not correctness, for every backend-side claim.

Three gaps in the text itself, each minor and each a question rather than a
finding:

- The rule is stated for `publish` and `draft`. Does `draft promote` re-run it?
  A draft saved before the gate existed would otherwise promote script-carrying
  HTML through a path the doc says is covered.
- `element replace` is documented separately (`:200-201`) and it is not clear
  whether that path returns the same `html_contains_javascript` code or a
  different one. If it is the same, saying so would save a round-trip.
- `asset add` is not mentioned. If an uploaded SVG asset is scanned too, that
  deserves a sentence — SVG is the one asset type where "media upload" and
  "script" overlap.

---

## 5. Verdict

**APPROVE.** The change closes a real footgun: a hard backend gate that was
previously discoverable only by a failed publish, in a skill whose entire
audience is agents generating HTML. The actionable core — don't emit script, fix
every violation in one pass, don't retry unchanged, don't try to smuggle script
past the gate — is correct and well placed. All five findings are P2; none of
them needs to hold up the merge. I would take the commit-message/PR-body fix and
the `truncated` rewording in a follow-up commit on this branch if it is cheap,
and leave the spec-description mirroring to a separate PR.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #131 (octo-cli)

## Summary

Docs-only change to `skills/octo-html/SKILL.md` (+29/−1, version bump 0.1.0 → 0.2.0). It states the backend's no-JavaScript gate for published/draft HTML documents: the stable `400 html_contains_javascript` code, the banned-construct list, CSS-only interaction guidance, and the `details.violations` rejection report. The prose matches everything verifiable inside this checkout: the `html_contains_javascript` code and the backend's `error.details` field pass through the CLI's error normalization intact (`internal/output/errors.go:182`), and the pinned strings in `skills/skills_test.go:76` are untouched (the only deletion is the version line).

## Verification

Static analysis only at head `461d65969376183c4f038e88246d86766287f0ed`; build and tests not executed in this environment.

- ✅ Diff scope verified: one file, three hunks (frontmatter version, new §1 no-JS section, new Errors bullet); merge-base `0134bcb`.
- ✅ Error-code path traced: backend body → `parseBackendError` layer 1 (`internal/output/errors.go:182`) → `newBackendError` stores the whole body as `Detail` (`internal/output/errors.go:289`) → serialized as `error.detail` (`internal/output/envelope.go:116`). `html_contains_javascript` is unmapped in `backendErrorMapping`, so the literal code reaches the agent unchanged.
- ✅ Test contract: all canonical-create strings pinned by `skills/skills_test.go:76` survive; docs-only commits carry no CHANGELOG entry by precedent (`0eb40d1`, `45b8de3`).
- The gate itself lives in the octo-doc backend, outside this checkout; its rules are taken as the documented contract.

## Findings

No P0/P1 issues. Two non-blocking P2 doc-contract items.

### P2 — `details.violations` is not at that path in the CLI envelope (`skills/octo-html/SKILL.md:74`)

The new text says the rejection carries `details.violations`, but the CLI envelope — documented in this same skill's Auth & space section and emitted by the renderer — is `{ok:false,error:{type,code,message,hint,detail}}`. The backend payload containing `error.details.violations` is embedded verbatim under `error.detail` (`internal/output/errors.go:289`, `internal/output/envelope.go:116`). An agent parsing CLI output with `.error.details.violations` gets null; the actual accessor is `.error.detail.error.details.violations`. Add one clause here and in the Errors bullet at `:197` pointing readers at `error.detail` (the raw backend payload).

### P2 — §1 CSS guidance collides with the stricter element-replace gate (`skills/octo-html/SKILL.md:200`)

§1 (changed lines `:63-72`) now recommends CSS-only interaction — `:hover`, `:target`, transitions/animations, none expressible via inline `style` alone, so a `<style>` block is implied — and states CSS is allowed at publish. This unchanged bullet says element replacement rejects `scripts/styles`, while the replace spec at `internal/registry/specs/html.json:1058` lists only `scripts / inline event handlers / javascript: URLs`, omitting styles — a pre-existing skill-vs-spec mismatch that the new bullet inserted just above (`:196-199`) now sits next to. The documented workflow (publish a CSS-interactive doc per §1, then `html element replace` a fragment containing a `<style>` block) can 400 on this gate with no explanation in §1. Reconcile the replace-gate wording between `:200` and `html.json:1058`; if `<style>` is indeed rejected in replacement fragments, say so in §1 (e.g. use inline `style` attributes inside replacement fragments).

- Considered and dismissed: allowing `<meta http-equiv=refresh>` while banning non-empty `srcdoc` (`:69-72`) is consistent with the stated threat model. The gate targets stored script execution ("script that got stored would execute for every reader"); `srcdoc` nests a full HTML document and is a script-smuggling vector, meta-refresh cannot execute script, and plain `<a href>` already allows arbitrary navigation. Flagging only in case the final pass reads the backend policy differently.
- "list is complete (capped at 50, with `details.truncated` …)" (`:76`) reads slightly self-contradictory, but the same sentence documents the cap and the overflow flag; the fix-all-and-republish loop stays correct.
- No repo-wide doc contradicts the new rule (README/CLAUDE.md/docs/CHANGELOG grepped).

## Verdict: APPROVED

Both P2s are non-blocking clarity fixes; this PR changes no runtime contract.

yujiawei:
Addendum to my review above — one more P2 I missed on the first pass, plus a correction to something I overstated. Verdict is unchanged: **still APPROVE**, both items are non-blocking.

### P2 — the new CSS guidance in §1 collides with the stricter element-replace gate

§1 now tells a generating agent to express interaction with `:hover`, `:target`, `:checked` + sibling selectors and transitions (`skills/octo-html/SKILL.md:66-68`). None of those are expressible through an inline `style` attribute — they need a rule block. §1's allow-list then says "CSS, inline `style` … are all still allowed" (`:71-72`), which is true for publish/draft, the section it sits in.

But the Errors section two screens down says element replacement rejects "multiple top-level elements, **scripts/styles**, inline event handlers, and `javascript:` URLs" (`:200-201`). So a replacement fragment that introduces *new* rule-based CSS — exactly the kind §1 now steers agents toward — is refused, while §1 gives no hint that the fragment path is narrower than the publish path.

Two things worth separating here:

1. **The failure mode is narrower than it first looks.** `element replace` swaps one artifact and republishes; the document-level `<style>` in `<head>` survives. So ordinary content edits keep working, and only a fragment that needs to bring its own new rules trips the gate. Worth stating precisely rather than as a blanket workflow break.
2. **The skill and the spec disagree about whether `<style>` is refused at all.** `internal/registry/specs/html.json:1058` describes the same gate as "exactly one top-level element and free of scripts / inline event handlers / `javascript:` URLs" — no mention of styles. The skill has claimed styles are rejected since before the #125 rewrite (`<script>/<style>` in the previous revision). One of the two is wrong about the backend and I can't tell which from this repository.

Suggested fix, both cheap: scope §1's allow-list sentence to publish/draft and point at the fragment rule, and reconcile `:200-201` with `html.json:1058` in whichever direction matches the server. If `<style>` really is refused in fragments, saying "use inline `style` attributes inside replacement fragments" in §1 closes the loop.

### Correction to my P2 on the allow-list

I suggested `<meta http-equiv=refresh>` deserves a security caveat under the new top-level-serving rationale. On reflection that is weaker than I made it sound: meta refresh cannot execute script, and a plain `<a href>` already lets a document point a reader anywhere, so the incremental exposure is automatic-vs-clicked navigation rather than a new capability. The `<iframe>` half of that point is likewise about reader expectations, not a same-origin script escape — an embedded page runs its own script in its own origin, which the gate was never claiming to prevent.

The part of that finding I do stand behind is the first half: the shipped rationale and the PR description / commit message still say opposite things about whether stored script would run, and the commit message is the copy that lands in `main`.

yujiawei (approved):
# Code Review — PR #131 (octo-cli)

Reviewed at head `35ebe380` (`Merge branch 'main' into docs/html-no-javascript`). Prior approvals on this PR were issued against `461d5969`; this verdict is issued independently against the current head.

**Merge state verified:** `git merge-base 35ebe380 origin/main` = `df54c370`, i.e. main's tip is fully contained in the branch, and `git diff df54c370..35ebe380` is exactly `skills/octo-html/SKILL.md | 30 +++-` (+29/−1). The merge commit pulled in #133 (mail) but contributes nothing of its own — the PR's true delta against main is still one markdown file.

**Verdict: APPROVE.** No P0/P1. The change documents a hard backend gate that was previously discoverable only by a failed publish, and the actionable core — don't emit script, fix every violation in one pass, don't retry unchanged, don't try to encode around the gate — is correct and well placed (at the point of use in §1, plus a cross-referenced entry in the Errors list). Four non-blocking items below.

## 1. Scope

- Diff does what the title says and nothing more: one markdown file, no changes under `cmd/`, `internal/`, or `internal/registry/specs/`. No command, flag, or output change.
- Skill frontmatter bumped `0.1.0` → `0.2.0` (`skills/octo-html/SKILL.md:3`). Minor bump is right for purely additive guidance.
- §1 rule and the Errors entry cross-reference each other correctly (`SKILL.md:56-79`, `SKILL.md:196-200`).

**Testing claims verified locally at head**, not taken on trust:

```
go build ./...          # clean
go test ./... -count=1  # all packages pass (incl. skills + registry)
gofmt -l .              # clean
```

## 2. Findings (all P2 / non-blocking)

### P2 — `details.violations` is the backend wire path, not what an octo-cli caller sees

`SKILL.md:74` and `SKILL.md:197` tell a generating agent to read `details.violations`. That is accurate for the raw backend envelope, but this skill's audience consumes the **CLI** envelope, and the CLI does not forward `details` — it stuffs the entire backend body into a single `detail` field:

- `internal/output/errors.go:294-300` — `newBackendError(...)` sets `Detail: json.RawMessage(body)` (the whole response body).
- `internal/output/envelope.go:115-121` — the error envelope emits `error.detail` (singular) with that raw body verbatim.
- `internal/client/client.go:1073-1078` — `html` is not the `loop` service, so it goes through exactly this path.

So an agent that follows the sentence literally and runs `jq '.error.details.violations'` on `--output json` gets `null`. The real path is `.error.detail.error.details.violations`. This matters more than usual here because the same file already states the CLI envelope shape at `SKILL.md:52-55` (`{ok:false,error:{type,code,message,hint,detail}}`), so a careful reader hits a contradiction between two paragraphs of the same skill. Suggest naming the CLI-visible path explicitly, e.g. "the backend's `details.violations`, surfaced by the CLI under `error.detail`".

### P2 — "the list is complete" and "capped at 50" contradict each other in one sentence

`SKILL.md:74-76`:

> Fix all of them in one pass — the list is complete (capped at 50, with `details.truncated` set when it overflows).

The parenthetical retracts the claim the sentence just made. An agent reading the imperative and skipping the parenthesis will assume one pass always suffices. Suggest inverting: "the list is complete unless `details.truncated` is set; if it is, expect a second pass after fixing these."

### P2 — no test pins the new rule, against this repo's own convention

`skills/skills_test.go:70` (`TestOctoHTMLSkillCanonicalCreateContract`) exists precisely to pin load-bearing octo-html contract strings — it asserts on `idempotency_key`, `data.slug == data.doc_id`, `Never invent a slug`, and so on, and also asserts that stale guidance is *absent*. The main branch this PR merges in did the same thing one commit earlier: `df54c370` (#133) edited `skills/octo-mail/SKILL.md` and extended `skills/skills_test.go` (+18) in the same commit.

A no-JavaScript gate that returns a hard `400` is at least as load-bearing as the slug contract. Adding `html_contains_javascript` (and, say, `no JavaScript`) to that assertion list would cost two lines and prevent the rule being silently dropped by a future rewrite of §1.

### P2 — no `CHANGELOG.md` entry

The repo keeps a Keep-a-Changelog file and does record skill-documentation-only corrections there — e.g. "The `octo-drive` skill and a validation hint prescribed a `-r` flag the CLI does not implement" (`CHANGELOG.md:297`) and "Corrected the `drive` role contract in schema, docs and skill" (`CHANGELOG.md:381`). Documenting a previously undocumented hard backend rejection is comparable in reader impact. A one-line entry under `[Unreleased] → Added` (or `Changed`) would keep the convention intact.

## 3. Nits

- **`SKILL.md:71-72` vs `SKILL.md:200`.** §1 says "CSS, inline `style`, `<iframe>`, `<noscript>` and `<meta http-equiv=refresh>` are all still allowed", while the Errors section says element replacement "rejects multiple top-level elements, scripts/styles, inline event handlers, and `javascript:` URLs". These describe different endpoints (publish/draft vs element replacement) and are not actually in conflict, but nothing in §1 says so. Half a clause — "for publish and draft; element replacement is stricter (see below)" — removes the ambiguity.
- **PR description is stale relative to the shipped text.** The description still argues the rule from the opposite premise: "these documents render inside an iframe with `sandbox="allow-same-origin"` and **no** `allow-scripts`. Script never ran... silently inert." The shipped text says the reverse at `SKILL.md:68-71`: script is *not* merely inert, because the document's own version URL serves the stored HTML as a top-level page. The shipped wording is the more conservative and more useful framing, but if this repo squash-merges, the description becomes the commit message and the permanent record will contradict the file it landed. Worth a quick edit to the body before merge.

## 4. Coverage note / what I could not verify

Everything this diff asserts is a claim about a backend I cannot read: `Mininglamp-OSS/octo-docs-html` and `Mininglamp-OSS/octo-docs-backend` both return 404 for this reviewer, and an org-wide code search for `html_contains_javascript` returns nothing outside this PR. Specifically unverified:

1. The stable error code `html_contains_javascript` and its `400` status.
2. The exact rejected-construct list (`<script>` incl. inside `<svg>`, `on*`, `javascript:`/`vbscript:`/scriptable `data:` in `href`/`src`/`xlink:href`/`action`/`formaction`/`object[data]`, non-empty `srcdoc`).
3. The `details.violations` shape (`kind`/`tag`/`attr`/1-based `line`), the cap of 50, and `details.truncated`.
4. The allow-list (CSS, inline `style`, `<iframe>`, `<noscript>`, `<meta http-equiv=refresh>`).
5. The security framing at `SKILL.md:68-71` — that a version URL serves stored HTML top-level.

Items 1–3 fail safe if slightly wrong (an agent still avoids script and still reads whatever the error carries). Item 4 is the risky direction: an over-broad allow-list would have agents generating documents that then get rejected, and it is the part most likely to drift as the server gate tightens. Item 5 is a security assertion whose conservative direction is correct regardless. Since nothing in this repo binds the doc to the server's actual gate, this text can drift silently — which is another argument for the assertion in `skills_test.go` above, and for a follow-up link to the server-side change once that repo is reachable.

Approving: none of the above is something the author must fix before this can merge, and the change is a net improvement over the status quo of learning the rule by rejection.

PR #129, opened 2026-08-11
feat(mail): add Bot-bound Agent Mail commands

Liuxy20:
## Summary

Add Bot-bound Agent Mail authorization and command support through the unified OCTO gateway.

## Changes

- Add PKCE `mail auth login` and `mail auth status` flows with authoritative Bot identity verification.
- Store mailbox credentials and pending authorization proof in dedicated encrypted files, bound to RobotID, SpaceID, the normalized Octo API origin, and the authorizing Bot token fingerprint.
- Add metadata-driven Mail commands for mailbox identity, messages, threads, attachments, controlled send, versioned Drafts, delivery state, and JMAP changes.
- Select Bot and Mail credentials at separate transport boundaries, suppress Bot Space headers for mailbox operations, and disable retries for ambiguous Mail side effects.
- Keep explicit `--bot-id` selection fail-closed, expose an unverified environment claim as `identity.robot_id_claimed`, and keep Mail `--dry-run` fully network-free.
- Add `auth update` with the same API-origin validation as `auth login`, the embedded `octo-mail` skill, documentation, and regression tests.

## Motivation

Fixes #128

Depends on Mininglamp-OSS/octo-mail#48 and the external `/agent-mail-api` routing in Mininglamp-OSS/octo-web#1315.

The CLI calls the device and token bootstrap endpoints without attaching Bot or Mail credentials. A real authorization login verifies the active Bot through `/v1/bot/register` before creating the device flow. Live end-to-end authorization was not exercised; the external gateway must admit those public PKCE bootstrap endpoints while stripping unrelated credentials.

## Testing

- [x] `go test ./... -count=1` passes
- [x] `go test -race -shuffle=on -count=1 ./...` passes
- [x] `go vet ./...` passes
- [x] `go build ./...` passes
- [x] `gofmt -l .` is clean
- [x] `git diff --check` passes

## Checklist

- [x] Code is in English (comments, commit messages, variable names)
- [x] New commands added to README Command Reference table
- [x] CLAUDE.md command list updated if commands changed
- [x] New features include tests
- [x] No unrelated changes included

PR #133, opened 2026-08-17
fix(mail): align Agent commands with owner authorization

### LINKED ISSUE #132: fix(mail): align Agent commands with owner authorization
## Problem / Background

The Agent Mail command registry exposes owner-only message and Draft mutation operations. Those operations are rejected by the backend for Agent mailbox credentials. The attachment schema also uses a field name that does not match the backend send-intent contract.

## Goal

Align the octo-cli Agent Mail surface and embedded Skill with the authorization model enforced by octo-mail.

## Current state

- Implemented locally with registry, command, and Skill tests.
- The backend authorization model remains unchanged.

## Scope

- Keep Agent operations for reading, flags, delivery state, send intents, Agent Draft creation, and reply-Draft preparation.
- Remove direct send, reply, reply-all, forward, permanent message deletion, and owner Draft mutation commands from the Agent registry.
- Remove the obsolete Agent self-confirmation instructions.
- Direct owners to OCTO Web for Draft review, send, edit, and deletion.
- Use the backend attachment field "content" for base64 payloads.
- Update embedded Skill documentation and command-surface tests.

## Out of scope

- Backend authorization changes.
- OCTO Web behavior.
- Credential storage changes.

## Acceptance criteria

- [ ] Owner-only operations are absent from octo-cli mail.
- [ ] Agent send-intent, create-agent, and reply-draft workflows remain available.
- [ ] The Skill directs owner-only actions to OCTO Web.
- [ ] Attachment send-intent requests use the backend "content" field.
- [ ] Go tests, vet, build, and golangci-lint pass.

## Verification

- Registry and command tests cover allowed and excluded operations.
- Skill tests reject stale confirmation-token instructions.
- Attachment request test verifies the serialized backend field.

Liuxy20:
## Summary

Align the octo-cli Agent Mail command surface and Skill with the authorization contract enforced by octo-mail.

## Changes

- Remove owner-only direct message and Draft mutation operations from the Agent command registry.
- Keep Agent-safe read, flag, delivery, send-intent, Agent Draft, and reply-Draft workflows.
- Direct owner review, editing, sending, and deletion to OCTO Web in the embedded Skill.
- Serialize attachment payloads with the backend `content` field and update command-surface tests.

## Motivation

The removed operations require Owner authorization and are not executable with Agent mailbox credentials. This keeps the CLI surface limited to operations the Agent can complete.

Closes #132

## Testing

- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] `go build ./...` passes
- `go test -race ./... -count=1` — passed
- `golangci-lint run ./...` — 0 issues

## Checklist

- [x] New commands added to README Command Reference table (not applicable; no new commands)
- [x] CLAUDE.md command list updated if commands changed

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -101,11 +101,11 @@ octo-cli mail      me
                mailbox list
                address list
                thread get
-               message list|read|raw|send|send-intent|reply|reply-draft
-                       reply-all|forward|flag|delete|delivery|auto-reply-context
+               message list|read|raw|send-intent|reply-draft
+                       flag|delivery|auto-reply-context
                        state|changes
                        attachment download
-               draft list|create|create-agent|update|send|delete
+               draft list|create-agent
 
 octo-cli auth      login | status | update | logout | list
 octo-cli schema [--list [domain] | <operation-id>]
--- a/cmd/service/mail_test.go
+++ b/cmd/service/mail_test.go
@@ -1,6 +1,7 @@
 package service
 
 import (
+	"encoding/json"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -69,16 +70,14 @@ func TestMailRegistrySurface(t *testing.T) {
 	want := map[string]string{
 		"mail.address.list":                "/agent-mail-api/webapi/v0/addresses",
 		"mail.draft.create_agent":          "/agent-mail-api/webapi/v0/agent-drafts",
-		"mail.draft.update":                "/agent-mail-api/webapi/v0/drafts/{id}",
+		"mail.draft.list":                  "/agent-mail-api/webapi/v0/drafts",
 		"mail.me":                          "/agent-mail-api/webapi/v0/identity",
 		"mail.message.auto_reply_context":  "/agent-mail-api/webapi/v0/messages/{id}/auto-reply-context",
 		"mail.message.list":                "/agent-mail-api/webapi/v0/messages",
 		"mail.message.raw":                 "/agent-mail-api/webapi/v0/messages/{id}/raw",
 		"mail.message.reply_draft":         "/agent-mail-api/webapi/v0/messages/{id}/reply-draft",
-		"mail.message.send":                "/agent-mail-api/webapi/v0/messages",
 		"mail.message.send_intent":         "/agent-mail-api/webapi/v0/agent-send-intents",
 		"mail.message.read":                "/agent-mail-api/webapi/v0/messages/{id}",
-		"mail.message.reply":               "/agent-mail-api/webapi/v0/messages/{id}/reply",
 		"mail.message.delivery":            "/agent-mail-api/webapi/v0/messages/{id}/delivery",
 		"mail.message.attachment.download": "/agent-mail-api/webapi/v0/messages/{id}/attachments/{partId}",
 		"mail.thread.get":                  "/agent-mail-api/webapi/v0/threads/{id}",
@@ -115,18 +114,9 @@ func TestMailRegistrySurface(t *testing.T) {
 	}
 
 	for _, id := range []string{
-		"mail.message.send",
 		"mail.message.send_intent",
-		"mail.message.reply",
 		"mail.message.reply_draft",
-		"mail.message.reply_all",
-		"mail.message.forward",
-		"mail.message.delete",
-		"mail.draft.create",
 		"mail.draft.create_agent",
-		"mail.draft.update",
-		"mail.draft.send",
-		"mail.draft.delete",
 	} {
 		op, ok := reg.GetOperation(id)
 		if !ok {
@@ -147,7 +137,7 @@ func TestMailRegistrySurface(t *testing.T) {
 	}
 }
 
-func TestMailConfirmationTokensAreSecret(t *testing.T) {
+func TestMailRegistryExcludesOwnerOnlyOperationsAndDeletedConfirmationFlow(t *testing.T) {
 	reg := registry.MustNew()
 	for _, operationID := range []string{
 		"mail.message.send",
@@ -155,27 +145,24 @@ func TestMailConfirmationTokensAreSecret(t *testing.T) {
 		"mail.message.reply",
 		"mail.message.reply_all",
 		"mail.message.forward",
+		"mail.draft.create",
+		"mail.draft.update",
 		"mail.draft.send",
 		"mail.draft.delete",
 	} {
-		op, ok := reg.GetOperation(operationID)
+		if _, ok := reg.GetOperation(operationID); ok {
+			t.Errorf("owner-only operation %s must not be exposed by Agent Mail", operationID)
+		}
+	}
+	for _, info := range reg.ListOperations("mail") {
+		op, ok := reg.GetOperation(info.ID)
 		if !ok {
-			t.Errorf("missing operation %s", operationID)
-			continue
+			t.Fatalf("mail operation %s disappeared", info.ID)
 		}
-		var confirmation *registry.ParamInfo
-		for i := range op.Parameters {
-			if op.Parameters[i].Name == "X-Octo-Confirmation" && op.Parameters[i].In == "header" {
-				confirmation = &op.Parameters[i]
-				break
+		for _, parameter := range op.Parameters {
+			if parameter.Name == "X-Octo-Confirmation" || parameter.FlagName == "confirmation-token" {
+				t.Errorf("%s still exposes the deleted Agent self-confirmation flow", info.ID)
 			}
-		}
-		if confirmation == nil {
-			t.Errorf("%s has no X-Octo-Confirmation header", operationID)
-			continue
-		}
-		if !confirmation.Secret {
-			t.Errorf("%s confirmation token is not marked secret", operationID)
 		}
 	}
 }
@@ -224,6 +211,51 @@ func TestMailSendIntentUsesPolicyEndpointAndDoesNotRetry(t *testing.T) {
 	}
 }
 
+func TestMailSendIntentAttachmentUsesBackendContentField(t *testing.T) {
+	var gotContent string
+	mailServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var body struct {
+			Attachments []struct {
+				Content string `json:"content"`
+			} `json:"attachments"`
+		}
+		if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
+			t.Fatalf("decode body: %v", err)
+		}
+		if len(body.Attachments) != 1 {
+			t.Fatalf("attachments = %#v", body.Attachments)
+		}
+		gotContent = body.Attachments[0].Content
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusAccepted)
+		_, _ = w.Write([]byte(`{"outcome":"accepted","status":"accepted","messageId":"E1"}`))
+	}))
+	defer mailServer.Close()
+
+	tf := cmdutil.NewTestFactory()
+	cfg := &config.Config{APIBaseURL: mailServer.URL, BotToken: "app_test", Format: "json"}
+	tf.SetConfig(cfg)
+	tf.SetCredential(&credential.BotCredential{Token: "app_test", Source: "test"})
+	mailCredential := &credential.MailCredential{Token: "omb_mail_secret", Source: "test"}
+	tf.SetMailCredential(mailCredential)
+	tf.SetMailClient(client.NewMail(cfg, mailCredential, client.Options{ErrOut: io.Discard}))
+	tf.RegistryFunc = registry.MustNew
+
+	root := &cobra.Command{Use: "octo-cli", SilenceUsage: true, SilenceErrors: true}
+	RegisterServiceCommands(root, tf.Factory)
+	root.SetArgs([]string{
+		"mail", "message", "send-intent",
+		"--data", `{"to":["recipient@example.com"],"subject":"Attachment","attachments":[{"filename":"image.png","contentType":"image/png","content":"aW1hZ2U="}]}`,
+		"--idempotency-key", "attachment-intent-0001",
+	})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("execute attachment send intent: %v\n%s", err, tf.ErrOut.String())
+	}
+	if gotContent != "aW1hZ2U=" {
+		t.Fatalf("attachment content = %q", gotContent)
+	}
+}
+
 func TestMailAttachmentDownloadWritesOutput(t *testing.T) {
 	mailServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.URL.Path != "/agent-mail-api/webapi/v0/messages/E1/attachments/1.2" {
@@ -260,41 +292,5 @@ func TestMailAttachmentDownloadWritesOutput(t *testing.T) {
 	}
 	if string(content) != "attachment body" {
 		t.Fatalf("downloaded content = %q", content)
-	}
-}
-
-func TestMailSendForwardsServerConfirmationToken(t *testing.T) {
-	var gotConfirmation string
-	mailServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		gotConfirmation = r.Header.Get("X-Octo-Confirmation")
-		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusAccepted)
-		_, _ = w.Write([]byte(`{"messageId":"E1","submissionIds":["S1"]}`))
-	}))
-	defer mailServer.Close()
-
-	tf := cmdutil.NewTestFactory()
-	cfg := &config.Config{APIBaseURL: mailServer.URL, BotToken: "app_test", Format: "json"}
-	tf.SetConfig(cfg)
-	tf.SetCredential(&credential.BotCredential{Token: "app_test", Source: "test"})
-	mailCredential := &credential.MailCredential{Token: "omb_mail_secret", Source: "test"}
-	tf.SetMailCredential(mailCredential)
-	tf.SetMailClient(client.NewMail(cfg, mailCredential, client.Options{ErrOut: io.Discard}))
-	tf.RegistryFunc = registry.MustNew
-
-	root := &cobra.Command{Use: "octo-cli", SilenceUsage: true, SilenceErrors: true}
-	RegisterServiceCommands(root, tf.Factory)
-	root.SetArgs([]string{
-		"mail", "message", "send",
-		"--to", "recipient@example.com",
-		"--subject", "Confirmed",
-		"--text", "Body",
-		"--confirmation-token", "omc_once",
-	})
-	if err := root.Execute(); err != nil {
-		t.Fatalf("execute confirmed send: %v\n%s", err, tf.ErrOut.String())
-	}
-	if gotConfirmation != "omc_once" {
-		t.Fatalf("X-Octo-Confirmation = %q", gotConfirmation)
 	}
 }
--- a/internal/registry/loader_test.go
+++ b/internal/registry/loader_test.go
@@ -48,7 +48,7 @@ func TestAllDomainOperationCounts(t *testing.T) {
 		"drive":       42,
 		"html":        21,
 		"marketplace": 44,
-		"mail":        24,
+		"mail":        15,
 		"summary":     4,
 		"loop":        126,
 	}
--- a/internal/registry/specs/mail.json
+++ b/internal/registry/specs/mail.json
@@ -39,31 +39,6 @@
           { "name": "offset", "in": "query", "schema": { "type": "integer", "default": 0 } }
         ],
         "responses": { "200": { "description": "Message page" } }
-      },
-      "post": {
-        "operationId": "mail.message.send",
-        "summary": "Send a message through the legacy confirmation flow",
-        "x-octo-risk": "write",
-		"x-octo-retry": "never",
-		"parameters": [
-		  { "name": "X-Octo-Confirmation", "in": "header", "x-octo-flag": "confirmation-token", "x-octo-secret": true, "description": "One-time server confirmation token", "schema": { "type": "string" } }
-		],
-        "requestBody": {
-          "required": true,
-          "content": {
-            "application/json": {
-              "schema": {
-                "$ref": "#/components/schemas/SendMessage"
-              }
-            }
-          }
-        },
-        "responses": {
-          "202": {
-            "description": "Accepted for delivery",
-            "content": { "application/json": { "schema": { "$ref": "#/components/schemas/AcceptedWrite" } } }
-          }
-        }
       }
     },
     "/agent-mail-api/webapi/v0/messages/{id}": {
@@ -98,17 +73,6 @@
           }
         },
         "responses": { "200": { "description": "Updated" } }
-      },
-      "delete": {
-        "operationId": "mail.message.delete",
-        "summary": "Permanently delete a message",
-        "x-octo-risk": "high-risk-write",
-        "x-octo-retry": "never",
-        "parameters": [
-		  { "name": "id", "in": "path", "required": true, "schema": { "type": "string" } },
-		  { "name": "X-Octo-Confirmation", "in": "header", "x-octo-flag": "confirmation-token", "x-octo-secret": true, "description": "One-time server confirmation token", "schema": { "type": "string" } }
-        ],
-        "responses": { "204": { "description": "Deleted" } }
       }
     },
     "/agent-mail-api/webapi/v0/messages/{id}/raw": {
@@ -163,84 +127,6 @@
                 "schema": { "type": "string", "format": "binary" }
               }
             }
-          }
-        }
-      }
-    },
-    "/agent-mail-api/webapi/v0/messages/{id}/reply": {
-      "post": {
-        "operationId": "mail.message.reply",
-        "summary": "Reply to a message",
-        "x-octo-risk": "write",
-        "x-octo-retry": "never",
-        "parameters": [
-		  { "name": "id", "in": "path", "required": true, "schema": { "type": "string" } },
-		  { "name": "X-Octo-Confirmation", "in": "header", "x-octo-flag": "confirmation-token", "x-octo-secret": true, "description": "One-time server confirmation token", "schema": { "type": "string" } }
-        ],
-        "requestBody": {
-          "required": true,
-          "content": { "application/json": { "schema": { "$ref": "#/components/schemas/ReplyMessage" } } }
-        },
-        "responses": {
-          "202": {
-            "description": "Accepted for delivery",
-            "content": { "application/json": { "schema": { "$ref": "#/components/schemas/AcceptedWrite" } } }
-          }
-        }
-      }
-    },
-    "/agent-mail-api/webapi/v0/messages/{id}/reply-all": {
-      "post": {
-        "operationId": "mail.message.reply_all",
-        "summary": "Reply to all recipients",
-        "x-octo-risk": "write",
-        "x-octo-retry": "never",
-        "parameters": [
-		  { "name": "id", "in": "path", "required": true, "schema": { "type": "string" } },
-		  { "name": "X-Octo-Confirmation", "in": "header", "x-octo-flag": "confirmation-token", "x-octo-secret": true, "description": "One-time server confirmation token", "schema": { "type": "string" } }
-        ],
-        "requestBody": {
-          "required": true,
-          "content": { "application/json": { "schema": { "$ref": "#/components/schemas/ReplyMessage" } } }
-        },
-        "responses": {
-          "202": {
-            "description": "Accepted for delivery",
-            "content": { "application/json": { "schema": { "$ref": "#/components/schemas/AcceptedWrite" } } }
-          }
-        }
-      }
-    },
-    "/agent-mail-api/webapi/v0/messages/{id}/forward": {
-      "post": {
-        "operationId": "mail.message.forward",
-        "summary": "Forward a message",
-        "x-octo-risk": "write",
-        "x-octo-retry": "never",
-        "parameters": [
-		  { "name": "id", "in": "path", "required": true, "schema": { "type": "string" } },
-		  { "name": "X-Octo-Confirmation", "in": "header", "x-octo-flag": "confirmation-token", "x-octo-secret": true, "description": "One-time server confirmation token", "schema": { "type": "string" } }
-        ],
-        "requestBody": {
-          "required": true,
-          "content": {
-            "application/json": {
-              "schema": {
-                "type": "object",
-                "required": ["to"],
-                "properties": {
-                  "to": { "type": "array", "items": { "type": "string" } },
-                  "text": { "type": "string" },
-                  "attachments": { "type": "array", "items": { "$ref": "#/components/schemas/Attachment" } }
-                }
-              }
-            }
-          }
-        },
-        "responses": {
-          "202": {
-            "description": "Accepted for delivery",
-            "content": { "application/json": { "schema": { "$ref": "#/components/schemas/AcceptedWrite" } } }
           }
         }
       }
@@ -344,66 +230,6 @@
         "summary": "List drafts",
         "x-octo-risk": "read",
         "responses": { "200": { "description": "Drafts" } }
-      },
-      "post": {
-        "operationId": "mail.draft.create",
-        "summary": "Create a draft",
-        "x-octo-risk": "write",
-        "x-octo-retry": "never",
-        "requestBody": {
-          "required": true,
-          "content": { "application/json": { "schema": { "$ref": "#/components/schemas/SendMessage" } } }
-        },
-        "responses": { "201": { "description": "Created" } }
-      }
-    },
-    "/agent-mail-api/webapi/v0/drafts/{id}/send": {
-      "post": {
-        "operationId": "mail.draft.send",
-        "summary": "Send an existing draft",
-        "x-octo-risk": "write",
-        "x-octo-retry": "never",
-        "parameters": [
-		  { "name": "id", "in": "path", "required": true, "schema": { "type": "string" } },
-		  { "name": "X-Octo-Confirmation", "in": "header", "x-octo-flag": "confirmation-token", "x-octo-secret": true, "description": "One-time server confirmation token", "schema": { "type": "string" } }
-		],
-        "requestBody": {
-          "required": false,
-          "content": { "application/json": { "schema": { "$ref": "#/components/schemas/SendDraft" } } }
-        },
-        "responses": {
-          "202": {
-            "description": "Accepted for delivery",
-            "content": { "application/json": { "schema": { "$ref": "#/components/schemas/AcceptedWrite" } } }
-          }
-        }
-      }
-    },
-    "/agent-mail-api/webapi/v0/drafts/{id}": {
-      "patch": {
-        "operationId": "mail.draft.update",
-        "summary": "Replace a draft with an updated immutable version",
-        "x-octo-risk": "write",
-        "x-octo-retry": "never",
-        "parameters": [
-          { "name": "id", "in": "path", "required": true, "schema": { "type": "string" } }
-        ],
-        "requestBody": {
-          "required": true,
-          "content": { "application/json": { "schema": { "$ref": "#/components/schemas/UpdateDraft" } } }
-        },
-        "responses": { "200": { "description": "Replacement draft id and version" } }
-      },
-      "delete": {
-        "operationId": "mail.draft.delete",
-        "summary": "Delete a draft",
-        "x-octo-risk": "high-risk-write",
-        "x-octo-retry": "never",
-        "parameters": [
-		  { "name": "id", "in": "path", "required": true, "schema": { "type": "string" } },
-		  { "name": "X-Octo-Confirmation", "in": "header", "x-octo-flag": "confirmation-token", "x-octo-secret": true, "description": "One-time server confirmation token", "schema": { "type": "string" } }
-        ],
-        "responses": { "204": { "description": "Deleted" } }
       }
     },
     "/agent-mail-api/webapi/v0/addresses": {
@@ -419,11 +245,11 @@
     "schemas": {
       "Attachment": {
         "type": "object",
-        "required": ["filename", "contentType", "data"],
+        "required": ["filename", "contentType", "content"],
         "properties": {
           "filename": { "type": "string" },
           "contentType": { "type": "string" },
-          "data": { "type": "string", "description": "Base64-encoded content" }
+          "content": { "type": "string", "description": "Base64-encoded content" }
         }
       },
       "SendMessage": {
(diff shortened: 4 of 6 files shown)


Jerry-Xin (dismissed):
This in-scope PR correctly aligns Agent Mail commands with owner authorization boundaries.

✅ Highlights

- Owner-only operations are fully removed from the registry and documentation.
- Attachment payloads now use the backend-compatible `content` field.
- Regression tests cover excluded operations and obsolete confirmation tokens.
- Skill guidance clearly routes owner actions to OCTO Web.
- `go test ./...` and `go vet ./...` pass.
- No security, correctness, performance, or architectural blockers found.

yujiawei (approved):
# Code Review — PR #133 (octo-cli)

Reviewed `e627723..7a862c1` (6 files, +106 / −333) against the acceptance criteria in #132. All verification below was run locally against the PR head.

| Criterion (#132) | Status | Evidence |
|---|---|---|
| Owner-only operations absent from `octo-cli mail` | ✅ | 24 → 15 operations in `internal/registry/specs/mail.json` |
| `send-intent`, `create-agent`, `reply-draft` still available | ✅ | present at head, all three keep `x-octo-retry: never` |
| Skill directs owner-only actions to OCTO Web | ✅ | `skills/octo-mail/SKILL.md:43-47, 199-202, 218-219, 226-228` |
| Attachment send-intent uses the backend `content` field | ✅ | `mail.json` `Attachment.required = [filename, contentType, content]` |
| build / vet / test / lint pass | ✅ (lint via CI) | see below |

I enumerated `operationId`s at base and at head: the delta is **exactly** the nine operations named in the issue — `mail.message.send`, `reply`, `reply_all`, `forward`, `delete` and `mail.draft.create`, `update`, `send`, `delete`. Nothing else was removed, no operation was added, and no behavior outside the stated scope changed. **Nothing missing, nothing extra, no deviation from the issue.**

`CLAUDE.md:120` asks that `skills/` be kept in sync when command shapes change; this PR does that, which is why the skill rewrite belongs in the same change rather than a follow-up.

## 2. Correctness verification

A spec deletion of this size tends to leave orphans behind, so I checked for them specifically:

- **No dangling `$ref`.** At head, the schemas defined and the set of `#/components/schemas/*` references are identical: `{Attachment, SendMessage, ReplyMessage, AgentDraftResult, AgentSendIntentResult}`. `AcceptedWrite`, `SendDraft` and `UpdateDraft` were removed together with their only referents. Worth confirming explicitly because `internal/registry/loader.go` only unmarshals and checks duplicate operation IDs — it does not globally validate refs, so an orphan would surface at runtime rather than at load.
- **No stale references to the removed operation IDs** remain in Go, Markdown, or the other specs. `README.md` and `CLAUDE.md` were already free of the removed commands or are updated here.
- **The generated command tree matches the updated docs.** Built the binary and enumerated: `mail message` = `attachment, auto-reply-context, changes, delivery, flag, list, raw, read, reply-draft, send-intent, state`; `mail draft` = `create-agent, list`. That is exactly the tree now documented at `CLAUDE.md:99-108`. `state` / `changes` are hand-written in `cmd/mail_jmap.go` and are correctly unaffected.
- **`go build ./...`, `go vet ./...`, `gofmt -l .`, and `go test -race -count=1 ./...` are all clean** at `7a862c1`. (`golangci-lint` is not installed in my environment; relying on CI for that one.)

**The removal costs no reachable capability.** This was the claim I most wanted to check independently, because deleting nine commands is only safe if the CLI can never hold an owner credential. It cannot — `cmd/mail_auth.go:247` rejects any stored mail token that is not `omb_`-prefixed:

```go
if err := json.Unmarshal(raw, &token); err != nil || !strings.HasPrefix(token.AccessToken, "omb_") || token.MailboxAddress == "" {
```

There is no path by which `octo-cli` authenticates as the human owner, so the removed endpoints were unreachable in practice rather than merely discouraged. That makes the "align the surface with the authorization contract" framing accurate.

**The attachment rename is load-bearing, and the new test proves it.** `attachments` has no promoted flag — `send-intent --help` exposes only `--to/--cc/--bcc/--subject/--text/--html/--data` — so `--data` is the sole attachment path, and `--data` is *not* raw passthrough: `bodySchemaValidator` (`cmd/service/run.go:664-712`) recurses into arrays of objects and enforces `required`. I confirmed `TestMailSendIntentAttachmentUsesBackendContentField` is non-vacuous by reverting only the `Attachment` schema back to `data` and re-running it:

```
--- FAIL: TestMailSendIntentAttachmentUsesBackendContentField
    mail_test.go:252: VALIDATION_ERROR: request body is missing required field(s): attachments[0].data
```

So the test does bind the schema to the wire, and `octo-cli schema mail.message.send_intent` now advertises `content` to callers. The three write operations that take attachments (`send_intent`, `draft.create_agent`, `reply_draft`) all funnel through the single `Attachment` definition, so there is no second copy left on the old field name.

## 3. Findings

No P0 or P1. Five P2s, none of them merge-blocking.

### P2-1 — no `CHANGELOG.md` entry, and these commands shipped in a release

`git tag --contains e627723` returns `v0.12.0` (tagged 2026-08-14), so `mail message send|reply|reply-all|forward|delete` and the `draft create|update|send|delete` family exist in a released binary. The project follows Keep a Changelog and has used `### Removed` for exactly this situation before (`CHANGELOG.md:428`, `:495`), but `## [Unreleased]` gains nothing here — a user upgrading from v0.12.0 gets `unknown command` with no note pointing at OCTO Web. Suggested entry:

```markdown
### Removed
- **Owner-only Agent Mail commands** — `mail message send|reply|reply-all|forward|delete`
  and `mail draft create|update|send|delete`, plus the `--confirmation-token` flow.
  They require Owner authorization and are rejected for Agent mailbox credentials.
  Use `mail message send-intent`, `mail draft create-agent`, and `mail message reply-draft`;
  review, edit, send, and deletion are completed by the owner in OCTO Web.
```

For fairness: the `mail` domain is absent from `CHANGELOG.md` entirely today (`grep -i mail CHANGELOG.md` → 0 hits), so this is a pre-existing gap this PR is a natural moment to close, not a regression it introduced.

### P2-2 — the skill guard covers only half the removed surface

`skills/skills_test.go:188-193` forbids `octo-cli mail message delete` and `octo-cli mail draft send`, but not `reply-all`, `forward`, `octo-cli mail draft update`, `octo-cli mail draft delete`, or the bare `octo-cli mail message send` / `octo-cli mail draft create`. The guard's stated purpose is to keep stale owner-only instructions out of the embedded skill; as written, a future edit reintroducing `octo-cli mail message forward <id> --to ...` into `SKILL.md` would pass CI. Extending the slice is cheap and makes the guard match its intent.

Two of those need care: `octo-cli mail message send` and `octo-cli mail draft create` must be matched with a trailing space (or a stricter check) so they do not also match `send-intent` and `create-agent`.

### P2-3 — the skill routes attachments through `--data` but never gives the object shape

`SKILL.md:182` says "For complex messages or attachments, pass JSON through `--data @file.json`", and that is now the only way to attach anything — yet the skill never shows the attachment object, and never mentions `octo-cli schema <operation-id>` as a discovery path. Since the field name just changed and the CLI now hard-rejects the old one, an agent working from prior knowledge of `data` hits a validation error with nothing in-skill to correct it. Three lines close the loop:

```markdown
An attachment is `{"filename": ..., "contentType": ..., "content": "<base64>"}` —
`content` is the backend field name. Run `octo-cli schema mail.message.send_intent`
for the authoritative body schema.
```

### P2-4 — brittle assertion string

`skills/skills_test.go:191` forbids the literal `"confirmation_required without"`, which is a fragment of the sentence this PR deleted rather than a stable token. If that sentence were ever reworded instead of removed, the assertion would silently stop matching. `--confirmation-token` and `omc_` on the same list already carry the real signal; consider dropping the fragment.

### P2-5 — `t.Fatalf` from inside the httptest handler goroutine

`cmd/service/mail_test.go:223` and `:226` call `t.Fatalf` from the HTTP handler, which runs on the server's goroutine, not the test's. `testing` documents `Fatal` as test-goroutine-only: it marks the test failed and then calls `runtime.Goexit()`, which kills the handler mid-request, so the client sees a bare `EOF` and the failure surfaces at `root.Execute()` (line 252) as a transport error rather than "decode body: …". The test still fails — there is no false-pass here — but the message points at the wrong thing.

The sibling handler directly below (`:261-266`) already uses `t.Errorf` for exactly this reason. Suggest matching it:

```go
if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
    t.Errorf("decode body: %v", err)
    return
}
```

Flagging this as a P2 rather than higher because it is a pre-existing convention in the repo, not something this PR invented — `cmd/mail_auth_test.go:473` does the same thing — and `go vet` does not catch it (its `testinggoroutine` analyzer only sees `go` statements written inside the test function).

## 4. Notes, not findings

- **`TestMailRegistrySurface`'s `want` map is a subset check, not set equality** (`cmd/service/mail_test.go:70-84`; it has always omitted `mail.mailbox.list` and `mail.message.flag`). Pre-existing, and adequately backstopped here by the new exclusion test plus `loader_test.go:51`'s `"mail": 15`. Asserting set equality would lock both directions — but note the count assertion alone is coarse, since swapping one operation for another preserves 15.
- **This is a surface alignment, not an authorization control.** `send-intent` still accepts arbitrary `to`/`cc`, so an agent could approximate reply-all or forward (losing proper `In-Reply-To` / `References` threading) if it chose to. That is fine — server policy remains the authority, which is exactly the direction this PR converges on — but the removal should not be read as closing a hole.
- `SKILL.md:46` still says "Do not retry an owner-only endpoint", referring to endpoints the CLI no longer exposes. Harmless as a general guardrail.
- The new attachment test's handler asserts neither the request path nor the `Authorization` header, unlike its neighbours. I looked at whether that leaves a hole and concluded it does not: the send-intent path is already pinned by `TestMailSendIntentUsesPolicyEndpointAndDoesNotRetry` (`mail_test.go:175-177`) and the mail bearer token by `TestMailCommandUsesMailEndpointAndCredential` and the download test. Adding the two lines would be tidier, but nothing is currently unguarded.

## 5. Coverage — what I could not verify

- **The backend contract itself.** Both load-bearing premises — that the nine endpoints reject Agent mailbox credentials, and that the attachment field is `content` — rest on the assertion in #132. The mail backend is not reachable from this repo and an org-wide code search for `agent-send-intents` returned nothing. The CLI-side change is internally consistent and the `omb_`-only credential check makes the authorization premise very plausible, but nothing in this repo pins either one. If the mail service publishes an OpenAPI document, a contract test against it would make this class of drift — which is what produced `data` vs `content` in the first place — impossible rather than caught-later.
- `golangci-lint` was not available locally; taking the PR's claim plus CI for that check.
- No live mailbox was exercised; verification was static analysis plus the repo's own `httptest` harness.

## Verdict

**APPROVED.** The scope matches the linked issue exactly, the deletions leave no orphaned refs or stale references, the command tree and docs agree with the spec, the full test suite passes under `-race`, and the one behavioral change (`data` → `content`) is genuinely covered by a test I confirmed fails without the fix. The five P2s are documentation and test-hygiene improvements; the `CHANGELOG` entry (P2-1) is the one I would most like to see land, ideally in this PR since the commands shipped in v0.12.0, but none of them should block merge.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #133 (octo-cli)

## Summary

This PR aligns the Agent Mail command surface with the authorization contract enforced by octo-mail: it removes nine owner-only operations (direct send/reply/reply-all/forward, permanent message delete, and owner Draft create/update/send/delete) from `internal/registry/specs/mail.json`, deletes the obsolete Agent self-confirmation-token instructions from the embedded `octo-mail` skill, renames the attachment payload field `data` → `content` to match the backend wire contract, and updates the command-surface tests and CLAUDE.md accordingly.

I verified every removed and kept operation against the actual octo-mail backend source (read-only). All nine removed operations were already unusable with Agent mailbox credentials, and every kept operation is genuinely Agent-accessible, so the reduction removes dead/broken surface without breaking any working path. No P0/P1 issues; one pre-existing P2 doc drift and two nits below.

## Verification

Static analysis only at head `7a862c1`; build and tests not executed in this environment. Cross-repo contract verification was done read-only against `Mininglamp-OSS/octo-mail` @ HEAD:

- ✅ **Removed ops are Agent-unusable in the backend** — `mail.message.send/delete/reply/reply_all/forward` and `mail.draft.send/delete` sit behind `hAgentConfirmed`, which returns 403 `owner_required` for Agent credentials lacking an `X-Octo-Automation` header; the CLI never wired that header (grep at base and head: zero hits), and the old `X-Octo-Confirmation` token flow does not exist anywhere in octo-mail (0 code hits for the header and for `confirmation_required`). `createDraft`/`updateDraft` hard-reject Agents (`owner_required`) in every branch.
- ✅ **Kept ops are Agent-accessible** — plain-auth routes for list/read/raw/flag/delivery/auto-reply-context/reply-draft/thread/draft-list/address-list/identity, plus `agent-send-intents`/`agent-drafts` guarded by `requireAgentDraftCredential`. `flag` is limited to non-owner keywords, but the skill's examples (`$seen`, `$flagged`) are in the allowed set.
- ✅ **Attachment rename matches the backend** — octo-mail's `attachment` struct uses `Content string \`json:"content"\`` (base64), decoded in `compose()`; the old `data` field was silently mismatched, so this is a fix. The CLI's `bodySchemaValidator` enforces the new `required` list recursively, so the contract is checked client-side before send.
- ✅ **Response contract matches** — backend returns 202 `{outcome:"accepted", messageId, submissionIds, senderAddress}` and 201 `agentDraftInfo{outcome:"owner_confirmation_required", status, draftType, draftId, draftVersion, draftSubject, ...}`, exactly the shapes in the spec.
- ✅ **Spec/tests/docs self-consistency** — 15 operationIds remain, matching `loader_test.go` count 24→15; no dangling `$ref`s (`SendMessage`/`ReplyMessage`/`Attachment` still consumed by send-intent/create-agent/reply-draft); zero repo-wide references to removed commands or the confirmation flow outside tests; skills_test required/forbidden string assertions all pass against the new SKILL.md.

## Findings

No P0/P1 issues; one P2 (pre-existing) and two nits below.

### P2 — `owner_review_required` outcome is documented but never returned (`skills/octo-mail/SKILL.md:179`)

The skill instructs Agents to interpret `owner_review_required` as a structured send-intent outcome, and the spec declares it in the enum (`internal/registry/specs/mail.json:295`). In octo-mail, a policy review block is returned as HTTP 409 with error code `outbound_review_required` and a policy projection — an error response, not a 2xx outcome. An Agent following the skill sees an error instead of the documented outcome. This is fail-safe (nothing is sent; the 409 message states the email was not sent and owner review is required) and the drift already existed at base — this PR only reworded the bullet — so it is not a merge blocker. Fix direction: document the 409 error shape in the skill (and/or drop the unreachable enum value), so Agents branch on the real wire signal.

### Nit — `state|changes` subcommands listed but nonexistent (`CLAUDE.md:106`)

`mail message ... state|changes` remains in the command list, but no such operations exist in the mail registry at base (24 ops) or head (15 ops). Pre-existing drift adjacent to the lines this PR edits; clean up or annotate as aspirational.

### Nit — no CHANGELOG entry for removed user-visible commands

Nine `octo-cli mail` subcommands disappear and the attachment field renames, but `[Unreleased]` (CHANGELOG.md:8) gains no entry, despite the file being maintained per-feature. Consider a Removed/Changed entry.

- Idempotency contract: `--idempotency-key` → `X-Octo-Idempotency-Key` required on all three Agent write ops; backend enforces 8–200 printable-ASCII keys, exactly what the skill documents.
- Retry semantics: kept write ops remain `x-octo-retry: never`; the skill's `RESULT_UNKNOWN` guidance is consistent.
- Registry loader: embedded-spec load path, duplicate-operationId guard, and `octo-cli schema` consumers unaffected by the removals.
- Security direction: the PR strictly narrows the Agent surface, adds explicit fail-safe guidance (never retry owner-only endpoints, never ask for raw owner credentials, never claim a prepared Draft was sent), and introduces no new capability or endpoint.

## 数据流回溯

1. **附件 `content`**：用户 `--data` JSON → `resolveBody` (`cmd/service/run.go:528`) → `bodySchemaValidator` 递归校验 Attachment `required:[filename,contentType,content]` (`cmd/service/run.go:627`) → HTTP body → octo-mail `decodeStrict` → `sendRequest.Attachments` → `compose()` 里 `base64.StdEncoding.DecodeString(at.Content)`。字段名两端一致，数据真实流到消费点（base 的 `data` 字段则会被后端忽略 → 空附件，本 PR 修复此断点）。
2. **被删操作**：`mail.json`（embed）→ `registry.New()` 解析 → 命令树按 operation 注册；被删 id 无 entry → cobra 层即 unknown command。`TestMailRegistryExcludesOwnerOnlyOperationsAndDeletedConfirmationFlow` 断言缺失 + 全量 mail 操作无 `X-Octo-Confirmation`/`confirmation-token` 残留，与代码路径一致。
3. **send-intent 结果**：backend 202/201 响应体（见 Verification）→ CLI 按 `AgentSendIntentResult`/`AgentDraftResult` schema 呈现 → SKILL.md 分支指导与真实 outcome 匹配（唯一例外是上文 P2 的 review-required 409 形态）。
4. **idempotency key**：flag → header → backend `validOutboundIdempotencyKey`（8–200 可见 ASCII），与 SKILL.md 描述一致。

## 盲点 Checklist（C1–C6）

- **C1 双路径 parity**：clear。只删不增，无 add/remove 不对称；grep 确认被删操作在 spec/tests/docs/CLAUDE.md 两端同步移除，无悬空引用。
- **C2 control-flow ordering / 复用**：clear。命令生成与 body 校验为全 domain 复用逻辑，本 PR 未改其执行顺序；`content` 改名只经 schema 数据影响 mail 写操作，验证器递归数组项的行为已确认。
- **C3 授权边界**：clear（本 PR 核心）。CLI 暴露面与 octo-mail 的 Agent 授权边界逐端点核对一致；`X-Octo-Automation` 服务端自动化通道（auto-reply / owner-confirmed-draft）在 base 与 head 的 CLI 均未暴露，无新增能力泄漏。
- **C4 授权生命周期 / 容器级联**：N/A。CLI 无容器-成员状态查询路径；Agent binding 撤销（backend DELETE /agent-mailboxes/{id}/binding）不在本 PR 范围且本就未暴露。
- **C5 build/note ≠ 运行期**：未在本环境执行 build/test（按审查纪律）；已对 registry→命令注册→schema 校验→HTTP→backend handler 全链做静态运行期推演，并逐行核对 backend 源码，不以"作者称测试通过"作缓解。
- **C6 治理/文档自洽**：clear（除上述 P2 与两条 nit）。SKILL.md ↔ mail.json ↔ CLAUDE.md ↔ README 相互一致；安全边界表述（不重试 owner-only、不索凭证、不虚报已发送）自洽且方向 fail-safe。

## 跨轮 Blocker 复检（R6）

N/A — 本 PR 首轮审查，无既往 blocker。

## Verdict: APPROVED

收缩方向正确且逐端点核实：被删的九个操作在 Agent 凭证下本就被后端 403 拒绝（CLI 从未接线自动化 header，确认令牌流程在后端根本不存在），保留面全部 Agent 可达，附件字段改名与后端 wire contract 精确对齐，测试与文档同步更新、无悬空引用。唯一 P2 是 base 即存在的 review-required 文档漂移（fail-safe，不阻塞）。建议合入；P2/nit 可随后跟进。

[Octo-Q] verdict: APPROVE — 无 P0/P1；被删操作均为 Agent 凭证下已不可用的死面，附件 content 改名经 backend 源码核实为修复，仅 1 条 pre-existing P2 文档漂移 + 2 nit。

yujiawei:
## Follow-up: backend contract verified

In the review above I flagged one gap — that the two load-bearing premises (the nine endpoints being unusable with Agent credentials, and the attachment field being `content`) rested on the issue's assertion and could not be checked from this repo. I've since read the mail service source directly (`Mininglamp-OSS/octo-mail` @ `2b541d5`, read-only). **Both premises hold**, and my approval stands unchanged.

**Premise 1 — the nine removed operations really are Agent-unusable.** Seven of them route through `hAgentConfirmed` (`protocol/webapi/webapi.go:124,127,133,134,135,144,145`), which rejects any Agent credential that does not carry an `X-Octo-Automation` header:

```go
// protocol/webapi/confirmation.go:38
authorizationErr = errStatus(http.StatusForbidden, "owner_required", "this Agent Mail action must be completed by the human owner gateway")
```

The only accepted automation values are `auto-reply` and `owner-confirmed-draft` (`confirmation.go:14-17`), both server-authorized paths that the CLI has never wired. The remaining two, `createDraft` and `updateDraft`, are on plain routes but reject Agents inline (`drafts.go:87`, `:178-193`). So the removal takes away nothing that worked.

One near-miss worth recording, because it reads the other way at first glance: `deleteDraft` contains a branch that *permits* an Agent to delete a server-prepared Agent Draft —

```go
// protocol/webapi/drafts.go:444-445
if a.agentCredentialID > 0 && !hasAgentDraft {
    return errStatus(http.StatusForbidden, "owner_required", "an Agent may delete only a server-prepared Agent Draft")
}
```

— which would make `mail draft delete` a genuine Agent capability rather than dead surface. It isn't reachable: `hAgentConfirmed` returns 403 before the handler runs. No capability is lost, but the inner allowance and the outer gate disagree in spirit, so it's worth knowing the outer gate is what's binding.

Also note the `X-Octo-Confirmation` header this PR deletes has **no reader in the mail service's production code at all** — it survives only in that service's own test helpers. The flow the CLI was documenting was never implemented on the other side, which makes removing it straightforwardly correct rather than merely a scope reduction.

**Premise 2 — the attachment rename is a fix, not a cosmetic change.**

```go
// protocol/webapi/compose.go:20-24
type attachment struct {
	Filename    string `json:"filename"`
	ContentType string `json:"contentType,omitempty"`
	Content     string `json:"content"` // base64
}
```

The previous `data` field would have been silently dropped by the strict decoder, i.e. attachments were being lost, not rejected. Good catch in this PR.

---

### One additional P2 the backend read surfaced (pre-existing, not introduced here)

**`owner_review_required` is documented and declared, but is not reachable as a send-intent outcome.**

`skills/octo-mail/SKILL.md:179` tells agents to interpret `owner_review_required` as a structured outcome, and it is declared in the enum at `internal/registry/specs/mail.json:295`. In the mail service, that string is an *internal* policy-decision constant (`mailflow/outboundpolicy/policy.go:17`), and the web layer converts that decision into an HTTP **409** carrying a different code:

```go
// protocol/webapi/outbound_policy.go:85-88
return &statusError{
	status: http.StatusConflict,
	code:   "outbound_review_required",
	msg:    "email was not sent; owner review is required",
	policy: info,
}
```

Tracing `createAgentSendIntent` end to end, the only outcomes that can appear in a 2xx body are `accepted` (`agent_drafts.go:132`) and `owner_confirmation_required` (`agent_drafts.go:417`). An agent that follows the skill will branch on an outcome it can never observe, and will instead receive an error response it has no guidance for.

This is fail-safe — nothing is sent, and the 409 message says so explicitly — and the drift predates this PR (the enum is untouched here; only the bullet's wording changed). So it does not affect the verdict. Fix direction when convenient: document the 409 `outbound_review_required` shape in the skill so agents branch on the real wire signal, and either drop the unreachable enum value or annotate it as a policy-internal state.

**Verdict unchanged: APPROVED.**

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #133 (octo-cli)

## Summary

This PR removes the nine owner-only operations from the Agent Mail CLI surface — `mail.message.send`, `reply`, `reply-all`, `forward`, `delete`, and `mail.draft.create`, `update`, `send`, `delete` — together with the Agent self-confirmation token flow that gated them (`X-Octo-Confirmation` header, `--confirmation-token` flag, `omc_` tokens). The registry spec (`internal/registry/specs/mail.json`) drops from 24 operations to 15; because the CLI is fully metadata-driven (every command is generated from the embedded spec), deleting the operations deletes the commands. The octo-mail skill and CLAUDE.md are re-aligned to the remaining owner-sanctioned paths: `send-intent`, `draft create-agent`, and human completion in OCTO Web. The PR also renames the Attachment schema field `data` → `content` to match the backend payload. A clean, well-scoped authorization-alignment change; no P0/P1 issues found.

## Verification

Local checks at head `bd04b949`:
- `go test ./cmd/service/ ./internal/registry/ ./skills/ ./cmd/` — all pass (includes the rewritten surface tests)
- Operation count: `mail.json` carries exactly 15 operationIds, matching `internal/registry/loader_test.go:51`; base had 24 and the 9 removed are exactly the owner-only set asserted in `cmd/service/mail_test.go:143`
- Every `$ref` in `internal/registry/specs/mail.json` resolves; the deleted `SendDraft` / `UpdateDraft` / `AcceptedWrite` schemas leave no dangling references
- Repo-wide grep: no production references to the removed operation IDs, `X-Octo-Confirmation`, `confirmation-token`, or `omc_` — the only hits are tests asserting their absence
- `cmd/service/service.go:28` (`RegisterServiceCommands`) builds every mail command from the registry, and `rejectUnknownSubcommand` makes removed verbs fail loudly, so scripted calls to deleted commands cannot exit 0 silently
- CLAUDE.md mail tree cross-checked against hand-wired commands: `mail auth` (`cmd/mail_auth.go:57`) and `message state|changes` (`cmd/mail_jmap.go:56`) really exist, so the docs surface is accurate
- `skills/octo-mail/SKILL.md` contains all strings required by `skills/skills_test.go` and none of the forbidden ones; both outcome enums (`owner_confirmation_required` at `skills/octo-mail/SKILL.md:177`, `owner_review_required` at `:179`) are documented

## Findings

No P0/P1 issues; two nits below.

### Nit — attachment payload shape undocumented (`skills/octo-mail/SKILL.md:182`)

The skill tells agents to pass complex messages/attachments through `--data @file.json` but never shows the attachment object shape — which this PR just changed (`data` → `content` in the Attachment schema). Agents must discover it via `octo-cli schema mail.message.send_intent`. One short JSON example (`{"filename": "...", "contentType": "...", "content": "<base64>"}`) would stop agents from guessing the old `data` field. Non-blocking.

### Nit — substring tripwire in skill content test (`skills/skills_test.go:227`)

The forbidden entry `"confirmation_required without"` overlaps with the required string `owner_confirmation_required`; any future legitimate sentence such as "owner_confirmation_required without …" would fail `TestOctoMailSkillEmbeddedAndSafe`. Consider forbidding the exact stale sentence (or adding surrounding context) rather than the bare fragment. Non-blocking.

## Human-verify

1. The octo-mail backend is not part of this repo (octo-server only proxies owner-session traffic through `/v1/mail-gateway`), so the actual authorization boundary could not be verified from here. Please confirm that Agent credentials (`omb_…`) are rejected server-side on the removed owner-only endpoints (message send/reply/reply-all/forward/delete and draft create/update/send/delete). If the backend still honors them, this PR is defense-in-depth and prompt hygiene rather than hard capability removal — worth a follow-up server-side issue. **Not a merge blocker for this PR.**
2. Confirm the backend expects `attachments[].content` (base64) in `send-intent` / `draft create-agent` payloads. The renamed spec and the new runtime-path test agree, but the backend is the source of truth. **Not a merge blocker.**

- Removed commands fail loudly: `rejectUnknownSubcommand` turns scripted calls to removed verbs into errors instead of silent exit 0.
- The `octo-cli api` escape hatch cannot carry mail credentials: `--service` only routes `matters | dmworkim` (`cmd/api.go:114`), and `omb_` credentials are wired exclusively through the mail factory path — the removed surface is not reachable through the generic api command.
- `TestMailSendIntentAttachmentUsesBackendContentField` exercises the real production path (cobra root + real registry + httptest capture of the outgoing body), not hand-fed data.
- `mail.message.flag` stays a write without `x-octo-retry: never`; keyword add/remove is idempotent so retry is safe, and the skill's retry wording ("Send intents disable transport retries") now matches the spec exactly.
- Draft lifecycle asymmetry (Agent can create but not edit/send/delete) is intentional and documented; owners manage drafts in OCTO Web.

## Verdict: APPROVED

The PR consistently removes the owner-only surface across spec, tests, skill docs, and repo docs. The registry-driven CLI guarantees the removed operations disappear at runtime with loud failure, and all changed-package tests pass. The two nits and two human-verify items are non-blocking.

---

## 补充核验记录（内部，供终审复核）

**数据流回溯**（每个被消费数据 → 上游来源 → 是否真流到消费点）：
- registry operations → cobra 命令树：`RegisterServiceCommands` 逐 operation 生成（`cmd/service/service.go:35-44`）；spec 删除 = 命令删除，测试经真实 cobra root 执行验证。✅
- `attachments[].content` → HTTP body：`--data` 原样透传（无字段翻译层，prod 代码无任何 `"data"` attachment marshaling）；新测试在 httptest 端直接解码真实请求体断言 `content`。✅
- loader 计数 15：逐条枚举新 `mail.json` operationId = 15，与 `loader_test.go:51` 一致；base = 24，差集恰为 9 个 owner-only。✅
- `skills_test.go` 期望串 ↔ SKILL.md 实际内容：逐条 grep 验证 + `go test ./skills/` 通过。✅

**盲点 checklist**：
- C1 双路径 parity：clear。create-agent↔delete 的不对称是设计（owner 在 Web 完成 delete），SKILL.md 明示；flag add/remove 同一 operation 内对称；registry 是命令唯一注册路径（aliases 仅 matter），无旁路注册等价写法。
- C2 control-flow/复用：clear。loader 通用、无顺序依赖；confirmation flow 为纯 spec 删除，无残留代码路径。
- C3 授权边界≠能力边界：hit→已处置。CLI 侧能力经凭证模型封死（`api` 命令不带 omb_ 凭证）；后端边界不可验证 → Human-verify #1。
- C4 授权生命周期/级联：N/A，无鉴权中间件改动。
- C5 build/note≠运行期：clear。未信文档声明——实跑 4 个受影响包测试并核对命令生成机制。
- C6 治理/文档自洽：clear。CLAUDE.md ↔ spec ↔ SKILL.md ↔ skills_test 四方交叉一致。

**R3 分布式副作用**：N/A — 无通知/推送/计数/emit 改动；被删命令 fail-loud 属预期对齐。

**R6 跨轮 blocker 复检**：N/A（本 issue 首轮审查，无历史 comment，无既往 blocker）。

**R2 diff-scope**：两个 nit 均非阻塞——attachment shape 文档缺口属 pre-existing（旧流程同样无示例，仅字段改名使其更显眼）；tripwire 为本 PR 新增但仅 test-only。

[Octo-Q] verdict: APPROVE — 无 P0/P1。owner-only 命令面在 spec/测试/skill/repo 文档四处一致移除，registry 驱动机制保证运行时命令真删除且 fail-loud，受影响包测试全绿；唯后端 owner 授权强制与 content 字段契约无法在本仓验证，已列 Human-verify，不阻塞合入。

yujiawei (approved):
# Code Review — PR #133 (octo-cli)

Reviewed at head `bd04b949fe4b8b2091a3d5afe8b335f7f7d52ba5` against merge-base `0134bcb` (6 files, +106/−333).

Note: the three earlier approvals on this PR were submitted against head `7a862c1`, which the branch has since been force-pushed past — that object is no longer reachable, so those approvals cannot be shown to apply to the current tree. Everything below is an independent pass over `bd04b949`; nothing is carried over.

---

## 1. Spec compliance

**Spec: ✅**

Checked line-by-line against the acceptance criteria in #132.

| Criterion (#132) | Status | Evidence |
|---|---|---|
| Owner-only operations absent from `octo-cli mail` | ✅ | All nine (`mail.message.send/.reply/.reply_all/.forward/.delete`, `mail.draft.create/.update/.send/.delete`) gone from `internal/registry/specs/mail.json`; asserted absent in `cmd/service/mail_test.go:140-155`; op count 24→15 in `internal/registry/loader_test.go:51`. I counted the surviving `operationId`s independently: 15. |
| Agent send-intent / create-agent / reply-draft remain | ✅ | `mail.json:147` (`reply_draft`), `:181` (`send_intent`), `:207` (`draft.create_agent`) — all still `x-octo-retry: never` with a required `X-Octo-Idempotency-Key`. |
| Skill directs owner-only actions to OCTO Web | ✅ | `skills/octo-mail/SKILL.md:44-48`, `:177-180`, `:199-203`, `:213-214`, `:224-226`. |
| Attachment send-intent uses backend `content` | ✅ | `mail.json:246-253`. See the mutation check below — this one is genuinely covered, not just asserted. |
| `go test` / `vet` / `build` / `golangci-lint` pass | ✅ | I re-ran all four locally at this SHA, plus `go test -race ./cmd/service/`. All green; `golangci-lint run ./...` → `0 issues`. |

**Over-build: none.** The diff touches exactly six files, all in the mail surface. `SendDraft`, `UpdateDraft` and `AcceptedWrite` are removed alongside their only consumers — I grepped the whole repo and found no dangling `$ref`, no orphan schema, and no residual reference to a removed operationId, to `--confirmation-token`/`X-Octo-Confirmation`, or to the old Attachment `data` field. The declared out-of-scope items (backend authorization, OCTO Web, credential storage) are untouched.

**Deviation: none.** Two things I specifically tried to break and could not:

- **The `data` → `content` rename is actually covered by a test.** `TestMailSendIntentAttachmentUsesBackendContentField` passes the payload through `--data`, which at first read looks like a raw passthrough that would make the assertion tautological. It isn't: `bodySchemaValidator` in `cmd/service/run.go:716-968` walks `--data` against the request schema and enforces `required` at every nesting depth. I confirmed by mutation — reverting `mail.json` to `required: [filename, contentType, data]` and re-running the test:

  ```
  --- FAIL: TestMailSendIntentAttachmentUsesBackendContentField
      execute attachment send intent: VALIDATION_ERROR: request body is
      missing required field(s): attachments[0].data
  ```

  The test fails on the old schema and passes on the new one. Real coverage.

- **No generated-flag collision from the rename.** `attachments` is an array of objects and gets no promoted flag, so `content` never reaches the cobra flag set. `octo-cli mail message send-intent --help` registers only `--to/--cc/--bcc/--subject/--text/--html/--data/--idempotency-key`. `octo-cli schema mail.message.send_intent` does surface the nested `content` field, so the shape stays discoverable at runtime.

One architectural point in the PR's favour that's worth stating explicitly, since "why remove it client-side when the server already rejects it?" is the obvious objection: `internal/authstore/authstore.go:180-182` documents that a mailbox authorization belongs to the Bot credential and Space. There is no owner-credentialed path into this command tree, so these operations were unreachable for *every* caller of `octo-cli mail`, not merely discouraged. Removing them is the correct layer, not defence-in-depth theatre.

I also checked the obvious bypass — the generic `octo-cli api` passthrough — and it cannot reach the deleted endpoints, though not for the reason one might assume. The `--service` override is irrelevant here: it defaults to `OCTO_API_BASE_URL`, which is exactly the base URL the mail spec declares, so `octo-cli api POST /agent-mail-api/webapi/v0/messages` does resolve to the right host. What stops it is the credential dispatch. `cmd/api.go:91` calls `f.Client()` (the active Bot credential), whereas every mail operation goes through `f.ClientForCredential(ctx, "mail")` at `cmd/service/run.go:1066`, and `internal/cmdutil/factory.go:426-436` shows `"mail"` is the only key that returns the mailbox transport. The passthrough would therefore present a Bot token to the mail gateway, not the mailbox credential.

---

## 2. Code quality

**Quality: Approved** — no P0/P1. The items below are P2 / nits and none should block merge.

### P2-1 — `t.Fatalf` called from the httptest handler goroutine

`cmd/service/mail_test.go:223` and `:226`:

```go
mailServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
        t.Fatalf("decode body: %v", err)     // <- server goroutine
    }
    if len(body.Attachments) != 1 {
        t.Fatalf("attachments = %#v", body.Attachments)
    }
```

`t.Fatal*` may only be called from the goroutine running the test function. From a handler it calls `runtime.Goexit` on the server goroutine: the response is never written, the test does not stop, and the real failure surfaces later as an opaque client-side EOF instead of your message. The sibling test right below (`TestMailAttachmentDownloadWritesOutput:262-266`) already uses `t.Errorf` for exactly this reason, and repo-wide this is the only other occurrence besides `cmd/mail_auth_test.go:474`. Suggest `t.Errorf` + `w.WriteHeader(http.StatusBadRequest); return`, and assert on the captured value in the test body where you already do.

### P2-2 — `TestMailRegistrySurface` is a subset assertion, not an exact-set one

`cmd/service/mail_test.go:70-84` pins 13 operations, but the registry has 15 — `mail.mailbox.list` and `mail.message.flag` are unpinned. The count check in `loader_test.go:51` catches a net change in size but not a swap: dropping `mail.mailbox.list` while adding something else keeps both tests green. Since pinning the command surface is the whole point of this PR, consider asserting set equality against `reg.ListOperations("mail")` and deleting the separate count constant.

### P2-3 — Skill regression guard covers 2 of the 9 removed commands

`skills/skills_test.go:224-232` forbids `octo-cli mail message delete` and `octo-cli mail draft send`. Documentation that reintroduced `message send`, `reply`, `reply-all`, `forward`, or `draft create`/`update`/`delete` would pass. A stronger and cheaper invariant: extract every `octo-cli mail …` command string from `SKILL.md` and assert each resolves to a registered operation (or to one of the hand-written `mail message state`/`changes`/`auth …` commands). That catches all nine removals plus any future drift, without maintaining a denylist. I ran that check ad hoc against this head and all 17 documented commands resolve, so the skill is currently in sync — the test just doesn't hold it there.

### P2-4 — The confirmation rule became conditional on something the agent cannot evaluate

`SKILL.md:31-33`:

```
- Before a send intent that may immediately deliver mail, show the exact
  recipients, subject, content, and attachments and obtain user confirmation.
```

The previous wording was unconditional. Whether a send intent delivers immediately is decided server-side by the mailbox's outbound mode, and nothing on the CLI surface exposes that mode — so "that may immediately deliver mail" is a predicate the agent cannot evaluate before acting, which invites skipping confirmation whenever it assumes manual mode. The prose at `:36-41` still states the rule unconditionally ("do not call it until the user has explicitly approved the exact recipients, subject, body, and attachments"), so this is an internal inconsistency rather than an open hole — but in a security-boundary bullet list the bullet is what gets followed. Suggest dropping the qualifier.

### P2-5 — No CHANGELOG entry for nine removed public commands

`CHANGELOG.md:3` states "All notable changes to `octo-cli` are documented in this file" and follows Keep a Changelog, but `[Unreleased]` is untouched.

Weighing this honestly: the convention is not strictly enforced here — no `fix(…)` commit in the last 25 touched `CHANGELOG.md`, `CONTRIBUTING.md` imposes no requirement, and the mail domain has never appeared in the changelog at all (#129, which introduced this entire surface, also skipped it). So this is advisory, not a policy violation. It is still worth an entry, because this is the first *removal* rather than an addition, and a `### Removed` line is the only place a user scripting `octo-cli mail message send` would find out why it stopped working.

### Nits

- **Migration UX for removed commands.** `octo-cli mail draft send D1` gives a good error (`unknown subcommand for "octo-cli mail draft"; available: create-agent, list`), but the realistic invocation `octo-cli mail message send --to a@b.c` degrades to `a flag in the command line was not recognised`, because cobra parses `send` as a positional of the group. That is pre-existing engine behaviour rather than something this PR introduced, but this PR is what makes users hit it. A hidden stub command per removed verb, erroring with a pointer to `send-intent` / OCTO Web, would turn nine confusing failures into nine useful ones.
- **`SKILL.md:182`** tells the agent to pass attachments via `--data @file.json` but never shows the attachment object shape — the exact thing this PR renames. Discoverable via `octo-cli schema mail.message.send_intent`, so this is minor, but a three-line `{"filename","contentType","content"}` example in the skill would remove the guesswork.
- **Brittle substring guard.** `skills/skills_test.go:227` forbids the bare fragment `confirmation_required without`, which is a substring of `owner_confirmation_required without …`. Since `owner_confirmation_required` is simultaneously a *required* string (`:212`) and is discussed at `SKILL.md:177`, a perfectly legitimate future sentence like "do not report `owner_confirmation_required` without showing the Draft id" would fail `TestOctoMailSkillEmbeddedAndSafe`. Confirmed by substring check. Forbid the full stale sentence, or anchor on `` `confirmation_required` without performing ``, rather than the bare fragment.
- **Diff churn.** `TestMailAttachmentDownloadWritesOutput` is moved below the new test without being modified, which inflates the diff for no reason. Harmless, just noisier to read.
- **Nothing can clean up a bad Agent draft.** `reply-draft` / `create-agent` can create server-side drafts that the CLI can no longer delete. That is a direct and intended consequence of the authorization model in #132, and `X-Octo-Idempotency-Key` is required on both, so duplicates are bounded — noting it only so the follow-up cost is visible if owners start complaining about draft clutter.

---

## 3. Verdict

**APPROVED.** The change does exactly what #132 asked, nothing more, and the one claim that looked unverifiable from the diff (the attachment field rename) holds up under a mutation check. The findings above are all P2 or nits; I'd take P2-1 and P2-4 in a follow-up, and the rest at leisure.

---

## 4. What I did not verify

- **The backend contract itself.** I cannot confirm from this repo that `content` is octo-mail's actual attachment field name — there is no linked backend commit and no contract test spanning the two services. If it's wrong, every attachment send fails server-side with no client-side signal, and the CLI's own tests will stay green. A link to the octo-mail definition in the PR description would close this.
- **Live behaviour against a real Agent mailbox.** All tests here are `httptest` fakes; no end-to-end run against a real mailbox, and no verification that OCTO Web actually offers the owner draft edit/send/delete workflow the skill now redirects to.
- **External consumers.** No data on whether anyone scripts the removed commands; the removal is immediate with no deprecation window.
- **Generated artifacts outside the repo** (shell completions, published npm package contents) — `npm/` contains no checked-in command list, so nothing there needs updating, but I did not inspect a built release artifact.

Jerry-Xin (approved):
Re-review verified at head bd04b949fe4b — approving.

This PR aligns the mail Agent command set with owner authorization:

- The CLI command tree is fully registry-driven (service.go iterates ListOperations); no hardcoded handlers or generic dispatch escape hatch. The 9 removed owner-only ops (send, reply, reply_all, forward, delete, draft create/update/send/delete) are unreachable from the CLI.
- All 15 remaining ops audited: 11 read-only, flag (body schema restricted to addKeywords/removeKeywords, unknown fields rejected), send_intent (retry: never + idempotency key), reply_draft, create_agent, draft.list. No X-Octo-Confirmation residue in the repo.
- Credentials stay account-scoped Agent Mail credentials with private transport conversion; owner enforcement remains on the octo-mail server side.
- Fail-closed guards intact: empty path values, dot segments, empty required fields, unknown fields, and unknown subcommands are all rejected.
- go vet + go test fully green at this head.

LGTM.

PR #61, opened 2026-06-25
fix(release): clean build artifact after publish

liuooo:
## What
The release workflow now builds GoReleaser output once and passes it through the
run-scoped `goreleaser-dist` Actions artifact. The `release-upload` job has no
checkout, so `gh release upload` cannot infer the repository from `.git`. The
same temporary artifact was also retained for GitHub's minimum retention window
after both consumers had already finished.

## How
- **.github/workflows/release-publish.yml** — pass `--repo "$GITHUB_REPOSITORY"`
  to `gh release upload` so Release asset upload works without a checkout.
- **.github/workflows/release-publish.yml** — add `cleanup-dist-artifact`, which
  waits for `release-upload` and `publish-npm`, then deletes the run-scoped
  `goreleaser-dist` artifact through the Actions artifact API.
- **.github/workflows/release-publish.yml** — document that `retention-days: 1`
  is GitHub's minimum and the cleanup job is what makes the artifact temporary.

## Tests
- `actionlint .github/workflows/release-publish.yml`
- `git diff --check`

Closes #60

PR #134, opened 2026-08-18
feat(docs): add the docs invite command family

### LINKED ISSUE #130: feat(docs): 给 bot 增加文档邀请链接命令族（docs invite create/accept，对齐 drive invite）
## 背景

CLI 的 `docs` 域目前只有 `create | list | search | get | rename | delete | forward-grant`，**没有文档邀请链接相关命令**。而服务端 bot 侧接口已存在（docs-backend #61 加了 bot 专用接受口 `POST /v1/bot/docs/invites/:token/accept`），bot 缺的只是 CLI 封装。

对照参照：`drive` 域已经有完整的 invite 四件套（`invite create / list / revoke / accept`），docs 域应当对齐。

## 需求

给 bot（以及 user token 场景）在 `docs` 域补邀请链接命令族，至少覆盖两个核心场景：

1. **bot 生成文档邀请链接**：对指定文档创建 invite（可选 role / 有效期），返回 invite token/链接，便于把文档分享给群成员或其他 bot。
2. **bot 通过邀请链接加入文档**：`docs invite accept <invite-token>`，bot 凭自身 token 接受邀请、按链接预设角色成为文档成员（服务端 `POST /v1/bot/docs/invites/:token/accept`）。

## 参考实现

- `drive invite create/list/revoke/accept`（design.md 91–94 号命令）的命名、role 校验、token 脱敏（share/invite token 在 --verbose/--dry-run 里 mask）都可以直接对齐。
- 权限矩阵与 role 枚举复用 docs 现有角色体系（owner/admin/editor/commenter/reader）。

## 验收建议

- bot 身份下 `docs invite create <doc> --role <role>` 能产出可分享链接。
- bot 身份下 `docs invite accept <token>` 成功后，`docs get` 可见该文档、角色与链接预设一致。
- invite token 不出现在 trace/dry-run 明文里（沿用现有 secret-masking 机制）。

## 安全提醒（写进实现时注意）

接受邀请接口目前只校验 token 有效性（历史调研发现 accept 事务无 space 校验，可能留下跨 space 脏成员行，见 octo-docs-backend 相关讨论）——CLI 侧无法修服务端，但 issue 里先标注，实现时建议顺带跟进服务端校验。

Ranwanglc:
## Summary

Adds the `docs invite` command family so a bot can hand out a document invite link, and — the half that was actually missing — **join a document by accepting one**:

```
octo-cli docs invite create <docId> [--role writer] [--expiresInDays 3] [--maxUses 0]
octo-cli docs invite list   <docId>
octo-cli docs invite revoke <docId> --invite-token <token>
octo-cli docs invite accept <token>            # or the whole invite link
```

All four are declared in `internal/registry/specs/docs.json` and generated by the existing registry engine. The docs backend's bot-side invite endpoints (`invitesRouter` on the bot chain, plus the bot accept route from docs-backend #61) were already live, so **this PR changes no backend and no frontend** — the gap was purely the CLI wrapper.

`invite accept` also takes the whole web invite link (`{octo web origin}/docs/invite/<token>`), because that is what a person actually has in hand.

## Linked Spec

Fixes #130. The issue's own sketch is followed except where the backend contradicted it — see COMPREHENSION (a).

## How verified

- `go build ./...`, `go vet ./...`, `gofmt -l ./cmd ./internal`, `go test ./... -count=1` — all clean.
- End-to-end against a live deployment with two bot credentials: bot A `invite create` → `invite list` → bot B `invite accept <full link>` → bot B's `docs get` shows the granted role → bot A's `members list` shows the new row with `source=invite` → repeat accept is still 200 (idempotent) → `invite revoke` → `list` empty → accept now `410 invite_invalid`. Repeated for `--role commenter`.
- Negative paths, locally, no request sent: a link on a foreign host is `INVALID_INVITE_URL`; `--role bogus` is `ENUM_NOT_ALLOWED`; `--dry-run` renders the token as `***REDACTED***`.
- Mutation-tested the new guards (each change made the intended test fail, then reverted): turning the missing-leaf `panic` back into a silent `return`; narrowing the role enum back to three values; renaming the `x-octo-flag` to `invite-id`; restoring the fail-open `return nil` in the flag read.

## COMPREHENSION

**(a) What this does to the load-bearing path (before → after).**

Before, `docs` had no invite surface at all — 32 operations, none of them invite — so a bot could only be added to a document by an admin who already knew its uid (`docs members set`). There was no way for a bot to join a document *itself*, which is the whole point of a link invite. After, four operations exist and the accept path produces a `doc_member` row with `source=invite` at the invite's role.

Two things about the shape are load-bearing and are **not** copies of `drive invite`:

1. **Role vocabulary.** `reader | commenter | writer | admin` — the full docs member ladder. An earlier revision of this branch pinned `reader|writer|admin` and documented "an unknown value is silently downgraded to writer". Both were wrong: they were transcribed from a local backend checkout ~86 commits behind `origin/main`. On `origin/main`, `parseRole` (`src/api/routes/invites.ts`) accepts all four and answers **400** for anything else; only an *omitted* role takes the `writer` default. A narrow enum here would be worse than useless, because the CLI's enum gate is a hard local exit 2 — it would have refused a call the backend accepts, with no way through but `octo-cli api`. Note `commenter` ranks *between* reader and writer (`roleRank` 10/20/30/40) even though its stored code is `4`; the number is historical, the rank is not.
2. **Revoke is addressed by the token**, not an id — the docs backend stores no separate invite id, unlike drive. So the flag is `--invite-token`, and following the drive recipe would hand the token to a flag that does not exist. The lifetime field is likewise `expiresInDays` (1–7, clamped silently) and not drive's `expires_in_seconds`.

The link-accepting behaviour is a *value* normalisation on the `inviteToken` slot only: the generated leaf still builds and sends the request, so path, mount, risk annotation and the secret list all keep coming from the spec.

**(b) What could break, and how.**

- **`drive share access` / `download` / `create`** — the real blast radius. `assertSameOrigin` and `assertShareIDSegment` gained a `linkPolicy` parameter so the invite surface could share them instead of holding a second copy that a future hardening would miss. The rule bodies are unchanged and `shareLinkPolicy` reproduces the previous strings exactly, so drive keeps `INVALID_SHARE_URL` and its wording; the existing share tests cover this and still pass.
- **Invite tokens are credential-equivalent.** Both token-addressed operations put the value in the URL path, and `--dry-run`/`--verbose` print paths. Both parameters are declared `x-octo-secret`, and every rejection path in the link parser is written not to echo the input — the last path segment of a nearly-correct link *is* the token. `invite create`'s success envelope necessarily contains the token in full; that output is not safe to paste anywhere.
- **The wrapper detaching silently.** `registerDocsInviteAcceptURL` finds its target by walking `docs → invite → accept`. It used to return quietly if any step came up empty, which made link support silently optional: rename the operationId or change the flag and a pasted link would be sent to the backend verbatim, coming back as a confusing 410. It now derives the expectation from the registry — spec doesn't declare the operation ⇒ nothing to wrap, return; spec declares it but the tree lacks the leaf / a runnable `RunE` / a string `--invite-token` ⇒ `panic`, the same policy `registry.MustNew` applies to embedded data. This is a build-time invariant: the command tree is built from the registry alone (not from the credential), and `withholdDisabledServices` runs *after* registration, so disabling `docs` removes the service after the wrapper has already attached. It cannot be provoked by user input, configuration or the network, and every test that builds a root command runs into it.
- **Reading a stale backend checkout.** The failure mode that produced the wrong enum in the first place. The `enum_vocab_test.go` entry now records the remote authority, per that file's own rule.

**(c) How I know it works.**

The end-to-end run in "How verified" is against a live deployment, not a mock — including the `source=invite` member row, idempotent re-accept, and the `410` after revoke. `commenter` was exercised specifically because it is the value the earlier revision wrongly excluded. The guards are mutation-tested, so each one is known to fail when the thing it protects regresses, rather than merely being present.

## Deployment prerequisites

None. No new environment variable, no migration, no backend or frontend change. `OCTO_API_BASE_URL` must be the Octo origin (already required by every other command) because the invite-link parser compares the link's host against it; passing a bare token does not need it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **`octo-cli docs invite` family** (4 commands) — link invites for an online
+  document, generated from `docs.json`: `invite create` mints a token granting
+  `reader` | `commenter` | `writer` | `admin` (default `writer`) for 1–7 days
+  (`--expiresInDays`, default 3) with an optional accept cap (`--maxUses`,
+  `0` = unlimited),
+  `invite list` shows a doc's live invites with their usage counters,
+  `invite revoke` kills one, and `invite accept` claims one for the credential in
+  use. Three details differ from the `drive invite` family and are pinned by
+  tests: the role vocabulary is the full docs member ladder
+  `reader|commenter|writer|admin` (docs-backend's invite `parseRole` accepts all
+  four — only an *omitted* role takes the `writer` default, and an unrecognised
+  one is a `400`, never a downgrade), the lifetime is `expiresInDays` and not
+  `expires_in_seconds`, and **revoke is addressed by the invite token**
+  (`--invite-token`) because the docs backend stores no separate invite id.
+  Accept is idempotent — repeating it, or accepting into a document you already
+  belong to, is still `200` with the resulting role and never downgrades a higher
+  role already held; an unknown, revoked, expired or exhausted token is
+  `410 invite_invalid`.
+  - **`invite accept` also takes the whole invite link.** The web app serves
+    invites at `{octo web origin}/docs/invite/<token>`, so pasting that link
+    works. It is parsed locally under the same strict same-origin rules as
+    `drive share access` and is **never fetched**: the host is only compared
+    against `OCTO_API_BASE_URL`, and a link on another host, with embedded
+    credentials, a downgraded scheme, a percent-encoded path or extra segments is
+    refused with `INVALID_INVITE_URL` (exit 2) before any request.
+  - **Invite tokens are credential-equivalent** and declared `x-octo-secret`, so
+    they are masked in `--dry-run` and `--verbose` traces and never echoed in a
+    rejection message. `invite create`'s success envelope necessarily carries the
+    token in full — it is the deliverable — so that output is not safe to paste
+    into a ticket.
 - **`octo-cli marketplace expert` / `squad` families** — CRUD, `mine` lists,
   taxonomy (`expert-category` / `expert-tag`), viewable `skillmd get`, presigned
   `expert-skill-upload create` / `skill-download` for the Expert Marketplace
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -84,6 +84,7 @@ octo-cli docs      create | list | search | get | rename | delete | forward-gran
                scene    get|edit|export
                members  list|set|remove
                share    get|set
+               invite   create|list|revoke|accept
                comments list|add|edit|delete
                versions list|create|state|rename|delete|restore
                attachments presign|get|resolve
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Key properties:
 
 | Domain    | Ops | Purpose                                                        |
 |-----------|-----|----------------------------------------------------------------|
-| `docs`    | 32  | Documents, spreadsheets & whiteboards — lifecycle, full-text search, body content, sheet cells (paged read), board scenes, members, comments, versions, attachments |
+| `docs`    | 36  | Documents, spreadsheets & whiteboards — lifecycle, full-text search, body content, sheet cells (paged read), board scenes, members, invite links, comments, versions, attachments |
 | `html`    | 20  | Interactive HTML documents (octo-doc, **separate backend** from `docs`) — publish immutable versions, drafts, per-doc share codes & per-uid grants, media assets, inline comments, agent element read/replace |
 | `drive`   | 42  | Network drive — spaces & members, folder/file tree, two-phase blob upload & signed download, online-document mounts, share links, invites, IM-attachment transfer. Plus 3 composite commands (`upload file`, `download file`, `share create`) for 45 leaves total |
 | `matter`  | 14  | Todos/tasks — **temporarily withheld** while the backend API stabilizes |
--- a/cmd/docs_inviteurl.go
+++ b/cmd/docs_inviteurl.go
@@ -1,0 +1,242 @@
+package cmd
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/Mininglamp-OSS/octo-cli/cmd/service"
+	"github.com/Mininglamp-OSS/octo-cli/internal/cmdutil"
+	"github.com/Mininglamp-OSS/octo-cli/internal/config"
+	"github.com/Mininglamp-OSS/octo-cli/internal/output"
+)
+
+// docsInvitePathPrefix is where the Octo web app serves a document invite link:
+// {octo web origin}/docs/invite/<inviteToken>.
+const docsInvitePathPrefix = "/docs/invite/"
+
+// inviteLinkPolicy is the invite-link spelling of the shared link rules: the
+// same enforcement as a share link, reported as INVALID_INVITE_URL with a hint
+// naming the two accepted forms.
+var inviteLinkPolicy = linkPolicy{noun: "invite link", fail: invalidInviteURL}
+
+// docsInviteAcceptOp is the operation this file wraps. It is named once, here,
+// because the wrapper's correctness depends on it matching the spec: see
+// registerDocsInviteAcceptURL for what happens when it stops matching.
+const docsInviteAcceptOp = "docs.invite.accept"
+
+// registerDocsInviteAcceptURL lets `docs invite accept` take the whole invite
+// link in place of the bare token.
+//
+// The leaf itself stays the generated one — flags, risk gate, secret masking and
+// `octo-cli schema docs.invite.accept` all keep coming from the spec. Only the
+// *value* of the inviteToken slot is normalised first, because that is the whole
+// difference: the backend endpoint is unchanged, and what a person actually has
+// in hand is the link the web app gave them, not the token buried in it.
+//
+// Wrapping rather than detaching is deliberate. A hand-written composite (the
+// shape `drive share access` needs, because it takes an argument the engine
+// cannot express at all) would have to re-derive the path, the mount, the risk
+// annotation and the secret list, and would drift from the spec the first time
+// one of them changed. Here the engine still builds and runs the request.
+//
+// # Why the shape is asserted rather than tolerated
+//
+// This function used to return quietly whenever it could not find what it was
+// looking for. That made the whole feature silently optional: rename the
+// operationId, regroup the leaf, or change the flag, and link support would
+// vanish with nothing failing — and the first symptom would be a user pasting a
+// link and getting a backend 410 for a "token" that is really a whole URL.
+//
+// So the expectation is derived from the registry instead of assumed. If the
+// embedded spec does not declare the operation at all, there is genuinely
+// nothing to wrap and returning is correct. If it *does* declare it, the command
+// tree must carry the matching leaf with a runnable RunE and a string
+// invite-token flag, and anything else is a wiring bug in embedded data — the
+// same class registry.MustNew panics on (internal/registry/loader.go:96, used by
+// cmdutil's default RegistryFunc on the stated grounds that "specs are embedded;
+// parse failure is a build-time bug"). It panics for the same reason: it cannot
+// be provoked by user input or by the network, only by a change in this repo,
+// and every test that builds a root command runs straight into it.
+func registerDocsInviteAcceptURL(root *cobra.Command, f *cmdutil.Factory) {
+	reg := f.Registry()
+	if reg == nil {
+		return
+	}
+	if _, declared := reg.GetOperation(docsInviteAcceptOp); !declared {
+		// The spec does not offer this operation (a fixture registry, or the
+		// operation withdrawn on purpose). Nothing to wrap, and nothing lost.
+		return
+	}
+
+	accept := docsInviteAcceptLeaf(root)
+	if accept == nil {
+		panic(docsInviteAcceptOp + " is declared in the embedded spec but `docs invite accept` is not in the " +
+			"command tree: the invite-link wrapper in cmd/docs_inviteurl.go has nothing to attach to, so pasting " +
+			"an invite link would send the whole URL to the backend as a token. Re-point the wrapper at the leaf " +
+			"the spec now produces.")
+	}
+	if accept.RunE == nil {
+		panic(docsInviteAcceptOp + " leaf has no RunE, so the invite-link wrapper would replace nothing " +
+			"(cmd/docs_inviteurl.go).")
+	}
+	if flag := accept.Flags().Lookup(docsInviteTokenFlag); flag == nil || flag.Value.Type() != "string" {
+		panic(docsInviteAcceptOp + " leaf has no string --" + docsInviteTokenFlag + " flag, which is the slot the " +
+			"invite-link wrapper rewrites (cmd/docs_inviteurl.go). Check x-octo-flag on the inviteToken path " +
+			"parameter in internal/registry/specs/docs.json.")
+	}
+
+	generated := accept.RunE
+	accept.RunE = func(cmd *cobra.Command, args []string) error {
+		if err := normalizeInviteTokenArg(cmd, f, args); err != nil {
+			return failErr(f, err)
+		}
+		return generated(cmd, args)
+	}
+	accept.Long += docsInviteLinkHelp
+}
+
+// docsInviteAcceptLeaf walks root to the generated `docs invite accept` leaf, or
+// returns nil if any step of the path is missing. Split out so a test can assert
+// the wrapper landed on the same command a user reaches.
+func docsInviteAcceptLeaf(root *cobra.Command) *cobra.Command {
+	cur := root
+	for _, name := range []string{"docs", "invite", "accept"} {
+		if cur == nil {
+			return nil
+		}
+		cur = service.FindChild(cur, name)
+	}
+	return cur
+}
+
+// docsInviteTokenFlag is the flag form of the inviteToken path parameter
+// (x-octo-flag in docs.json). The wrapper rewrites this slot, so the name is a
+// contract between this file and the spec.
+const docsInviteTokenFlag = "invite-token"
+
+// docsInviteLinkHelp is the help text the wrapper appends. It is a package-level
+// constant so a test can assert it is present on the assembled leaf and absent
+// from a bare generated one — i.e. that the wrapper is what put it there.
+const docsInviteLinkHelp = "\n\n<invite-token> may also be given as the whole invite link " +
+	"(" + docsInvitePathPrefix + "<token> on the configured Octo origin). The link is " +
+	"parsed locally and never fetched: only the token is extracted, and the request goes " +
+	"to the configured Octo API."
+
+// normalizeInviteTokenArg rewrites the inviteToken value in place when the
+// caller supplied a link instead of a token. A bare token is left untouched, so
+// the ordinary path is byte-identical to the generated leaf's.
+func normalizeInviteTokenArg(cmd *cobra.Command, f *cmdutil.Factory, args []string) *output.ExitError {
+	if cmd.Flags().Changed(docsInviteTokenFlag) {
+		raw, err := cmd.Flags().GetString(docsInviteTokenFlag)
+		if err != nil {
+			// Fail closed. The caller set the flag, so a value exists and this
+			// code cannot read it; forwarding regardless would hand an
+			// unvalidated value — possibly a whole URL — straight to the
+			// backend, which is exactly the check this wrapper exists to make.
+			// registerDocsInviteAcceptURL already refuses to attach unless the
+			// flag is a string, so reaching here means the leaf changed shape
+			// under us at runtime.
+			return invalidInviteURL("the --" + docsInviteTokenFlag + " value could not be read for checking")
+		}
+		token, verr := inviteTokenFromValue(f, raw)
+		if verr != nil {
+			return verr
+		}
+		if token != raw {
+			if err := cmd.Flags().Set(docsInviteTokenFlag, token); err != nil {
+				return invalidInviteURL("the invite link could not be reduced to its token")
+			}
+		}
+		return nil
+	}
+	if len(args) == 0 {
+		return nil // the engine reports the missing-value error, naming both forms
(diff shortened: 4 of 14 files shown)


Jerry-Xin (approved):
Relevant, well-designed extension of octo-cli’s spec-driven `docs` commands.

✅ Highlights

- URL normalization preserves the generated command path and security controls (`cmd/docs_inviteurl.go:50`).
- Strict origin, path-segment, and credential checks prevent unsafe invite-link handling (`cmd/docs_inviteurl.go:205`).
- Invite tokens are correctly marked secret for path tracing and dry-run redaction (`internal/registry/specs/docs.json:914`).
- Registry contracts accurately model create, list, revoke, and accept behavior.
- Tests cover request shapes, role validation, redaction, foreign origins, wrapper wiring, and regressions.
- Full `go test ./... -count=1` passes.

No blocking correctness, security, performance, or architectural issues found.

yujiawei (approved):
# Code Review — PR #134 (octo-cli)

Reviewed at head `919d940`, merge-base `e627723`. Single commit, 14 files, +1121/−45.

Verified locally on the PR branch: `go build ./...`, `go vet ./...`, `gofmt -l ./cmd ./internal`, and `go test ./... -count=1` are all clean. I also drove the built binary against a local stub server to confirm the invite-link behaviour end to end rather than reading it off the tests.

---

## 1. Scope / spec compliance

**Verdict: pass.**

Issue #130 asked for at least `invite create` and `invite accept`, aligned with the existing `drive invite` four-command family. The PR delivers all four (`create` / `list` / `revoke` / `accept`), plus the link-accepting form. Nothing outside that scope is added:

- The generated tree grows by exactly 4 operations (`internal/registry/loader_test.go:47`, 32 → 36) and `cmd/service/docs_test.go:52` pins the subtree shape.
- The only change outside the invite surface is the `linkPolicy` parameterisation of `assertSameOrigin` / `assertShareIDSegment`, which is required to share the boundary rather than fork it, and a prose correction on `docs.forward-grant` (`internal/registry/specs/docs.json:797-798`) where the summary said `reader/writer` while the enum on line 812 already allowed `commenter`. That is a doc-vs-reality fix in the touched file, not new behaviour.

**Deviations from the issue text, both justified in the PR body:**

- Role vocabulary is `reader|commenter|writer|admin` rather than the ladder sketched in the issue. See "needs human verification" below — this is the one load-bearing claim I cannot check from this repository.
- `revoke` is addressed by `--invite-token` rather than an invite id, because the docs backend stores no separate id. `cmd/service/docs_test.go` asserts `--invite-id` does *not* exist, so a caller following the `drive` recipe fails loudly instead of silently passing the token to an unknown flag. Good call.

## 2. Code quality

**Verdict: approve.** No P0/P1 survived verification. Everything below is P2 / advisory.

### What holds up

The SSRF boundary is correct, and I checked it rather than trusting the comment:

- `parseInviteURL` (`cmd/docs_inviteurl.go:203`) only ever *compares* the link's origin; the request goes to the configured API. A foreign-host link is refused locally with no socket opened — confirmed live (`INVALID_INVITE_URL`, exit 2) and pinned by `cmd/drive_shareurl_test.go:394`.
- Sharing `assertSameOrigin` / `assertShareIDSegment` through `linkPolicy` instead of copying them is the right call, and the drive strings are reproduced exactly: `p.noun` is `"share link"` for `shareLinkPolicy`, so every previous message renders byte-identically. All existing share tests pass unchanged.
- `--dry-run` masks the token in the URL path — verified live: `.../invites/***REDACTED***/accept`. This works because `inviteToken` is an `x-octo-secret` *path parameter*, which `collectSecrets` (`cmd/service/run.go:290`) does read.
- The dangerous downstream shapes are already handled by the engine: `.`/`..` path values are refused (`cmd/service/run.go:273`), and supplying a path value both positionally and by flag is a hard error (`cmd/service/run.go:191-198`).
- Risk annotations match the `drive` invite family exactly (`create`=write, `list`=read, `revoke`=high-risk-write, `accept`=write) — I diffed them against `drive.json:820-885`.
- The tests are unusually good. `TestParseInviteURL_Rejects` mirrors the share case list so neither surface can quietly enforce less, and `TestNormalizeInviteTokenArg_FailsClosedOnAnUnreadableFlag` pins the deny side of a branch that used to fail open. These pin behaviour, not shape.

### P2 — bare token is classified trimmed but forwarded untrimmed

`cmd/docs_inviteurl.go:174-178`:

```go
trimmed := strings.TrimSpace(raw)
if !strings.Contains(trimmed, "/") {
    return raw, nil
}
```

The whitespace decision is made on `trimmed` but `raw` is returned. Reproduced against a stub server:

```
$ octo-cli docs invite accept "  abc123  "
→ POST /v1/bot/docs/invites/%20%20abc123%20%20/accept
```

The backend answers `410 invite_invalid`, which reads as "your invite is dead" for what is really a copy-paste artefact — the single most likely way a token arrives from a chat window. The link form *is* trimmed, so the two spellings disagree. `parseShareURL` trims at `cmd/drive_shareurl.go:62`; suggest `return trimmed, nil`, and rejecting whitespace-only input. `TestParseInviteURL_Accepts:231` pre-trims its own input, so it does not currently cover this.

### P2 — the "never echoes the token" invariant has a gap in the shared helper

`invalidInviteURL` is documented as never carrying the token, and `TestParseInviteURL_NeverEchoesTheToken` asserts it. But `assertSameOrigin` echoes the scheme and the host verbatim (`cmd/drive_shareurl.go:167`, `:176`). If the value lands in either position, it is printed:

```
$ octo-cli docs invite accept "Ab3cDeFgHiJkLmN0pQrS://octo.example.com/docs/invite/Ab3cDeFgHiJkLmN0pQrS"
message: unsupported scheme "ab3cdefghijklmn0pqrs"

$ octo-cli docs invite accept "https://Ab3cDeFgHiJkLmN0pQrS/docs/invite/Ab3cDeFgHiJkLmN0pQrS"
message: the invite link host "Ab3cDeFgHiJkLmN0pQrS" is not the configured Octo origin "octo.example.com"
```

This runs before `collectSecrets`, so no later redaction applies. Not blocking, for two reasons: the input shape is contrived (whoever constructs it already holds the token), and it is **pre-existing** — `drive share access` prints the identical message today from the same helper, so this PR inherits the gap rather than creating it. Still worth closing in `assertSameOrigin` so both surfaces benefit; the realistic trigger is a bot that templates the link with `{token}` and `{origin}` transposed and then logs the error.

### P2 — `x-octo-secret` on response properties is inert

`x-octo-secret` is declared on the `inviteToken` *response* properties (`internal/registry/specs/docs.json:855`, `:875`). Redaction is value-based and sourced from request inputs only (`collectSecrets` reads path/query/header/body flags); response bytes are emitted unchanged. `ResponseSchema` is parsed but its only consumer is `octo-cli schema` output.

The consequence — `docs invite list` printing every live token in the clear — is intentional and is exactly what the operation is for, and both the spec description and `skills/octo-docs/common.md` say so plainly. So this is not a disclosure bug. The concern is narrower: a future reader may take the marker as evidence that output redaction happens. Consider a one-line note at the declaration saying it is advisory metadata for `schema` consumers, not an output control.

### P2 — an opaque absolute URL slips past link validation

`https:deadbeef` is a valid absolute URL with no `/`, so the shortcut classifies it as a bare token and forwards it:

```
$ octo-cli docs invite accept "https:deadbeef"
→ POST /v1/bot/docs/invites/https:deadbeef/accept
```

Not a security issue — it goes to the configured API like any other token — but it is malformed input reaching the network instead of failing locally, which is the case the parser exists to prevent. Cheap fix: also treat a value with a `:` before any `/` as a link.

### P2 — the registration panic is reachable through the exported factory seam

The panics at `cmd/docs_inviteurl.go:74-88` are argued as build-time invariants, and for the shipped binary that holds: the registry comes from `go:embed` via `registry.New()`, so no user input, env var, or config can reach them. But `cmd` is an importable package and `Factory.RegistryFunc` is exported and replaceable, so a Go embedder — or a test stub — can produce the missing-leaf panic. `TestRegisterDocsInviteAcceptURL_PanicsWhenTheLeafIsGone` demonstrates precisely this. The reasoning for panicking is sound and I would not change it; I would soften the "cannot be provoked by configuration" claim in the comment to name the embedder seam, so the next person does not treat it as an absolute.

### Nit

`--expiresInDays` / `--maxUses` are camelCase where most flags are kebab-case. I checked before raising it: this matches existing `docs.json` practice (`--folderId`, `--docType`, `--parentId`, `--fileName`, `--pageSize` on `docs.list`), so it is consistent with the domain even though `x-octo-flag` could produce kebab-case as `docs.search` does. Not worth changing in this PR; worth a domain-wide sweep someday.

---

## 3. Needs human verification

Flagging these because the change is security-adjacent and I could not close them from this repository:

1. **The role enum rests on an unverifiable claim.** The whole `reader|commenter|writer|admin` decision — and the story about a local checkout being ~86 commits behind — depends on `parseRole` in `octo-docs-backend/src/api/routes/invites.ts` on `origin/main`. That repository is not reachable from here, so I verified only that the CLI is internally consistent (spec, `enum_vocab_test.go`, skill docs, CHANGELOG all agree). Someone with backend access should confirm the four values and that an unrecognised role is a `400` rather than a downgrade. If it is a downgrade, the "never downgraded" wording in `internal/registry/specs/docs.json:837` and `skills/octo-docs/common.md` is misleading in a way that matters for an `admin`-granting link.
2. **Cross-space accept.** Issue #130's own security note says the accept transaction has no space validation, which may leave cross-space member rows. `docs.invite.accept` carries no `docId` and sends no `X-Space-Id` (pinned by `TestDocsInviteAccept_PathAndSpaceHeader`), so a credential in one space accepting a token minted in another is entirely a backend-side decision. The CLI cannot fix this and the endpoint was already reachable via `octo-cli api`, so it is not a blocker here — but this PR is what makes it a first-class command, and the PR description does not carry the note forward. Worth tracking backend-side.
3. **Origin assumption.** The invite-link parser requires the link host to equal `OCTO_API_BASE_URL`, i.e. it assumes the web app and the API share an origin. That is the existing repo-wide convention (`webOrigin`, used by `drive share` today), so it is not new — but if any deployment serves the web app on a different host, `invite accept <link>` will refuse legitimate links there while the bare-token form keeps working.
4. **The end-to-end run in the PR description** (live deployment, `source=invite` member row, idempotent re-accept, `410` after revoke, `commenter` specifically) is not reproducible from here. I verified request shape, path, body field names, and local rejection paths against a stub server; I did not verify backend semantics.

---

## Verdict

**APPROVED.** The security-relevant parts — the same-origin boundary, secret masking on the two token-addressed operations, refusal without a network call, and the absence of any regression to `drive share` — hold up under direct testing, not just under the tests shipped with the PR. Everything I found is P2: a whitespace-trimming inconsistency worth fixing before this gets used from chat-pasted tokens, a pre-existing token-echo gap in the shared helper, and two comment/annotation claims that are stated more absolutely than they hold. None of them should block merge.

The item I would most like a second pair of eyes on is the role enum (§3.1) — not because the code is wrong, but because its correctness lives in a repository this review cannot see, and the value at stake includes `admin`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #134 (octo-cli)

> Reviewer: Octo-Q (automated review) — head `919d940ae3013e32f98126d849a4d0e6344e876f`, merge-base `e6277239ac7d347fd9463d9e8dbcb72124aaf9dc`

## Summary

This PR adds the `docs invite` command family — `create` / `list` / `revoke` / `accept` — generated from four new operations in `internal/registry/specs/docs.json`, plus a hand-written wrapper (`cmd/docs_inviteurl.go`) that lets `docs invite accept` take the whole web invite link (`{origin}/docs/invite/<inviteToken>`) in place of the bare token. The link surface reuses the drive share-link SSRF boundary: `assertSameOrigin` / `assertShareIDSegment` are parameterized with a new `linkPolicy` type so both surfaces share one enforcement copy, and all existing drive call sites are re-pointed with wording-preserving `shareLinkPolicy`. The PR also fixes stale `forward-grant` summary text (the `commenter` role was already in the base enum) and updates CHANGELOG, README, CLAUDE.md and the `octo-docs` skill docs. Test coverage is dense and integration-level, driving the real command tree against httptest backends.

## Verification

Static analysis only at head `919d940a`; build and tests not executed in this environment.

- ✅ **Invite-link SSRF boundary matches the share boundary** — `parseInviteURL` (`cmd/docs_inviteurl.go:205`) runs the same origin/userinfo/scheme/percent-encoding/segment checks as `parseShareURL` (`cmd/drive_shareurl.go:59`) through the shared helpers; all seven call sites of the two helpers were enumerated and pass an explicit policy (`cmd/drive_share.go:283`,`:382`, `cmd/drive_shareurl.go:80`,`:87`,`:99`, `cmd/docs_inviteurl.go:220`,`:230`), and `shareLinkPolicy` reproduces the previous error strings verbatim, so share behavior is byte-identical.
- ✅ **The link is compared, never fetched** — the host is only matched against the origin derived from `OCTO_API_BASE_URL` (`webOrigin`, `cmd/drive_shareurl.go:232`); the request goes to the configured API with the extracted token. `TestDocsInviteAccept_TakesALinkOrAToken` asserts a foreign-host link produces no network call.
- ✅ **Wrapper fails closed on wiring drift** — registration panics when the spec declares `docs.invite.accept` but the tree shape does not match (same policy as `registry.MustNew` for embedded data), an unreadable flag value is refused rather than forwarded (`cmd/docs_inviteurl.go:136-142`), and `TestDocsInviteAcceptURL_WrapperIsInstalledOnTheSpecLeaf` pins the live wiring.
- ✅ **Flag/positional semantics** — `resolvePathValues` (`cmd/service/run.go:168`) takes the flag when set, else the positional, and rejects supplying both; the wrapper normalizes either slot before the generated `RunE` runs.
- ✅ **Secret handling end to end** — `inviteToken` is `x-octo-secret` on both token-addressed path params; `collectSecrets` (`cmd/service/run.go:297`) gathers the resolved path value after normalization, the verbose trace redacts the URL (`internal/client/client.go:966`), dry-run masking is tested (`TestDocsInvite_TokensAreMaskedInDryRun`), and no rejection path echoes the token (`TestParseInviteURL_NeverEchoesTheToken`).
- ✅ **Spec, code and docs agree** — role ladder `reader|commenter|writer|admin` in `docs.json` for invite create and members set matches every doc edit; `expiresInDays` (not drive's `expires_in_seconds`) is pinned by `TestDocsInviteCreate_BodyAndPath`; the `-q '.data.inviteToken'` recipe in `skills/octo-docs/common.md:205` matches the `WriteSuccess` envelope (`internal/output/envelope.go:40`); op counts agree (loader test 36, README 36); service-level `x-octo-space-header:false` (`internal/registry/specs/docs.json:10`) covers accept.

## Findings

No P0/P1 issues; no P2. One non-blocking Nit below.

### Nit — Bare tokens bypass the local validation links get (`cmd/docs_inviteurl.go:177`)

`inviteTokenFromValue` returns a value containing no `/` exactly as given — untrimmed, charset-unchecked — while link-shaped values are trimmed and fully validated. The asymmetry is deliberate: the comments at `cmd/docs_inviteurl.go:166-174` and `:127-130` document that a bare token must stay byte-identical to the generated leaf's pre-PR path, and `TestInviteTokenFromValue_BareTokenPassesThrough` pins it (value returned unchanged without even resolving config). Tightening it here would be a behavior change, not a fix, and could reject tokens a future backend mints; if local validation of bare tokens is ever wanted, it belongs engine-wide for all path values, not in this wrapper. Non-blocking.

## Human-verify

1. Backend contract claims — `parseRole` vocabulary, silent `expiresInDays` clamp to [1,7], `410 invite_invalid` semantics, idempotent accept, and "no stored invite id" — are asserted in the spec descriptions and pinned against httptest stubs in this PR, not against a live octo-docs-backend. The `enum_vocab_test.go` entry records verification against octo-docs-backend origin/main (including a prior stale-checkout correction), which is the right process; a spot-check of `invites.ts` on origin/main is still worth a human glance. **Not a merge blocker for this PR.**
2. The web route shape — `docsInvitePathPrefix = "/docs/invite/"` (`cmd/docs_inviteurl.go:18`) — assumes the octo-web frontend serves invite acceptance at that path. That route lives in another repo and cannot be verified from this diff. **Not a merge blocker for this PR.**

- **Data flow**: the normalized token genuinely reaches the wire — `cmd.Flags().Set` in the wrapper updates the same storage `resolvePathValues` reads (`pathFlag.strVal` bound via `StringVar`, `cmd/service/flags.go:270`), and `collectSecrets` runs after normalization so masking sees the final bare token; both flag-form and positional-form links produce `POST /v1/bot/docs/invites/<token>/accept` in tests.
- **Paired-path parity (share↔invite, create↔revoke)**: one shared enforcement helper pair, per-surface error codes (`INVALID_SHARE_URL` / `INVALID_INVITE_URL`), and the full family (`create|list|revoke|accept`) pinned by `TestDocs_TreeShape` and `TestDocs_RegistryShape`.
- **Input-shape edges**: empty, whitespace-padded, percent-encoded (`%2F` smuggling), dot-segment, multi-segment, userinfo, foreign host/port, and non-http(s) scheme inputs are all refused; query strings on invite links are discarded exactly as on blob share links.
- **Authorization boundary**: no new capability surface — the CLI calls existing `/v1/bot/docs/*` endpoints under the configured bot credential; needs-admin enforcement stays server-side, and the link surface is compare-only.
- **Disabled-service path**: `withholdDisabledServices` (`cmd/root.go:106`) removes the whole `docs` subtree after the wrapper attaches, so a disabled service leaves no dangling wrapped leaf.
- **forward-grant wording fix**: the `commenter` enum entry already existed at the merge base; this PR only aligns the stale summary/description with it.
- **No local length caps introduced**: token validation is charset-only, matching the pre-existing share behavior; no byte-budget paths were added.

## Verdict: APPROVED

Zero P0/P1/P2 defects. The invite-link surface extends the CLI's existing "compare, never contact" SSRF boundary by sharing the exact enforcement helpers with the share path instead of forking them, secret handling is wired end to end (spec declaration → path collection → verbose/dry-run redaction, all tested), the wrapper fails closed on every drift scenario, and the docs/CHANGELOG/skill edits match the implemented behavior point for point. The single Nit concerns an intentional, documented, test-pinned pass-through. Per the review rubric, a Nit-only result maps to approve.

[Octo-Q] verdict: APPROVE — 无 P0/P1/P2；仅 1 条针对既有刻意设计的 Nit；SSRF 边界经共享 helpers 正确扩展，secret  masking 端到端连通，文档与实现逐条一致。

PR #127, opened 2026-08-10
feat(marketplace): add expert & squad CLI commands

l-s-c:
## Summary

CLI-ify the Expert Marketplace (专家市场): experts (专家) and squads (专家团). Adds 19 auto-registered operations to the marketplace OpenAPI spec so the metadata-driven registry exposes `octo-cli marketplace expert …` and `… squad …` command trees — no handler code.

## Changes

- **`internal/registry/specs/marketplace.json`** (+19 ops): for both `expert` and `squad` — `list` / `mine list` / `get` / `create` / `update` / `delete`, plus `skillmd get` and `skill-download`; and the shared `expert-category list`, `expert-tag list`, `expert-skill-upload create`.
- Expert lists use **offset pagination** (`--page` / `--page-size`); `--page-all` is deliberately **not** wired (distinct from the cursor-based `skill` / `mcp` lists), since the backend returns `{total,page,page_size}` with no cursor.
- Skill packages are whole Agent-Skill `.zip/.skill` bundles: the presign → PUT → reference-`upload_object_key` flow is documented for agents.
- **Docs**: new `skills/octo-marketplace/expert.md` workflow reference; `SKILL.md` routing row + version bump (0.3.0→0.4.0); README skill blurb.
- **Tests**: updated the registry op-count (25→44), marketplace shape, command-tree, and skill-embed guards; added an expert-workflow doc test.

## Motivation

Backend contract: `octo-marketplace` `docs/api/expert-v1.md` (branch `feat/expert-marketplace`), whose consumers list `octo-cli`. This wires the CLI consumer. Verified every spec entry (method, path, query/path params, body fields) against the backend `internal/api/handler/expert/handler.go`.

> Note: the backend `feat/expert-marketplace` branch is not yet merged to its `main`; this consumer rides ahead of it. If the contract shifts in review, the spec follows.

## Testing

- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] `go build ./...` passes
- Dry-run confirmed URL, query-param, and `--data` body shaping for expert/squad commands.

## Checklist

- [x] Code is in English (comments, code, commit message; Chinese only in example payloads, matching existing `octo-summary` / `octo-docs` skill docs)
- [ ] New commands added to README Command Reference table — N/A: README has no per-command table and `marketplace` is not in the Domains table (pre-existing); updated the `octo-marketplace` skill blurb instead
[...]

PR #124, opened 2026-08-10
feat(drive): add the drive command domain with local contract enforcement

aunknown:
## Summary

Adds a `drive` service domain to the CLI: 45 leaf commands across `space`,
`member`, `folder`, `file`, `blob`, `upload`, `download`, `doc`, `share`,
`invite` and `im-transfer`, together with the engine features a
dual-identity domain needs.

The domain talks to two different mounts depending on who is calling, so
credential resolution and route selection are handled once in the engine
rather than per command.

### Credentials and routing

- `OCTO_TOKEN` becomes the high-priority credential slot. The existing
  `OCTO_BOT_TOKEN` order is preserved, so a caller that only sets the
  latter behaves exactly as before.
- Requests are routed by token kind, so a user API key and a bot token
  reach their own mount.
- The drive error envelope is mapped onto the CLI taxonomy: an
  unsupported token kind or malformed input exits 2, a missing or
  invalid credential exits 3.

### Local contract enforcement

Spec `enum` values were previously only interpolated into help text and
never read again, so an out-of-vocabulary value was sent upstream and
failed there — sometimes with a decode error that leaked an internal
struct name.

- `enum` is now enforced while the request is built, for both body and
  query parameters, including nested objects and array elements. A
  rejected value produces `ENUM_NOT_ALLOWED`, exit 2, and **no HTTP
  request**.
- Comparison uses a type-tagged canonical form rather than `==`: the
  same wire value arrives as `int`, `float64` or `json.Number` depending
  on the flag kind, while spec literals come from `encoding/json`.
  `1`, `1.0` and `json.Number("1")` are equivalent; `"1"` does not match
  a numeric vocabulary.
- Only flags the caller actually set are checked, so an unset integer
  flag does not fail on its zero value.
- Non-scalar values are rejected against scalar vocabularies, so a
  hand-written body cannot smuggle a shape the backend then fails to
  decode.
- Structural violations keep their existing `VALIDATION_ERROR` code and a
  byte-identical envelope.

### Spec corrections found while wiring the above

- `drive.doc.mount` declared a single `source` value while the service
  accepts two. Enforcing the narrower list locally would have rejected a
  valid call, so it is widened, and
  `TestEnum_DriveVocabulariesMatchTheBackend` now pins all eight drive
  vocabularies — a future narrowing fails in this repository instead of
  at a caller.
- `drive.im-transfer.create` now requires `im_channel_type`. It selects
  the upstream message-read route and is the first segment of the
[...]

PR #123, opened 2026-08-08
feat(loop): add Fleet public API support

liuooo:
## What

The Loop CLI needs to use the Fleet public v1 API through the unified gateway,
including daemon-injected task credentials and the stable public product
vocabulary. The existing Autopilot operations were also backed by generic
request schemas, and the Task lifecycle contract omitted `in_review`.

## How

- Route Loop business commands through the Fleet public v1 API.
- Add task credential mode and restrict it to the Loop command surface.
- Normalize configured API URLs to the gateway origin and reject service paths.
- Add typed Autopilot request, response, run, trigger, and delivery schemas.
- Expose `dispatch_mode` and `task_title_template` without legacy internal fields.
- Define the complete Task lifecycle, including `in_review` for work awaiting review.
- Expand the `octo-loop` skill with current workflows and status semantics.

## Tests

- `go test ./...`
- `go test ./internal/registry ./skills`
- `git diff --check`

All commands are green locally.

## Checklist

- [x] PR description is in English
- [x] Tests added or updated
- [x] Documentation and skill guidance updated
- [x] Conventional Commits used

Closes #116

PR #135, opened 2026-08-18
feat(html): promote the display title to a first-class --title flag

Ranwanglc:
## Summary

Give the `html` domain a first-class `--title` flag on its write operations. Spec-only change: no engine code, no server change.

## Changes

- `internal/registry/specs/html.json`: declare a top-level `title` string on `html.publish`, `html.draft.create` and `html.draft.save`; give `html.draft.promote` the optional `requestBody` it never had (one optional `title`). `meta` is left untouched on publish/draft-create and its description now labels it the legacy fallback. `title` is not added to any `required` list and `x-octo-body-variants` is unchanged.
- `cmd/service/service_test.go`: 3 tests merged into the existing file — `--title` promotes a top-level wire field on all four operations (and is not re-wrapped into `meta`); `html draft promote` body optionality; `--data` vs `--title` precedence.
- `skills/octo-html/SKILL.md` + `README.md`: teach `--title` as the way to set a display name, keeping the identity guidance intact (title is metadata, never identity; no same-name republish).

## Motivation

`html` was the only write domain without `--title`. `docs create` / `docs rename` / `loop task create` / `matter create` / `summary create` all get one for free, because flags are auto-generated from the request body: `registerBodyFlags` walks `requestBody.properties`, and `promotableKind` promotes primitives only — an `object` property is skipped outright. `html.json` declared the display name solely as a nested property of the `meta` object, so the whole object was skipped and neither `--title` nor `--meta` was ever registered. An Agent had to hand-write `--data '{"html":...,"meta":{"title":"..."}}'`, which is easy to omit — and when it is omitted the sidebar file list shows a row with no usable label at all.

The server side already expects this. `readPublishBody` resolves `title := raw.Title; if title == "" && raw.Meta != nil { title = raw.Meta.Title }` — top-level wins, `meta.title` is the documented back-compat path for the CLI's `meta.json`. `handlePromote` likewise decodes an optional `{"title"}`. The capability existed on the wire with no way to reach it from the CLI.

## Testing

- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] `go build ./...` passes

Also run locally: `gofmt -l .` (empty), `make lint` → `0 issues`, `go test -race -count=1` on `cmd/service` / `internal/registry` / `skills`, and `go mod tidy` leaves `go.mod`/`go.sum` unchanged.

Binary checks against a real build:

- `--title` present in `--help` for `html publish`, `html draft create`, `html draft save`, `html draft promote`.
- `html publish --html '<h1>x</h1>' --title 'Runbook Q3' --dry-run` → body `{"html":"<h1>x</h1>","idempotency_key":"<generated>","title":"Runbook Q3"}`.
- `html draft promote <ref> --title 'Renamed' --dry-run` → body `{"title":"Renamed"}`.
- `html draft promote <ref> --dry-run` → byte-identical to the same command built from `main`; no body, no `Content-Type`.

Fail-before evidence:

- Removing the promote `requestBody` / publish `title` → `unknown flag: --title`, the new tests go red.
- Adding `title` to promote's `schema.required` → `bare promote sends no body` fails with `VALIDATION_ERROR: request body is missing required field(s): title`.
- Flipping only `requestBody.required` to `true` changes nothing, because the validator short-circuits an empty body when the schema declares no required fields. The regression guard pins the boundary that actually matters, and the test comment says so rather than crediting the `required: false` flag.

## COMPREHENSION

**(a) What it does to the load-bearing path.** Before: `registerBodyFlags` iterated `html.publish` / `html.draft.create` properties, hit `meta` as `type: object`, `promotableKind` returned `(0,false)`, and the property was skipped — so no `--title`. `html.draft.save` declared only `html`. `html.draft.promote` declared no `requestBody`, so the engine registered neither `--title` nor `--data`, and its server-side title override was unreachable. After: `title` is a promoted top-level `string` on all four, emitted as the wire key `title`. Existing call shapes are untouched: `meta.title` via `--data` still works and still reaches the server, and a bare promote still sends a genuinely bodyless POST (`resolveBody` returns `nil` for an empty merged base, `client.Do` only marshals a non-nil body).

**(b) What could break.** The only behavioural risk is `html.draft.promote`, which gained a request body where it previously had none — every existing bodyless promote must stay bodyless, or callers would start sending `{}` or hitting validation. Guarded by an explicit test asserting zero body bytes *and* an empty `Content-Type`, plus a byte-diff against a binary built from `main`. Dependents checked for spec-shape coupling: `cmd/service/enum_vocab_test.go` (html enum vocab — untouched, no title enum), `internal/registry/loader_test.go` (`html.publish` operation lookup), `skills/skills_test.go` (asserts required substrings present and stale identity strings absent in the embedded skill — still satisfied), `cmd/service/flagcollision_test.go` (`title` collides with no reserved, query, header or path flag on these four commands). Edge inputs verified: `--data '{}'` stays bodyless, `--data 'null'` still rejects, `--title ""` emits `{"title":""}` which the server treats as "keep the current title".

**(c) How I know it works.** The three new tests plus the fail-before/restore cycle above; `--help` and `--dry-run` against a real build for all four operations; the promote byte-diff against `main`; and full `build` / `vet` / `test -count=1` / `test -race` / `golangci-lint` green.

## Checklist

- [x] New commands added to README Command Reference table — no new commands; the existing README html examples now show `--title`
- [x] CLAUDE.md command list updated if commands changed — no command list change (flag only)

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,8 +175,8 @@ octo-cli docs export board-7 --export-format png -o ./board.png
 # HTML docs (octo-doc) — a SEPARATE backend from `docs`. Publish self-contained
 # interactive HTML as immutable versions, then edit a single stamped artifact.
 # Canonical create has no doc reference: omit --slug. The CLI generates a key.
-octo-cli html publish --html '<h1>hi</h1>' \
-  --mount-type group --group-no <group_no> --data '{"meta":{"title":"Launch page"}}'
+octo-cli html publish --html '<h1>hi</h1>' --title 'Launch page' \
+  --mount-type group --group-no <group_no>
 # Save data.slug from the publish response. For a new document data.slug == data.doc_id, mounted or
 # unmounted. Every later operation uses data.slug. Old documents keep their legacy
 # slug as data.slug; do not infer this from mount_type or doc_id being non-empty.
@@ -184,8 +184,9 @@ octo-cli html publish --html '<h1>hi</h1>' \
 # legacy slug is rejected and cannot create a document.
 octo-cli html list
 octo-cli html versions <doc-ref>
-octo-cli html draft create --html '<h1>wip</h1>'
-octo-cli html draft save <doc-ref> --data '{"html":"<h1>wip</h1>"}'   # then: html draft promote <doc-ref>
+octo-cli html draft create --html '<h1>wip</h1>' --title 'WIP page'
+octo-cli html draft save <doc-ref> --html '<h1>wip</h1>'             # then: html draft promote <doc-ref>
+# --title sets the display name (metadata, never identity) on publish/draft create/save/promote.
 octo-cli html share <doc-ref>                                        # mint a reader share code
 octo-cli html grant add <doc-ref> --data '{"uid":"u-1"}'             # per-uid authorization
 octo-cli html element get --slug <doc-ref> --aid <content-hash>      # wire flag remains named slug
--- a/cmd/service/service_test.go
+++ b/cmd/service/service_test.go
@@ -11,6 +11,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"testing"
 
@@ -563,6 +564,172 @@ func TestHTMLCanonicalDraftCreate(t *testing.T) {
 	}
 	if key, _ := gotBody["idempotency_key"].(string); gotPath != "/docs-html/v1/docs/draft" || key == "" {
 		t.Errorf("draft create request: path=%q body=%#v", gotPath, gotBody)
+	}
+}
+
+// TestHTMLTitleFlagPromotesTopLevelTitle pins the html domain's first-class
+// --title flag on every write operation. Before this, title existed only as a
+// nested property of the `meta` OBJECT, and registerBodyFlags/promotableKind
+// skip object properties outright — so html was the one domain with no --title
+// while the server has always read (and preferred) a top-level `title`.
+//
+// The assertion is deliberately two-sided: the emitted body must carry a
+// top-level "title", and it must NOT wrap it back into a `meta` object. The CLI
+// promotes the wire field as declared; the legacy meta.title path stays
+// reachable only through an explicit --data.
+func TestHTMLTitleFlagPromotesTopLevelTitle(t *testing.T) {
+	tests := []struct {
+		name     string
+		args     []string
+		wantPath string
+	}{
+		{
+			name:     "publish",
+			args:     []string{"html", "publish", "--html", "<h1>doc</h1>", "--title", "X"},
+			wantPath: "/docs-html/v1/docs",
+		},
+		{
+			name:     "draft create",
+			args:     []string{"html", "draft", "create", "--html", "<h1>wip</h1>", "--title", "X"},
+			wantPath: "/docs-html/v1/docs/draft",
+		},
+		{
+			name:     "draft save",
+			args:     []string{"html", "draft", "save", "doc-1", "--html", "<h1>wip</h1>", "--title", "X"},
+			wantPath: "/docs-html/v1/docs/doc-1/draft",
+		},
+		{
+			name:     "draft promote",
+			args:     []string{"html", "draft", "promote", "doc-1", "--title", "X"},
+			wantPath: "/docs-html/v1/docs/doc-1/draft/promote",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var gotPath string
+			var gotBody map[string]any
+			root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+				gotPath = r.URL.EscapedPath()
+				_ = json.NewDecoder(r.Body).Decode(&gotBody)
+				w.Header().Set("Content-Type", "application/json")
+				_, _ = w.Write([]byte(`{"data":{"slug":"doc-1","doc_id":"doc-1"}}`))
+			})
+			root.SetArgs(tt.args)
+			if err := root.Execute(); err != nil {
+				t.Fatalf("execute: %v", err)
+			}
+			if gotPath != tt.wantPath {
+				t.Errorf("path = %q, want %q", gotPath, tt.wantPath)
+			}
+			if gotBody["title"] != "X" {
+				t.Errorf("body must carry a top-level title: %#v", gotBody)
+			}
+			if _, wrapped := gotBody["meta"]; wrapped {
+				t.Errorf("--title must not be wrapped into meta: %#v", gotBody)
+			}
+		})
+	}
+}
+
+// TestHTMLDraftPromoteBodyOptionality is the regression guard for adding a
+// requestBody to html.draft.promote purely so --title becomes available. A bare
+// promote must still send NO body at all — same bytes on the wire as before the
+// spec grew the field. A regression here would start sending `{}` (or fail
+// validation) on every existing promote call.
+//
+// What actually protects that is the EMPTY `required` list on the schema, not
+// the `required: false` flag on the requestBody: resolveBody returns a nil body
+// whenever the merged base is empty, so the only way to break a bare promote is
+// to make a field required. Flipping requestBody.required to true alone changes
+// nothing (bodySchemaValidator finds no required field to complain about); it is
+// adding `title` to schema.required that turns a bare promote into a
+// VALIDATION_ERROR. This test pins that boundary, so keep it green rather than
+// relying on the `required: false` declaration to hold the line.
+func TestHTMLDraftPromoteBodyOptionality(t *testing.T) {
+	tests := []struct {
+		name      string
+		args      []string
+		wantBody  string
+		wantCType string
+	}{
+		{
+			name:     "bare promote sends no body",
+			args:     []string{"html", "draft", "promote", "doc-1"},
+			wantBody: "",
+		},
+		{
+			name:      "promote with --title sends only title",
+			args:      []string{"html", "draft", "promote", "doc-1", "--title", "X"},
+			wantBody:  `{"title":"X"}`,
+			wantCType: "application/json",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var gotRaw []byte
+			var gotCType string
+			root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+				gotRaw, _ = io.ReadAll(r.Body)
+				gotCType = r.Header.Get("Content-Type")
+				w.Header().Set("Content-Type", "application/json")
+				_, _ = w.Write([]byte(`{"data":{"slug":"doc-1","version":3}}`))
+			})
+			root.SetArgs(tt.args)
+			if err := root.Execute(); err != nil {
+				t.Fatalf("execute: %v", err)
+			}
+			if string(gotRaw) != tt.wantBody {
+				t.Errorf("request body = %q, want %q", string(gotRaw), tt.wantBody)
+			}
+			if gotCType != tt.wantCType {
+				t.Errorf("Content-Type = %q, want %q", gotCType, tt.wantCType)
+			}
+		})
+	}
+}
+
+// TestHTMLTitleFlagOverridesData pins the engine's documented precedence
+// (architecture §5.2: individual flags override --data) for the new field, and
+// records the observed shape of the legacy path: --data's `meta` object is left
+// untouched next to the promoted top-level title, because the CLI merges by
+// wire key and meta.title is a different key. The server resolves the two
+// (top-level title wins, meta.title is the back-compat fallback), so shipping
+// both is safe; only the same-key case is a CLI-side override.
+func TestHTMLTitleFlagOverridesData(t *testing.T) {
+	tests := []struct {
+		name     string
+		args     []string
+		wantMeta any
+	}{
+		{
+			name:     "legacy meta.title kept alongside the promoted title",
+			args:     []string{"html", "publish", "--data", `{"html":"x","meta":{"title":"legacy"}}`, "--title", "flag"},
+			wantMeta: map[string]any{"title": "legacy"},
+		},
+		{
+			name: "same-key --data title loses to the flag",
+			args: []string{"html", "publish", "--data", `{"html":"x","title":"from-data"}`, "--title", "flag"},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var gotBody map[string]any
+			root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+				_ = json.NewDecoder(r.Body).Decode(&gotBody)
+				w.Header().Set("Content-Type", "application/json")
+				_, _ = w.Write([]byte(`{"data":{"slug":"doc-1","doc_id":"doc-1"}}`))
+			})
+			root.SetArgs(tt.args)
+			if err := root.Execute(); err != nil {
+				t.Fatalf("execute: %v", err)
+			}
+			if gotBody["title"] != "flag" {
+				t.Errorf("--title must win over --data: %#v", gotBody)
+			}
+			if !reflect.DeepEqual(gotBody["meta"], tt.wantMeta) {
+				t.Errorf("meta = %#v, want %#v", gotBody["meta"], tt.wantMeta)
+			}
+		})
 	}
 }
 
--- a/internal/registry/specs/html.json
+++ b/internal/registry/specs/html.json
@@ -102,9 +102,13 @@
                     "type": "integer",
                     "description": "optional explicit version; omit to auto-increment"
                   },
+                  "title": {
+                    "type": "string",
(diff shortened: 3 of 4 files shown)


Jerry-Xin (approved):
✅ **APPROVE** — head `aab436a3`.

Spec-only change promoting the html display title to a first-class `--title` flag on all four write operations (`html publish`, `html draft create`, `html draft save`, `html draft promote`). Independently verified against the registry engine plus live dry-run probes on a real build.

**Validation rigor (#125 pattern preserved)**
- `title` is declared optional everywhere: it is not added to any schema `required` list in `internal/registry/specs/html.json`, and promote's new `requestBody` is `required: false` with no schema-level `required`. Ops without the field keep fail-open.
- Declared-required fields still fail closed: `html publish --title X` without `--html` → local `VALIDATION_ERROR: request body is missing required field(s): html` (probe-verified); same for `html draft save`.
- `--title` passes through byte-for-byte like every other promoted string flag (no trimming in `applyBodyFlags`, `cmd/service/run.go`): a whitespace-only title goes out as-is. Consistent with the engine convention; per the description the server resolves an empty title as "keep the current title".

**Backward compatibility (probe-verified)**
- Bare `html draft promote <ref>` still sends no body and no Content-Type — pinned by `TestHTMLDraftPromoteBodyOptionality` (zero body bytes + empty Content-Type) and confirmed by dry-run probe (no body key emitted).
- `--data '{}'` on promote stays bodyless; `--data` carrying the legacy `meta.title` still works — the meta object ships untouched next to the promoted top-level title (`TestHTMLTitleFlagOverridesData`), server prefers the top-level field.

**Idempotency / body variants intact**
- Create variant still auto-generates `idempotency_key` alongside `--title` (probe: body carries both); republish variant (slug present) still omits the key as forbidden by the variant. `x-octo-body-variants` is unchanged and both directions were probe-verified. The #125 error-envelope key annotation is engine code and untouched by this PR.

**Tests & docs**
- Full `go test ./... -count=1` green (12 packages); `go build` / `go vet` clean.
- The flag-collision guard (`cmd/service/flagcollision_test.go`) is registry-wide, so the newly promoted `--title`/`--data` on promote are covered and pass.
- README + `skills/octo-html/SKILL.md` match observed behavior; the identity guidance (title is metadata, never identity; no same-name republish) is preserved.

💬 Non-blocking notes
- `--title ""` on promote emits `{"title":""}` rather than no body (probe-verified). That is documented as "keep the current title" server-side; a future test could pin this shape too.
- No `maxLength` on title, so no client-side overlong guard — consistent with every other html string field; the server stays the authority on length.

Clean, minimal, well-pinned addition. Nothing blocking.

yujiawei (approved):
# Code Review — PR #135 (octo-cli)

Reviewed at head SHA `aab436a36b2c39793c74043e1dcb2656eba9f942` against `main` (`df54c37`).

Independently verified locally: `go build ./...`, `go vet ./...` and `go test ./cmd/... ./internal/... ./skills/...` all pass, and every behavioural claim in the description that is reachable from this repository was re-checked byte-for-byte against a build of this branch and a build of `main`.

---

## 1. Spec compliance

**Spec: ✅**

There is no linked design doc, so the scope is the PR's own stated goal: give the `html` domain a first-class `--title` flag on its write operations, spec-only.

- **Missing work:** none. `title` is declared on all four write operations — `html.publish` (`internal/registry/specs/html.json:105`), `html.draft.create` (`:232`), `html.draft.save` (`:448`), `html.draft.promote` (`:495`) — and each one registers a working `--title`.
- **Over-build:** none functional. `title` is added to no `required` list, `x-octo-body-variants` is untouched (`html.json:55-74`), and no engine code changed. The only thing beyond the strict scope is a handful of adjacent doc-example rewrites (`README.md:187` switching `html draft save` from `--data` to `--html`; `SKILL.md:84` switching republish to `--slug`/`--html`). They are correct and an improvement, but they are not "`--title`" — worth a line in the description rather than sitting under "No unrelated changes included".
- **Deviation:** none. The promoted field is emitted under the wire key `title`, not re-wrapped into `meta`, which is what the schema declares.

Verified against real builds of this branch and of `main`:

| check | result |
|---|---|
| `html draft promote doc-1 --dry-run` | byte-identical to the same command from `main` — no `body`, no `Content-Type` |
| `html draft promote doc-1 --title Renamed --dry-run` | `{"title":"Renamed"}` |
| `html draft promote doc-1 --data '{}' --dry-run` | still bodyless |
| `html publish --title X --dry-run` (no `--html`) | `VALIDATION_ERROR: missing required field(s): html` — required list unchanged |
| `html publish --slug doc-1 --html … --title …` | no `idempotency_key` generated — the `republish` variant still behaves |
| `--title` in `--help` | present on all four commands |

The `requestBody` added to `html.draft.promote` is the one real behavioural risk in the diff, and it is correctly neutralised: `resolveBody` returns `nil` for an empty merged map (`cmd/service/run.go:595-597`) and `validateRequiredBodyFields` short-circuits when the body is empty and not required (`run.go:731-733`). The regression test's comment is right that the empty `required` list — not the `required: false` flag — is what holds the line, and it says so instead of crediting the wrong mechanism. That is unusually honest test documentation.

---

## 2. Code quality

**Quality: Approved** — no P0/P1. Five P2 items below, none of which blocks merge.

### P2-1 — The backend contract is asserted in the description but never recorded, and it is thinnest for `draft save`

The description cites server code for two of the four operations: `readPublishBody` (publish) and `handlePromote` (promote). No handler is cited for `html.draft.create` or `html.draft.save`.

For `draft create` there is at least indirect in-repo evidence: like `publish`, it already carried a `meta.title` path before this PR (`html.json:113`, `:240`), so the server clearly reads a title on that endpoint. **`html.draft.save` has neither.** Its schema declared only `html` before this change (that block is now `html.json:443-457`); there was no `meta`, no `title`, no title path of any kind. So it is the one operation where the flag rests entirely on an uncited claim.

If that handler ignores an unrecognised `title`, `octo-cli html draft save <ref> --html … --title 'X'` silently does nothing, while `--help` and `skills/octo-html/SKILL.md:64` ("Retitle while saving") tell an agent it renamed the document — which is precisely the failure mode this PR exists to remove, just relocated. A caller can notice via `html get` (which returns `title`, `html.json:326`), but nothing in the CLI surfaces it.

Concretely, please: (a) cite the `draft create` and `draft save` handlers the same way you cited the other two, and (b) record the minimum backend version. This repo already has that habit — `CHANGELOG.md:146-147` gates the previous `html` change on `octo-docs-backend#166` and `octo-docs-html#33` being merged and deployed. "The capability existed on the wire" is the right claim; it just isn't written down anywhere a releaser will see it.

### P2-2 — No `CHANGELOG.md` entry

`CHANGELOG.md:3` states that all notable changes are documented there, and a new user-facing flag on four commands qualifies. The immediately preceding change to this same domain (`0134bcb`, #125) added both an entry and the rollout-dependency note above; `#123`, `#124`, `#127` did too. An `### Added` bullet under `[Unreleased]` is also the natural home for P2-1.

### P2-3 — `TestHTMLTitleFlagPromotesTopLevelTitle` asserts the path but not the HTTP method

`cmd/service/service_test.go:612` captures `r.URL.EscapedPath()` only. `draft create` is `POST /docs-html/v1/docs/draft` and `draft save` is `PUT /docs-html/v1/docs/{doc_id}/draft`; the mock accepts any verb, so a spec edit that flipped a method would leave this test green. One extra field in the table closes it.

### P2-4 — `--title ""` sends `{"title":""}` on an unverified server behaviour

Confirmed by dry-run: `html draft promote doc-1 --title '' --dry-run` emits `{"title":""}`. The description says the server reads that as "keep the current title"; that is not verifiable here, and if it is instead stored, an unset shell variable (`--title "$NAME"`) silently blanks a document's display name. This is engine-wide behaviour, not something this PR introduced — `rejectEmptyRequiredValue` (`cmd/service/run.go:509`) only guards *required* values — so the fix is documentation, not code: state the observed empty-string behaviour in `skills/octo-html/SKILL.md`, since that file is what agents actually read.

### P2-5 — `html draft promote` gained a `--data` escape hatch (note, not a defect)

Adding a `requestBody` also registers `--data` on the operation for the first time (`cmd/service/flags.go:305-311`). Verified: `html draft promote doc-1 --data '{"evil":"x","html":"<b>"}' --dry-run` forwards both keys, because the body validator does not enforce closed objects outside the `loop` domain (`run.go:730`). That is consistent with every other bodied operation in the CLI and is a consequence of the engine's design rather than a problem with this diff — recording it so it is a known consequence rather than a surprise.

### What is good here

Worth saying explicitly: the regression test asserts **zero body bytes *and* an empty `Content-Type`** rather than "a body that decodes to nothing", which is the only assertion shape that actually distinguishes "no body" from `{}`. Combined with the fail-before/restore evidence and the byte-diff against `main`, the highest-risk part of this change is pinned properly. The precedence test recording that `meta` survives *next to* the promoted `title` (rather than pretending the CLI resolves them) is also the accurate thing to encode.

One second opinion suggested relaxing the byte-exact body comparison at `service_test.go:678` as brittle. I disagree and would keep it: parsing the body into a map before comparing would destroy the exact property the test exists to pin.

---

## 3. Verdict

**APPROVE**

Spec ✅ and Quality Approved. The five items above are advisory. P2-1 and P2-2 are the two worth doing before release rather than after — ideally as one commit that adds the `CHANGELOG` entry with the backend dependency spelled out.

---

## 4. Suggestions

1. Add the `### Added` `CHANGELOG` bullet with a **Minimum rollout dependency** line naming the backend PRs, matching the shape at `CHANGELOG.md:146-147`.
2. Quote the `draft create` / `draft save` handlers in a PR comment so the four-endpoint claim is symmetric with the two you already evidenced.
3. Add `wantMethod` to the `TestHTMLTitleFlagPromotesTopLevelTitle` table.
4. Document the `--title ""` behaviour in `skills/octo-html/SKILL.md` once it is confirmed server-side.

## 5. Additional observations (outside this PR's scope)

- **There is no pure retitle path.** `publish` and `draft save` both require `html` (`html.json:85-87`, `:440-442`), and `draft promote` mints a new immutable version. So an agent that only wants to rename a document must either resend the full HTML or cause a version bump; the domain has no `html.update` / `html.rename` operation (all 21 operation ids in `html.json`). Given that the motivation here is "the sidebar shows a row with no usable label", a rename-only operation is the obvious follow-up.
- **Downstream doc generation.** `Mininglamp-OSS/octo-docs` generates `reference/api/html.mdx` from this exact spec file ("DO NOT EDIT — generated by scripts/gen-api.mjs from … octo-cli/internal/registry/specs/html.json"), and that page is already stale relative to `main` — it still documents `PUT /v1/docs/{slug}/draft` and lists only `meta`. Pre-existing drift, not caused by this PR, but this PR widens it; a `npm run gen:all` in that repo would be a good companion change.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #135 (octo-cli)
Head: `aab436a36b2c39793c74043e1dcb2656eba9f942` · merge-base: `df54c370` (origin/main) · 4 files, +226/−14

## Summary

This PR promotes the html domain's display title from a nested `meta.title` property to a first-class `--title` flag on all four title-bearing write operations (`html publish`, `html draft create`, `html draft save`, `html draft promote`). Since octo-cli is metadata-driven, the change is spec-only: a top-level `title` string property is added to three request bodies in `internal/registry/specs/html.json`, a new optional `requestBody` is added to `html.draft.promote` (which previously had none), plus README/skill doc updates and three new engine tests. The legacy `meta.title` path remains reachable via explicit `--data`.

The review traced the full spec→registry→cobra→wire path in the engine (unchanged by this PR) and verified the new fields flow correctly at runtime, including the empty-body edge on bare promote. No blocking issues.

## Verification

Static analysis only at head `aab436a`; build and tests not executed in this environment.

- ✅ **Flag wiring** — `title` is a top-level string property, so `promotableKind` (`cmd/service/flags.go:408`) promotes it and `registerBodyFlags` (`cmd/service/flags.go:291`) registers `--title`; `title` is not in `reservedFlagNames` and collides with no path/query/header flag; `TestFlags_NoSpecCollidesWithEngineFlags` guards this class for all specs.
- ✅ **Bare promote unchanged on the wire** — `resolveBody` returns nil when the merged base is empty (`cmd/service/run.go:595`), promote has no `x-octo-auto-idempotency-key` and `RequestBodyRequired=false`, and the client only marshals when `req.Body != nil` (`internal/client/client.go:867`). `octo-cli html draft promote <doc-ref>` sends zero bytes exactly as before.
- ✅ **Tests exercise the real production path** — `rootWithService` wires `registry.MustNew` (embedded specs) and builds the full command tree (`cmd/service/service_test.go:88`); no hand-fed stub registry.

## Findings

No P0/P1 issues.

- All four title-bearing write ops got the field (spec map: publish/draft create/save/promote); other html write ops (grant, asset, comment, reply, element) don't set display titles — parity is complete, no operation was missed.
- Precedence: `applyBodyFlags` (`cmd/service/run.go:691`) applies only `Changed` flags after `--data` parsing, so `--title` overrides a same-key `--data` title and leaves a `meta.title` object untouched alongside; pinned by `TestHTMLTitleFlagOverridesData` (`cmd/service/service_test.go:698`).
- Publish/draft-create body variants (create-vs-republish) reference only `slug`/`idempotency_key`; adding `title` doesn't interact with them.
- `draft save` keeps `html` required as before — retitle-only save is impossible, matching the skill doc ("Retitle while saving").
- README and `skills/octo-html/SKILL.md` updated consistently; no stale `meta.title` examples remain elsewhere; `docs/octo-cli-design.md` explicitly predates the html domain, so no sync gap there.
- Two-sided test assertions: top-level `title` present AND not re-wrapped into `meta` (`cmd/service/service_test.go:580`); promote optionality pinned at byte level including Content-Type (`cmd/service/service_test.go:648`).

## 数据流回溯

| 被消费数据 | 上游来源 | 是否真流到消费点 |
|---|---|---|
| `--title` 值 → HTTP body | cobra flag (Changed) → `applyBodyFlags` 写入 `base["title"]` → `req.Body` → `json.Marshal` (`internal/client/client.go:867`) | ✅ 四操作均验证；仅 Changed 时写入，未设 flag 不覆盖 `--data` |
| spec `title` property → cobra flag | `operationRequestBody` (`internal/registry/loader.go:599`，required=false 也解析) → `registerBodyFlags` | ✅ 无碰撞、无 reserved 名、`meta`(object) 被 `promotableKind` 正确跳过 |
| bare promote 的空 body | `resolveBody` `len(base)==0 → nil` (`cmd/service/run.go:595`) | ✅ 与 PR 前字节一致；promote 无 auto-idempotency-key，不会生成隐藏 body |
| "top-level title wins" 优先级 | **octo-docs-html 后端契约（另一仓库）** | ⚠️ 本 checkout 内不可验证；spec/skill/test 一致声明该契约，作为跨仓假设记录（见额外发现） |

## 盲点 checklist

- **C1 双路径 parity**：clear — 四个设置 title 的写操作全部加齐；不存在 title 的 remove/unset 对称操作；守卫等价写法覆盖由 `flagcollision_test.go` 对所有 spec 全量断言。
- **C2 control-flow ordering / 嵌套复用**：clear — `registerBodyFlags`/`resolveBody` 被所有操作复用，本 PR 唯一的新增结构是给 promote 加 optional requestBody；空 body 路径（ordering 关键点）已验证字节不变。
- **C3 授权边界**：N/A — 无权限/凭证/tool 暴露改动；title 是元数据，identity 仍为 slug/doc_id。
- **C4 授权生命周期 / 容器级联**：N/A — 无鉴权或容器-成员状态改动。
- **C5 build 过 ≠ 运行期正确**：clear（有保留）— 未以"静态审过"缓解：对运行期路径（flag 注册、body 合并、空 body、wire 序列化）逐段做了代码级推演并锚定行号；测试经 httptest 真服务模拟运行期。本环境未执行 build/test，如实标注。
- **C6 文档自洽**：clear — README/SKILL/spec 三处描述互相一致（fallback 仅 publish/create 有 meta，与 schema 吻合；"omitting --title keeps current name" 与引擎行为吻合）。

## 跨轮 blocker 复检（R6）

N/A — 本 PR 首轮审查。

## 额外发现

1. **Nit — 缺 CHANGELOG 条目**：四个命令新增用户可见 `--title` flag，但 `CHANGELOG.md` 的 `[Unreleased]` 未记录；同级别 html 域契约变更（canonical create，#125）有记录先例。建议补一条（含 meta.title legacy fallback 说明）。非阻塞，按合并策略不进入对外 gating findings。
2. **跨仓契约假设**：「both present 时 top-level title wins」依赖 octo-docs-html 后端行为，本仓库无法验证。spec/skill/测试注释一致声明该契约（"the server has always read (and preferred) a top-level title"），无相反证据；若终审需要绝对把握，可向后端仓库交叉确认，但不应阻塞本 CLI PR。

## 建议

- 在 `[Unreleased]` Added 下补一条 `--title` 条目（一行即可）。
- 其余保持现状：测试设计与回归守卫（尤其 `TestHTMLDraftPromoteBodyOptionality` 对"required 列表才是真防线"的注释）质量很高。

## Verdict: APPROVED

Spec-only change on a metadata-driven CLI; every runtime contract the new field touches (flag registration, collision safety, --data precedence, empty-body promote, wire serialization) was traced through the engine and shown preserved, with regression tests running against the real embedded specs. Zero P0/P1 findings across both review legs; the only observation is a missing changelog entry, which is non-blocking.

[Octo-Q] verdict: APPROVE — 无 P0/P1；spec→flag→wire 全链路验证通过，bare promote 字节级不变，测试走真实嵌入 spec。唯一观察项为 CHANGELOG 缺条目（Nit，非阻塞）。跨仓假设"top-level title wins"建议终审知悉但不阻塞。

PR #141, opened 2026-08-19
feat(drive): expose full-text file search in the CLI

### LINKED ISSUE #140: feat(drive): expose full-text file search in the CLI
## Problem Statement

The `octo-drive` backend already implements full-text file search — `POST /v1/{bot,user}/drive/search` is live (handler `internal/search/api.go`, wired in `cmd/drive/main.go` under all three auth mounts), backed by OpenSearch with name + blob/doc-body (Tika-extracted) indexing, per-space ACL push-down, breadcrumb resolution and per-user rate limiting. But `octo-cli` exposes **no** way to call it: the `drive` command tree covers browse / folder / file / blob / upload / download / doc / share / invite / im-transfer, yet has no `search`. An agent (or a person on a `uk_` key) that wants to locate a file by name or content has to fall back to paging `drive browse` per folder, which cannot match body text at all.

This is the drive-side analogue of #117 (`feat(docs): expose full-text document search in the CLI`), which closed the same gap for the docs domain.

## Proposed Solution

Add a `drive search` command by declaring the existing endpoint in the embedded spec (`internal/registry/specs/drive.json`) — no Go logic, consistent with the metadata-driven architecture. Shape mirrored from the backend contract (`internal/search/query.go` `Request` / `service.go` `Result`):

```
octo-cli drive search --q "报销单"                                  # all visible spaces
octo-cli drive search --q "预算" --scope space --space-id "$SPACE"
octo-cli drive search --q "合同" --data '{"filters":{"type":"blob","owner_scope":"me"}}'
```

- Scalar flags: `--q` (required), `--scope` (all|space, enum-gated locally), `--space-id`, `--page-index` (0-based), `--page-size`.
- Advanced `--filters` object via `--data` (type / owner_scope / updated_after / size_min / size_max), matching the `message.search` convention for opaque filter objects.
- `x-octo-mount-by-token-kind` routes `uk_` → `/v1/user/drive/search` and bots → `/v1/bot/drive/search`, same as every other drive op. Results are always scoped to the caller's own space membership.
- Response `SearchResult` schema: `total`, `truncated`, `items[]` with `file_id` / `parent_id` as lossless decimal strings (uint64), `path` breadcrumb, `type`, `highlights.{name,body}` with `<mark>` fragments, `content_meta`.

## Alternatives Considered

- **Page `drive browse` client-side** — cannot match body/OCR text, O(folders) requests, no relevance ranking. Not a real substitute.
- **`octo-cli api` raw passthrough** — works today but bypasses flag ergonomics, local enum/required validation, and uint64-lossless output shaping; not discoverable.

## Additional Context

- Requires the server's drive search backend (OpenSearch) to be configured; otherwise the endpoint returns `503 unavailable` (spec §8.3, no SQL-LIKE fallback) — the CLI surfaces this as an `unavailable` error envelope.
- Backend caps `page_size` to its configured max (default 50).
- I have an implementation ready (spec + guard-test updates + skill docs) and have run it end-to-end against a live test environment (name / body / CJK-body hits, filters, scope, pagination, enum + required validation — all green). Happy to open a PR referencing this issue.

mou-ren:
## Summary

Add a `drive search` command that exposes octo-drive's existing full-text file search endpoint through the CLI.

## Changes

- **`internal/registry/specs/drive.json`**: declare `POST /v1/bot/drive/search` (`operationId: drive.search`, `x-octo-risk: read`) + a `SearchResult` response schema. Spec-only — the metadata-driven engine generates the command, flags, enum gate, uint64-lossless output shaping, and `x-octo-mount-by-token-kind` routing (`uk_` → `/v1/user/drive/search`, bots → `/v1/bot/drive/search`). No Go logic added.
- Request: `--q` (required), `--scope` (all|space, locally enum-gated), `--space-id`, `--page-index` (0-based), `--page-size`; advanced `--filters` object (type / owner_scope / updated_after / size_min / size_max) via `--data`, matching the `message.search` convention.
- `items[].file_id` / `items[].parent_id` are lossless decimal strings (uint64); `size` is a plain integer with `format: uint64` (files are gateway-capped well below 2^53, so no lossless string needed — consistent with `DriveEntry`).
- **Guard tests** updated: drive op count 42→43 (`loader_test.go`), command-tree (`cmd/drive_test.go` 45→46 with `search` under the top-level group), `RegistryShape` path pin (`cmd/service/drive_test.go`), and the `scope` request-side enum vocabulary with provenance (`cmd/service/enum_vocab_test.go`).
- **Docs**: README command-reference table (43 ops / 46 leaves), CHANGELOG, `cmd/drive.go` package doc comment, and `skills/octo-drive/SKILL.md` (usage + `--filters` + highlights + 503 note).

## Motivation

The octo-drive backend already implements this search (`internal/search`, OpenSearch-backed, name + Tika-extracted body, per-space ACL, breadcrumb, rate limiting) and mounts it on all three auth paths, but the CLI had no way to call it — the only alternative was paging `drive browse` per folder, which can't match body/OCR text. This is the drive-side analogue of the docs search command from #117.

Fixes #140

## Testing

Verified end-to-end against a live test environment (user-API-key identity), real HTTP + real byte uploads + real index pipeline (drive → Kafka → indexer → OpenSearch), with all test data cleaned up afterward:

- **Search e2e — 20/20 PASS**: name hit + `<mark>` highlight, English body hit (Tika), **CJK body hit** (`预算审批`), path breadcrumb, owner_uid; filters type=blob / type=folder-exclude / owner_scope me / owner_scope others-exclude; scope=all / scope=space; page_size; empty-`q` reject, invalid-scope `ENUM_NOT_ALLOWED`, missing-`space_id` reject.
- **Full drive regression — 35/35 PASS**: space / member / folder / file (get·move·copy·rename) / blob / upload (prepare→PUT→confirm) / download (file·url) / doc (candidates·mount·list·unmount) / share (create·list·access·download·revoke) / invite (create·list·revoke). (member-add / invite-accept / im-transfer skipped for lack of a second identity or a real IM file message — not faked.)

- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] `go build ./...` passes

## Checklist

- [x] New commands added to README Command Reference table
- [x] CLAUDE.md command list updated if commands changed (CLAUDE.md has no per-command list; command usage lives in `skills/octo-drive`, which is updated)
- [x] New features include tests (four registry/command guard tests updated for the new op)


Jerry-Xin (approved):
The PR is in scope and correctly integrates drive search with the metadata-driven CLI architecture.

💬 Non-blocking

- 🟡 Warning — `internal/registry/specs/drive.json:262`: `filters` lacks property schemas. Consequently, nested enum/type/uint64 validation and detailed schema introspection are unavailable. Consider declaring its five supported properties explicitly.
- 🔵 Suggestion — Add a focused command test covering request construction, user-key path routing, enum rejection, and lossless normalization of `items[].file_id`/`parent_id`. Current additions primarily guard registry shape and command counts.

✅ Highlights

- Correctly inherits token-kind routing and ACL-aware server behavior.
- Lossless uint64 response paths follow existing drive conventions.
- CLI help, command inventory, changelog, README, and skill documentation are consistently updated.
- `go test ./...` and `git diff --check` pass.

yujiawei (approved):
# Code Review — PR #141 (octo-cli)

Reviewed at head `e810183`, merge-base `f9d9f42` (single commit, clean linear diff — no sibling-PR interaction in this repo right now).

I reviewed this as a spec-only change against the generation engine rather than by reading the JSON alone: I built the binary at this SHA and exercised the generated `drive search` leaf against a local mock backend to confirm what the engine actually does with each new declaration. Findings below are marked **[verified]** where I observed the behaviour, **[static]** where it comes from reading the engine.

---

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. Issue #140 asked for the drive full-text search endpoint to be reachable from the CLI; `POST /v1/bot/drive/search` is declared with the full request surface (`q` / `scope` / `space_id` / `page_index` / `page_size` / `filters`) and a complete `SearchResult` response schema.
- **Over-build:** none. No Go logic added; no new flags, extensions or endpoints beyond the one operation. `x-octo-risk: read` is correct for a read-only POST and matches `message.search` (also POST + `read`).
- **Deviation:** none material. `space_id` is declared `type: string` (`drive.json:259`), which is consistent with all 13 other drive operations that take a space id; `parent_id`/`file_id` stay `integer/format: uint64`. The counts in `loader_test.go:48`, `cmd/drive_test.go:141`, `cmd/service/drive_test.go:71` and `enum_vocab_test.go:122` are all updated consistently. `go build`, `go vet ./...` and `go test ./... -count=1` are green at this SHA — I ran them.

### Behaviour I verified end-to-end

| Check | Result |
|---|---|
| Mount routing `uk_` → `/v1/user/drive/search` | ✅ server received `/v1/user/drive/search` |
| Mount routing `bf_` / `app_` → `/v1/bot/drive/search` | ✅ |
| `loop_credential` rejected locally | ✅ `TOKEN_KIND_NOT_ALLOWED`, exit 2, no HTTP |
| `scope` enum gate (flag path and `--data` path) | ✅ both return `ENUM_NOT_ALLOWED` before any request |
| Missing `q` | ✅ `VALIDATION_ERROR: request body is missing required field(s): q` |
| `--data` deep-merges with promoted flags | ✅ `--q x --data '{"filters":{…}}'` → `{"filters":{…},"q":"x"}`; the flag is **not** dropped |
| `items[].file_id` / `items[].parent_id` lossless | ✅ `18446744073709551615` and `9007199254740995` emitted as exact decimal strings |
| `size` above 2^53 as a plain number | ✅ `9007199254740993` survives byte-exact in `json`, `table`, `csv` and `ndjson` — the `UseNumber` decode makes the "no lossless string needed" decision safe as argued |

The `size`-as-plain-integer call in the PR description is sound, and for a stronger reason than the one given: it holds even if a file ever *did* exceed 2^53, because the row extractor decodes with `UseNumber` regardless of whether a field is declared lossless.

---

## 2. Code quality

**Quality: Approved** — no P0/P1. Five P2s, all cheap.

### P2-1 — `--q` deviates from the `--keyword` convention both sibling search commands use, and collides with the `-q` global shorthand

`internal/registry/specs/drive.json:257`

`docs.search` declares the identical backend field as `"q": { … "x-octo-flag": "keyword" }` (`docs.json:121`) and `message.search` names its query field `keyword` (`message.json`). This PR exposes the raw property name, so the drive command is the only search in the tree spelled `--q`.

That matters more than a naming preference here, because `-q` is already registered globally as the shorthand for `--jq`:

```
  -q, --jq string        filter output with a jq expression
```

So the natural short form silently binds to the wrong flag — **[verified]**:

```
$ octo-cli drive search -q "报销单" --dry-run
{"ok":false,"error":{"code":"VALIDATION_ERROR",
 "message":"request body is missing required field(s): q"}}
```

The value went to `--jq`. The failure is loud rather than silent, so this is not a correctness bug — but the error message points at `q` while the actual problem is the flag the caller reached for, which is exactly the kind of thing that costs an agent a retry loop to diagnose.

This is the one item I would most encourage fixing *before* merge rather than after: it is a one-line spec change now, and a breaking change to an agent-facing surface later.

```json
"q": { "type": "string", "x-octo-flag": "keyword", "description": "…" }
```

The wire key stays `q`; only the CLI surface changes. `skills/octo-drive/SKILL.md:161-165` would need the same edit.

### P2-2 — `filters` is an untyped object, so its documented vocabulary is unenforceable

`internal/registry/specs/drive.json:262`

`filters` is declared `{"type": "object"}` with the field list in prose. Nothing inside it is validated — **[verified]**:

```
$ octo-cli drive search --q x --data '{"filters":{"type":"nonsense","bogus_key":1}}' --dry-run
{"filters":{"bogus_key":1,"type":"nonsense"},"q":"x"}
```

To be explicit about severity: **this is the established convention in this repo**, not a new sin — all seven `message.search*` operations declare `filters` exactly this way. That is why I am scoring it P2 and not blocking on it.

The sharper version of the point is about this repo's own guard discipline rather than about type safety. `cmd/service/enum_vocab_test.go:12-35` states the rule that "a spec author cannot add an enum — and with it a new hard gate — without transcribing the backend's set and recording where it came from." The author clearly *knows* `type` is `folder|doc|blob` and `owner_scope` is `me|others` — both are written out at `drive.json:262` and `SKILL.md:167`. Declaring them as nested `enum`s instead of prose would put them under that guard, get them gated locally for free, and cost nothing:

```json
"filters": {
  "type": "object",
  "properties": {
    "type":        { "type": "string", "enum": ["folder", "doc", "blob"] },
    "owner_scope": { "type": "string", "enum": ["me", "others"] },
    "updated_after": { "type": "string" },
    "size_min":    { "type": "integer", "format": "uint64" },
    "size_max":    { "type": "integer", "format": "uint64" }
  }
}
```

The body walker recurses at every depth and `--data` goes through it, so this would gate the nested values immediately. `size_min`/`size_max` as `format: uint64` additionally picks up the range check. Note this would also require corresponding `enum_vocab_test.go` entries with provenance — which is the point.

### P2-3 — Command-count drift introduced by this PR (4 sites)

The PR updates the count in four places but misses four others that were *correct before this commit* and are wrong after it:

| File:line | Says | Should say |
|---|---|---|
| `skills/octo-drive/SKILL.md:13` | `45 commands over one backend` | 46 |
| `cmd/drive_test.go:98` | `pins the 45-command surface: 39 generated leaves` | 46 / 40 |
| `docs/octo-cli-design.md:211` | `## Domain 8: drive (octo-drive, 45 commands)` | 46 |
| `docs/octo-cli-design.md:512` | `\| drive \| 45 \| 45 … \| 45 \| 0 \|` | 46 |

`SKILL.md:13` is the one I would fix regardless of the others: this PR edits that file, and the result is a file that says "45 commands" in its header and then lists 46 in its own command tree at line 243.

Separately, `docs/octo-cli-design.md:229-273` is a numbered per-command table (`| 61 | octo-cli drive browse | GET | {mount}/browse | read |`) with no row for `drive search`. If that table is still considered live, a row belongs next to `browse`.

For fairness: the *grand totals* in that same design doc (`| Total | 101 |`) and `CLAUDE.md:44` ("308 operations", actual is 334) are already stale independently of this PR. I am not asking this PR to fix those — only the four sites it made stale.

### P2-4 — Two documented preconditions are backend-only, and read as if they were local

`internal/registry/specs/drive.json:257` and `:259`

- `q` is described as "Required and non-empty", but only presence is gated. `--q ""` is forwarded with an empty body value — **[verified]**. `minLength` would not help: `validateString` only runs under `enforcePublicAPIConstraints` (`cmd/service/run.go:802-812`), which is Loop-only, so a `minLength: 1` here would be inert. **[static]**
- `space_id` is "required when `--scope=space`", but `--scope space` with no `--space-id` is forwarded — **[verified]**. The engine has no conditional-required mechanism for this shape (`BodyVariant` exists but covers required/forbidden sets, not value-conditioned ones).

Both are genuinely backend-enforced and the PR's e2e run reports the server rejecting them, so this is a wording issue, not a gap: the descriptions currently read like CLI contracts. Suggest making the enforcement boundary explicit, e.g. "must be non-empty (rejected by the server)" — so nobody later reads the text as evidence of a local gate that never existed.

### P2-5 — `--page-all` is unavailable here, unlike every sibling search

`drive search` offers no `--page-all` / `--page-limit`, while `docs search` and `message search` both do. This is architecturally forced, not an oversight: `PaginationInfo` (`internal/registry/loader.go:256-268`) is cursor-based only (`CursorParam` / `CursorField`), and this endpoint is offset-paged. **[static]**

No change requested. `SKILL.md:161` already says "`--page-index` here IS a real page (0-based)", which is the right disclosure. Worth one added clause that walking all pages means looping `--page-index` yourself — an agent that learned `--page-all` from `docs search` will reach for it here.

---

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. The design decisions that carry real risk — uint64 losslessness, mount routing by token kind, enum gating through both the flag and `--data` paths — are all correct, and I confirmed each against a running binary rather than trusting the description. Everything above is P2.

---

## 4. For a human reviewer — security-sensitive label

This PR is labelled `needs-human-review`, and the CLI side of it is clean: no credential handling, no new auth path, no local trust decision. The routing gate works (`loop_credential` refused locally, `uk_`/`bf_`/`app_` mounted correctly) and `x-octo-risk: read` is accurate.

The risk this change carries is not in the diff — it is in what the diff makes reachable, and it cannot be verified from this repository. **Two items I would want checked against octo-drive before merge:**

1. **Is search gated by space membership, or by per-file role?** The operation description (`drive.json:124`) and `SKILL.md:161` both say results are "gated by the spaces the caller is a member of" — membership, not role. But drive's role vocabulary includes `preview_only` and `downloader` (`enum_vocab_test.go:99-107`), which exist precisely to distinguish who may read content. `highlights.body` returns extracted document text (`drive.json:190-195`). If the backend filter is membership-only, a `preview_only` member gets body-text fragments — including OCR'd content — of files they are not permitted to download, and gets them without touching the download path that would normally enforce the role. This is the one question I would not merge without an answer to. It is a backend property, so the answer does not change this diff either way; but it changes whether this CLI surface should ship now.

2. **Materially broader read primitive for bots.** Before this change a bot's read surface was per-folder `browse` — names only. After it, any bot added to a space can full-text search **file bodies** across every space it belongs to, in one call, and receive content excerpts without downloading anything. That is the intended feature and the identity scoping looks right; it is worth a conscious sign-off rather than an implicit one, since "bot was added to a space" now grants meaningfully more than it did last week.

Two lower-priority notes:

3. `<mark>` fragments (`drive.json:190-195`) are backend-generated markup wrapping user-controlled file content. For this CLI the output is JSON on stdout, so there is no injection surface here. It is worth confirming the search backend HTML-escapes extracted text before wrapping it, for the benefit of any consumer that renders these fragments — and worth noting that an agent consuming `highlights.body` is reading attacker-influenceable text, which is a prompt-injection consideration rather than an XSS one.
4. The documented 503 for "search backend not configured" is a permanent condition, but 503 is on the transport's retryable list (`internal/client/client.go:1296-1304`), so calls on an unconfigured server walk the full backoff ladder before failing. Generic transport policy, not this PR's doing — flagged only so it is not mistaken for a hang.

---

## 5. Coverage — what this review did not cover

Stated explicitly so the gaps are not read as clearances:

- **The octo-drive backend is not in this workspace.** Every claim about ACL scoping, empty-`q` rejection, `scope` validation, server-side page caps, and 503 semantics comes from the PR description or the spec text. I verified none of it. The PR's "20/20 e2e PASS" against a live environment is the only evidence for those, and I could not independently confirm it.
- **The `filters` vocabulary was not checked against the backend.** I confirmed the CLI forwards anything; I did not confirm `folder|doc|blob` and `me|others` are the backend's actual sets. If they are narrower or wider than the prose, that is invisible from here — which is the argument for putting them under the `enum_vocab_test.go` provenance guard.
- **No real-network run.** All runtime evidence is against a local mock plus `--dry-run`; retry, rate-limit and pagination behaviour under a real server were not exercised.
- **`docs/octo-cli-design.md`'s live status is unclear.** Its totals are stale by ~230 operations independently of this PR, so I could not tell whether the drive command table is maintained or abandoned. I raised the two lines this PR made wrong and left the rest alone.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #141 (octo-cli)

## Summary

This PR exposes octo-drive's new full-text search as a spec-driven CLI leaf: one new `drive.search` operation in `internal/registry/specs/drive.json` (POST `/v1/bot/drive/search`, required `q`, optional `scope`/`space_id`/`page_index`/`page_size` flags plus a `--data`-only `filters` object, and a `SearchResult` response schema) together with the count/doc updates in CHANGELOG, README, the octo-drive SKILL.md, and the test tables that pin operation counts, registry shape, and enum vocabularies. No engine code changes at all — the new leaf rides entirely on existing machinery (token-kind mount routing, body-flag promotion, enum/required pre-flight, lossless-id normalization), so the review focused on proving the operation plugs into each piece of that machinery correctly. No P0/P1 issues; two non-blocking P2 items below.

## Verification

Static analysis only at head `e8101836`; build and tests not executed in this environment.

- ✅ **Token-kind routing** — the spec-level `x-octo-allowed-token-kinds` / `x-octo-mount-by-token-kind` (`internal/registry/specs/drive.json:11`) cover the new operation; `swapMount` (`cmd/service/identity.go:105`) rewrites `/v1/bot/drive/search` to `/v1/user/drive/search` for `uk_` tokens, exactly as the operation description claims, and a path matching no declared mount fails loudly (`MOUNT_PREFIX_UNKNOWN`).
- ✅ **Lossless ids** — `x-octo-lossless-id-fields: ["items[].file_id", "items[].parent_id"]` uses the same `entries[].` array syntax `drive.browse` already ships, and `walkSegments` (`internal/output/normalize.go:100`) resolves `a[].b` across every array element, rewriting JSON numbers to decimal strings.
- ✅ **Pre-flight** — `required: ["q"]` is enforced on the merged body including `--data` (`cmd/service/run.go:953`); the `scope` enum `[all, space]` is gated by `checkEnum` before any HTTP and pinned with backend provenance in `cmd/service/enum_vocab_test.go:122`.
- ✅ **Flag generation** — only Changed flags enter the body (`cmd/service/run.go:691`), so spec defaults (`scope=all`, `page_size=20`) apply server-side instead of being overridden by zero values; `filters` is a non-promotable object and correctly rides `--data` only.
- ✅ **Count/doc parity** — 43 operations / 46 leaves / 40 generated agree across `CHANGELOG.md:28`, `README.md:42`, `cmd/drive.go:3`, `cmd/drive_test.go:111`, `cmd/service/drive_test.go:71`, `internal/registry/loader_test.go:48`; the new SKILL.md examples are enforced against the real command tree by `TestSkills_EveryDocumentedFlagExists`, and flag-collision/vocabulary tests walk every operation generically.

## Findings

No P0/P1 issues; two non-blocking P2 items below.

### P2 — Spec promises more pre-flight than the engine enforces (`internal/registry/specs/drive.json:257`)

`q` is described as "Required and non-empty", but the generic validator treats required as presence-only (`cmd/service/run.go:960` counts `""` as present) and MinLength is enforced only for Loop-service operations — so `octo-cli drive search --q ""` passes pre-flight and is only refused by octo-drive's `validate()`. The same holds for the conditional pair at `:258-259`: `--scope space` without `--space-id` ships `{"scope":"space"}` and fails server-side. The consequence is a worse error location, not a broken path, and `docs.search` has the identical shape, so this is precedent-consistent. Reword the descriptions to point at server-side rejection, or land an engine follow-up enforcing MinLength / dependent-required for all services.

### P2 — `filters` bypasses every local gate (`internal/registry/specs/drive.json:262`)

`filters` is declared as a bare object, so its documented sub-fields (`type` folder|doc|blob, `owner_scope` me|others, `updated_after` RFC3339, `size_min`/`size_max` uint64) are forwarded unvalidated; a typo'd `owner_scope` or a float/oversized size only fails at the backend. Keeping it opaque is defensible today — per this repo's vocabulary discipline a locally-declared enum must never be narrower than the backend, and the filter set may still evolve — but once octo-drive's filter vocabulary stabilizes, declare the properties here (enum + `format: uint64`) so the existing ENUM_NOT_ALLOWED / uint64 gates catch agent typos before a round-trip.

## Human-verify

1. The CLI contract presumes octo-drive serves `POST /v1/bot/drive/search` and `/v1/user/drive/search` with results scoped to the caller's space membership (503 `unavailable` when OpenSearch is not configured). In-checkout evidence supports this (`cmd/service/enum_vocab_test.go:122` cites octo-drive `internal/search/query.go` ScopeAll/ScopeSpace), but the server side is outside this checkout — confirm the endpoints are deployed and membership-scoped. This is **not** a merge blocker for this CLI PR.
2. OpenSearch query construction from `q`/`filters` (injection safety, cost caps on very long queries) is entirely server-side; confirm octo-drive sanitizes and bounds before building queries. This is **not** a merge blocker for this CLI PR.

- Mount prefix contract: the new path starts with exactly one declared mount; the user-key swap produces `/v1/user/drive/search`.
- No flag collisions: `q`, `scope`, `space-id`, `page-index`, `page-size` avoid engine and global names (`--space` ≠ `--space-id`); `-q` stays bound to `--jq` as documented for other `q` params.
- `x-octo-risk: read` is informational; default transport retry on a POST read is safe, and no `x-octo-secret` values exist to mask.
- Response `size` stays a plain JSON number under the documented sub-2^53 assumption; table/csv row extraction decodes with UseNumber, so no float rounding in any output format.
- CHANGELOG edits the still-Unreleased drive-domain bullet in place — correct Keep-a-Changelog convention for an unreleased domain.

## Verdict: APPROVED

No correctness, security, or contract-breaking issues in the CLI diff: the new leaf plugs correctly into identity routing, validation, lossless-id handling, and flag generation, all pre-existing and test-backed. The two P2s are non-blocking polish. Server-side existence and authorization of the search endpoints are out of scope for this checkout and listed under Human-verify.

yujiawei:
**Correction to my review above — P2-1.**

I wrote that "the drive command is the only search in the tree spelled `--q`." That is wrong, and it overstated the convention argument. Enumerating every flag across the built command tree at this SHA, three commands expose a `--q` flag:

```
drive search
loop project search
loop skill search
```

So `--q` has precedent here; the repo is genuinely split (`docs.search` renames `q` → `--keyword` via `x-octo-flag`, Loop's two searches keep `--q`). This PR follows one of two existing patterns rather than departing from a settled one, and my "deviates from convention" framing was not accurate.

What does hold up, verified on all three commands, is the ergonomic half: `-q` is registered globally as the shorthand for `--jq`, so the short form never reaches the query field.

```
$ octo-cli loop project search -q "test" --dry-run
{"ok":false,"error":{"message":"required flag(s) \"q\", \"workspace-id\" not set"}}
```

That is pre-existing behaviour this PR inherits, not something it introduces. It is also not documented anywhere as a known consequence — what is documented (README:328, `skills/octo-shared/SKILL.md:126`) is only that `-q` means `--jq`.

Net effect on the recommendation: weaker than I stated. Renaming to `--keyword` would align this with `docs.search` and sidestep the shorthand trap, and it is still cheapest before release — but it is a consistency call for the maintainers between two live patterns, not a defect in this PR. Everything else in my review stands, and the verdict is unchanged.

PR #148, opened 2026-08-26
feat(docs): add portable board export and comment anchors

wwwzzzcccc:
## Summary

- extend `docs scene export --image-format excalidraw` to save the authoritative live Board scene as a portable Excalidraw v2 file, embedding referenced attachments as guarded data URLs while preserving PNG/SVG behavior
- extend `docs comments add` with `--point` and repeatable `--element-id` flags for point, single-element, and multi-element Board anchors while preserving the existing `--data` and reply flows
- normalize bound text to its container and resolve point-aware linear-element geometry
- document both commands and register their schemas
- enforce the final 25 MiB import boundary before atomic output commit

## Linked Spec

Fixes #147

## How verified

- `go test ./... -count=1`
- `go vet ./...`
- `go build ./...`
- `git diff --check upstream/main..HEAD`
- real local-stack E2E: 52-element Board with Frame, embed, image, rich text, and connectors exported and re-imported with live element JSON preserved; image data embedded and re-uploaded
- real comment E2E: point, single element, multi-element, bound-text normalization, curved linear geometry, reply/edit/resolve/reopen/delete/list flows
- focused regressions cover 18 MiB portable round-trip, 19 MiB base64 overflow rejection, atomic no-output behavior, and POST-to-list anchor round-trips

## COMPREHENSION

**What does this change actually do?**

Before, the CLI could read/edit raw Board scenes, import `.excalidraw`, and export PNG/SVG, but could not produce a portable editable Board file or construct Web-compatible Board comment anchors. After, callers can save a live Board to a self-contained `.excalidraw` file and create point/element/multi-element comments without hand-building anchor payloads.

**What could break?**

The main risks are exporting a file that cannot be imported, downloading unsafe attachment URLs, losing attachment portability, or calculating anchors differently from Web. The implementation reuses the import size boundary and transfer guard, writes atomically, resolves anchors from the authoritative live scene, and has explicit regression coverage for size expansion, missing/deleted targets, bound text, and linear geometry. Existing import and comment flows remain compatible, and PNG/SVG scene export behavior is preserved.

**How do you know it works?**

In addition to the full Go test/vet/build gates, the commands were exercised against the local authenticated stack with a complex live Board. Export → import preserved the live scene, and each comment anchor form was created then read back through the comments API. The final branch is one commit directly on current `upstream/main`.

### Latest review fixes

- unified portable export under `docs scene export --image-format excalidraw` and Board anchors under `docs comments add`
[...]

PR #150, opened 2026-08-27
feat(docs): support Space-qualified members

### LINKED ISSUE #149: feat(docs): support Space-qualified member operations
## Problem Statement

Docs memberships are keyed by both user UID and principal Space. The CLI currently cannot select that Space for `docs members set`, and `docs members remove` omits the `principalSpaceId` query parameter now required by the backend. This makes removal fail and leaves same-UID, multi-Space memberships ambiguous.

## Proposed Solution

- Add optional `principalSpaceId` to the `docs.members.set` request body and expose `--principal-space-id`; omission keeps the Home Space behavior.
- Add required `principalSpaceId` to the `docs.members.remove` query and expose the same required flag.
- Document that the flag is mandatory when removing a UID that can exist in multiple Spaces.
- Preserve the existing member-list response fields and keep forward-grant bound to the document Space.
- Cover generated flags and wire payloads with tests.

## Alternatives Considered

Inferring the Space during removal is unsafe because a UID can have multiple membership rows. Extending forward-grant is unnecessary because forwarding intentionally grants within the document Space.

## Additional Context

The backend contract defaults member set to the document Home Space when `principalSpaceId` is omitted, but requires `principalSpaceId` for member removal.

wwwzzzcccc:
## Summary

- make Docs member mutations explicitly Space-qualified in the embedded OpenAPI contract and CLI documentation
- for Bot `docs members set`, omit `--principal-space-id` to use the authenticated Bot Space; explicitly selecting that same Space supports add/upsert, while a foreign Space can only update an existing exact membership and never performs a directory lookup
- require and send `--principal-space-id` for `docs members remove`, making deletion target the exact `(docId, uid, principalSpaceId)` row
- include `principalSpaceId` in the `docs members list` response schema and document the complete member-role vocabulary, including `commenter`
- use isolated command roots in set/remove tests so Cobra persistent flag state cannot leak between cases; leave document-Space-bound forward-grant behavior unchanged

## Linked Issue

Fixes #149

## How verified

- `go generate ./...` (no generated changes)
- `make fmt`
- `make vet`
- `golangci-lint v2.12.2 run`
- focused Docs member tests
- `make test` (`go test -race -count=1 ./...`)
- `make build`
- `git diff --check`
- built-CLI dry runs: set omitted / set qualified / remove omitted / remove qualified
- schema/help inspection: member-list items expose `principalSpaceId`; set help describes authenticated-Bot-Space default and foreign-Space exact-update behavior

## COMPREHENSION

1. **What invariant is load-bearing?** A Docs membership is identified by `(docId, uid, principalSpaceId)`, not by uid alone. Set and remove must preserve that qualified identity on the wire.
2. **What is the Bot set contract?** Omitting `principalSpaceId` delegates to the authenticated Bot Space. Supplying that same Space explicitly permits add/upsert. Supplying a foreign Space performs no directory lookup and can only update an already-existing exact row; an absent row returns `404`.
3. **Why is remove stricter?** A uid may have memberships in multiple Spaces, so Bot deletion must require `principalSpaceId` and delete only the exact qualified row. The CLI now fails locally before HTTP when the flag is absent.
4. **What failure would indicate a regression?** Emitting the optional set field when omitted, treating a foreign Space as directory-addable, allowing unqualified removal, sending the remove selector outside the query, or omitting `principalSpaceId` from list schema would violate the backend contract. Shape and wire tests pin these boundaries with fresh command roots.

## Compatibility / rollout

- `docs members set` remains source-compatible for callers that omit `--principal-space-id`; its server-side default is the authenticated Bot Space.
- `docs members remove` is intentionally breaking for unqualified calls: callers must add `--principal-space-id` to avoid ambiguous deletion.
- This CLI change depends on the matching Docs Backend contract in MR !110 and Server principal lookup behavior in PR #819; deploy those backend changes before relying on cross-Space exact updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,6 +128,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `--role` flags front the `shareScope` / `shareRole` wire keys.
 
 ### Changed
+- **`octo-cli docs members remove` now requires `--principal-space-id`** — Bot
+  membership deletion is an exact Space-qualified mutation. Requiring the
+  principal Space prevents an ambiguous uid from deleting the wrong row when
+  the same uid has memberships in multiple Spaces.
 - **`octo-cli html` adopts canonical create and document references** — create
   omits `slug`; the CLI generates `idempotency_key` once per invocation and
   reuses it across transport retries (an explicit key remains supported). It returns `data.doc_id` plus
--- a/README.md
+++ b/README.md
@@ -162,7 +162,11 @@ octo-cli docs export doc-123 --export-format pdf -o ./notes.pdf
 octo-cli docs scene export board-7 --image-format excalidraw -o ./board.excalidraw
 octo-cli docs comments add board-7 --body "Review this" --point 120,240
 # For --element-id anchors, dry-run still reads the live board scene to resolve geometry.
-octo-cli docs members set doc-123 --data '{"uid":"u-1","role":"writer"}'
+octo-cli docs members set doc-123 --uid u-1 --role writer
+# Member mutations are Space-qualified. Set defaults to the authenticated Bot
+# Space when omitted; removal always requires the exact principal Space.
+octo-cli docs members set doc-123 --uid u-1 --role writer --principal-space-id space-2
+octo-cli docs members remove doc-123 u-1 --principal-space-id space-2
 
 # Fleet/Loop uses the same gateway under /fleet/api/v1.
 octo-cli loop task list --workspace-id <workspace-id>
--- a/cmd/service/docs_test.go
+++ b/cmd/service/docs_test.go
@@ -122,6 +122,86 @@ func TestDocs_RegistryShape(t *testing.T) {
 			t.Errorf("%s: first segment must be the service; got %q", id, segs[0])
 		}
 	}
+
+	list, ok := reg.GetOperation("docs.members.list")
+	if !ok || list.ResponseSchema == nil {
+		t.Fatal("docs.members.list response schema missing")
+	}
+	items, ok := list.ResponseSchema.Properties["items"]
+	if !ok || items.Items == nil {
+		t.Fatalf("docs.members.list items schema = %#v, want array item schema", items)
+	}
+	principalSpaceID, ok := items.Items.Properties["principalSpaceId"]
+	if !ok || principalSpaceID.Type != "string" {
+		t.Fatalf("docs.members.list item principalSpaceId = %#v, want string", principalSpaceID)
+	}
+}
+
+func TestDocsMembersSet_PrincipalSpaceIDIsOptionalBodyField(t *testing.T) {
+	cases := []struct {
+		name      string
+		args      []string
+		wantSpace string
+	}{
+		{name: "defaults to authenticated Bot Space", args: []string{"docs", "members", "set", "doc-1", "--uid", "user-1", "--role", "writer"}},
+		{name: "sends explicit principal Space", args: []string{"docs", "members", "set", "doc-1", "--uid", "user-1", "--role", "writer", "--principal-space-id", "space-2"}, wantSpace: "space-2"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			var body map[string]any
+			root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+				if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
+					t.Fatalf("decode body: %v", err)
+				}
+				_, _ = w.Write([]byte(`{"ok":true}`))
+			})
+			root.SetArgs(tc.args)
+			if err := root.Execute(); err != nil {
+				t.Fatalf("execute: %v", err)
+			}
+			got, present := body["principalSpaceId"]
+			if tc.wantSpace == "" {
+				if present {
+					t.Fatalf("omitted --principal-space-id must leave Bot Space resolution to the backend: %v", body)
+				}
+				return
+			}
+			if got != tc.wantSpace {
+				t.Fatalf("principalSpaceId = %v, want %q", got, tc.wantSpace)
+			}
+		})
+	}
+}
+
+func TestDocsMembersRemove_RequiresAndSendsPrincipalSpaceID(t *testing.T) {
+	t.Run("requires principal Space", func(t *testing.T) {
+		requests := 0
+		root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+			requests++
+		})
+		root.SetArgs([]string{"docs", "members", "remove", "doc-1", "user-1"})
+		if err := root.Execute(); err == nil || !strings.Contains(err.Error(), "required flag") {
+			t.Fatalf("missing --principal-space-id error = %v, want required flag error", err)
+		}
+		if requests != 0 {
+			t.Fatalf("missing required flag sent %d requests, want 0", requests)
+		}
+	})
+
+	t.Run("sends exact principal Space", func(t *testing.T) {
+		var gotQuery url.Values
+		root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+			gotQuery = r.URL.Query()
+			_, _ = w.Write([]byte(`{"ok":true}`))
+		})
+		root.SetArgs([]string{"docs", "members", "remove", "doc-1", "user-1", "--principal-space-id", "space-2"})
+		if err := root.Execute(); err != nil {
+			t.Fatalf("execute: %v", err)
+		}
+		if got := gotQuery.Get("principalSpaceId"); got != "space-2" {
+			t.Fatalf("principalSpaceId query = %q, want space-2", got)
+		}
+	})
 }
 
 // TestDocsSceneExport_ImageFormatFlagNoGlobalShadow pins the fix for the
@@ -543,18 +623,22 @@ func TestDocsMembersSet_PutUpsertBody(t *testing.T) {
 
 // TestDocsMembersRemove_TwoPathArgs checks both positional args land in the path.
 func TestDocsMembersRemove_TwoPathArgs(t *testing.T) {
-	var gotMethod, gotPath string
+	var gotMethod, gotPath, gotSpace string
 	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
 		gotMethod, gotPath = r.Method, r.URL.Path
+		gotSpace = r.URL.Query().Get("principalSpaceId")
 		w.WriteHeader(200)
 		_, _ = w.Write([]byte(`{"ok":true}`))
 	})
-	root.SetArgs([]string{"docs", "members", "remove", "d1", "u9"})
+	root.SetArgs([]string{"docs", "members", "remove", "d1", "u9", "--principal-space-id", "s1"})
 	if err := root.Execute(); err != nil {
 		t.Fatalf("execute: %v", err)
 	}
 	if gotMethod != "DELETE" || gotPath != "/v1/bot/docs/d1/members/u9" {
 		t.Errorf("got %s %s, want DELETE /v1/bot/docs/d1/members/u9", gotMethod, gotPath)
+	}
+	if gotSpace != "s1" {
+		t.Errorf("principalSpaceId = %q, want s1", gotSpace)
 	}
 }
 
--- a/internal/registry/specs/docs.json
+++ b/internal/registry/specs/docs.json
@@ -668,6 +668,7 @@
                         "type": "object",
                         "properties": {
                           "uid": { "type": "string" },
+                          "principalSpaceId": { "type": "string", "description": "principal Space id that qualifies this membership" },
                           "role": { "type": "string" },
                           "source": { "type": "string" },
                           "grantedBy": { "type": "string" }
@@ -684,7 +685,7 @@
       "put": {
         "operationId": "docs.members.set",
         "summary": "Add or change a member's role by uid (upsert; needs admin)",
-        "description": "PUT-upsert: adds the member if absent or changes the role if present. The target uid must be a real Octo user or the backend returns 404 user_not_found.",
+        "description": "Space-qualified PUT mutation for Bot callers. When principalSpaceId is omitted, the backend uses the authenticated Bot Space. That Space may also be supplied explicitly and supports add or role update. A foreign principalSpaceId performs no directory lookup and can only update an existing exact (docId, uid, principalSpaceId) membership; a missing exact row returns 404.",
         "x-octo-risk": "write",
         "parameters": [
           { "name": "docId", "in": "path", "required": true, "schema": { "type": "string" } }
@@ -697,7 +698,8 @@
                 "type": "object",
                 "required": ["uid", "role"],
                 "properties": {
-                  "uid": { "type": "string", "description": "target user uid" },
+                  "uid": { "type": "string", "description": "uid in the Space-qualified membership key" },
+                  "principalSpaceId": { "type": "string", "x-octo-flag": "principal-space-id", "description": "principal Space id; for Bot calls, omission defaults to the authenticated Bot Space, an explicit matching value supports add/upsert, and a foreign value can only update an existing exact membership without a directory lookup" },
                   "role": { "type": "string", "enum": ["reader", "commenter", "writer", "admin"], "description": "role to grant. Matches octo-docs-backend isMemberRole (src/permission/role.ts); commenter is a stored role ranked between reader and writer, so omitting it made the local enum gate reject a call the backend accepts" }
                 }
               }
@@ -710,11 +712,13 @@
     "/v1/bot/docs/{docId}/members/{uid}": {
       "delete": {
         "operationId": "docs.members.remove",
-        "summary": "Remove a member (needs admin; the owner cannot be removed)",
+        "summary": "Remove a Space-qualified member (needs admin; the owner cannot be removed)",
+        "description": "Bot exact-delete mutation for the membership identified by (docId, uid, principalSpaceId). principalSpaceId is required because one uid can have distinct memberships in multiple Spaces.",
         "x-octo-risk": "write",
         "parameters": [
           { "name": "docId", "in": "path", "required": true, "schema": { "type": "string" } },
-          { "name": "uid", "in": "path", "required": true, "schema": { "type": "string" } }
+          { "name": "uid", "in": "path", "required": true, "schema": { "type": "string" } },
+          { "name": "principalSpaceId", "in": "query", "required": true, "x-octo-flag": "principal-space-id", "schema": { "type": "string" }, "description": "principal Space id of the membership to remove; required to disambiguate the same uid across Spaces" }
         ],
         "responses": { "200": { "description": "Removed" } }
       }
--- a/skills/octo-docs/common.md
+++ b/skills/octo-docs/common.md
@@ -110,9 +110,10 @@ Schema: `octo-cli schema docs.versions.restore`.
 ## Members & sharing
 
 ```bash
-octo-cli docs members list   <docId>                             # admin
-octo-cli docs members set    <docId> --uid <uid> --role writer   # add/upsert; role: reader|writer|admin
-octo-cli docs members remove <docId> <uid>                       # admin; owner cannot be removed
+octo-cli docs members list   <docId>                             # admin; items include principalSpaceId
+octo-cli docs members set    <docId> --uid <uid> --role writer   # authenticated Bot Space; role: reader|commenter|writer|admin
+octo-cli docs members set    <docId> --uid <uid> --role writer --principal-space-id <spaceId>
+octo-cli docs members remove <docId> <uid> --principal-space-id <spaceId> # required; owner cannot be removed
 
 # Grant-only forward access (never downgrades). Only reader|writer are grantable.
 octo-cli docs forward-grant  <docId> --uid <uid> --role reader
@@ -124,9 +125,15 @@ octo-cli docs share set <docId> --scope anyone_in_space --role read   # admin; e
 octo-cli docs share set <docId> --scope anyone_in_space --role edit   # admin; every space member gets edit
 ```
 
-`docs members set` is a PUT-upsert: it adds the member if absent or changes the
-role if already present. The target uid must be a real Octo user — a miss returns
-`404 user_not_found` and writes no ghost member. Schema: `octo-cli schema docs.members.set`.
+`docs members set` is a Space-qualified Bot mutation. Omit
+`--principal-space-id` to use the authenticated Bot Space, or pass that same
+Space explicitly; either form can add a member or update an existing role. A
+foreign Space is never queried through the directory and can only update an
+existing exact `(docId, uid, principalSpaceId)` row; if that row is absent, the
+backend returns `404`. A uid can have distinct member rows in multiple Spaces,
+so `docs members remove` always requires `--principal-space-id` and deletes the
+exact qualified row. Schemas:
+`octo-cli schema docs.members.set` and `octo-cli schema docs.members.remove`.
 
 `docs share set` changes the space-level share scope. Two scopes: `restricted`
 (only the owner and explicit members reach the doc — the default) and


Jerry-Xin (dismissed):
APPROVE — Space-qualified docs memberships are implemented correctly and fail-closed.

Verification at head 9dc49ae (byte-level review of all 4 files + local test run):

✅ Flags & fail-closed guards
- `docs.members.remove`: `principalSpaceId` declared `in: query, required: true` → cobra MarkFlagRequired fires ("required flag(s) … not set") when omitted, and the engine's `rejectEmptyRequiredValue` guard also rejects `--principal-space-id ""` locally, so the flag gate cannot be satisfied with an empty value.
- `docs.members.set`: optional body field — `applyBodyFlags` only merges Changed flags, so omission keeps the field off the wire and the backend Home Space default intact.
- No new leaf commands (existing operations only gain flags/params), so the registry-driven tree wiring is untouched; specs load at runtime via go:embed.

✅ Wire behavior (tests run locally, `-race`, all green)
- `TestDocsMembersSet_PrincipalSpaceIDIsOptionalBodyField`: body omits the field when the flag is absent; carries `space-2` when set.
- `TestDocsMembersRemove_RequiresAndSendsPrincipalSpaceID`: missing flag → required-flag error; set → `?principalSpaceId=space-2`.
- `TestDocsMembersRemove_TwoPathArgs` updated for the query param; full `go test -race -count=1 ./...` passes (cmd/service 25.4s).

✅ Backwards compatibility
- `members set` unchanged for existing callers (flag optional). `members remove` gains a required flag — a deliberate breaking change, but per #149 the backend already rejects removal without the query parameter, so previously-working invocations do not exist; the CLI now fails earlier with an actionable message instead of a backend error. README + embedded skill docs both updated.

💬 Non-blocking
- 🟡 skills/octo-docs/common.md:114 lists roles as `reader|writer|admin` while the schema enum also accepts `commenter` (pre-existing wording, not introduced here) — consider listing all accepted roles.
- 🟡 CI `check-sprint` is red: linked issue #149 has no Sprint set on the Octo board (needs a maintainer to assign Sprint W35, then re-run). `build & test (1.24.x)` was still pending at review time.

Jerry-Xin (changes requested):
Re-review at head 7e2c18c — delta vs previously approved 9dc49ae is a single-line description change at internal/registry/specs/docs.json:701 (`docs.members.set` requestBody `principalSpaceId`). That one line introduces a blocking self-contradiction in the published contract.

### 🔴 Blocking

- 🔴 Contract self-contradiction on `docs.members.set` `principalSpaceId` (internal/registry/specs/docs.json:701). The changed description now reads "optional for Bot calls and defaults to the authenticated Bot Space; an explicit value must match that Space". Every other statement about this same flag at this head says the opposite:
  - docs.json:687 (PUT operation description, unchanged): "Omit principalSpaceId to target the document's Home Space; pass it when updating an existing membership for the same uid in another Space."
  - README.md:166-168: "Select the principal Space explicitly for a non-Home-Space membership", with an example passing `--principal-space-id space-2`.
  - skills/octo-docs/common.md:129-131: "pass it to address an existing membership for the same uid in another Space."
  - Issue #149 (linked via Fixes): "The backend contract defaults member set to the document Home Space when principalSpaceId is omitted."
  - The PR body itself: "preserving the backend's Home Space default when omitted" (and COMPREHENSION #1: "set may omit it only to intentionally select the document Home Space").

  Under the new wording the flag is useless for Bot calls (it may only repeat the default), and agents consuming `octo-cli schema docs.members.set` now get guidance that directly contradicts the README, the embedded skill doc, and the linked issue — exactly the ambiguity this PR exists to remove. Please verify the true backend contract and reconcile all surfaces: either restore the Home-Space-default/other-Space-permitted wording at :701, or — if the backend genuinely scopes Bot calls to the authenticated Bot Space and rejects other values — update the operation description (docs.json:687), README, common.md, and PR body to match. Note: octo-docs-backend is currently unreachable (404) from this environment, so the backend semantics could not be byte-verified here; the author should confirm against the backend and align the texts in one direction.

### 💬 Non-blocking

- 🟡 skills/octo-docs/common.md:114 still lists `role: reader|writer|admin`, omitting the `commenter` role that the spec enum supports (pre-existing; carried from the previous round).
- 🟡 check-sprint CI is failing — process gate: a maintainer needs to assign the Sprint label before merge.

### ✅ Re-verified at this head (no regression from the delta)

- Delta vs 9dc49ae is exactly 1 line (the description above); no schema/required/enum/flag-wiring changes, so runtime behavior is unchanged.
- Fail-closed guards intact: `docs.members.remove` `principalSpaceId` required (cobra MarkFlagRequired + engine rejectEmptyRequiredValue); `docs.members.set` stays optional and merges only explicitly set fields.
- Local `go test -race -count=1 ./...` fully green (12 packages); CI build & test and npm test pass.

yujiawei (changes requested):
# Code Review — PR #150 (octo-cli)

Reviewed at head `7e2c18c862f6fbd19e7af774ef22482ac15986ad` against merge-base `afb2859`.

Verified locally on the PR head: `go build ./...`, full `go test ./...`, `go test -race ./cmd/service/`, and `go vet ./cmd/service/` are all green. I also built the binary and exercised the four wire paths with `--dry-run`, and ran three mutation tests against the new assertions (details below).

---

## 1. Spec compliance

**Spec: ❌**

Checked against the linked issue #149 (`feat(docs): support Space-qualified member operations`).

**Missing:** none of the five items in the issue's Proposed Solution are unimplemented. All are present and behave as claimed — see the wire verification below.

**Out of scope / added beyond the ask:** none. No extra flags, endpoints, or fields; the change touches only `docs.members.set` and `docs.members.remove`.

**Divergence:** two.

- **D1 — the shipped contract text for `principalSpaceId` states a different rule than the issue and the docs added in the same PR.** The issue says the flag selects the Space for a same-UID membership; `README.md:168` and `skills/octo-docs/common.md:115` show `--principal-space-id space-2`. But the property description that actually reaches users says the opposite. See P1-1.
- **D2 — the issue says "preserve the existing member-list response fields", and the PR does preserve them, but then adds a doc claim about a field that response does not declare** (`skills/octo-docs/common.md:113`). See P1-2.

**Wire behaviour verification (all confirmed, `--dry-run` on a build of this head):**

| Command | Result |
|---|---|
| `docs members set d1 --uid u1 --role writer` | body `{"role":"writer","uid":"u1"}` — field correctly absent |
| `... --principal-space-id s2` | body `{"principalSpaceId":"s2",...}` — correct body placement |
| `docs members remove d1 u1` | `VALIDATION_ERROR: required flag(s) "principal-space-id" not set` |
| `... --principal-space-id s2` | `DELETE /v1/bot/docs/d1/members/u1?principalSpaceId=s2` — correct query placement |
| `... --principal-space-id ""` | rejected by the existing empty-required-value guard (`cmd/service/run.go:446`) |

The mechanism is right: `x-octo-flag` is honoured on both a query parameter (`cmd/service/flags.go:154`) and a promoted body property (`cmd/service/flags.go:341`), and `required: true` on a query parameter becomes `MarkFlagRequired` (`cmd/service/flags.go:143`).

**Test quality — verified non-vacuous by mutation.** I reverted each half of the change and confirmed the new tests fail:

- dropping `"required": true` at `internal/registry/specs/docs.json:720` → `TestDocsMembersRemove_RequiresAndSendsPrincipalSpaceID` fails (`missing --principal-space-id error = <nil>`)
- dropping `x-octo-flag` at `internal/registry/specs/docs.json:701` → `TestDocsMembersSet_PrincipalSpaceIDIsOptionalBodyField` fails (`unknown flag: --principal-space-id`)

---

## 2. Code quality

### P1-1 — `principalSpaceId` ships two contradictory contracts, and the one users actually see is the restrictive one

`internal/registry/specs/docs.json:687` (operation description):

```
Omit principalSpaceId to target the document's Home Space; pass it when
updating an existing membership for the same uid in another Space.
```

`internal/registry/specs/docs.json:701` (property description):

```
target principal Space id; optional for Bot calls and defaults to the
authenticated Bot Space; an explicit value must match that Space
```

These are incompatible on two axes:

1. **Different defaults.** One says an omitted value targets *the document's Home Space*; the other says it defaults to *the authenticated Bot Space*. For a Bot whose Space is not the document's Home Space these are different membership rows, so `docs members set` with no flag has an undefined target — which is precisely the ambiguity this PR exists to remove.
2. **Different rules for an explicit value.** If "an explicit value must match that Space" is accurate, then a Bot in `space-1` cannot pass `space-2` at all, and the headline capability on the `set` side does not exist. `README.md:168` and `skills/octo-docs/common.md:115` both advertise exactly that call.

What makes this blocking rather than cosmetic is **which of the two texts reaches a user**. The registry never reads an operation's `description` — `internal/registry/loader.go:380` and `:460` read only `op["summary"]`, and there is no `op["description"]` anywhere in the loader. Only parameter and schema descriptions are read (`:562`, `:826`). Confirmed empirically on a build of this head:

```
$ octo-cli docs members set --help
      --principal-space-id string   target principal Space id; optional for Bot calls and
                                    defaults to the authenticated Bot Space; an explicit
                                    value must match that Space
```

`octo-cli schema docs.members.set` prints the same property description and no operation description. So the reconciling sentence added at `:687` is invisible to every consumer, and the only text an agent or operator can see is the one that contradicts the README, the skill, and the linked issue. The same applies to the new operation description at `:715` for `remove` (the updated `summary` there does surface, so that half is fine).

**Ask:** settle on one contract and put it in the property description. If the backend accepts an arbitrary Space id, correct `:701`. If it really requires the value to equal the authenticated Bot Space, then `README.md:168` and `skills/octo-docs/common.md:115` are wrong and the `set` half of the feature needs rescoping.

### P1-2 — `remove` now requires a value the CLI's own contract offers no way to obtain

`skills/octo-docs/common.md:113`, added in this PR:

```
octo-cli docs members list <docId>    # admin; items include principalSpaceId
```

The response schema for that operation declares four fields and `principalSpaceId` is not among them (`internal/registry/specs/docs.json:670-673`):

```json
"uid": { "type": "string" },
"role": { "type": "string" },
"source": { "type": "string" },
"grantedBy": { "type": "string" }
```

Confirmed on a build of this head — `octo-cli schema docs.members.list` returns exactly those four under `response_schema.properties.items.items.properties`.

This matters because `remove` is now hard-gated on the value, and `members list` is the only documented way to discover it. The same paragraph in the skill tells agents to introspect with `octo-cli schema`, which will not show the field. Either the claim is false and agents following the shipped skill hit a dead end, or the claim is true and the machine-readable contract this repo treats as the source of truth is now stale. Both are shipped defects, and both fixes are one line.

**Ask:** add `"principalSpaceId": { "type": "string" }` to the `members.list` item schema and pin it with a spec-shape assertion, or drop the claim from `common.md:113`.

### P2-1 — a breaking command-shape change with no changelog entry or stated compatibility rationale

`internal/registry/specs/docs.json:720` makes `--principal-space-id` mandatory, so `octo-cli docs members remove <docId> <uid>` — valid before this PR — now exits 2 locally. `CHANGELOG.md` has an `[Unreleased]` section and the project states it follows Keep a Changelog; a breaking change to an existing command surface is exactly what `### Changed` is for. Recent feature commits have also skipped the changelog, so this is partly house style — but a breaking change is the case worth the exception.

There is also a real ergonomic asymmetry: `set` lets a caller add a Home Space member without knowing any Space id, while `remove` forces them to fetch one. That is only justified if the backend genuinely rejects `DELETE` without the parameter. The PR body asserts this; nothing in the diff evidences it. **Please cite the backend contract** — if `DELETE` also defaults safely to the Home Space, `required: true` should be dropped for backward compatibility.

### P2-2 — the new `set` test silently depends on execution order

`cmd/service/docs_test.go:127-150` reuses one `root` across two `Execute()` calls. cobra flag state is sticky across `Execute()` on the same command, so the assertion that an omitted flag produces no body field only holds because the omitting case runs *first*. I verified this by swapping the two calls:

```
docs_test.go:148: omitted --principal-space-id must preserve backend Home Space
default: map[principalSpaceId:space-2 role:writer uid:user-1]
```

The current order is correct and the test is sound as written. But the invariant is load-bearing and uncommented, and a future third case appended after the "with flag" call would fail for a reason unrelated to the behaviour under test. A one-line comment, or a fresh root per case, would make it obvious. `cmd/service/docs_test.go:152-172` has the same latent coupling.

### P2-3 (nit, pre-existing) — required query flags are not marked `(required)` in `--help`

`cmd/service/flags.go:358-360` appends `(required)` to a required *body* flag's help; `registerQueryFlags` (`cmd/service/flags.go:98-146`) does not do the same for a required *query* flag. So `--uid` and `--role` on `set` read `(required)` while `--principal-space-id` on `remove` does not. This is an engine asymmetry that predates this PR — not a change request here, but this PR is the first `docs` command to depend on a required query flag, so it is worth a follow-up issue.

---

## 3. Overall verdict

**CHANGES_REQUESTED**

The mechanism, the wire behaviour, and the tests are all correct and I could not break them. Both blockers are contract-text defects rather than logic defects — but in a metadata-driven CLI whose consumers are agents introspecting `--help` and `octo-cli schema`, the spec text *is* the API. Shipping a flag whose help says "must match the authenticated Bot Space" alongside a README that passes a different Space is a defect in the deliverable, and it is roughly two lines to clear.

## 4. Suggested changes

1. Pick one contract for `principalSpaceId` on `set` and write it into the property description at `docs.json:701`. Move anything load-bearing out of the operation `description`, which no consumer reads.
2. Add `principalSpaceId` to the `members.list` item schema (`docs.json:670-673`), or remove the claim at `common.md:113`.
3. Add a `### Changed` changelog entry noting that `docs members remove` now requires `--principal-space-id`, and state in the PR description which backend behaviour makes it mandatory for `DELETE` but optional for `PUT`.
4. Comment the ordering invariant in `docs_test.go:127-150`, or build a fresh root per case.

## 5. Additional observations

- The `check-sprint` check is red, but for a process reason — the linked issue has no Sprint assigned on the org board. Not a code problem; a maintainer needs to set it.
- Every other check is green, and I reproduced build, test, race and vet locally on this head.
- `docs members set` moving from `--data '{...}'` to `--uid/--role` in `README.md:165` is a genuine readability improvement, independent of the rest of this change.

**Coverage note — what I could not verify:** backend semantics are outside this repository, so both blockers ultimately hinge on a contract I cannot read: whether `DELETE /members/{uid}` truly rejects a request without `principalSpaceId`, whether `PUT` accepts an arbitrary Space id, and whether the live `GET /members` response carries `principalSpaceId`. I also did not search for consumers of the old two-argument `remove` form outside this repository (other services, the published npm wrapper, operator runbooks); the breaking change could reach them.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #150 (octo-cli)
Head SHA: `7e2c18c862f6fbd19e7af774ef22482ac15986ad` | Base: `main` (merge-base `afb2859a`) | Round: 1

## Summary

This PR adds Space-qualified membership support to the `docs` domain: `docs members set` gains an optional `--principal-space-id` body field (`principalSpaceId`, Home Space default preserved when omitted), and `docs members remove` gains a required `--principal-space-id` query parameter so a uid with memberships in multiple Spaces is unambiguous. The change is spec-driven — only `internal/registry/specs/docs.json`, wire-level tests, README, and the embedded docs skill change; no Go engine code is touched. Per linked issue #149 the backend already rejects removal without `principalSpaceId`, so the new required flag fixes a path that was already broken server-side rather than breaking a working one.

## Verification

Local checks at head `7e2c18c8` (executed in this environment):
- `go test ./... -count=1` — all packages pass
- `go test ./cmd/service/ -run 'TestDocsMembers|TestDocs_RegistryShape|TestFlags_NoSpecCollides'` — pass
- ✅ **Spec→runtime propagation** — loader parses `required`/`x-octo-flag` for both shapes (`internal/registry/loader.go:563` params, `:827` body props); flags register via `cmd/service/flags.go` and values flow in `cmd/service/run.go`.
- ✅ **Test realism** — `rootWithService` wires the real binary registry (`cmd/service/service_test.go:87`, `registry.MustNew`), so the new tests exercise the production command-build path, not a stub.

## Findings

### P2 — Role enum comment omits `commenter` (`skills/octo-docs/common.md:114`)

Diff-scope: pre-existing wording on a line this PR edits (base already read `role: reader|writer|admin`; PR changed only the `# Home Space` prefix). The `docs.members.set` spec accepts `["reader","commenter","writer","admin"]` (`internal/registry/specs/docs.json:702` area, role property), and its description explicitly notes `commenter` is a stored rank between reader and writer. An agent following the skill doc would believe `--role commenter` is invalid even though the CLI and backend accept it. Not R1: no working path becomes unusable and no wrong data is produced. Fix: update the comment to `reader|commenter|writer|admin` while the line is already being touched.

## Suggestions

- Fold the `commenter` enum fix above into this PR (one-line change).
- Optional, out of scope: a required string flag set to the empty value (`--principal-space-id ""`) passes cobra's required check and would send an empty query param; this is engine-generic behavior for all required query flags and the backend rejects it, so no action needed here.

## Additional findings

- README (`README.md:165-169`) and `skills/octo-docs/common.md:113-134` are mutually consistent with the new spec; `skills/skills_test.go:53` content assertions still hold (suite passes).
- CHANGELOG.md was not updated, consistent with the two immediately preceding feature PRs (#139, #148); flagged as an observation only.
- Note: the dual-model fan-out scripts referenced by the review skill installation are absent in this environment; this report is the single-leg result.

## Data flow traceback

- `principalSpaceId` (set body): spec property → loader `schemaInfoFromNode` (`internal/registry/loader.go:827`) → `registerBodyFlags` (`cmd/service/flags.go:337` region, kindString, flag from `x-octo-flag`) → `applyBodyFlags` merges only `Changed` flags (`cmd/service/run.go:693`) → key genuinely absent on the wire when the flag is omitted. Test `TestDocsMembersSet_PrincipalSpaceIDIsOptionalBodyField` asserts both branches.
- `principalSpaceId` (remove query): spec param (`in=query, required=true`) → loader `appendParams` (`internal/registry/loader.go:563`) → `registerQueryFlags` + `MarkFlagRequired` (`cmd/service/flags.go:145` region) → `buildQuery` emits only `Changed` flags (`cmd/service/run.go:415`). Test asserts the `required flag` error and the exact query value `space-2`.
- Wire key stays the spec name `principalSpaceId` in both body and query; the flag alias is only CLI-facing. No empty-array/early-return/gating short-circuit anywhere in the chain; runtime data provably reaches the consumption points.

## Blind-spot checklist

- **C1 dual-path parity (set↔remove)**: hit → cleared. Asymmetry is intentional and documented in three places (spec descriptions, README, common.md): set may default to Home Space because upsert has a safe default target; remove must not guess because the membership key is `(docId, uid, principalSpaceId)`. `rg` confirms no other in-repo caller of `docs members remove` that the new required flag could break.
- **C2 control-flow ordering / reuse**: N/A — no shared engine code changed; the spec is consumed per-operation by unchanged loader/flag machinery. Flag name `principal-space-id` checked against `reservedFlagNames`/engine/path/query/header collisions; `TestFlags_NoSpecCollidesWithEngineFlags` passes.
- **C3 authorization boundary**: cleared — no auth surface changed; the spec states an explicit `principalSpaceId` must match the authenticated Bot Space (backend-enforced), so the CLI gains no cross-Space capability.
- **C4 authorization lifecycle / cascade**: N/A — no status/container-membership state changes.
- **C5 build/note pass ≠ runtime path**: hit → cleared — full `go test ./...` executed at head (not just static review) plus code-level trace of the generated-flag path above.
- **C6 governance/docs self-consistency**: hit → cleared except the P2 `commenter` omission; README/common.md/spec cross-checked against each other.
- **R3 distributed side effects**: N/A — no notification/event/count/cache changes.

## Cross-round blocker re-check

N/A — first review round for this PR; no prior blockers to re-verify.

## Verdict: APPROVED

The spec→loader→flags→wire chain is proven end-to-end at head `7e2c18c8` with the full test suite passing, the required-flag behavior on remove matches the backend contract documented in issue #149, and docs are synchronized. The only finding is a P2 docs nit; recommended verdict for final sign-off: APPROVED.

yujiawei (approved):
# Code Review — PR #150 (octo-cli)

Reviewed at head `6e3c4a6249ee4baf492d9a7673bae464b0df072a` against merge-base `afb2859`. This is the third review round; the two blockers raised at `7e2c18c` are both fixed, and I verified each fix on a build of this head rather than by reading the diff alone.

**Verification run locally at this head:** `go build ./...`, `go vet ./...`, `gofmt -l` (clean), `go test ./...` (12 packages green), `go test -race -count=1 ./cmd/service/` (green, 34s), `go generate ./...` (no diff). I also built the binary and exercised every wire path with `--dry-run`, and mutation-tested all three new assertions.

---

## 1. Spec compliance

**Spec: ✅**

Checked against linked issue #149 (`feat(docs): support Space-qualified member operations`).

**Missing: none.** All five items of the issue's Proposed Solution are implemented:

| Issue requirement | Where | Verified |
|---|---|---|
| optional `principalSpaceId` on `members.set` body + `--principal-space-id` | `internal/registry/specs/docs.json:702` | body omits the key when the flag is absent; carries it when set |
| required `principalSpaceId` on `members.remove` query + required flag | `internal/registry/specs/docs.json:721` | `MarkFlagRequired` fires; value lands in the query string |
| document that the flag is mandatory for removal | `README.md:167,169`, `skills/octo-docs/common.md:116,134`, `CHANGELOG.md:131` | present |
| preserve existing member-list response fields | `internal/registry/specs/docs.json:668-674` | `uid` / `role` / `source` / `grantedBy` all retained; `principalSpaceId` added alongside |
| cover generated flags and wire payloads with tests | `cmd/service/docs_test.go:126-207,623-645` | see mutation results below |

**Out of scope: none.** No new leaf commands, endpoints, or unrelated flags; the change touches only `docs.members.list` / `.set` / `.remove`. Forward-grant is left document-Space-bound as the issue requires.

**Divergence: none blocking.** One wording note, see Q4.

**Wire behaviour — all confirmed with `--dry-run` on a build of this head:**

| Command | Observed |
|---|---|
| `docs members set d1 --uid u1 --role writer` | `PUT /v1/bot/docs/d1/members` body `{"role":"writer","uid":"u1"}` — field correctly absent |
| `... --principal-space-id s2` | body `{"principalSpaceId":"s2","role":"writer","uid":"u1"}` — correct body placement |
| `docs members remove d1 u1` | `VALIDATION_ERROR: required flag(s) "principal-space-id" not set`, **0 requests issued** |
| `... --principal-space-id s2` | `DELETE /v1/bot/docs/d1/members/u1?principalSpaceId=s2` — correct query placement |
| `... --principal-space-id ""` | rejected by the empty-required-value guard with an actionable hint |
| `--data '{"principalSpaceId":"sX",...}'` | raw-body path still works; flag and `--data` compose as documented |

**Test quality — verified non-vacuous by mutation.** I reverted each half of the change and confirmed the matching test fails:

- delete `principalSpaceId` from the list item schema (`docs.json:671`) → `TestDocs_RegistryShape` fails at `docs_test.go:136`
- flip `"required": true` → `false` on the remove query param (`docs.json:721`) → `TestDocsMembersRemove_RequiresAndSendsPrincipalSpaceID/requires_principal_Space` fails at `docs_test.go:184`
- drop `x-octo-flag` from the set body property (`docs.json:702`) → `TestDocsMembersSet_PrincipalSpaceIDIsOptionalBodyField/sends_explicit_principal_Space` fails with `unknown flag: --principal-space-id`

All three fail on mutation and pass on the real spec, so the assertions are load-bearing.

---

## 2. Code quality

**Quality: Approved**

### Previous-round blockers — both cleared, verified empirically

**B1 — contradictory `principalSpaceId` contract → fixed.** `docs.json:688` (operation) and `:702` (property) now state the same rule, and `README.md:166-167` and `skills/octo-docs/common.md:128-135` match it. More importantly the reconciled text now lives in the **property** description, which is the surface a consumer actually sees:

```
$ octo-cli docs members set --help
      --principal-space-id string   principal Space id; for Bot calls, omission defaults to the
                                    authenticated Bot Space, an explicit matching value supports
                                    add/upsert, and a foreign value can only update an existing
                                    exact membership without a directory lookup
```

**B2 — `members list` did not declare `principalSpaceId` → fixed.** `docs.json:671` adds it, and `octo-cli schema docs.members.list` now returns it under `response_schema.properties.items.items.properties`. The claim at `common.md:113` is now backed by the machine-readable contract, and `docs_test.go:126-137` pins it so it cannot silently regress. The discovery path for the value `remove` now requires is therefore closed.

**Also cleared:** `CHANGELOG.md:129-133` documents the breaking change with a rationale; the new tests build a fresh command root per subtest (`docs_test.go:152,179,193`), so the cobra-flag-state ordering coupling flagged last round is gone; `common.md:114` now lists `commenter` alongside the other roles, matching the enum at `docs.json:703`.

### P2 — non-blocking

**Q1 — the `set` summary still advertises an unconditional upsert.** `docs.json:687` reads `Add or change a member's role by uid (upsert; needs admin)` and the response at `:709` is `"Upserted"`. Under the contract this PR ships, that is only true for the authenticated Bot Space; a foreign `principalSpaceId` is update-only and 404s on a missing row. The summary is the headline line of `octo-cli docs members set --help` and of `octo-cli schema docs.members.set` — the flag description immediately below does qualify it, but the two prominent strings pull in different directions. Suggest narrowing the summary, e.g. `Add or update a member's role (Space-qualified; foreign Space is update-only)`.

**Q2 — operation-level `description` is dead text in this CLI.** The loader only reads `op["summary"]` (`internal/registry/loader.go:380` and `:460`); there is no `op["description"]` read anywhere, and parameter/schema descriptions come from separate paths (`:562`, `:826`). So the newly added descriptions at `docs.json:688` and `:716` never reach `--help` or `octo-cli schema`. Not a defect here — the load-bearing text was correctly duplicated into the property and parameter descriptions this round — but worth knowing so future contract text is not parked in a field no consumer reads.

**Q3 — required *query* flags are not marked `(required)` in help.** `cmd/service/flags.go:359` appends `(required)` for a required body property, but the query-parameter registration path does not. Result on this PR's own command:

```
$ octo-cli docs members remove --help
      --principal-space-id string   principal Space id of the membership to remove; ...
```

— no `(required)` marker, while `--uid` and `--role` on `set` both carry one. This is a pre-existing engine asymmetry, not something introduced here, but `docs members remove` is the first `docs` command to hard-gate on a required query flag, so the gap is now user-visible. Worth a follow-up issue against `flags.go` rather than a change in this PR.

**Q4 — the linked issue's wording is now stale relative to the shipped contract.** Issue #149 says set "defaults to the Home Space when `principalSpaceId` is omitted"; this PR consistently says the **authenticated Bot Space** on all four surfaces. The PR description explains why (dependent backend contract MR !110 and server PR #819) and the four CLI surfaces agree with each other, so this is internally coherent — but a reader arriving from #149 will see two different defaults. Suggest a one-line note on #149 recording the clarified semantics.

**Q5 — `README.md:168` omits the foreign-Space caveat.** The README shows `docs members set doc-123 --uid u-1 --role writer --principal-space-id space-2` under a comment that only says mutations are Space-qualified. A reader could reasonably take that example as adding a member into `space-2`, which the contract says it cannot do unless the row already exists. `common.md:128-135` states this correctly; one clause in the README comment would close the gap.

**Q6 — the relationship between `--space` and `--principal-space-id` is undocumented.** Both `docs.members.set` and `.remove` report `space_header_set: true`, so for a platform-scoped bot the acting Space arrives via the `X-Space-Id` header from `--space`, while the membership Space arrives via `--principal-space-id`. Since the documented default for the latter is "the authenticated Bot Space", the two are related but never connected in prose. One sentence in `common.md` would prevent an operator from setting `--space` and expecting it to qualify the membership.

### On the breaking change to `remove`

Making `--principal-space-id` mandatory turns a previously-accepted invocation into a local exit-2, which normally deserves scrutiny. I do not think it blocks here: issue #149 states the backend **already** requires the query parameter and that unqualified removal currently *fails*, so there is no working invocation to break — the CLI now fails earlier with an actionable message instead of surfacing a backend error, and `CHANGELOG.md:131` records it. I checked for other consumers inside the repo (`README.md`, `skills/`, `docs/`, `npm/`): only the two doc surfaces reference the command, and both are updated. The residual risk is that this rests on the issue text rather than on backend code I can read — see the coverage note.

---

## 3. Overall verdict

**APPROVE**

Both blockers from the previous round are genuinely fixed, not papered over, and I confirmed each on a running binary rather than from the diff. The mechanism is correct on the wire in all six cases I exercised, the new assertions all fail under mutation, and the full suite is green under `-race`. Everything remaining is P2 documentation polish.

## 4. Suggested changes (none blocking)

1. Narrow the `set` summary at `docs.json:687` so the headline does not promise an unconditional upsert (Q1).
2. Add one clause to the `README.md:166-167` comment noting that a foreign Space can only update an existing membership (Q5).
3. Open a follow-up issue for `(required)` on required query flags in `cmd/service/flags.go` (Q3).
4. Note the clarified Bot-Space-vs-Home-Space default on issue #149 (Q4).

## 5. Additional observations

- `check-sprint` is red for a process reason — the linked issue has no Sprint on the board. Not a code problem; a maintainer needs to set it before merge.
- `build & test (1.24.x)` was still pending at review time; I reproduced build, vet, test, race and `go generate` locally on this head and all are green.
- The skills package tests assert only that certain command strings appear in the embedded markdown (`skills/skills_test.go:52`); they do not validate the examples against the registry. So skill-doc drift is caught by review, not by CI. The spec-shape assertion added in this PR is a good precedent for closing that class of gap.

**Coverage note — what was not verified.** Backend semantics live outside this repository and could not be read from this environment, so three claims rest entirely on the spec text and the linked issue: that an omitted `principalSpaceId` resolves to the authenticated Bot Space, that a foreign Space is update-only with no directory lookup, and that `DELETE` genuinely rejects unqualified calls. The live `GET /members` response was likewise not observed, so the added list-schema field is documentation rather than an observed contract (harmless either way — the CLI does not filter response fields). I also did not look for consumers outside this repository (other services, operator runbooks); the `remove` change would reach them. Automated analysis passes over this diff covered the schema/contract surface and the breaking-change question but did not run the test suite, mutation-test the assertions, or exercise the built binary; those three were done by hand here.

Jerry-Xin (approved):
Third-round re-review at head 6e3c4a6249ee. Both round-2 blockers are resolved; APPROVE.

## Resolution of prior blockers

### 🔴 (r2, Jerry-Xin) — principalSpaceId wording contradiction — RESOLVED
At the previous head, `internal/registry/specs/docs.json` described `docs.members.set` defaulting to the authenticated Bot Space while every other surface said the document's Home Space. This head harmonizes ALL surfaces to the finalized backend contract, byte-verified:
- `internal/registry/specs/docs.json` — operation description, body property description, and remove description/required query param all describe: omission defaults to the authenticated Bot Space; an explicit matching Space supports add/upsert; a foreign Space is lookup-free and can only update an existing exact (docId, uid, principalSpaceId) row, 404 when absent.
- `README.md`, `skills/octo-docs/common.md`, `CHANGELOG.md` — same wording family, no residual "Home Space" occurrences anywhere in the tree (grep-verified).
- Cross-checked against the authoritative backend contract (octo-docs-backend MR !110, "keep the finalized PUT behavior unchanged: Bot own-Space targets use exact bearer validation and may upsert even on a foreign-Home document; explicit foreign-Space targets are lookup-free and update-only") and its members route code, which computes the Bot default from the bearer-authenticated Space. The harmonized wording is the true contract; issue #149's "Home Space" sentence reflects the superseded draft contract (see non-blocking notes).

### 🔴 (r2, yujiwei) — member-list field declared but absent from response contract — RESOLVED
At the previous head, docs declared `members list` items include `principalSpaceId` while the `docs.members.list` response schema omitted it. This head adds `principalSpaceId` to the list item response schema and pins it in `TestDocs_RegistryShape` (fails if the field is missing or not a string). Independently verified the real backend list handler already returns `principalSpaceId` per item (and on the synthesized owner item), unchanged by MR !110 — so the schema declaration matches the actual response.

## Guards re-verified (no regression)
- `docs.members.remove` principalSpaceId remains required: spec `required: true` on the query param drives cobra `MarkFlagRequired`, and `rejectEmptyRequiredValue` rejects an empty value that would otherwise slip past the required-flag gate. Tests pin: missing flag errors locally with zero requests; explicit value sent exactly as the query parameter.
- `docs.members.set` principalSpaceId remains optional (body requires only uid + role); `applyBodyFlags` merges only Changed flags, so omission sends nothing and the backend default applies. Tests pin both the absent-key and explicit-value wire payloads, now on isolated Cobra roots to prevent persistent-flag leakage between cases.

## Verification
- Full `go test ./...` green at this head (12 packages); `-race` green for the Docs member tests. Build clean.
- CI: build & test (1.24.x) pass, npm test pass, scans pass. `check-sprint` fails as a process gate only — linked issue #149 is not assigned to the current sprint on the Octo Board; needs a Maintainer, not a code change.

💬 Non-blocking
- 🔵 Issue #149's body still carries the superseded "defaults to the document Home Space" sentence. Not a shipped surface, but updating it (or adding a pointer to the finalized contract) would prevent future confusion.
- 🔵 Merge-order reminder (already disclosed in the PR body): the documented cross-Space semantics depend on octo-docs-backend MR !110 and octo-server #819, both still open; deploy those backend changes before relying on them. The CLI wire behavior (omit unless explicitly passed) is safe on the current backend either way.
- 🔵 Consider `minLength: 1` on the set body's `principalSpaceId` so an explicit empty string fails locally; today remove already rejects empty via the required-value gate, while set would send an empty string that the backend treats as omission.

mochashanyao (approved):
[Octo-Q · automated review]

---

PR: Mininglamp-OSS/octo-cli #150 — feat(docs): support Space-qualified members
Head SHA: `6e3c4a6249ee4baf492d9a7673bae464b0df072a` ｜ merge-base: `afb2859a6a420308da48d589a393c2d359eb718c` (origin/main)
变更面: 5 files, +116/−13（spec/docs/tests，无手写命令代码，docs 服务为 spec 驱动）

---

# Code Review — PR #150 (octo-cli)

## Summary

This PR makes Docs member mutations Space-qualified in the embedded OpenAPI contract (`internal/registry/specs/docs.json`), the CLI documentation, and the tests: `docs members set` gains an optional `principalSpaceId` body field (flag `--principal-space-id`; omission delegates to the authenticated Bot Space), `docs members remove` gains a required `principalSpaceId` query parameter so deletion targets the exact `(docId, uid, principalSpaceId)` row and fails locally before any HTTP when the flag is missing, and the `docs members list` response schema now exposes `principalSpaceId`. Because the docs service is spec-driven, all of this is picked up by the generic engine — the core review question is whether the declared contract actually reaches the wire. It does.

## Verification

Local checks at head `6e3c4a62`:
- `go build ./...` — clean
- `go test -race -count=1 ./cmd/service/ -run 'TestDocs'` — pass, including the new `TestDocsMembersSet_PrincipalSpaceIDIsOptionalBodyField` and `TestDocsMembersRemove_RequiresAndSendsPrincipalSpaceID`
- `go test -race -count=1 ./...` — all packages pass

Contract traced from spec through engine to wire: `internal/registry/loader.go:563`/`:827` parse `x-octo-flag` on parameters and body properties; `cmd/service/flags.go:143` marks the remove query flag required via `MarkFlagRequired`; `buildQuery`/`applyBodyFlags` in `cmd/service/run.go` emit only Changed flags, so an omitted set field is genuinely absent from the body rather than sent empty.

## Findings

### P2 — Breaking-change entry omits the backend deployment-order dependency (`CHANGELOG.md:131`)

The remove entry documents the new required `--principal-space-id` but not that exact `(docId, uid, principalSpaceId)` deletion only takes effect once octo-docs-backend MR !110 and octo-server PR #819 land; before that, an upgraded CLI sends a query param an old backend may ignore, so an operator can believe they performed an exact qualified delete while the backend still deletes by unqualified `(docId, uid)`. Repo convention already carries such notes in the same file (`CHANGELOG.md:126`, docs.share.set: "backend lands in octo-docs-backend#68; merge this after that endpoint ships"). Add the same dependency/ordering sentence to the remove entry.

- Set omission path: omitted `--principal-space-id` leaves `principalSpaceId` absent from the PUT body (verified by httptest wire capture), preserving the backend Bot-Space default.
- Remove local gate: missing flag errors before any HTTP request (`requests == 0`), and `--principal-space-id ""` is rejected by `rejectEmptyRequiredValue`.
- No flag collision: `principal-space-id` is not in `reservedFlagNames`, and the two new declarations sit on separate leaves (set body vs remove query).
- `docs members list` response schema exposes `principalSpaceId` (`internal/registry/specs/docs.json:671`) and the registry-shape test pins it.
- Docs parity: `README.md:165`-`169` examples and `skills/octo-docs/common.md:113`-`136` match the spec's flag names, role enum (`reader|commenter|writer|admin`), and required-flag semantics; forward-grant behavior is untouched.
- Existing `TestDocsMembersRemove_TwoPathArgs` was updated to pass the now-required flag, so no stale test asserts the old unqualified call.

## Verdict: APPROVED

The contract is declared, wired, and wire-tested end to end; the intentional breaking change to remove is documented, locally enforced, and justified. The single P2 is a CHANGELOG completeness note and is not blocking.

---

## 1. 验证结论

- ✅ 契约声明→线上请求全链路贯通：`x-octo-flag: principal-space-id` 在 set（body 可选）与 remove（query 必填）均被通用引擎真实消费，非空转声明（证据见下方数据流回溯）。
- ✅ 新增测试全部通过（-race，httptest 真实抓 body/query/path，非 hand-feed）。
- ✅ 全仓 `go build` + `go test -race ./...` 通过。
- ⚠️ `CHANGELOG.md:131` breaking-change 条目缺后端部署顺序依赖说明（P2，见上）。

## 2. 发现问题（diff-scope 三问 + 定级）

**P2 — CHANGELOG remove 条目缺部署顺序说明**（`CHANGELOG.md:131`）
1. new/amplified/pre-existing：**new**（本 PR 新增条目）。
2. 本 PR 是否改了该路径行为契约：是——本 PR 有意将 remove 从"不带限定删除"改为"必填 --principal-space-id 的精确删除"，条目记录了 breaking 本身，但漏记精确语义依赖后端（MR !110 / server #819）先上线。
3. 降级判断：不满足 R1（不让任何现有路径不可用、不产生错误数据——CLI 先升级时 remove 仍按旧后端语义工作，只是用户可能误以为已精确删除）→ 文档完备性问题，P2 非阻塞。

无 P0/P1。R1 逐项核对：旧的"省略 principalSpaceId 的 set"线上行为完全不变（body 不带该键）；remove 的 breaking 是有意、有文档、有本地前置校验、有测试的契约变更，属设计内变更而非回归；list 为纯增量 schema。R3（分布式副作用）：无通知/推送/计数/缓存改动，N/A。

## 3. 建议

- 在 `CHANGELOG.md:131` remove 条目末尾补一句部署顺序说明，套用同文件 `:126`（docs.share.set）的既有写法："backend lands in octo-docs-backend MR !110 + octo-server PR #819; deploy those before relying on exact qualified deletion"。set 的行为变化说明可一并补。
- 无需代码改动：引擎接线正确，测试已钉住 wire 行为。

## 4. 额外发现

- README 示例从 `--data '{"uid":...}'` 改为 `--uid/--role` flag 形式——两种形式均有效（flag override --data，架构 §5.2），非缺陷。
- `role` enum 含 `commenter` 为本 PR 之前的既有状态（diff 中为 context 行），README/common.md 的角色词表与之一致。

## 5. 数据流回溯

| 被消费数据 | 上游来源 | 是否真流到消费点 |
|---|---|---|
| set body `principalSpaceId` | `docs.json:702` 声明 → `loader.go:827` 解析 FlagName → `flags.go` registerBodyFlags 注册 bodyFlag{apiName:"principalSpaceId"} → `run.go` applyBodyFlags 仅 Changed 时写入 | ✅ 有：带 flag 时 body 含该键；省略时键不存在（httptest 抓包双向验证，测试过） |
| remove query `principalSpaceId` | `docs.json:721`（required:true, in:query）→ `loader.go:563` 解析 Required+FlagName → `flags.go:143` MarkFlagRequired → `run.go` buildQuery 仅 Changed 写入 + rejectEmptyRequiredValue 拒空 | ✅ 有：带 flag 时 `?principalSpaceId=space-2`；缺 flag 本地报错且 0 请求；空值被拒（测试过） |
| list item `principalSpaceId` | `docs.json:671` → `loader.go:513` firstSuccessSchema → ResponseSchema.Properties["items"].Items.Properties | ✅ 有：TestDocs_RegistryShape 断言存在且为 string（过） |
| docId/uid 位置参数 | path params（无 x-octo-flag → 纯位置参数，与本 PR 前一致） | ✅ 有：TwoPathArgs 测试验证 `DELETE /v1/bot/docs/d1/members/u9` |

未发现提前 return / 门控短路 / 空数组陷阱：引擎对 docs 无特判，全为通用路径。

## 6. 盲点 checklist（C1–C6）

- **C1 双路径 parity**：hit & clear——set↔remove 成对核对：两者均声明 `x-octo-flag: principal-space-id`，wire 名一致；set 可选 / remove 必填的不对称是后端契约的有意设计（set 缺省回退 Bot Space，remove 必须消歧），两端均有 wire 级测试，无漏边。
- **C2 control-flow ordering / 嵌套复用**：clear——引擎注册顺序（path→query→header→body）未变；新 flag 名过碰撞门（非 reserved、无同 leaf 冲突）；flagcollision/pathflag 套件随全量测试通过。
- **C3 授权边界 ≠ 能力边界**：N/A——无新 endpoint/tool 暴露，同一 bot 鉴权端点；外部 Space 的授权语义归后端（MR !110），CLI 未放大任何能力。
- **C4 授权生命周期 / 容器-成员级联**：N/A（CLI 层）——级联校验属后端职责（MR !110 / server #819），CLI 仅透传限定符。
- **C5 build/test 过 ≠ 运行期路径正确**：clear（有运行期证据）——测试经真实 cobra root + 真实内嵌 registry + httptest 抓真实 HTTP 请求（body/query/path/method），非 hand-feed 数据；-race 全量通过。
- **C6 治理/文档自洽性**：hit——README/common.md/CHANGELOG/spec 逐条交叉比对，语义一致；唯一缺口即上述 P2（CHANGELOG 漏部署顺序说明）。

## 7. 跨轮 blocker 复检（R6）

N/A——本 issue 无先前评论，这是本 head SHA 的首轮automated review。提示终审：GitHub 侧该 PR 的 reviewDecision 现显示 CHANGES_REQUESTED（可能针对更早 head），本腿按约束未读 GitHub 旧评论，请终审在发对外 verdict 前对当前 head 独立复核并 reconcile 先前 review 状态。

---

[Octo-Q] verdict: APPROVE — 契约声明→引擎接线→线上行为全链路验证通过（真实构建 + -race 全量测试 + httptest wire 抓包），无 P0/P1；仅 1 条 P2（CHANGELOG 缺后端部署顺序说明，非阻塞）。按 R4：仅 P2/nit → APPROVE。

yujiawei:
Follow-up to my approving review — one additional non-blocking suggestion I missed, verdict unchanged.

**P2 — the `remove` changelog entry omits the backend deployment-order note this file otherwise carries.**

`CHANGELOG.md:131-134` documents the breaking change itself but not its backend dependency. That file already has this convention twice:

- `CHANGELOG.md:125-126` — "This targets the planned `/v1/bot/docs/{docId}/share` endpoint (backend lands in octo-docs-backend#68); merge this after that endpoint ships."
- `CHANGELOG.md:150-151` — "**Minimum rollout dependency:** release only after the canonical-create changes in octo-docs-backend#166 and octo-docs-html#33 are merged and deployed."

The PR description already names the dependency ("depends on the matching Docs Backend contract in MR !110 and Server principal lookup behavior in PR #819; deploy those backend changes before relying on cross-Space exact updates"), so this is just moving a sentence you have already written into the file that a future operator will actually read.

One thing worth pinning down while you add it: the PR description and the linked issue currently imply different backend states. Issue #149 says the backend **already** requires `principalSpaceId` on removal ("`docs members remove` omits the `principalSpaceId` query parameter now required by the backend. This makes removal fail"), which would mean the unqualified call is already broken today. The PR description instead frames the backend work as a pending dependency. Both cannot be true for the `remove` path, and which one holds determines whether this CLI change is a fix for an already-broken command or a change that must land after the backend. A sentence in the changelog entry stating which would settle it.

Not blocking — the code, wiring, and tests are unaffected either way.

PR #151, opened 2026-09-01
fix(loop): rename expert team evaluation

liuooo:
## What

The Loop registry generated `task team-evaluation record` from a singular, weakly typed endpoint contract.

## How

- Rename the operation to `expert_team.evaluate` and use the plural `/tasks/{task_id}/expert_team_evaluations` resource.
- Type `outcome` as `action | no_action | failed` and expose `reason` as a normal generated flag.
- Verify the generated `loop expert-team evaluate` command, request path, workspace header, and JSON body.

## Tests

- `GOCACHE=/private/tmp/octo-go-build-cache go test ./cmd ./internal/...`
- `GOCACHE=/private/tmp/octo-go-build-cache go run ./cmd/octo-cli loop expert-team evaluate --help`
- `git diff --check`

## Dependency

This is a stacked PR targeting `test/autopilot-public-contract-integration`, which contains the pending Loop Public API contract required by this operation.

### DIFF
--- a/cmd/service/service_test.go
+++ b/cmd/service/service_test.go
@@ -167,6 +167,50 @@ func TestLoopCommandUsesUnifiedGatewayAndModulePath(t *testing.T) {
 	}
 }
 
+func TestLoopExpertTeamEvaluateCommandShape(t *testing.T) {
+	var gotMethod, gotPath, gotWorkspace string
+	var gotBody map[string]any
+	root, _, _ := rootWithService(t, func(w http.ResponseWriter, r *http.Request) {
+		gotMethod = r.Method
+		gotPath = r.URL.Path
+		gotWorkspace = r.Header.Get("X-Workspace-ID")
+		if err := json.NewDecoder(r.Body).Decode(&gotBody); err != nil {
+			t.Fatalf("decode request body: %v", err)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusCreated)
+		_, _ = w.Write([]byte(`{"id":"evaluation-1"}`))
+	})
+	loop := findCmd(root, "loop")
+	expertTeam := findCmd(loop, "expert-team")
+	if expertTeam == nil || findCmd(expertTeam, "evaluate") == nil {
+		t.Fatal("loop expert-team evaluate command not registered")
+	}
+	task := findCmd(loop, "task")
+	if task != nil && findCmd(task, "team-evaluation") != nil {
+		t.Fatal("legacy loop task team-evaluation hierarchy must not be generated")
+	}
+
+	root.SetArgs([]string{
+		"loop", "expert-team", "evaluate", "task-1",
+		"--workspace-id", "workspace-1",
+		"--outcome", "no_action",
+		"--reason", "nothing to dispatch",
+	})
+	if err := root.Execute(); err != nil {
+		t.Fatalf("loop expert-team evaluate: %v", err)
+	}
+	if gotMethod != http.MethodPost || gotPath != "/fleet/api/v1/tasks/task-1/expert_team_evaluations" {
+		t.Fatalf("request = %s %s", gotMethod, gotPath)
+	}
+	if gotWorkspace != "workspace-1" {
+		t.Fatalf("X-Workspace-ID = %q", gotWorkspace)
+	}
+	if gotBody["outcome"] != "no_action" || gotBody["reason"] != "nothing to dispatch" {
+		t.Fatalf("request body = %#v", gotBody)
+	}
+}
+
 func TestLoopTaskWorkspaceHeaderFlag(t *testing.T) {
 	var gotPath, gotWorkspace string
 	gateway := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
--- a/internal/registry/loader_test.go
+++ b/internal/registry/loader_test.go
@@ -2,6 +2,7 @@ package registry
 
 import (
 	"reflect"
+	"slices"
 	"strings"
 	"testing"
 )
@@ -160,7 +161,7 @@ func TestLoopExtendedBusinessContract(t *testing.T) {
 		"loop.skill.list",
 		"runtime.update_request.get",
 		"autopilot.delivery.replay",
-		"task.team_evaluation.record",
+		"expert_team.evaluate",
 	} {
 		op, ok := r.GetOperation(id)
 		if !ok {
@@ -170,6 +171,24 @@ func TestLoopExtendedBusinessContract(t *testing.T) {
 		if op.Service != "loop" || !strings.HasPrefix(op.Path, "/fleet/api/v1/") {
 			t.Errorf("%s: got service=%q path=%q", id, op.Service, op.Path)
 		}
+	}
+
+	evaluation, ok := r.GetOperation("expert_team.evaluate")
+	if !ok {
+		t.Fatal("expert_team.evaluate: not found")
+	}
+	if evaluation.Path != "/fleet/api/v1/tasks/{task_id}/expert_team_evaluations" {
+		t.Fatalf("expert_team.evaluate path = %q", evaluation.Path)
+	}
+	if evaluation.RequestBody == nil {
+		t.Fatal("expert_team.evaluate request body missing")
+	}
+	outcome, ok := evaluation.RequestBody.Properties["outcome"]
+	if !ok || !slices.Equal(outcome.Enum, []any{"action", "no_action", "failed"}) {
+		t.Fatalf("expert_team.evaluate outcome schema = %+v", outcome)
+	}
+	if _, ok := evaluation.RequestBody.Properties["reason"]; !ok {
+		t.Fatalf("expert_team.evaluate reason schema missing: %+v", evaluation.RequestBody.Properties)
 	}
 
 	op, ok := r.GetOperation("label.list")
--- a/internal/registry/specs/loop.json
+++ b/internal/registry/specs/loop.json
@@ -6338,14 +6338,14 @@
         }
       }
     },
-    "/fleet/api/v1/tasks/{task_id}/expert_team_evaluation": {
+    "/fleet/api/v1/tasks/{task_id}/expert_team_evaluations": {
       "post": {
-        "summary": "Record expert team evaluation",
-        "description": "Record expert team evaluation in the authenticated workspace.",
+        "summary": "Evaluate a task as expert team leader",
+        "description": "Record the authenticated Expert Team Leader's evaluation outcome for the current Task execution.",
         "tags": [
           "tasks"
         ],
-        "operationId": "task.team_evaluation.record",
+        "operationId": "expert_team.evaluate",
         "parameters": [
           {
             "name": "task_id",
@@ -6362,7 +6362,24 @@
           "content": {
             "application/json": {
               "schema": {
-                "$ref": "#/components/schemas/FlexibleRequest"
+                "type": "object",
+                "additionalProperties": false,
+                "required": [
+                  "outcome"
+                ],
+                "properties": {
+                  "outcome": {
+                    "type": "string",
+                    "enum": [
+                      "action",
+                      "no_action",
+                      "failed"
+                    ]
+                  },
+                  "reason": {
+                    "type": "string"
+                  }
+                }
               }
             }
           }


Jerry-Xin (approved):
This in-scope PR correctly updates the Loop contract and generated command without blocking issues.

✅ Highlights

- Correct command rename to `loop expert-team evaluate`.
- Correct plural endpoint and workspace header propagation.
- Request schema requires `outcome`, exposes `reason`, and documents valid outcomes.
- Tests cover registry metadata, legacy-command removal, request method/path/header/body.
- Focused test suite, help generation, and `git diff --check` pass.

yujiawei (approved):
# Code Review — PR #151 (octo-cli)

Reviewed at head `f13491a`, against base `test/autopilot-public-contract-integration` (merge-base `3de3fc4`). 3 files, +86/-6.

## 1. Spec compliance

**Spec: ✅**

The PR body states three deliverables; all three land, and nothing beyond them was touched.

| Stated deliverable | Status | Evidence |
|---|---|---|
| Rename to `expert_team.evaluate`, plural resource | done | `internal/registry/specs/loop.json:6341,6348` |
| Type `outcome` as `action \| no_action \| failed`, expose `reason` as a generated flag | done | `internal/registry/specs/loop.json:6365-6382` |
| Verify generated command, path, workspace header, JSON body | done | `cmd/service/service_test.go:170-213`, `internal/registry/loader_test.go:176-192` |

- **Missing:** none.
- **Out of scope / unrequested:** none. The diff is confined to one spec path node plus its two tests.
- **Deviation:** none.

No dangling references to the old identifiers remain anywhere in the tree — `grep -rn 'team.evaluation\|team_evaluation\|team-evaluation'` returns only the new spec node and the two tests.

## 2. Code quality

**Quality: Approved** (no P0/P1; four P2s below)

### Verification performed

- `go test ./...` — all packages green, including `TestAllDomainOperationCounts` (`loop: 126`, unchanged by a rename).
- `gofmt -l ./cmd ./internal` — clean.
- `go run ./cmd/octo-cli loop expert-team evaluate --help` — command registers as expected; `--outcome`, `--reason`, `--workspace-id`, `--data` all present.
- `--dry-run` with `--workspace-id ws-1` and no `--outcome` → `request body is missing required field(s): outcome`, exit 2. The `required: ["outcome"]` declaration is enforced locally (`cmd/service/run.go:253-266`), so the required-ness is not help-text-only.
- `--dry-run` with `--outcome` set and `--reason` omitted → body is exactly `{"outcome":"..."}`. The `Changed()` gate at `cmd/service/run.go:202` keeps unset optional flags off the wire, so the optional `reason` cannot leak an empty string.
- Wire shape confirmed: `POST {base}/fleet/api/v1/tasks/task-1/expert_team_evaluations`, header `X-Workspace-ID: ws-1` (inherited from the path-level `WorkspaceIDHeader` `$ref`).

### P2 — `--outcome` enum is help-text only; invalid values reach the wire

`internal/registry/specs/loop.json:6371-6377` declares the enum, but nothing validates it client-side:

```go
// cmd/service/run.go:243
func validateBodySchema(schema *registry.SchemaInfo, value any, path string) string {
	switch schema.Type {
	case "object":  return validateObjectSchema(schema, value, path)
	case "array":   return validateArraySchema(schema, value, path)
	}
	return ""   // primitives: no enum / minLength / pattern check
}
```

Reproduced:

```
$ octo-cli loop expert-team evaluate task-1 --workspace-id ws-1 --outcome bogus --dry-run
{"data":{"body":{"outcome":"bogus"}, ...},"ok":true}
```

**Not a blocker, and not introduced here.** This is a repo-wide characteristic of the generator: there are 15 body-level enums across 12 pre-existing operations (`docs.members.set role`, `html.reply status`, `message.search sort`, `matter.transition status`, …) and none of them is validated either. The declared `400 ValidationError` response covers it server-side. Flagging it because this is the first typed enum in `loop.json`, so the spec now promises more than the CLI enforces. If local enum rejection is wanted, it belongs in a separate change to `validateBodySchema` that fixes all 12 operations at once — not in this PR.

### P2 — `additionalProperties: false` is silently dropped by the loader

`registry.SchemaInfo` (`internal/registry/loader.go:170-186`) has no `AdditionalProperties` field, and `resolveSchemaWithDepth` never reads the key. Consequences:

- `octo-cli schema expert_team.evaluate` emits the object without the constraint, so an agent introspecting the operation cannot see that extra keys are rejected.
- `--data '{"outcome":"action","typo":true}'` is forwarded verbatim; the rejection only happens server-side.

The constraint is therefore a server-contract statement, not an enforced one. That is fine as documentation — just be aware it buys nothing on the client today.

### P2 — `evaluate` sits under `expert-team` but takes a task id

```
$ octo-cli loop expert-team --help
  create        Create expert team          <expert-team-id>-shaped siblings
  delete        Delete expert team
  evaluate      Evaluate a task as expert team leader   <-- takes <task-id>
  get           Get expert team
  update        Update expert team
```

`expert_team.evaluate` is the only operation in `loop.json` whose operationId prefix does not match its URL root resource — all 34 `task.*` operations map to `/tasks/...` and all other `expert_team.*` operations map to `/expert_teams/...`; this one is `expert_team.*` over `/tasks/{task_id}/...`. For a CLI whose primary consumer is an agent, having `evaluate <task-id>` adjacent to `get <expert-team-id>` / `update <expert-team-id>` under the same parent is an easy id mix-up, and the error surfaces as a server 404 rather than a local validation error.

`task.expert_team_evaluation.create` (→ `loop task expert-team-evaluation create <task-id>`) would match the sibling convention of `task.comment.create` / `task.metadata.set`. The new test at `cmd/service/service_test.go:189-192` explicitly forbids that hierarchy, so this looks deliberate — if the grouping was chosen for prompt parity with the expert-team-leader flow, that rationale is worth a one-line comment in the spec node so the exception is not later "fixed" back.

### P2 — no `description` on either body property

`outcome` and `reason` have no `description`, which produces semantically empty help:

```
      --outcome string        (one of: action, no_action, failed) (required)
      --reason string
```

`--reason` in particular tells an agent nothing about when to supply it (e.g. whether it is expected when `outcome=failed`). The larger specs in this repo do carry descriptions on inline body properties (docs.json 39/39, html.json 21/25, message.json 43/58). Two short strings would close the gap.

### Note on the new tests

`TestLoopExpertTeamEvaluateCommandShape` and the `loader_test.go` additions do real work — they pin the generator wiring (operationId → command path, path-param positional, path-level workspace header, body assembly) and the negative assertion that the legacy `loop task team-evaluation` subtree is gone is a genuinely good guard. Be clear-eyed about what they do *not* establish: both assert that the CLI matches a spec this same PR edits, so they are silent on whether the Fleet backend actually serves the plural resource. That confidence has to come from the server side (see below).

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. The change is well scoped, the tests pass, the rename leaves no dangling references, and the four findings above are all advisory.

## 4. Suggestions

1. Add `description` to `outcome` and `reason` in the spec node.
2. If the `expert_team.*` grouping over a `/tasks/...` path is intentional, add a one-line note in the spec so the exception survives future consistency passes.
3. Before this branch reaches `main`, confirm against the Fleet deployment that the resource is plural (`/expert_team_evaluations`) and that `action | no_action | failed` is the server's actual accepted set. Nothing in this repository can establish that — the spec is hand-maintained with no sync tooling.
4. Optional follow-up (not this PR): teach `validateBodySchema` to check `enum` on primitives, which would benefit all 12 operations that currently declare enums without enforcement.

## 5. Additional observations (outside the diff)

- **No CHANGELOG entry exists for the entire `loop` domain.** `git log origin/main..origin/test/autopilot-public-contract-integration -- CHANGELOG.md` is empty, yet the branch adds 126 operations and a new `octo-loop` skill. This predates PR #151 and is not its job to fix, but it should be closed before the integration branch merges. Note the precedent already in `CHANGELOG.md` for the `summary` domain: the spec ships behind `x-octo-disabled` with an explicit "withheld until the backend PR is merged and deployed" note. If the plural Fleet route is not yet live, that mechanism is the established way to land the spec safely.
- **`CLAUDE.md:37` still says "9 domains, 105 operations."** With `loop` the branch is at 10 domains / 260 operations. Again a base-branch gap, not this PR's.
- **The new command is absent from `skills/octo-loop/SKILL.md`.** Since the operation is explicitly agent-facing ("Record the authenticated Expert Team Leader's evaluation outcome"), an agent following the skill will not discover it. The old singular version was equally undocumented, so this is not a regression — but the rename is a natural moment to add it.

## 6. Review coverage and its limits

Two independent assistive analysis passes were run alongside this review; both completed. Where they disagreed with the evidence, the evidence won:

- Both passes independently flagged the rename as a **breaking change requiring a migration notice**. **Refuted.** The entire `loop` domain is unreleased: `git tag --contains d78bd6c` returns zero tags, and `git merge-base --is-ancestor d78bd6c origin/main` is false. The old `task.team_evaluation.record` operation exists only on this unmerged integration branch (added in `d78bd6c`, one of the 10 commits this branch carries that `main` does not). There are no shipped consumers to break, and pre-release is precisely the right time to fix the name.
- One pass argued `additionalProperties: false` would break clients sending undocumented extra fields. **Refuted** on the same no-consumers grounds, and additionally because the loader discards the constraint (see P2 above).
- One pass raised that the test omits the `--reason`-absent case and might hit zero-value mapping. **Refuted empirically** — the `Changed()` gate at `cmd/service/run.go:202` omits unset flags entirely; the dry-run body contains only `outcome`. Still worth one extra assertion for regression safety, but there is no defect today.

**Not covered:** live Fleet backend behavior (path plurality, accepted `outcome` values, whether `reason` is conditionally required when `outcome=failed`, and the 201 response body shape) — the Fleet service is not in this org and could not be inspected; sibling-PR interaction on the integration branch beyond the single spec node this PR edits; and the runtime/deployment sequencing between this CLI branch and the backend rollout.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #151 (octo-cli)

## Summary

This PR renames the Loop expert-team evaluation operation in the CLI's embedded OpenAPI registry and tightens its contract: the path moves from `/fleet/api/v1/tasks/{task_id}/expert_team_evaluation` to the plural `.../expert_team_evaluations`, the operationId moves from `task.team_evaluation.record` to `expert_team.evaluate`, and the request body switches from the generic `FlexibleRequest` ref to a strict schema — required `outcome` enum (`action` | `no_action` | `failed`), optional `reason`, `additionalProperties: false`. Two test updates pin the new contract: registry-level assertions on the parsed schema (`internal/registry/loader_test.go:176`) and an end-to-end shape test of the generated `loop expert-team evaluate` command including a negative assertion that the legacy `loop task team-evaluation` hierarchy is gone (`cmd/service/service_test.go:170`).

The change is a clean, complete rename: I traced the new spec data through the metadata-driven generator end to end and found no defect introduced or exposed by this diff.

## Verification

Static analysis only at head `f13491a`; build and tests not executed in this environment. Note: the second-model (GLM-5) review leg timed out twice at the gateway; per degradation policy this review proceeded on the single (Octo-Q) leg.

- ✅ **Command tree generation** — `expert_team.evaluate` splits to `["expert_team","evaluate"]`; the first segment differs from service `loop`, so it becomes the `expert-team` group (`cmd/service/service.go:87`), reusing the existing expert-team CRUD group via `findChild`; leaf is `evaluate <task-id>`.
- ✅ **`--workspace-id` reachability** — the operation's own parameter list only has `task_id`, but the path item declares `#/components/parameters/WorkspaceIDHeader` at path level, and `operationParameters` merges path-level parameters into every operation (`internal/registry/loader.go:384`), so the flag exists and `X-Workspace-ID` is emitted when set. This is the one spot where a surface read of the diff suggests a missing flag; the merge proves it wired.
- ✅ **Required-body enforcement** — `validateRequiredBodyFields` runs before the empty-body short-circuit, so `loop expert-team evaluate <id>` without `--outcome` fails locally as a validation error (`cmd/service/run.go:222`); partial writes cannot reach the server.
- ✅ **201 handling** — spec declares `201 → Resource` and the client treats any 2xx as success (`internal/client/client.go:306`), matching the shape test's 201 handler.
- ✅ **No stale references** — repo-wide search for `team_evaluation.record`, the singular path, and the old command hierarchy matches only the negative test assertion. `FlexibleRequest` still has 21 other consumers, so no orphaned component.
- ✅ **No cross-spec collision** — `expert_team.evaluate` exists only in `loop.json`; go 1.24 covers the new `slices` import in the test.

## Findings

No P0/P1 issues; no P2 issues either. Zero findings on the completed review leg.

### Additional observations (non-blocking, no action required)

- Enum values and `additionalProperties: false` are not enforced client-side (`validateBodySchema` only checks required fields and nested object/array shape). An invalid `--outcome` is forwarded and rejected server-side with a 400. This is the pre-existing design for every enum-bearing operation — the flag help lists the allowed values — so it is not a defect of this PR, just where the enforcement boundary lives.
- `task_id` is declared inline on the operation instead of `$ref`-ing the shared `TaskID` component that sibling task paths use. Behaviorally identical (the loader handles both); a future tidy-up at most.

## Data-flow trace

Each consumed datum in the diff traced to its upstream source:

1. `evaluation.RequestBody.Properties["outcome"].Enum` (loader test) ← inline `requestBody.content.application/json.schema` parsed by `resolveSchemaWithDepth` (`internal/registry/loader.go:590`). No `$ref` hop needed; data flows.
2. `X-Workspace-ID` on the wire (service test) ← path-item-level `WorkspaceIDHeader` param → `operationParameters` merge → `registerHeaderFlags` → `buildHeaders` (set-only-when-flag-set) → `client.attempt` header loop. Data flows; the operation-level-only view is the trap here.
3. Request body `{outcome, reason}` ← `--outcome`/`--reason` flags (Changed-only) merged over any `--data` base by `applyBodyFlags`, then required-field validated, then JSON-marshalled in `client.Do`. Data flows.
4. Success envelope from 201 ← client 2xx gate. Flows.
5. Absence of legacy `task team-evaluation` commands ← the old operation no longer exists in the registry, and `attachOperation` only walks present operations. Verified.

## Blind-spot checklist

- **C1 dual-path parity** — cleared. Rename is atomic: old operationId, singular path, and old CLI hierarchy all removed together; grep confirms no asymmetric remainder.
- **C2 control-flow ordering / reuse** — cleared. The shared generator (`attachOperation`/`registerBodyFlags`) consumes the new spec data like every other operation; the pre-existing `expert-team` group command is reused without duplicate registration; no security control is involved that would need non-canonical probing.
- **C3 authorization boundary** — cleared. No capability exposure changes; the workspace selector header stays `required: true`; authorization itself is server-side and untouched.
- **C4 authorization lifecycle / cascade** — N/A. No container/member-state logic in a CLI spec rename.
- **C5 build/test ≠ runtime** — runtime path was reasoned explicitly (flag registration → body assembly → local validation → request → status gate) rather than relying on "tests pass"; tests/build were not executed in this environment.
- **C6 governance/docs consistency** — cleared. No governance docs touched; `skills/octo-loop/SKILL.md` and docs never referenced the old evaluation command, so no drift is introduced.

## Cross-round blocker re-check (R6)

N/A — first review round for this PR.

## Verdict: APPROVED

The rename is complete repo-wide, the tightened request schema is enforced client-side for its required field, the generated command shape is pinned by an end-to-end test that exercises the real embedded registry, and the one subtle wiring point (path-item-level workspace header) is proven intact. Nothing in this diff makes a working path unavailable, produces wrong data, or regresses prior behavior.

yujiawei:
### Review addendum — one more P2 (verdict unchanged: still APPROVED)

Following up on my review above with one consistency point I missed on the first pass. Non-blocking, but it is slightly more than cosmetic, so it is worth a line.

**`task_id` is declared inline instead of `$ref`-ing the shared `TaskID` component.**

`loop.json` defines a shared parameter component:

```json
"TaskID": {
  "name": "task_id", "in": "path", "required": true,
  "description": "Stable task identifier.",
  "schema": { "type": "string", "format": "uuid" }
}
```

Of the 21 places a `task_id` path parameter is declared across `/tasks/{task_id}/...`, 20 use `$ref: "#/components/parameters/TaskID"`. `expert_team.evaluate` (`internal/registry/specs/loop.json:6350-6359`) is the only one that hand-rolls it — and the hand-rolled copy drops the component's `description`.

That is observable, not just stylistic, because the description reaches the agent-facing introspection output:

```
$ octo-cli schema task.comment.create      # sibling, uses $ref
  { "description": "Stable task identifier.", "in": "path", "name": "task_id", ... }

$ octo-cli schema expert_team.evaluate     # this PR, inline
  { "in": "path", "name": "task_id", "required": true, "type": "string" }
```

Behaviour is identical (the loader resolves both forms), so nothing breaks. But swapping the inline block for the `$ref` is a strictly smaller diff than what is there now, restores parity in `schema` output, and folds into suggestion #1 in my review — the same root cause, which is that the new spec node was authored without the descriptions its siblings carry.

Suggested fix, replacing the inline `parameters` block:

```json
"parameters": [
  { "$ref": "#/components/parameters/TaskID" }
]
```

PR #152, opened 2026-09-01
docs: guide agents to ingest external document images

luoshu0211:
## Summary

Document the safe Agent workflow for inserting external images, including SVG, into rich-text Octo documents.

## Changes

- Route attachment URL-ingest tasks from the top-level octo-docs skill to common.md.
- Require external image URLs to be re-hosted through `/attachments/ingest` before writing an image node.
- Show the durable `attachId` image-node shape with a positive `width` compatibility default.
- Tell agents to degrade failed ingests to links instead of persisting broken external image nodes.
- Extend embedded-skill tests to lock the new guidance.

## Motivation

An Agent can use `docs content edit` to persist an image containing only an external `src`. Off-whitelist hosts then render as **Image unavailable**. Separately, the current client can lay out an attachment-backed image with `width: null` as `0x0`. The skill should teach the complete ingest → attachId → image-node flow instead of leaving these constraints implicit.

This is documentation and regression coverage for Agent guidance; it does not change the frontend layout implementation or add a generated `docs attachments ingest` command.

## Linked Spec

No separate issue. The guidance follows the existing docs attachment-ingest and image-node contracts.

## Testing

- [x] `go vet ./...` passes
- [x] `go build ./...` passes

## Checklist

- [x] Code and documentation are in English
- [x] N/A — no new command was added to the README command table
- [x] N/A — no command list changed in CLAUDE.md
- [x] New guidance includes embedded-skill regression checks

## COMPREHENSION

The load-bearing rule is that `src` is not a durable substitute for an Octo attachment: external HTTP(S) images must first be ingested into the target document and referenced by that document-scoped `attachId`. Current clients also need a positive image `width` for predictable rendering; the skill uses the known width when available and `300` as a compatibility default. This depends on the existing backend ingest endpoint and frontend attachment resolver. It was verified against a real test document, where an external SVG produced **Image unavailable**, the ingested SVG resolved from Octo COS, and adding `width` changed its rendered box from `0x0` to `150x150`.

### DIFF
--- a/skills/octo-docs/SKILL.md
+++ b/skills/octo-docs/SKILL.md
@@ -25,7 +25,7 @@ All commands call `$OCTO_API_BASE_URL/v1/bot/docs/*`.
 | Read/edit a rich-text **document body** (`doc_type: doc`): incremental block ops | **`doc.md`** |
 | Read/edit a **whiteboard** (`doc_type: board`): scene elements/files, image export | **`board.md`** |
 | Continue from a searchable **HTML document** (`doc_type: html`): resolve its document reference, then use immutable versions/drafts/assets/comments | **`../octo-html/SKILL.md`** |
-| Cross-cutting features — **comments** (doc range or sheet cell), **versions** (snapshot/restore), **members & sharing**, **attachments** (presign/upload) | **`common.md`** |
+| Cross-cutting features — **comments** (doc range or sheet cell), **versions** (snapshot/restore), **members & sharing**, **attachments** (presign/upload and external-image ingest) | **`common.md`** |
 
 > The first four split by `doc_type` (what kind of document you're handling);
 > `common.md` holds the features that apply across kinds. Read a reference with
--- a/skills/octo-docs/common.md
+++ b/skills/octo-docs/common.md
@@ -6,7 +6,7 @@ you need:
 - [Comments](#comments) — add/list/reply/resolve, on a doc text range or a sheet cell
 - [Versions](#versions) — snapshot / preview / rename / delete / restore
 - [Members & sharing](#members--sharing) — roles, forward-grant
-- [Attachments](#attachments) — presign + upload file/image bytes
+- [Attachments](#attachments) — presign/upload local bytes and ingest external images
 
 All commands call `$OCTO_API_BASE_URL/v1/bot/docs/*`. Auth & space rules are in
 `SKILL.md`.
@@ -182,3 +182,46 @@ Once uploaded, reference the `attachId` from a document body's image node (see
 `doc.md`) or a whiteboard file ref (see `board.md`). NOTE: a spreadsheet's floating
 images do NOT use this attachment flow — they inline the base64 bytes in the
 drawing's `source` field (see `sheet.md`). Schema: `octo-cli schema docs.attachments.presign`.
+
+### External image URLs (including SVG)
+
+Never persist an arbitrary external URL as an image node's only `src`. Current
+Octo clients render images only from trusted storage hosts, so a node such as
+`{"type":"image","attrs":{"src":"https://third-party.example/image.svg"}}`
+can be stored but renders as **Image unavailable**.
+
+Re-host a public HTTP(S) image under the target document first. The ingest
+endpoint is not yet a generated `docs attachments` subcommand, so call it
+through the generic API passthrough:
+
+```bash
+octo-cli api POST /v1/bot/docs/<docId>/attachments/ingest \
+  --data '{"urls":["https://source.example/diagram.svg"]}'
+```
+
+Read the new document-scoped `attachId` from `data.mappings`; treat an entry in
+`data.notIngested` as a failed image import and keep it as a normal hyperlink or
+text instead of writing a broken image node. The backend fetches the source with
+SSRF protections, validates that the bytes are an allowed image, and stores the
+successful result in Octo object storage.
+
+After either upload or ingest succeeds, write the image node using its durable
+`attachId`. Omit `src` (or set it to `null`): the client resolves a fresh signed
+display URL from the attachment. Also provide a positive pixel `width`; current
+clients can collapse an agent-created image with `width: null` to `0x0`. Use the
+known display width when available, otherwise `300` is a safe compatibility
+default:
+
+```json
+{
+  "type": "image",
+  "attrs": {
+    "attachId": "att_xxx",
+    "width": 300,
+    "alt": "Diagram"
+  }
+}
+```
+
+The `attachId` must belong to the same target document; a foreign or unknown
+attachment is rejected with `422 attachment_not_found`.
--- a/skills/octo-docs/doc.md
+++ b/skills/octo-docs/doc.md
@@ -64,8 +64,35 @@ board/whiteboard/sheet), `413 too_many_ops` / `413 op_content_too_large` /
 attachment that is not this doc's (`attachment_not_found`), or content the
 schema rejects (`schema_incompatible`).
 
-Inserting an image into the body references an uploaded attachment — see
-`common.md` (Attachments) for the presign → upload → reference flow.
+## Image nodes
+
+Do not write a third-party image URL as the image node's only `src`: the Octo
+client rejects off-whitelist asset hosts and displays **Image unavailable**.
+First upload local bytes or ingest a public HTTP(S) URL into the target document
+as described in `common.md` (Attachments), then use the returned document-scoped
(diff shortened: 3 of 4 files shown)


Jerry-Xin (approved):
The PR is relevant, accurate, and consistent with the existing Octo document workflow.

💬 Non-blocking

- 🔵 Suggestion: In `skills/skills_test.go:53-56`, also assert guidance for `data.notIngested` fallback and omitting `src`; these are important safety rules currently not protected against regression.

✅ Highlights

- Correctly routes external-image ingestion to `common.md`.
- Clearly documents ingest → `attachId` → image-node handling.
- Safely degrades failed imports to hyperlinks or text.
- Documents the positive-width compatibility requirement.
- Relevant tests pass: `go test ./skills ./cmd/...`.

yujiawei (approved):
# Code Review — PR #152 (octo-cli)

Reviewed at head `9aa6d52eced3d87003f708db0c164aa631c8f088` against `main`. Docs-only change (3 skill markdown files + 1 test file), 79 additions / 6 deletions.

I verified the CLI-side claims against the source at this SHA rather than taking them on trust, and reproduced the stated test results locally: `go build ./...`, `go vet ./...`, and `go test ./skills/...` all pass.

---

**Spec: PASS**

There is no linked issue, so I checked the diff against the PR description's own five stated changes.

| Claimed change | Where | Status |
|---|---|---|
| Route attachment URL-ingest tasks from `SKILL.md` to `common.md` | `skills/octo-docs/SKILL.md:28` | Done |
| Require external URLs to be re-hosted via `/attachments/ingest` first | `skills/octo-docs/common.md:186-206` | Done |
| Show the durable `attachId` node shape with a positive `width` default | `common.md:208-224`, `doc.md:67-98` | Done |
| Degrade failed ingests to links instead of broken image nodes | `common.md:202-204` | Done |
| Extend embedded-skill tests to lock the new guidance | `skills/skills_test.go:46-57` | Done |

- **Missing:** none.
- **Extra / out of scope:** none. The PR states it does not add a generated `docs attachments ingest` command and does not touch the frontend, and the diff matches — no change to `internal/registry/specs/docs.json`, no new command file, no `cmd/` change at all.
- **Divergence:** none.

I also confirmed the change does not silently break an existing contract:

- `octo-cli api POST <path> --data '<json>'` is the real signature (`cmd/api.go:31-46`, `cobra.ExactArgs(2)`), and the path must start with `/` (`cmd/api.go:49`), which the example satisfies.
- The endpoint genuinely is absent from the embedded spec — `internal/registry/specs/docs.json` declares only `/attachments/presign`, `/attachments/{attachId}`, and `/attachments/resolve` — so "not yet a generated subcommand, use the passthrough" is accurate, and pointing at `api` is the escape hatch that command was designed for.
- The `data.mappings` / `data.notIngested` paths are correct for this CLI. `docs.json` declares no `x-octo-response-unwrap`, and `WriteSuccess` places a non-paginated body under `data` verbatim (`internal/output/envelope.go:64-65`), so the backend's top-level fields land exactly one level under `data`. This matches the existing `jq -r '.data.baseVersion'` idiom already in `doc.md:41`.
- `422 attachment_not_found` is consistent with the error list already documented at `doc.md:70`.
- The new bash example needs no explicit content type: `internal/client/client.go:872` sets `Content-Type: application/json` whenever a JSON body is present.

---

## 2. Code quality

**Quality: Approved with suggestions**

No P0 or P1. Everything below is advisory and non-blocking.

### P2-1 — Response shape of the ingest call is not documented

`common.md:202-203` tells the agent to read `attachId` "from `data.mappings`" and to treat entries in `data.notIngested` as failures, but never says what those values look like. An agent cannot tell whether `mappings` is a URL-keyed object (`{"https://…": "att_xxx"}`) or an array of records (`[{"url":…, "attachId":…}]`), and will either guess wrong or burn a turn inspecting the payload.

This is also inconsistent with the file's own convention two paragraphs up, which does enumerate the response:

```
# The response contains: attachId, uploadUrl, headers{...}, expiresInSec, ...
```
— `common.md:164`

Suggest adding a two-line example response block after the `octo-cli api` snippet.

### P2-2 — A document-body node shape is documented in the cross-cutting file

`common.md` is explicitly the surface-agnostic reference (`SKILL.md:28`), but the new section closes with "write the image node using its durable `attachId`" plus a ProseMirror `image` node (`common.md:208-224`). A whiteboard agent that lands here from the `SKILL.md` attachments route could copy that node into a scene, where the contract is a `files` map of `{attachId}` refs instead (`board.md:13`, and `422 board_file_invalid` at `board.md:98`).

The pre-existing sentence at `common.md:181-184` does route by surface, which is why I rate this P2 rather than higher — the hazard is confusion, not certain corruption. Still, one qualifier on `common.md:208` ("for a `doc_type: doc` body; whiteboards use the `files` map in `board.md`") would close it.

### P2-3 — A client-side layout defect is baked in as permanent guidance

`common.md:210-213` and `doc.md:92-98` instruct agents to always send `width: 300` when the real width is unknown, because current clients lay out `width: null` as `0x0`. That is the right short-term advice, but it is written as a durable rule with no expiry. Once the client is fixed, "always 300" becomes actively wrong — a 1200px screenshot will render at a quarter width, and nothing in the file will tell a future maintainer that the number was a workaround.

Suggest labeling it as such inline (e.g. "temporary client-compat workaround, tracked in <issue>") so it is greppable when the client lands the fix.

### P2-4 — No guidance on `height`

The text mandates a positive `width` but is silent on `height`. If the client derives height from the attachment's intrinsic aspect ratio, saying so explicitly prevents agents from inventing a `height` and distorting the image; if height is also required, the omission is a gap. One sentence either way.

### P2-5 — No documented caps on the ingest batch

`urls` is an array, but no URL count or byte limit is given. The neighbouring flows do document their limits — `presign` says the size "must be within the tier cap" (`common.md:161`), and the `attachments/resolve` schema says the batch is "capped by the backend batch limit" (`docs.json`). An agent batching a page's worth of images will discover the cap through an error.

### P2-6 — The passthrough is the one docs call path that still sends `X-Space-Id`

`docs.json` declares `"x-octo-space-header": false`, and generated commands honour it via `SuppressSpaceHeader: d.SpaceHeaderSet && !d.SpaceHeader` (`cmd/service/run.go:82`). `cmd/api.go:94-102` builds its `client.Request` without that field, so `internal/client/client.go:961-963` will attach `X-Space-Id` whenever the active credential carries a space — for the one docs call this PR now recommends.

This is **not** a functional break: `SKILL.md:50-52` states the bot mount "deliberately ignores any client-supplied space header (anti-spoof)". Flagging it because the guidance's correctness currently rests on that anti-spoof behaviour rather than on the CLI doing the right thing, and because it is another argument for P2-7.

### P2-7 — The contract has no in-repo anchor

Because the endpoint is not in `docs.json`, this flow is the only one in these skill files that cannot end with a `Schema: octo-cli schema <op>` pointer the way its neighbours do (`common.md:76`, `:106`, `:184`). It also means no schema help, no secret masking via `apiSecretsForRequest`, and no contract test to catch backend drift — the documented request/response shape can silently go stale.

Not blocking for a docs PR, but worth a follow-up to register the operation in `docs.json` so it becomes a generated `docs attachments ingest` command.

### Nit — the new test assertions are presence-only

`skills_test.go:46` and `:56-57` assert that literals introduced by this same PR are embedded. They lock the wording against accidental deletion, which is the existing convention in this file, but they validate no behaviour — in particular not that the documented command parses or that the response paths match the emitted envelope. Fine as-is; just don't read them as verification of the flow.

---

## 3. Verdict

**APPROVED**

Spec compliance passes and there are no P0/P1 findings. The guidance is accurate against the CLI as it exists at this SHA, the change stays inside its declared scope, and the build and tests are green. The seven P2s above are documentation-completeness improvements that can land in a follow-up.

## 4. Coverage — what I could not verify

Stated plainly rather than left implicit:

- **The backend ingest contract itself.** The docs backend is not in this repository, and the endpoint is absent from `internal/registry/specs/docs.json`. The request field name `urls`, the response fields `mappings` / `notIngested`, the SSRF protections, and the allowed-image validation are taken on the author's stated end-to-end verification against a real test document.
- **The frontend attachment resolver**, including the `width: null` → `0x0` layout behaviour and the asset-host whitelist that produces **Image unavailable**. Client repositories were not inspected.
- **Whether `422 attachment_not_found` is returned for ingest-created attachments specifically.** It is already documented for `docs content edit` at `doc.md:70`, so the claim is at least internally consistent.

PR #155, opened 2026-09-02
fix(loop): enforce public API media type

liuooo:
## What

Fleet selects its stable Public API only when Loop requests send
`Accept: application/vnd.octo+json`. Generic JSON enters the legacy `/api`
routing path and makes Public API routes return `404`.

The `dev/v0.14.1` client already uses the vendor media type by default, but an
explicit or generated `Accept: application/json` header could still override
that contract.

## How

- **internal/client/client.go** — normalize the `Accept` header after cloning
  request headers and enforce the Fleet Public API media type for every Loop
  request, including case-variant recipe headers.
- **internal/client/client_test.go** — assert the wire-level media type as a
  literal and cover generic JSON header override behavior.
- **cmd/service/service_test.go** — verify the generated Loop command sends the
  vendor media type through the full command-to-transport path.

## Tests

- `GOCACHE=/private/tmp/octo-cli-accept-gocache go test ./internal/client ./cmd/service -run '^(TestDo_Loop|TestLoopCommandUsesUnifiedGatewayAndModulePath)' -count=1`
- `make fmt`
- `go vet ./...`
- `golangci-lint run` with CI version `v2.12.2` (`0 issues`)
- `GOCACHE=/private/tmp/octo-cli-accept-gocache make test build`
- `git diff --check`

### DIFF
--- a/cmd/service/service_test.go
+++ b/cmd/service/service_test.go
@@ -129,11 +129,12 @@ func TestRegisterServiceCommands_TreeShape(t *testing.T) {
 }
 
 func TestLoopCommandUsesUnifiedGatewayAndModulePath(t *testing.T) {
-	var gotPath, gotAuth, gotSpace string
+	var gotPath, gotAuth, gotSpace, gotAccept string
 	gateway := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		gotPath = r.URL.Path
 		gotAuth = r.Header.Get("Authorization")
 		gotSpace = r.Header.Get("X-Space-Id")
+		gotAccept = r.Header.Get("Accept")
 		_, _ = w.Write([]byte(`{"data":{"task_id":"task-1"}}`))
 	}))
 	defer gateway.Close()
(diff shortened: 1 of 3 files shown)


Jerry-Xin (approved):
✅ **APPROVE** — Fleet Public API media type is now enforced for every Loop request.

## Verification summary

- **Enforcement point:** `Client.Do` unconditionally sets `Accept: application/vnd.octo+json` (`publicAPIMediaType`, `internal/client/client.go:34`) for all `Service == "loop"` requests after cloning headers (`internal/client/client.go:162-164`). The new `setHeader` helper deletes any case-insensitive variant of the header first, so an explicit or generated `accept: application/json` recipe header can no longer override the contract and push Loop traffic onto the legacy `/api` routing path.
- **Coverage:** `Client.Do` is the single transport chokepoint (only `http.NewRequestWithContext` call site is in `doWithRetry`, `internal/client/client.go:236`), and the registry-generated Loop commands (`internal/registry/specs/loop.json`) all flow through it — enforcement is applied at every Loop entry point, not just some.
- **Compatibility:** client-side only; no server contract change. The previous default already sent the vendor media type when no Accept was present; this closes the remaining override hole (fail-closed to the vendor type). Census of non-test call sites found no Loop caller relying on a custom Accept being preserved, and the removed `headerValue` helper has zero remaining callers. Other services are untouched.
- **Tests:** wire-level literal assertions at both boundaries — client level for the default case and for the lowercase `accept: application/json` override case (`internal/client/client_test.go`), plus a full command-to-transport assertion through a real test server (`cmd/service/service_test.go:168-170`). Both touched packages pass locally.

PR #156, opened 2026-09-03
fix(api): forward workspace context

liuooo:
## What

`octo-cli api` is the escape hatch for endpoints that are not represented in
the embedded registry, but it could not send workspace context. Calls to
workspace-scoped custom or legacy endpoints therefore could not provide the
required `X-Workspace-ID` header.

## How

- **cmd/api.go** — add an optional `--workspace-id` flag and forward its value
  as `X-Workspace-ID`. Omitted flags leave the header absent, preserving all
  existing passthrough behavior.
- **cmd/cmd_test.go** — cover both explicit header forwarding and default
  omission at the HTTP transport boundary.
- **README.md, CLAUDE.md, docs/architecture-design.md, skills/octo-shared/SKILL.md,
  CHANGELOG.md** — document the new command shape and workspace-scoped usage.

This change does not alter daemon task-mode command policy; task processes
remain restricted to the generated Loop command surface.

## Tests

- `go test -race -count=1 ./...`
- `make fmt`
- `go vet ./...`
- `go build -o /private/tmp/octo-cli-pr-verify ./cmd/octo-cli`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **`octo-cli api --workspace-id`** — generic passthrough requests can now send
+  explicit workspace context through the `X-Workspace-ID` header.
 - **`octo-cli docs search`** — permission-scoped full-text search across online
   documents, sheets, boards, and mounted HTML documents registered in docs-backend
   via `POST /v1/bot/docs/search`. Supports repeatable `--doc-type` filters, manual
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -70,7 +70,7 @@ octo-cli html      list | get | publish | versions | rm     (octo-doc HTML docs;
 
 octo-cli auth      login | status | logout | list
 octo-cli schema [--list [domain] | <operation-id>]
-octo-cli api <METHOD> <PATH> [--params ...] [--data ...] [--service ...]
+octo-cli api <METHOD> <PATH> [--params ...] [--data ...] [--service ...] [--workspace-id <uuid>]
 octo-cli config show
 octo-cli completion bash|zsh|fish|powershell
 octo-cli version
--- a/README.md
+++ b/README.md
@@ -185,6 +185,7 @@ octo-cli config show                # resolved config (token masked)
 # Generic passthrough for ops that aren't auto-registered.
 octo-cli api GET  /v1/messages --params '{"chat_id":"chat-1"}'
 octo-cli api POST /v1/messages --data @body.json
+octo-cli api GET  /fleet/api/v1/custom --workspace-id <workspace-uuid>
 ```
 
 ## Authentication
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -15,14 +15,14 @@ import (
 
 // newAPICmd implements the generic passthrough:
 //
-//	octo-cli api <METHOD> <PATH> [--params '{...}'] [--data '{...}']
+//	octo-cli api <METHOD> <PATH> [--params '{...}'] [--data '{...}'] [--workspace-id <id>]
 //
 // Auth, service URL routing, retry, envelope, and universal flags are reused
 // verbatim. No spec consultation, no flag auto-generation, no pagination
 // helper — this is the escape hatch when an endpoint isn't in the registry
 // yet or when a caller needs to exercise something bespoke.
 func newAPICmd(f *cmdutil.Factory) *cobra.Command {
-	var paramsSpec, dataSpec, service string
+	var paramsSpec, dataSpec, service, workspaceID string
 
 	cmd := &cobra.Command{
 		Use:   "api <METHOD> <PATH>",
@@ -33,6 +33,7 @@ Examples:
   octo-cli api GET /api/v1/matters --params '{"status":"open"}'
   octo-cli api POST /api/v1/matters --data '{"title":"test"}'
   octo-cli api GET /v1/bot/events --service dmworkim
+  octo-cli api GET /fleet/api/v1/custom --workspace-id 00000000-0000-0000-0000-000000000000
   octo-cli api POST /api/v1/matters --data @body.json
   octo-cli api POST /api/v1/matters --data @-       # read from stdin
 
@@ -74,12 +75,17 @@ flags are not auto-generated and --page-all is not available.`,
 				_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
 				return err
 			}
+			var headers map[string]string
+			if workspaceID != "" {
+				headers = map[string]string{"X-Workspace-ID": workspaceID}
+			}
 			respBody, err := cli.Do(cobraCmd.Context(), &client.Request{
 				Service: service,
 				Method:  method,
 				Path:    path,
 				Query:   q,
 				Body:    body,
+				Headers: headers,
 			})
 			if err != nil {
 				_ = f.EmitError(err) //nolint:errcheck // best-effort emit before returning err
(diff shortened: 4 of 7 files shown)


Jerry-Xin (approved):
The PR is relevant, correctly forwards optional workspace context, and preserves existing passthrough behavior.

✅ Highlights

- [cmd/api.go:78](cmd/api.go:78) adds the header only when explicitly requested.
- [cmd/cmd_test.go:238](cmd/cmd_test.go:238) covers both forwarding and default omission.
- Documentation is consistent across user and architecture references.
- Task-mode restrictions remain enforced by the root command policy.
- Full race-enabled test suite, `go vet`, build, and `git diff --check` all pass.

yujiawei (approved):
# Code Review — PR #156 (octo-cli)

Head SHA: `8ad61c5d2b3346f867ac6deb0f007518660e9730`
Change: `fix(api): forward workspace context`

Summary: Adds an optional `--workspace-id` flag to the generic `octo-cli api` passthrough command that, when set, is forwarded as the `X-Workspace-ID` header. When omitted, the header is not sent (zero change to existing behavior).

## 1. Spec compliance verdict
**Spec: ✅**
- Missing: none — `cmd/api.go` wires the flag, forwards the header only when non-empty, omits it by default; tests cover both forward and default-omit paths at the HTTP transport boundary; user-facing docs (README, CLAUDE.md, `docs/architecture-design.md`, `skills/octo-shared/SKILL.md`, `CHANGELOG.md`) are updated consistently.
- Extras: none — no new commands, no global flags, no behavior change outside `octo-cli api`.
- Deviations: none — flag name `--workspace-id` and header name `X-Workspace-ID` match the existing convention used by the Loop/loop task/workspace commands (see `cmd/service/service_test.go`), so this is consistent with the established CLI surface.

## 2. Code Quality verdict
**Quality: Approved**

What was checked:
- `internal/client/client.go:59` — the `Request.Headers map[string]string` field already exists and is the established extension point for per-request headers; no client changes needed.
- `internal/client/client.go:241-252` — user-supplied headers are applied via `httpReq.Header.Set(k, v)` **after** Authorization / `X-Space-Id` / Content-Type are set, so caller headers take precedence over defaults (correct and expected for an escape hatch); there is no pre-existing code path that injects `X-Workspace-ID`, so no collision risk.
- `internal/client/client.go:162-171,397-406` — `cloneHeaders` and `renderDryRun` correctly propagate the new header, so `--dry-run` displays `X-Workspace-ID` and the header persists across retries.
- `cmd/api.go:78-81` — when `--workspace-id` is omitted, `headers` is left as a nil map; `cloneHeaders(nil)` returns an empty allocated map, so no header is emitted (verified by `TestCmd_API_OmitsWorkspaceHeaderByDefault`).
- `cmd/cmd_test.go:238-288` — two targeted tests: one asserts the header is set to the flag value against an `httptest` server, one asserts it is absent by default. The test pattern mirrors the existing `TestCmd_API_POSTBody` setup.
- Ran `go vet ./...`, `go build ./...`, `go test -race -count=1 ./...` — all green.

Non-blocking observations (nits, not merge blockers):
- **No UUID-format validation on `--workspace-id`.** The flag help text says "workspace UUID" but the value is passed through verbatim. That matches the rest of the escape-hatch philosophy of `octo-cli api` (we don't validate `--service` against a strict allowlist either, and `--params` / `--data` are sent as-given), so I do not consider this a blocker. If you want a tighter UX later, consider validating with `github.com/google/uuid` or a simple regex and returning an `output.ErrValidation` before building the request — but that can ride a follow-up.
- **Doc example uses a deterministic zero-UUID (`00000000-0000-0000-0000-000000000000`).** That's fine as a placeholder; it's also the exact value used in `cmd/api.go`'s example block, so it's consistent across help text and docs.
- The `Headers: headers` field on the `client.Request` is a `map[string]string`; the current code constructs at most one entry. If future escape-hatch flags need to forward more headers (e.g. a generic `-H/--header` repeatable flag — not in this PR, just noting), this single-entry construction will need to grow. Out of scope here.

Cross-file / blind-spot checks:
- Global-flag conflict: no existing global `--workspace-id` (only the per-subcommand flag under `loop task|workspace`); adding it as a flag local to `api` does not shadow or collide.
- Service routing: when `--service dmworkim` / `--service matters` is used together with `--workspace-id`, the header is sent to whatever host `ServiceURL(service)` resolves to. That is the caller's responsibility for a passthrough command and matches the existing `--service` escape-hatch semantics; the PR body explicitly says "This change does not alter daemon task-mode command policy."
- Envelope / error output: no changes to how responses are emitted; the header only affects the outgoing request.
- Config / credential: no new config keys or env vars, so no migration concern.

## 3. Overall verdict
**APPROVE**

Small, surgical, well-tested, correctly scoped. Safe to merge to `dev/v0.14.1`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #156 (octo-cli)

## Summary

This PR adds a `--workspace-id` flag to the generic `octo-cli api` passthrough so escape-hatch requests can carry explicit workspace context as an `X-Workspace-ID` header (`cmd/api.go:79-86`, `cmd/api.go:101`). That brings the passthrough to parity with the spec-driven Loop commands, which already send the same header via the `WorkspaceIDHeader` parameter declared in `internal/registry/specs/loop.json`. The flag is opt-in: when omitted, no header is sent. Docs are updated consistently across README, CHANGELOG, CLAUDE.md, `docs/architecture-design.md`, and `skills/octo-shared/SKILL.md`, and two new tests cover both the forwarded and the default-omitted cases.

## Verification

Static analysis only at head `8ad61c5d2b3346f867ac6deb0f007518660e9730`; build and tests not executed in this environment.

- ✅ **Flag→wire data flow** — traced `--workspace-id` (`cmd/api.go:25`, `:101`) → non-empty gated map construction (`cmd/api.go:79-81`) → `client.Request.Headers` (`cmd/api.go:86`) → `cloneHeaders` (`internal/client/client.go:162`) → applied via `Header.Set` in `attempt` after Authorization/X-Space-Id/Content-Type (`internal/client/client.go:250-252`). The header genuinely reaches the outgoing request; `cloneHeaders(nil)` is a safe no-op when the flag is omitted.
- ✅ **Parity with the spec-driven path** — `loop.json` declares `X-Workspace-ID` (`x-octo-flag: workspace-id`); `registerHeaderFlags` + `buildHeaders` (`cmd/service/run.go:142`) emit it only when the user sets the flag, into the same `Request.Headers` consumed by the same `attempt()` code. Header name, flag name, and emit-only-when-set semantics match the new api-command path exactly.
- ✅ **Default behavior unchanged** — with the flag absent, `headers` is nil and request construction is byte-for-byte the pre-PR path; `TestCmd_API_OmitsWorkspaceHeaderByDefault` pins this.
- ✅ **Precedence and dry-run** — custom headers apply last, matching the documented contract "headers take precedence over client defaults" (`internal/client/client.go:47-48`); `renderDryRun` also includes them (`internal/client/client.go:390-406`).
- ✅ **Tests exercise the production path** — both new tests run the real cobra tree (`NewRootCmd`) with a real `client.New` against an httptest server (`cmd/cmd_test.go:238-288`); no hand-fed transport bypass.
- ✅ **No flag collision** — root persistent flags (`cmd/root.go:47-55`) don't define `workspace-id`; it is local to `api` and mirrors the Loop auto-generated flag of the same name.
- ✅ **Docs in sync** — all five changed doc surfaces agree on flag name, header name, and "no inference" semantics; `docs/octo-cli-design.md` carries no api-flag enumeration (no drift); `skills/octo-loop/SKILL.md` wording is consistent with the addition.

## Findings

No P0/P1/P2 issues found. Both independent review legs converged on zero findings.

- `OCTO_CREDENTIAL_MODE=task` still restricts daemons to the Loop surface (`cmd/root.go:76-94`) — pre-existing gating untouched by this PR; the new flag targets the normal-mode passthrough, where it works.
- No client-side UUID validation on the flag value — identical to the spec-driven Loop path (string flag, server validates membership; cf. `BOT_WORKSPACE_MEMBERSHIP_REQUIRED` in `docs/architecture-design.md:234`). Consistent with the escape-hatch design; the backend is the authorization boundary.
- Header-value injection is not possible: Go's net/http rejects invalid header values at write time, and the caller already owns the bearer token being sent.
- Verbose/dry-run logging exposes no secrets (workspace UUID is not a credential; Authorization stays masked in dry-run).

## Verdict: APPROVED

The change is minimal, opt-in, and correctly wired end-to-end from flag to wire with full parity against the existing spec-driven workspace-header path. Default behavior is provably unchanged, the new tests exercise the real client through the real command tree, and every doc surface that enumerates the api command was updated consistently. Nothing to fix.

[Octo-Q] verdict: APPROVE — 双模型独立审查均为零 finding：flag→wire 数据流完整、与 Loop spec 驱动路径完全对齐、默认行为不变、测试走真实生产链路、文档全量同步。

PR #157, opened 2026-09-03
fix(loop): clarify workspace UUID contract

liuooo:
## What

Generated Loop commands describe `--workspace-id` as a generic workspace
"selector", even though the Public API schema and Fleet contract require a
workspace UUID. This can lead callers to pass a slug or display name in
`X-Workspace-ID`, which Fleet rejects as `invalid workspace_id`.

## How

- **internal/registry/specs/loop.json** — describe the shared workspace header
  parameter explicitly as a UUID, updating generated help for every affected
  Loop operation.
- **skills/octo-loop/SKILL.md** — require the UUID returned by `workspace list`
  and use `<workspace-uuid>` consistently in command examples.
- **docs/architecture-design.md** — define the workspace context convention:
  Web human sessions use `X-Workspace-Slug`, CLI/Public API automation uses
  `X-Workspace-ID`, workspace resource paths use UUIDs, and credential-bound
  tasks cannot override their server-verified scope.
- **cmd/service/service_test.go** — pin the generated help text to the UUID
  contract.

The design intentionally does not overload `--workspace-id` with slug support.
Any future CLI slug capability must use a separate, mutually exclusive flag
and first be declared in the Public API specification.

## Tests

- `go test -race -count=1 ./...`
- `go test ./... -count=1` on the latest `dev/v0.14.1`
- `make fmt`
- `go vet ./...`
- `go build -o /private/tmp/octo-cli-workspace-id-verify ./cmd/octo-cli`
- `git diff --check`

Related: #156

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `--role` flags front the `shareScope` / `shareRole` wire keys.
 
 ### Changed
+- **Loop workspace flag terminology** — generated `--workspace-id` help and the
+  bundled Loop skill now state that the value is a workspace UUID, not a slug
+  or generic selector. The architecture guide records the separate Web slug,
+  Public API UUID, and credential-bound task conventions.
 - **Generated service commands now reject incomplete JSON bodies locally** —
   request-schema `required` fields and nested `minItems` constraints are
   validated after merging `--data` with promoted body flags, before any HTTP
--- a/cmd/service/service_test.go
+++ b/cmd/service/service_test.go
@@ -291,6 +291,21 @@ func TestLoopWorkspaceIDFlagIsRequired(t *testing.T) {
 	}
 }
 
+func TestLoopWorkspaceIDFlagDocumentsUUID(t *testing.T) {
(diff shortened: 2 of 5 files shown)


Jerry-Xin (approved):
The PR is relevant, internally consistent, and ready to merge. No blocking or non-blocking issues found.

✅ Highlights

- Shared UUID wording correctly propagates across Loop commands ([loop.json](internal/registry/specs/loop.json#L6526)).
- Documentation and skill examples consistently distinguish UUIDs from slugs.
- Regression coverage pins the generated help contract ([service_test.go](cmd/service/service_test.go#L293)).
- `go test -race -count=1 ./...`, focused `go vet`, JSON validation, and `git diff --check` pass.

Verification performed independently:
- The `--workspace-id` flag usage is generated directly from the spec parameter description (cmd/service/flags.go:136), so the new wording reaches every workspace-scoped Loop command.
- The new test `TestLoopWorkspaceIDFlagDocumentsUUID` passes locally, and the full `go test -race -count=1 ./...` suite is green.
- No stale "Workspace selector" / `<ws>` wording remains in the spec, docs, or skill examples; the CHANGELOG entry sits correctly under Unreleased → Changed.

PR #159, opened 2026-09-04
fix(auth): unify environment token precedence

liuooo:
## Summary

Unify `OCTO_TOKEN` and `OCTO_BOT_TOKEN` as compatible environment credential aliases, with `OCTO_TOKEN` taking precedence.

## Changes

- Normalize both aliases once through `config.ResolveEnvToken`, including the selected source.
- Keep explicit `--profile` / `--bot-id` selectors highest priority; otherwise prefer environment credentials over implicit stored profiles.
- Apply the same alias contract in normal and task credential modes while retaining Loop-only validation in task mode.
- Remove downstream alias-specific provider configuration and update authentication documentation and examples.
- Add and update tests for alias precedence, implicit and explicit profile selection, task compatibility, and fail-closed behavior.

## Motivation

Callers need to select different Octo bot identities per process without exposing or mutating stored profile configuration. Treating the variables as aliases avoids token-kind-specific behavior and keeps future credential policy independent of environment variable naming.

## Testing

- [x] `go vet ./...` passes
- [x] `go build ./...` passes
- [x] `gofmt` and `git diff --check` pass

## Checklist

- [x] No commands were added
- [x] Documentation was updated
- [x] New behavior includes tests


Jerry-Xin (dismissed):
## Verdict: APPROVE

Independent byte-level review of this security-sensitive credential-resolution change, plus `go vet`, `go build`, and `go test -race -count=1 ./...` on this head — all green.

### Old vs new resolution order — verified at every call site

Old: stored profile first, env fallback (`OCTO_TOKEN` → `OCTO_BOT_TOKEN`), with `AllowEmptyStoreBotIDFallback` letting `OCTO_BOT_ID` fall through on an empty store. New (internal/cmdutil/factory.go:258):

1. Explicit `--profile` / `--bot-id` → FileProvider only; mismatch is fail-closed (internal/credential/file_provider.go:49-80).
2. Otherwise a normalized environment credential wins over an implicit stored profile. `config.ResolveEnvToken` (internal/config/config.go:80) is the single alias normalizer: `OCTO_TOKEN` over `OCTO_BOT_TOKEN`, both trimmed, a blank canonical alias does not shadow the legacy one.
3. Otherwise the `OCTO_BOT_ID`-selected or sole implicit profile; ≥2 profiles with no selector and no env credential is still a hard ambiguity error.

Helper census: every env-token read now funnels through `ResolveEnvToken` (`config.Load`, `EnvProvider.Resolve` for both normal and task modes, `auth status` env_token_set, `defaultTokenSource`). The removed `TokenVar`/`TokenVars`/`AllowEmptyStoreBotIDFallback` knobs have no orphaned references, and the new `internal/credential` → `internal/config` import introduces no cycle.

### Fail direction — fail-closed confirmed

- Conflicting aliases: canonical wins and `identity.source` names the alias actually used (asserted in internal/config/config_test.go and internal/credential/provider_test.go, not just compile-green).
- Task mode: a non-Loop `OCTO_TOKEN` does **not** fall back to a valid `OCTO_BOT_TOKEN` — hard error (factory_test.go "invalid preferred alias does not fall back"); both aliases missing → `ErrNoCredential`; selectors/`OCTO_BOT_ID`/`--space` still rejected in task mode (internal/cmdutil/factory.go:340-357).
- Explicit selectors ignore environment credentials and fail closed (TestFactory_ExplicitBotIDAndEnvironmentClaimHaveDistinctEmptyStoreSemantics).
- `OCTO_BOT_ID` beside an environment credential stays an unverified claim: the envelope emits `robot_id_claimed` (internal/cmdutil/factory.go:593), and Mail paths verify via `/v1/bot/register` before binding, so a mismatched claim cannot select another bot's mailbox credential.

### Secrets handling

No new token surface in errors or logs: `Validate` names the variables but never values; `config show` masks (prefix + `***`); client verbose/dry-run redaction (`x-octo-secret`, masked `Authorization`) is untouched by this PR. Stale-credential scenario (leftover env token + healthy stored profile): the env alias now wins — observable via `identity.source` and failing with a clear 401 hint rather than silent wrong-profile use, matching the documented contract.

### Behavior change — documented, not silent

Environment credentials now override implicit stored profiles (the inverse precedence shipped in earlier releases; present as of the v0.13.0/v0.14.0 tags). The flip is explicitly documented in README, CLAUDE.md, CHANGELOG (Unreleased), docs/octo-cli-design.md, and the skill docs, and is covered by TestFactory_EnvironmentCredentialOverridesImplicitProfiles including the ambiguity-bypass case.

### 💬 Non-blocking

- 🔵 internal/cmdutil/factory.go:39-41 — stale GlobalOptions comment: `OCTO_BOT_ID` now labels an environment credential whenever one is present, not only "if the profile store is empty".
- 🔵 CHANGELOG.md — the rewritten entry previously described the profile-first precedence that already shipped; consider adding a standalone "Changed" note so upgraders notice that an ambient environment token now overrides an implicit stored profile.

Jerry-Xin (changes requested):
Round-2 re-review on head 392ef75 (rebased onto the updated base; delta from the round-1 head reviewed via range-diff, and every auth-resolution path re-verified byte-for-byte on this head).

### 🔴 Blocking

- 🔴 **`auth status` reports a different active identity than commands actually run under.** `newAuthStatusCmd` (cmd/auth.go:141) resolves directly through `store.ActiveProfile(...)` and never consults the environment credential. Under this PR's own precedence — an environment credential overrides an implicit stored profile — that produces wrong answers whenever `OCTO_TOKEN` / `OCTO_BOT_TOKEN` is set alongside stored profiles:
  - Sole stored profile → `auth status` reports that profile as `active` (profile name, masked stored token, `source: profile:<name>`), while every command actually runs under the env credential.
  - Two or more profiles, no selector → the runtime unambiguously runs under the env credential, yet `auth status` reports `active: null` with a "no bot selected" hint.

  The command's stated contract is exactly this question (`Short`: "Show the single active bot identity (whoami)", cmd/auth.go:134; in-code comment: "who would I act as right now?"), and its `StatusNone` branch already acknowledges the env credential via `env_token_set` — so the profile branches missing it looks like an oversight rather than a design choice. Agents use `auth status` to confirm who they act as before operating; a wrong answer in this PR's canonical scenario (install.sh now tells users `export OCTO_TOKEN=...`) defeats the identity-visibility goal the envelope's identity echo exists for. `config show` reports correctly because it prefers `f.Credential()` — the inconsistency is specific to `auth status`. Note: this was also present on the round-1 head; the round-1 review verified the runtime precedence thoroughly but did not cross-check `auth status`' reporting contract against it — calling that out plainly here.

  Suggested fix: mirror `Factory.buildCredential` — when no explicit `--profile` / `--bot-id` selector was supplied, resolve the environment credential first (via `config.ResolveEnvToken`) and report it as `active` with `source: env:<alias>`; keep explicit-selector behavior unchanged. Add tests for sole-profile and ambiguous-profile stores with each alias set.

### 💬 Non-blocking

- 🟡 Auth-failure hints still name only the legacy alias: `WrapCLIError` (internal/cmdutil/factory.go:409) advises "set OCTO_BOT_TOKEN ...", and the `UNAUTHORIZED` entry of `backendErrorMapping` (internal/output/errors.go:95) says "check OCTO_BOT_TOKEN ...". Both should name canonical `OCTO_TOKEN` alongside the legacy alias for consistency with the new precedence.
- 🔵 Round-1 note (a) addressed: the stale `GlobalOptions` selector comment (internal/cmdutil/factory.go) was rewritten and now accurately describes the explicit-selector semantics.
- 🔵 Round-1 note (b) substantively addressed: the CHANGELOG now has a standalone entry describing the precedence flip accurately ("Unified environment credential aliases"). Minor, optional: it sits under "Added"; since env credentials overriding implicit stored profiles is a behavior change for existing setups, "Changed" would classify it more precisely under Keep-a-Changelog.

### ✅ Re-verified on this head (byte-for-byte, not inherited)

- Precedence: explicit `--profile` / `--bot-id` (skip env entirely, fail-closed via FileProvider `StatusMissing`) > env credential (`OCTO_TOKEN` > `OCTO_BOT_TOKEN`, normalized once by `config.ResolveEnvToken`) > `OCTO_BOT_ID`-selected or sole implicit profile.
- All env-token reads funnel through `ResolveEnvToken` (the only `os.Getenv` calls for the token variables live inside it); zero orphan references to the removed `TokenVar` / `TokenVars` / `AllowEmptyStoreBotIDFallback`.
- Ambiguity (≥2 profiles, no selector, no env credential) still fails closed; task mode still rejects selectors, `OCTO_BOT_ID`, and `--space`, never opens the auth store, and fails closed on a missing token; the absence of a local task-credential kind check is inherited from the base (Fleet verifies server-side), not introduced here.
- Secrets are never echoed: every display path goes through `MaskToken`; `auth status` reports `env_token_set` only as a boolean.
- Tests protect the new semantics directly (env aliases override sole and ambiguous implicit profiles; explicit profile overrides env; task-mode alias preference plus the full rejection matrix; fail-closed paths). `go vet`, `go build`, and `go test -race -count=1 ./...` are all green.

PR #158, opened 2026-09-03
feat(docs): document spreadsheet P0 bot controls

luoshu0211:
## Summary

Teach the built-in Octo Docs skill and registry schema how to use the spreadsheet P0 capabilities exposed by the docs backend.

## Changes

- expose `freeze`, `filters`, and `dataValidations` in `docs sheet get/edit`
- document whole-row sorting with complete cell-object preservation
- document single-select `list` and multi-select `listMultiple` dropdown contracts
- document the unified validation contract: public REST uses per-sheet arrays for all rule families, while the Y.Map alone uses per-rule `${logicalId}!${uid}` keys
- document optimistic-concurrency behavior and `412 base_version_stale`
- add registry and embedded-skill regression coverage, including negative guards against the superseded flat-map / checkbox-only wording

## Motivation

Bots previously had no supported CLI contract for the spreadsheet P0 features, even though the product UI and backend now expose them. This change makes the capabilities discoverable to a cold-start bot.

Depends on backend MR !123: https://codex.mlamp.cn/dmwork/octo-docs-backend/-/merge_requests/123

For this delivery, MR !123 is the authoritative contract: REST reads and writes per-sheet arrays that may contain `list`, `listMultiple`, `checkbox`, and other validation families, while persistence expands those rules into per-rule Y.Map entries. The checkbox-only flat REST contract in backend MR !124 / CLI PR #154 is incompatible and must not be combined with this delivery without reconciliation.

## Testing

- [x] `jq empty internal/registry/specs/docs.json`
- [x] `go test ./internal/registry ./skills ./cmd/service -count=1`
- [x] `go test ./... -count=1`
- [x] `go test ./cmd/service -race -count=1`
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] negative mutations for the flat-map and checkbox-only contract guards fail as intended
- [x] rebuilt CLI exposes the revised schema and embedded Sheet skill
- [x] cold-start bot E2E passed against the real Docker backend, including stale-version rejection and restart persistence

## Checklist

- [x] Code is in English (comments, commit messages, variable names)
- [x] New commands added to README Command Reference table — N/A, no new CLI command
- [x] CLAUDE.md command list updated if commands changed — N/A, no command change
- [x] New features include tests
- [x] Embedded skill and registry documentation updated
- [x] No unrelated changes included

## COMPREHENSION

[...]

PR #154, opened 2026-09-02
feat(docs): support sheet checkbox validations

wwwzzzcccc:
## Summary

Expose Sheet checkbox validation rules through the metadata-driven Docs CLI and document the complete read, create, update, delete, recovery, and version-inspection workflow.

The change is additive: existing Sheet edit batches and error contracts remain available, while checkbox rules use the same optimistic `baseVersion` write path as cells and match the final frontend/backend contract.

## Changes

- Add `sheetDataValidations` to `docs.sheet.get` and `docs.versions.state` response schemas.
- Add the `dataValidations` object batch to `docs.sheet.edit`; callers pass it through `--data` because object-valued properties do not generate individual flags.
- Document rule keys, ranges, default numeric `1`/`0` values, custom string pairs, deletion, overlap validation, Sheet lifecycle, protected ranges, pagination, and repair-key behavior.
- Preserve complete backend error hints/details in the CLI envelope, including repair metadata, without exposing sensitive response fields.
- Keep PATCH whole-rule replacement semantics explicit so callers resend every field that must remain.
- Add registry and output coverage for request/response schema, full rule serialization, null deletion, and checkbox-specific 409/413 diagnostics.

## Motivation

A cell value alone does not create a checkbox. Agents need to write the matching validation rule, read it back from live and historical Sheet state, and recover bounded over-budget metadata without guessing the wire shape.

Closes #153

## Testing

- [x] `go test ./... -count=1`
- [x] `go vet ./...`
- [x] `go build -o /private/tmp/octo-cli-sheet-checkbox ./cmd/octo-cli`
- [x] `python3 -m json.tool internal/registry/specs/docs.json`
- [x] `git diff --check upstream/main`
- [x] Independent Codex review of the complete target diff (no blockers)

## Dependencies

- Docs Web implementation: https://codex.mlamp.cn/dmwork/octo-web-enterprise/octo-docs-module/-/merge_requests/76
- Docs backend API implementation: https://codex.mlamp.cn/dmwork/octo-docs-backend/-/merge_requests/124

The backend dependency must be deployed before checkbox writes are used. The Web dependency provides interactive rendering, editing, collaboration, and history behavior.

## COMPREHENSION

1. **What does this change do to the load-bearing path?** It extends the embedded registry contract so `dataValidations` travels with the existing `docs sheet edit --data` body, `sheetDataValidations` remains visible in live reads and version snapshots, and backend error metadata survives the output envelope.
[...]

PR #142, opened 2026-08-24
feat(marketplace): migrate CLI to the unified plugin API

l-s-c:
## Summary

Migrate the Marketplace CLI from the retired per-type endpoints to the unified `/plugins/*` API, then align it with the latest review-aware publishing lifecycle from `octo-marketplace` `upstream/main@50a1be3`.

The command surface now has 25 Marketplace operations, including explicit publish/delist and the complete review-request workflow. The metadata-driven engine also supports the backend's conditional `plugin_type` requirement and opt-in strict request-schema validation.

## Linked Spec

N/A — this repository has no matching `.octospec/tasks/*/brief.md` or linked GitHub issue. The implementation contract is the pinned `octo-marketplace` backend API at `upstream/main@50a1be3`; PR #142 is the change-tracking context.

## Changes

- Replace the legacy Skill/MCP/Expert/Squad command families with the unified `marketplace plugin` surface.
- Add `plugin publish`, `plugin delist`, and `plugin review-request create|list|get|approve|reject|cancel`.
- Let `plugin list --mode mine` omit `--plugin-type`, while catalog listings continue to require it.
- Add `x-octo-required-unless-query` and `x-octo-strict-request-schema` support to the generic registry/validation path.
- Enforce fixed-shape Marketplace write documents, nested enums, Unicode length limits, and relation count limits without narrowing intentionally opaque JSON payloads or tolerant review decoders.
- Document draft, private publication, Space review, delist, published-upgrade, retry-recovery, and relation ownership semantics.
- Update operation counts, command-tree tests, route/body/query assertions, enum provenance, and skill documentation contracts.

## How verified

- `go test ./... -count=1`
- `go vet ./...`
- `go build ./...`
- `python3 -m json.tool internal/registry/specs/marketplace.json`
- `git diff --check`
- Compared all 25 CLI method/path pairs against `octo-marketplace` `upstream/main@50a1be3`; no missing upstream route.
- Adversarially reviewed the generic validation engine, Marketplace OpenAPI metadata, and all four Marketplace skill documents.

### 1. What does this change actually do to the load-bearing path?

Before this change, the embedded Marketplace spec exposed only the earlier unified save/import surface and modeled `plugin_type` as unconditionally required. The backend now separates saving from publication: private publication lists immediately, Space publication freezes a review request, published Space upgrades must submit frozen content directly for review, and owners/admins have distinct delist/review actions.

[...]

PR #161, opened 2026-09-04
feat(docs): document spreadsheet P0 bot controls

luoshu0211:
## Summary

Teach the built-in Octo Docs skill and registry schema how to use the spreadsheet P0 capabilities exposed by the docs backend.

This replaces #158 with a fresh review surface and is rebased onto current `main` at `f908749` (including checkbox PR #154 and unified plugin API PR #142).

## Linked Spec

Depends on backend MR !123 and must merge after it:
https://codex.mlamp.cn/dmwork/octo-docs-backend/-/merge_requests/123

## Changes

- expose `freeze`, `filters`, and grouped `dataValidations` in `docs sheet get/edit`
- document whole-row sorting with complete `{v,f,s,p,t}` preservation and coordinate-bound resources
- document single-select `list`, multi-select `listMultiple`, and checkbox rules under one replace-per-sheet REST contract
- preserve the server-directed flat-key deletion-only repair path for over-budget historical checkbox metadata
- warn that splitting one sheet's validation array across retries drops earlier rules
- add an actionable `too_many_sheet_resources` mapping for freeze/filter key-count limits
- reconcile #154 wording, examples, transport tests, protected-range behavior, and checkbox-only overlap/repair constraints with the mixed-rule backend contract
- intentionally omit `sheetList`, `sheetHyperLinks`, and `sheetMerges` from `docs.versions.state`; that backend handler does not decode those fields

## How verified

On head `0e3e167c`:

- `jq empty internal/registry/specs/docs.json`
- `go test ./internal/output ./internal/registry ./skills ./cmd/service -count=1`
- negative mutation checks prove the safe validation-limit hints fail when reverted or removed
- `npm test` from `npm/`: 9/9 passed
- rebuilt CLI schema and embedded `octo-docs` skill expose the corrected contract
- one commit; staged tree and commit tree are identical (`4b2838b76a4423b43ca5245bd0d4b380b3b3d28f`)

GitHub CI passed on this exact head, including gofmt, go vet, go mod tidy, golangci-lint, race/coverage tests, build, and npm tests. A redundant local post-amend full-suite rerun was blocked during Go linking because the machine had only ~187 MiB free (`no space left on device`), not by a test assertion.

## COMPREHENSION

`dataValidations` is replace-style per logical sheet. Every edit must start from `sheetDataValidations` and resend all rules to keep; a later batch for the same sheet replaces an earlier batch rather than appending to it.

The internal Y.Map alone uses `${logicalId}!${uid}`. Public REST uses `{logicalId:[rules]}` except for an exact server-returned `repairDataValidationKey`, which is a deletion-only recovery path for checkbox-shaped stored metadata.

Backend !123 rejects checkbox-vs-checkbox overlap. It does not reject overlaps involving `list` or `listMultiple`; the CLI documents that distinction instead of claiming an unsupported all-family constraint.

## Checklist

- [x] Code and comments are in English
- [x] No new CLI command or flag
- [x] Embedded skill and registry documentation updated together
- [x] Current `main` checkbox and plugin changes reconciled
- [x] No unrelated files included


Jerry-Xin (dismissed):
The PR is relevant to octo-cli and consistently updates its Docs registry, embedded skill, examples, error guidance, and tests.

✅ Highlights

- Grouped validation arrays and the deletion-only legacy repair path are clearly distinguished.
- Sorting guidance correctly preserves `{v,f,s,p,t}` and warns about coordinate-bound resources.
- Registry and embedded-skill regression tests guard against obsolete checkbox-only guidance.
- `go test ./...`, `go vet ./...`, and `git diff --check` pass.
- No security, performance, error-handling, or compatibility blockers found.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #161 (octo-cli)

## Summary

This PR documents the P0 spreadsheet bot controls on the docs surface: freeze panes, shared filters, sorting semantics, and the full data-validation family (`list` / `listMultiple` / `checkbox`). `internal/registry/specs/docs.json` teaches the new `freeze`/`filters` request batches and the per-sheet `dataValidations` array (replacing the old `${logicalId}!${uid}` keyed write shape), extends the read schema with `sheetFreeze`/`sheetFilters`, makes `nextCursor` a nullable union, and documents the new `too_many_sheet_resources` gate. `skills/octo-docs/sheet.md`, `common.md`, `SKILL.md`, and the README follow the new contract, and the only production-code change is a generalized hint string in `internal/output/errors.go:153`. New loader/service/skill tests pin the spec-doc consistency. No P0/P1 issues; two non-blocking P2 items below.

## Verification

Static analysis only at head `13f874e9`; build and tests not executed in this environment.

- ✅ **Spec still loads safely through the registry loader** — the only schema-shape change is `nextCursor` → `["string","null"]` in the `docs.sheet.get` response; response properties parse via `schemaInfoFromNode` → `schemaType` (`internal/registry/loader.go:901`), which already resolves type unions to the first non-null member, and the parameter path using `stringOf` (`internal/registry/loader.go:567`) sees no array-typed `type` in this diff.
- ✅ **Shape migration is internally consistent** — README (`README.md:191`), `sheet.md`, the spec request/response schemas, and the `cmd/service/docs_test.go` transport tests all use the per-sheet array shape; a repo-wide grep finds no remaining old-shape writes outside the documented server-directed repair exception, and the new loader/skill tests explicitly guard against the superseded contract strings.
- ✅ **`--data` transport is pass-through** — no production Go code parses `dataValidations`/`freeze`/`filters`, so the documented shape change cannot break CLI runtime behavior; `TestDocsSheetEdit_SendsP0ResourcesAndIfMatch` pins the PATCH body plus If-Match header.
- ✅ **Hint rewrite is self-contained** — the `internal/output/errors.go:153` string change is mirrored in `internal/output/envelope_test.go:233`, has no other consumers, and hints are presentation-only (they cannot change exit codes).

## Findings

### P2 — 413 repair flow still described as checkbox-only (`internal/registry/specs/docs.json:384`)

This PR rewrote the `docs.sheet.get` description and widened data validations to every family (`list`, `listMultiple`, `checkbox`), but the 413 repair sentence kept its checkbox-scoped wording: "When checkbox metadata exceeds the size or inspection budget…". The same stale scope appears in `skills/octo-docs/sheet.md:264` ("bounded checkbox-rule inspection") and `:275` ("strictly shrinks the raw checkbox metadata"). An agent hitting a repair 413 on a dropdown-heavy sheet may conclude the repair path does not apply and give up, even though the server-directed delete-key flow is family-agnostic. Widen the wording to data-validation metadata in the spec description and both sheet.md spots.

### P2 — Newly documented `too_many_sheet_resources` gate has no error-mapping entry (`internal/output/errors.go:153`)

The rewritten edit contract (`internal/registry/specs/docs.json:479`) and `skills/octo-docs/sheet.md:480` now document `413 too_many_sheet_resources` as a P0 write gate, but `backendErrorMapping` has no entry for it, so the CLI surfaces a generic `api_error` with no actionable hint. This is a pre-existing pattern (`too_many_cells` / `cell_too_large` / `sheet_cell_invalid` / `protected_range` are also unmapped, and unknown codes still fall back safely), so it is not a blocker; since this PR already touches the adjacent `too_many_data_validations` entry, adding a one-line mapping would make the newly documented gate as actionable as its sibling.

- `docs.versions.state` dropping `sheetList`/`sheetHyperLinks`/`sheetMerges` from its response schema is intentional: `internal/registry/loader_test.go` pins the absence with an explicit "backend handler does not decode these" note and a removable guard.
- The expanded restore rollback list in `common.md` is consistent with the generic "forward reconcile" restore spec and contradicts nothing in the checkout.
- Add/remove parity holds for every new resource: freeze (`null` unfreezes), filters (`null` removes), validations (replace-style array, `[]`/`null` clears, server-directed repair-key deletion) — documented symmetrically in spec, `sheet.md`, and README.
- Sorting guidance (raw coordinate rewrite, formulas not re-anchored, coordinate-keyed resources staying put, filter-aware sparse permutation, "clearing the filter does not preserve the visible subset") matches between the spec descriptions and `sheet.md`.
- The CURSOR/PROGRESS family flagged on `errors.go` by coverage heuristics is a hint-string-only change; pagination/cursor handling is untouched by this diff.
- Backend wire-shape claims (per-sheet arrays, freeze/filter readback normalization) describe the server and cannot be verified from this checkout; spec, docs, and tests are mutually consistent, which is the reviewable contract here.

## Verdict: COMMENT

A documentation/spec PR with a coherent contract migration, comprehensive pinning tests, and a one-line safe production change. The two P2 items are wording/UX polish, not merge blockers.

yujiawei (changes requested):
# Code Review — PR #161 (octo-cli)

Reviewed at head `13f874e9`, merge-base `b86f638` (one commit, clean rebase onto current `main` including #154).

**Verdict: CHANGES_REQUESTED** — two blocking issues, both introduced by this diff, both on paths that silently destroy user data under the new replace-style validation contract.

Local verification on this head: `go test ./... -count=1` (12/12 packages pass), `go vet ./...` clean, `jq empty internal/registry/specs/docs.json` clean, `go build ./...` clean. I additionally dry-ran every new documented example through the CLI's pre-flight body validator (`--dry-run docs sheet edit` with the freeze, filters, dropdown-array, flat-key repair, and prefixed-dims payloads) — all are accepted and serialized verbatim, and `octo-cli schema docs.sheet.edit` now exposes `freeze` / `filters`. So the client side of this change is sound; the problems below are in the guidance itself.

---

## 1. Contract / scope compliance

The diff implements what the description claims: freeze, filters, and grouped `dataValidations` on both `docs sheet get` and `docs sheet edit`; the whole-row sorting contract with `{v,f,s,p,t}` preservation; single-select / multi-select / checkbox under one replace-per-sheet REST shape; the server-directed deletion-only repair path; and #154 reconciliation. Registry and embedded-skill docs move together, no new command or flag, no unrelated files. Two notes:

**1a. The `docs.versions.state` change is correct but undeclared.** `sheetList` is removed from that operation's response properties (`internal/registry/specs/docs.json:1021-1026`) and from the `common.md` enumeration. That is defensible — `internal/registry/loader_test.go:767-774` guards it with the reason "the version-state handler does not decode `sheetHyperLinks`, `sheetMerges`, or `sheetList`" — but it reverses a field #154 added one commit earlier, and the PR description does not mention it. Please call it out so a reader can tell a deliberate correction from an accidental drop.

**1b. The rule-overlap constraint is deleted from the spec entirely.** `internal/registry/specs/docs.json` now contains zero occurrences of "overlap". `b86f638` documented it twice — "Rules may not overlap the complete post-edit rule set" in the `docs.sheet.edit` description and "Rules must not overlap the complete post-edit rule set" in the `dataValidations` property. Only `skills/octo-docs/sheet.md:255` retains a *narrowed* form ("Checkbox ranges must not overlap the complete **checkbox** rule set"). Two problems: the spec is the surface the CLI itself points agents at (`invalid_argument` → "inspect the operation schema with `octo-cli schema <op>`"), so a schema-only consumer now has no way to anticipate a server-enforced `422 sheet_cell_invalid`; and the dropdown section never states whether `list` / `listMultiple` ranges may overlap each other or a checkbox range, which is exactly the question a mixed-family sheet raises. Restore the constraint in the spec, and state it for every family the array now accepts.

---

## 2. Implementation quality

### P1 — `too_many_data_validations` hint now recommends a data-losing retry

`internal/output/errors.go:153`

```go
"too_many_data_validations": {"validation", "split data-validation changes into smaller batches and retry"},
```

Before this PR the hint read "split **checkbox** validation changes into smaller batches and retry", and under the flat `${logicalId}!${uid}` contract that advice was safe: each key was an independent upsert, so N smaller batches composed to the same result as one large batch.

This PR makes `dataValidations` replace-style per logical sheet — `internal/registry/specs/docs.json:542`: "It replaces that sheet's entire rule set". Splitting one sheet's rules across batches is now destructive: batch 2 replaces batch 1's rules rather than adding to them, so the agent ends with only the final batch's rules and no error. The hint fires precisely when a sheet has too many rules, i.e. exactly the case where the most rules stand to be lost.

The wording change is also locked in by `internal/output/envelope_test.go:233`, so the fix needs both files.

Suggested replacement: split across *sheets*, never within one, and reduce the per-sheet rule count instead — e.g. `"send fewer rules per sheet; each sheet's array replaces its whole rule set, so splitting one sheet across retries drops rules"`.

### P1 — README checkbox quick-start silently wipes a sheet's other validations

`README.md:189-190`

```bash
octo-cli docs sheet edit sheet-9 --base-version "<token>" \
  --data '{"cells":{"default!1:0":{"v":0}},"dataValidations":{"default":[{"uid":"checkbox-a2",...}]}}'
```

At `b86f638` this same example used `"dataValidations":{"default!checkbox-a2":{...}}` — a single-key upsert that left every other rule alone. Converting it to the grouped array turns a copy-pasteable one-liner into "replace all validations on `default` with just this checkbox". Run it on a sheet that already has the dropdowns this PR introduces and they are gone. `--base-version` does not help: the guard proves nobody else moved the sheet, not that the caller meant to drop the rest.

`skills/octo-docs/sheet.md:279` got the caveat for the equivalent example ("The array below is complete only when the sheet has no other validation rules") and `sheet.md:172-177` spells out the read-merge-resend discipline. The README did not. Either mirror the caveat, or keep the README example non-destructive.

### P2 — `too_many_sheet_resources` is documented but unmapped, and the 413 fallback is actively misleading

`internal/registry/specs/docs.json:479` introduces `413 too_many_sheet_resources` (also asserted present in `skills/skills_test.go:59` and `skills/octo-docs/sheet.md:480`), but it has no entry in `backendErrorMapping` (`internal/output/errors.go:113-155`). Unmapped snake_case codes fall through to `hintFromStatus(413)` (`internal/output/errors.go:376-381`) → `codeFromStatus(413)` = `PAYLOAD_TOO_LARGE` → hint **"request body exceeds 1MB limit"**. That sends the agent off shrinking the payload when the actual cap is a resource-key count; a small body with many freeze/filter keys will hit it. Its sibling `too_many_data_validations` is mapped, so this is an inconsistency, not a deliberate omission. Add the mapping plus a regression test alongside `TestWriteError_DocsTooManyDataValidations`.

### P2 — the flat-key repair path shares a map whose `null` means "delete everything"

`skills/octo-docs/sheet.md:269-272` interpolates a server-supplied string as a key in the same map where `{logicalId: null}` is documented to remove every rule on that sheet (`internal/registry/specs/docs.json:479`). The doc correctly frames this as the one server-directed exception, but never says what the server does with a flat key sent *outside* that path, nor how the caller confirms the repair applied rather than silently no-op'ing. Worth one sentence, given the blast radius of the neighbouring semantic.

### P2 — the new regression tests assert prose, not behavior

`skills/skills_test.go:75-204` and the `sheet.md` section tests check for ~60 exact sentence fragments ("first scrollable", "recommended client-rendering inputs, not additional server-required fields", …). The negative guards against superseded flat-map / checkbox-only guidance are genuinely useful and I'd keep those. The positive assertions will fail on any legitimate rewording while not catching a semantically wrong rewrite that happens to retain the fragments — and notably none of them would have caught either P1 above. Consider trimming to the load-bearing contract statements.

---

## 3. Coverage and what I could not verify

- **Backend contract.** MR !123 is on a private host I cannot read. Every server-behavior claim here is unverified from this repo: that `p` / `t` satisfy the "at least one of v/f/s/p/t" set rule, the `-1` normalization of an unfrozen axis, `operator` stripping on `list` rules, the uid `1–128` / no-`!` charset rule, `filterColumns:[]` being omitted on readback, and the server accepting a flat key for the repair path. If backend tests cover these, linking them would close the gap.
- **Deployment ordering.** `internal/registry/specs/docs.json:479` tells the caller "do not write if the deployed read surface omits any validation family on that sheet" — i.e. the docs concede the read surface may not be deployed yet, while giving the agent no way to detect that. If this ships before the backend, a well-behaved agent following the replace-style instruction will drop rules it could not see. Please confirm the merge ordering against !123.
- **Rendering.** Whether `renderMode:2` plus a comma-separated `formula2` actually produces the colored dropdown described is a web-client behavior I have no way to exercise here.
- **Not in the diff.** `nextCursor` becoming `"type": ["string","null"]` is fine: the spec declares `openapi: 3.1.0`, `internal/registry/specs/summary.json:208` already uses a union, `schemaType` (`internal/registry/loader.go:901-915`) collapses it, and `octo-cli schema docs.sheet.get` still renders `nextCursor` as `type: "string"`. I checked this specifically and it is a non-issue.

---

Fix the two P1s and I'm happy to re-approve. The substance of the documentation work is good — the sorting contract, the coordinate-bound resource enumeration, and the `sheetDims` bare-vs-prefixed twin hazard are all things an agent would otherwise get wrong, and they are documented carefully.

yujiawei:
Follow-up to my review above — one additional P2 and a sharpened detail on an existing one. **Standing verdict is unchanged (CHANGES_REQUESTED); the two P1s are still the blockers.**

### P2 (new) — the 413 repair path is still described as checkbox-only

This PR deliberately de-checkbox-ified the surrounding terminology (`internal/output/errors.go:153` went from "split **checkbox** validation changes" to "split **data-validation** changes"), but three descriptions of the `repairDataValidationKey` flow kept the narrow scope:

- `internal/registry/specs/docs.json:384` — "When **checkbox** metadata exceeds the size or inspection budget, 413 includes baseVersion and repairDataValidationKey…"
- `skills/octo-docs/sheet.md:264` — "the server could not finish its bounded **checkbox-rule** inspection"
- `skills/octo-docs/sheet.md:275` — "The backend only permits this path when it strictly shrinks the raw **checkbox** metadata"

The repair flow is family-agnostic — the server hands back a key, the client deletes that key. An agent that hits a repair 413 on a dropdown-heavy sheet and reads any of these three sentences can reasonably conclude the path does not apply to it and give up, on a sheet it now cannot read at all. Please widen all three to data-validation metadata, matching what you already did to the error hint.

### Sharpening the `too_many_sheet_resources` P2

I claimed the unmapped code produces a misleading hint; here is the measured result rather than the inference. Feeding the backend envelope through `ParseBackendError` at head `13f874e9`:

```
code=too_many_sheet_resources   status=413 -> Type="validation" Hint="request body exceeds 1MB limit"
code=too_many_data_validations  status=413 -> Type="validation" Hint="split data-validation changes into smaller batches and retry"
```

So the fallback is not a graceful degradation to "no hint" — it is a confident misdiagnosis. A caller who trips the resource-key **count** cap is told to shrink a **byte** payload, and shrinking the body will not clear it. That is the argument for adding the one-line mapping rather than leaving it to the 413 default.

Worth noting the same fallback shape mis-describes other unmapped sheet codes too (`protected_range` at 403 resolves to "bot lacks permission; check space membership", which sends the caller to check space membership for a range-protection conflict). That part is pre-existing and out of scope here — I mention it only because it argues the 413 default is not a safe catch-all to rely on for a newly introduced code.

luoshu0211:
RC addressed at current head `0e3e167cdc03489866b7110068eaf48f494314c9`, rebased onto current `main` `f9087492`.

Blocking fixes:

- `too_many_data_validations` no longer recommends splitting one sheet across retries. The hint now says to reduce rules per sheet and explicitly warns that each array replaces that sheet's complete rule set.
- The README checkbox quick-start now states that its array is safe only when the sheet has no other validation rules, and otherwise requires read/merge/resend of every rule to keep.

Additional review follow-ups:

- Added an explicit `too_many_sheet_resources` mapping plus regression coverage; the hint now identifies the actual freeze/filter top-level key-count gate instead of the generic 1 MiB fallback.
- Restored overlap guidance in both the schema and embedded skill with the backend's exact behavior: checkbox-vs-checkbox overlap is rejected; overlaps involving `list` / `listMultiple` are accepted by the backend and must be intentional.
- Clarified that a flat repair key must be server-returned, deletes checkbox-shaped stored metadata only, must strictly shrink that metadata, and must be followed by a read.
- Updated the PR description to call out the intentional `docs.versions.state` omission and backend-first merge order.

I did not widen the repair wording to all validation families because backend !123 does not implement that contract. At !123 head `f26b3c42`, `validateSheetDataValidationDeletionRepairBatch` measures and permits only `isStoredCheckboxCandidate` entries and rejects non-checkbox flat-key deletions. Likewise, overlap resolution is applied only through the checkbox validator; non-checkbox candidates remain in `validateSheetDataValidationEntries`. The new wording follows that implementation rather than documenting a broader unsupported contract.

Verification on the amended tree:

- focused Go packages passed
- both new negative mutation checks failed for the intended reason, then passed after restoration
- schema JSON and diff checks passed
- npm package tests passed 9/9
- rebuilt CLI schema and embedded skill readback contain the corrected contract

The local post-amend full Go rerun reached the linker but the host had only ~187 MiB free and failed with `no space left on device`; no assertion failed. GitHub CI is rerunning the full suite on this exact head. Please re-review `0e3e167c`.

Jerry-Xin (approved):
# Re-review round 2 — head `0e3e167c` (round-1 APPROVE on `13f874e9` was dismissed; yujiwei's CR on that head was the gate)

I byte-verified every item of yujiwei's `CHANGES_REQUESTED` against the head it reviewed (`13f874e9`), then re-verified each fix byte-for-byte on this head. My round-1 APPROVE missed all of the items below; that was an under-call on my part — credit yujiwei for catching them.

## Adjudication of the round-1 CR items on `13f874e9` (all confirmed real)

- **`docs.versions.state` field removal undeclared.** At merge-base `b86f638` the operation both enumerated `sheetList` in its description and exposed it as a response property; `13f874e9` removed both. The removal is deliberate (the loader test guards it with the reason "the version-state handler does not decode `sheetHyperLinks`, `sheetMerges`, or `sheetList`"), but the old description never said so. Valid.
- **Overlap constraint deleted from the spec.** Merge-base `docs.json` stated the rule twice ("Rules may/must not overlap the complete post-edit rule set"); `13f874e9` had zero occurrences of "overlap" in the spec, leaving only a narrowed checkbox-only sentence in `sheet.md`. A schema-only consumer lost the ability to anticipate the server-enforced `422 sheet_cell_invalid`. Valid.
- **`too_many_data_validations` hint recommended a data-losing retry.** The old hint "split data-validation changes into smaller batches and retry" was safe under the previous flat-key upsert contract but destructive under this PR's replace-style contract: batch 2 replaces batch 1's rules rather than appending, so splitting one sheet's rules across retries silently keeps only the last batch. The hint fires exactly when a sheet has the most rules to lose. Valid.
- **README quick-start silently wiped a sheet's other validations.** Converting the checkbox example to the grouped array without the caveat `sheet.md` already carried turned a copy-pasteable one-liner into "replace all validations on `default` with just this checkbox". Valid.
- **`too_many_sheet_resources` documented but unmapped.** With no `backendErrorMapping` entry, a 413 with that code fell through to the generic payload hint ("request body exceeds 1MB limit"), sending agents off shrinking body size when the real cap is a freeze/filter key count. Valid.

## Fix verification on this head (`0e3e167c`), byte-for-byte

- **versions.state declaration.** The PR description now explicitly declares the intentional omission of `sheetList`/`sheetHyperLinks`/`sheetMerges` from `docs.versions.state` and the reason; the in-repo loader-test guard carries the same rationale. Resolved.
- **Overlap restored in the spec, family-accurate.** Both spec sites now state "Checkbox ranges must not overlap another checkbox rule after the complete edit; list and listMultiple overlaps are accepted by the backend and should be sent only when intentional"; `sheet.md` mirrors the distinction and both assertions are test-locked in the registry loader test and the skills contract test. The `422 sheet_cell_invalid` gate is documented in the same description, so schema-only consumers can anticipate it again. Caveat: backend !123 is on a host not readable from this repo (mirror stale), so the server-side reject behavior is adjudicated on spec+test coherence and the PR's stated !123 semantics, not byte-verified. Resolved.
- **Lossy hint replaced.** New hint: "send fewer rules per sheet; each sheet array replaces the whole rule set, so splitting one sheet across retries drops rules" — warns of the loss instead of advising it. Test-locked: I reverted the hint locally and `TestWriteError_DocsTooManyDataValidations` fails; restored, it passes. Resolved.
- **README caveat added.** The checkbox quick-start is now scoped to "a sheet that has no other validation rules" and points readers to read `sheetDataValidations` and resend every rule to keep — same discipline as `sheet.md`. Resolved.
- **`too_many_sheet_resources` mapped.** New mapping with the hint "send fewer freeze or filter sheet entries per edit; split only those top-level sheet keys across batches" — safe because freeze/filter entries are keyed per-logicalId upserts, not replace-style arrays — plus the requested regression test alongside the validations one. Resolved.
- **Repair path hardened.** Spec and `sheet.md` now both say flat keys are accepted only for deleting checkbox-shaped stored metadata and only when strictly reducing it, that callers must not invent a flat key, and that completion is confirmed by re-reading until no repair key returns. Resolved.

## Delta review (`13f874e9` → `0e3e167c`)

The new head is rebased from `b86f638` onto current main `f908749` (checkbox #154 + unified plugin API #142). I confirmed the marketplace spec / service / CHANGELOG / agent-guidance-doc churn in the range is byte-identical to main — pure rebase pickup, not new authorial surface. The author's own delta is confined to seven doc/test files, each addressing a CR item; `docs_test.go`, `SKILL.md`, and `common.md` are unchanged between the two heads.

## Local verification on this head

`jq empty internal/registry/specs/docs.json` clean; `go vet ./...` clean; `go build ./...` clean; `go test ./... -count=1` — 12/12 packages pass; negative mutation check on the lossy-hint guard fails when reverted.

## Carried, non-blocking

- The prose-assertion style note ("consider trimming ~60 exact-sentence assertions") is deliberately kept as-is; the two new overlap assertions are load-bearing contract statements.
- Server-side claims about !123 (overlap rejection scope, operator stripping, uid charset, flat-key repair acceptance) remain unverifiable from this repo; the PR description declares the required merge ordering ("must merge after" !123).

APPROVE — every blocking item from the round-1 CR is fixed and test-locked.

yujiawei (approved):
# Code Review — PR #161 (octo-cli)

Re-review at head `0e3e167c`, merge-base `f908749` (the branch has been rebased onto current `main`, which now includes the marketplace migration #142 — the description still says `b86f638`). The previous formal review was at `13f874e9` and requested changes on two P1 items; both are fixed. Verdict is **APPROVE**.

Local verification on this exact head:

- `go build ./...`, `go vet ./...` — clean
- `go test ./... -count=1` — 12/12 packages pass
- `gofmt -l .` — clean; `git diff --check f908749..0e3e167c` — clean; `jq empty internal/registry/specs/docs.json` — clean
- Rebuilt CLI: `schema docs.sheet.edit` exposes `freeze` / `filters`, and `schema docs.sheet.get` still renders `nextCursor` as `type: "string"` (the `["string","null"]` union collapses correctly — `openapi: 3.1.0` at `internal/registry/specs/docs.json:2`, union already used pre-diff at `docs.json:181` and `specs/summary.json:208`, and `schemaType` at `internal/registry/loader.go:982-996` drops the `null` member)
- Confirmed the only production change is presentation-only: `{"error":"too_many_sheet_resources"}` at 413 already resolved to `Type: "validation"` via `typeFromStatus` (`internal/output/errors.go:339-340`) before this diff, and `backendErrorMapping` reuses the same type (`errors.go:154`). Exit code stays 2 (`errors.go:48-56`) — no scripted-caller behavior change, only the hint text.
- Grepped the whole tree: **no** non-test Go code references `dataValidations`, `sheetDataValidations`, `sheetFreeze`, or `sheetFilters`. The CLI is a pure `--data` pass-through here, so the shape migration cannot break CLI runtime behavior.

---

## 1. Spec / scope compliance

**Spec: ✅**

- **Missing:** none. Both previously-blocking items are resolved.
  - `too_many_data_validations` no longer recommends a destructive split: `internal/output/errors.go:153` now reads `"send fewer rules per sheet; each sheet array replaces the whole rule set, so splitting one sheet across retries drops rules"`, mirrored in `internal/output/envelope_test.go:233`.
  - The README quick-start no longer silently wipes a sheet's other rules: `README.md:189-191` carries the replace-style caveat above the grouped-array example.
  - The rule-overlap constraint is back in the spec, in both places it was dropped from — `internal/registry/specs/docs.json:479` and `:542` — and it now states the per-family behavior instead of the old blanket sentence, matching `skills/octo-docs/sheet.md:170-173` and `:260`.
  - `too_many_sheet_resources` is now mapped (`internal/output/errors.go:154`) with a regression test (`internal/output/envelope_test.go:246`), so the newly documented 413 no longer resolves to the misleading `"request body exceeds 1MB limit"` fallback.
  - The flat-key repair path gained the two missing sentences: `internal/registry/specs/docs.json:479` (`"do not invent a flat key"`) and `skills/octo-docs/sheet.md:279-283` (plus a completion criterion: re-read until no repair key comes back).
- **Extra:** the `docs.versions.state` schema change (`docs.json:1021-1027`) is the one edit outside the stated "P0 bot controls" scope — it adds `sheetFreeze` / `sheetFilters` and removes `sheetList`. Adding the two is in scope. Removing `sheetList` is a reversal of what #154 added one commit earlier; see P2-1.
- **Divergence:** none. No new command or flag, registry and embedded skill move together, no unrelated files.

Note: the previous round asked for the `docs.versions.state` change to be called out in the description. The description is unchanged and now also stale on two facts (base is `f908749`, not `b86f638`; verification is recorded against `13f874e9`, not this head). That is bookkeeping, not a defect — please refresh it or fix it up in the squash message rather than pushing another commit just for this.

---

## 2. Code quality

**Quality: Approved** — no P0/P1. Five P2s, all fix-forward.

### P2-1 — `docs.versions.state` now contradicts merged #154 about `sheetList`, with no evidence either way

`internal/registry/specs/docs.json:1021-1027` drops `sheetList`, and `internal/registry/loader_test.go:821-827` pins the absence:

```go
// The version-state handler does not decode sheetHyperLinks, sheetMerges, or
// sheetList. Remove this guard when that backend response starts returning
// those fields and the schema is updated in the same change.
```

`#154` added that same field one day earlier with a specific purpose ("used with `sheetDataValidations` to exclude rules for deleted tabs"). Exactly one of the two is right, and neither is verifiable from this checkout. Consequence if the removal is wrong: a bot reading a sheet snapshot can no longer tell that tab names are available, so it cannot map `logicalId` → tab name or filter rules for deleted tabs. It is a read-only preview endpoint with no write path, hence P2 rather than P1 — but please state the evidence (a handler path, or a backend test) in the description or restore the field.

### P2-2 — the `too_many_sheet_resources` hint names only two of the three resource families

`internal/output/errors.go:154`: `"send fewer freeze or filter sheet entries per edit; split only those top-level sheet keys across batches"`.

The spec defines the resource set more broadly: `docs.json:479` says *"P0 worksheet resources are: freeze …, filters …, and dataValidations …"*, and the gate as *"worksheet-resource key count over the max → 413 too_many_sheet_resources"*. If validation sheet-keys also count toward that cap, an agent that trips it while writing `dataValidations` across many sheets is told to reduce freeze/filter entries it may not have sent at all. The remedy itself is safe (splitting *top-level sheet keys* is lossless for all three families, since each key is one whole sheet), so this is wording only — say "freeze, filter, or validation sheet entries", or drop the family list.

### P2-3 — the `too_many_data_validations` hint closes the wrong door but not the right one

`internal/output/errors.go:153`. The clause after the semicolon correctly rules out splitting one sheet across retries. What it does not say is what to do instead. Concretely: a sheet holds 40 rules, an agent reads them and resends 45, the cap is 42 → 413, the stored 40 are untouched. Read literally, "send fewer rules per sheet" resolves to *drop three of the rules you just read* — which is a silent loss the hint does not flag. The lossless option is missing: consolidate ranges so several ranges ride on one rule, since `ranges` is already an array.

An independent second-opinion pass rated this same line **P1/blocking**, on the reasoning that "send fewer rules" *necessarily* deletes rules under replace semantics. I read it as **P2** and I am not merging on the fence: the hint asserts nothing false, the 413 leaves the stored set intact so no write has lost data at the point the hint is shown, the type/exit-code are unchanged, and the surrounding surfaces (`docs.json:542`, `skills/octo-docs/sheet.md:161-168`) already spell out the read-merge-resend discipline. The gap is an omitted remedy, not an incorrect instruction. Recording the disagreement rather than averaging it away.

### P2-4 — the schema still cannot reject the obsolete flat write shape

`internal/registry/specs/docs.json:540-542` declares `dataValidations` as a bare `"type": "object"`. The whole migration — grouped array vs. flat `${logicalId}!${uid}` map — lives in prose plus substring assertions, so nothing structurally rejects a caller that sends the old shape, which is precisely the payload that silently replaces the wrong thing.

Worth knowing before acting on this: declaring `additionalProperties: {"type": ["array","null"]}` would today only help external OpenAPI consumers, **not** the CLI. `internal/registry/loader.go:918-923` reads only a boolean `additionalProperties` and deliberately ignores a schema-valued one ("Treat it as open until the validator can retain that nested schema…"). Structural client-side gating therefore needs a loader/validator change too — a reasonable follow-up issue, not something to bolt onto this PR.

### P2-5 — two doc gaps the new sorting contract leans on

- **Cell writes: replace or merge?** The sorting section (`skills/octo-docs/sheet.md:97-104`) instructs "carry every cell's complete `{v,f,s,p,t}` shape … so values, styles, rich text, and cell types stay together". That instruction only makes sense if a cell set replaces the whole stored cell object, and neither `sheet.md` nor `docs.json:510` ever says so. One sentence would close it.
- **The canonical sorting example drops `s` / `p` / `t`.** `skills/octo-docs/sheet.md:130-141` writes `{"v":"a"}` and `{"f":"=LEN(B2)"}`. It is defensible — the comment says every non-null value must come from the preceding `sheet get`, so these cells simply carry no style — but it is the snippet an agent copies, and read next to the paragraph above it, it demonstrates the opposite of the rule. Adding an `"s"` to one cell, or a note that these cells are unstyled, makes it self-consistent.

Also non-blocking: the residual overlap question. `sheet.md:170-173` covers checkbox-vs-checkbox (rejected) and list-vs-other-family (accepted) but never says whether two `list` rules may overlap each other; only the broader spec phrasing at `docs.json:479` implies yes.

---

## 3. Overall verdict

**APPROVE.** Spec ✅ and Quality Approved. The two blocking items from the previous round are genuinely fixed, not papered over, and the change now carries the strongest part of this work — the sorting contract, the coordinate-bound-resource enumeration, and the bare-vs-prefixed `sheetDims` twin hazard, all things an agent would otherwise get wrong.

A process note, since it is visible in the public timeline: counting #158, this is the tenth formal review round on the same body of work. Every remaining item above is a wording or follow-up-issue item. My recommendation is to land this and open a follow-up for P2-1 and P2-4 rather than spend an eleventh round on hint strings.

Separately, `check-sprint / check-sprint` is red on "Validate Sprint field on Project Board" — project-board metadata, not code, but it will need clearing before merge.

---

## 4. Suggested changes (all optional for this PR)

1. `internal/output/errors.go:153` — add the lossless remedy: `"…; consolidate ranges into fewer rules instead"`.
2. `internal/output/errors.go:154` — widen to "freeze, filter, or validation sheet entries".
3. Description — refresh the base SHA / verification head, and state the `sheetList` evidence.
4. Add a transport test for the server-directed repair payload (`{"dataValidations":{"default!checkbox-a2":null}}`). The old flat-key deletion assertion was removed from `cmd/service/docs_test.go:1142`, so the one flat-key shape that is still legal now has no coverage.
5. Follow-up issue: structural `dataValidations` validation, which needs the loader change described in P2-4.

## 5. Coverage — what I could not verify

- **Backend contract (MR !123).** Not reachable from this checkout. Every server-behavior claim here is unverified: the per-sheet array wire shape, `-1` normalization of an unfrozen axis, `operator` stripping on `list` rules, the uid `1–128` / no-`!` charset rule, `filterColumns:[]` being omitted on readback, `colId` being absolute rather than `ref`-relative, and — newly asserted in this round — that **`list` / `listMultiple` overlaps are accepted** (`docs.json:479`, `sheet.md:170`). That last one reverses what merged #154 documented as a blanket no-overlap rule; if the backend actually still rejects it, agents following this text will take a `422 sheet_cell_invalid`. Recoverable, no data loss, but linking the backend test would close it.
- **Deployment ordering.** `docs.json:542` tells callers "do not write if the deployed read surface omits any validation family" — i.e. the docs concede the read surface may lag, without giving the agent a way to detect it. If this lands before !123, a well-behaved agent following the replace-style instruction drops rules it could not see. Please confirm merge ordering.
- **`docs.versions.state` shape.** `docs.json:1026` now says that endpoint returns per-sheet arrays. That endpoint decodes a stored snapshot, and the same spec calls the stored form the *internal flat `${logicalId}!${uid}` Y.Map*. Whether the version-state handler normalizes to the public array shape is not checkable here, and it is the same handler whose field set P2-1 disputes.
- **Rendering.** Whether `renderMode: 2` plus a comma-separated `formula2` produces the colored dropdown described is web-client behavior I cannot exercise.
- **Second-opinion coverage, restated for symmetry.** One pass could not run the Go tests at all (sandbox limits) — I ran them here, 12/12 green. Neither pass examined the `docs.versions.state` field-set contradiction or the `additionalProperties` loader limitation; both are covered above from my own reading. Nothing in this review rests on a leg that failed or timed out.

mochashanyao (approved):
[Octo-Q · automated review]

---

PR: https://github.com/Mininglamp-OSS/octo-cli/pull/161
Head SHA: 0e3e167cdc03489866b7110068eaf48f494314c9

# Code Review — PR #161 (octo-cli)

## Summary

This PR documents the P0 spreadsheet bot controls and aligns every CLI surface with them: the embedded `docs.sheet.get`/`docs.sheet.edit` spec gains `freeze`, `filters`, and the replace-style per-sheet `dataValidations` array (list/listMultiple/checkbox), the skill doc `skills/octo-docs/sheet.md` is rewritten with freeze/filter/sorting/dropdown contracts, the README example moves to the array shape, and `internal/output/errors.go` updates the `too_many_data_validations` hint plus adds a `too_many_sheet_resources` mapping. Despite the `feat(docs)` title, the spec and error-map changes are runtime-visible because service commands are generated from the embedded registry; the test additions pin all of it against the real embedded artifacts. I found no P1/P2 issues: the new contract is internally consistent across spec, skill, README, error hints, and tests.

## Verification

Local checks at head `0e3e167c`:
- `go build ./...` — clean
- `go vet ./internal/registry/... ./internal/output/... ./cmd/service/...` — clean
- `go test ./...` — pass (full suite, including the new registry/service/skills/output tests)

Static cross-check of every documented claim against the embedded spec and command generation; no fabricated results above — all three commands were executed in this environment.

## Findings

No P0/P1/P2 issues.

- Contract propagation: `internal/registry/specs/docs.json` ships via `//go:embed` (`internal/registry/loader.go:18`) into `registry.New()` → `cmd/service.RegisterServiceCommands` (`cmd/root.go:67`); `--data` is an open transport with no hardcoded batch-key whitelist (only board `elements` index validation in `cmd/service/validate.go:23`), so the new `freeze`/`filters` keys flow to the PATCH unmodified. `cmd/service/docs_test.go` proves this end-to-end through the real command tree with If-Match assertion.
- Spec/skill parity: every new contract phrase the loader and skills tests pin exists verbatim in head `docs.json` and `sheet.md` (verified programmatically); `skills_test.go` reads the embedded skill FS, not a copy.
- Superseded-contract cleanup: no residual flat `${logicalId}!${uid}` write shape, `{v,f,s}`-only cell shape, or "only checkbox" wording remains outside the documented server-directed repair exception (`skills/octo-docs/sheet.md:148`, spec PATCH description) and negative test assertions.
- Apparent omission resolved: the PATCH description's "Existing surfaces are cells, dims, drawings, hyperlinks, merges, and sheets" is completed by the next sentence introducing freeze/filters/dataValidations as P0 resources, and the non-empty rule is centralized at operation level as the tests require.
- `docs.versions.state` dropping `sheetList` from its response schema is deliberate and pinned with an explanatory guard in `internal/registry/loader_test.go`; `skills/octo-docs/common.md`'s preview map list matches the schema exactly.
- Error hints match semantics: replace-style arrays make per-sheet splitting lossy, which the new `too_many_data_validations` hint states; `too_many_sheet_resources` guidance matches the per-logicalId freeze/filters shape (`internal/output/errors.go:153`).
- Sorting example in `sheet.md` is internally consistent (formula re-anchoring and explicit-null destination verified against the described coordinate rules).

## Verdict: APPROVED

Docs/spec/error-hint alignment PR with zero findings: the new P0 contract is mutually consistent across all six touched surfaces, the generated CLI transport carries the new batches without code changes, and the added tests pin the real embedded spec, skill text, and request path. Build, vet, and the full test suite pass at head.





[Octo-Q] verdict: APPROVE — zero P0/P1/P2 findings; the new P0 spreadsheet contract (freeze/filters/replace-style dataValidations) is mutually consistent across spec, skill, README, error hints, and tests; generated CLI transport carries the new batches unchanged; go build/vet + full test suite pass at head.

PR #163, opened 2026-09-07
fix(sheet): document sparse filter columns

luoshu0211:
## Summary

Expose the sparse sheet-filter column contract to bots and the built-in Octo Docs skill.

## Linked Spec

Depends on backend MR !128 and should merge after it:
https://codex.mlamp.cn/dmwork/octo-docs-backend/-/merge_requests/128

## How verified

- `go test ./...`
- `go vet ./...`
- schema, embedded-skill, and transport tests cover `enabledColumns`
- backend normalizes `[12,6,12]` to `[6,12]`; historical restore preserves it

## COMPREHENSION

Univer stores one continuous filter range per worksheet. `enabledColumns` records the exact absolute columns that should expose filter controls inside that range. For example, a G:M range with `enabledColumns:[6,12]` enables only G and M. Omitting the field is reserved for legacy snapshots where every column in the range is enabled.

This PR intentionally does not create or guess a sprint issue; the repository's `check-sprint` gate may require maintainers to link a real issue on the current board.


yujiawei (changes requested):
# Code Review — PR #163 (octo-cli)

Reviewed at head SHA `6083e623aa8f7365759858e288f732182c0a4067` against `main` (merge-base `7dc5aa6`). Diff is 5 files / +29 −12: OpenAPI description strings, the embedded `octo-docs` skill page, and the tests that guard both. No runtime Go code paths are touched.

I ran the affected packages locally at this SHA — `go test ./internal/registry/ ./skills/ ./cmd/service/` — all pass.

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. The stated goal (expose the sparse filter-column contract to bots and the built-in skill) is covered on all three surfaces: the write shape (`internal/registry/specs/docs.json:479` op description, `:538` `filters` property), the read shape (`:451` `sheetFilters`), and the human/agent-facing page (`skills/octo-docs/sheet.md:32,81-86,89-95`). Guard tests were added on each (`internal/registry/loader_test.go:729-734,779-784`, `skills/skills_test.go:59`, `cmd/service/docs_test.go:1246`).
- **Out of scope additions:** none. `docs.json:1025` (`docs.versions.state` → `sheetFilters`) and `skills/octo-docs/common.md:103` intentionally describe the map generically and do not enumerate the value shape, so leaving them untouched is consistent, not an omission.
- **Deviations:** none in scope. The transport test fixture was moved from a `A:B` range to a `G:M` range, which is a reasonable way to exercise the sparse case.

## 2. Code quality

### P1 — `skills/octo-docs/sheet.md:76`: prose now contradicts its own example

The paragraph still reads:

```
example below keeps rows whose column B raw value is `待处理`.
```

On `main` that was accurate — the example filtered `colId: 1` (column B). This PR moved the example to `ref` starting at column 6 with `filterColumns:[{"colId":6,...}]`, i.e. **column G**, but left the sentence pointing at column B.

This lands two lines above the sentence that exists specifically to stop index confusion ("`colId` is the absolute 0-based worksheet column, not an offset from `ref.startColumn`"). A reader trying to reconcile "column B" with `colId: 6` and `ref.startColumn: 6` gets no consistent reading, and the most available wrong inference — that `colId` is somehow relative — is exactly the misconception this section is meant to prevent. Since the skill page is the artifact bots consume, an incorrect column mapping here has the same practical weight as a wrong constant in code.

Fix is one phrase: `column B` → `column G` (and the surrounding sentence can name the range, e.g. "keeps rows whose column G raw value is `待处理` within the G:M range").

This is the only blocking item; everything below is advisory.

### P2 — `skills/octo-docs/sheet.md:32`: quick-reference notation is ambiguous

```
sheetFilters: { "logicalId": {ref,filterColumns?:[{colId,filters}],enabledColumns?:[colId]} },
```

Directly adjacent to `filterColumns?:[{colId,filters}]`, the bare `enabledColumns?:[colId]` reads like "array of objects with a `colId` field". The prose below disambiguates, but the cheat-sheet block is what gets skimmed and copied. Suggest matching the spec's own notation: `enabledColumns?:[<absolute 0-based column>]`, or literally `enabledColumns?:[6,12]`.

### P2 — interaction between `filterColumns` and `enabledColumns` is undefined

Neither `docs.json:451`/`:538` nor `sheet.md` says what happens when a column carries criteria in `filterColumns` but is absent from `enabledColumns` — rejected with `422 sheet_cell_invalid`, silently dropped, or accepted so that rows are hidden by a filter the user has no visible control to inspect or clear? That last outcome is a genuine user-facing trap, and it is the first combination a bot will produce once it starts building sparse ranges. One sentence stating the rule (and, if it is accepted, that it is intentional) closes it.

### P2 — read-modify-write on `filters` now has a silent-widening footgun

`filters` is a replace-style per-`logicalId` snapshot. A bot that reads `sheetFilters`, changes only the criteria, and resends the object without `enabledColumns` will — per the documented legacy rule — silently re-enable every column in `ref`. The `dataValidations` surface carries an explicit warning for exactly this shape ("read `sheetDataValidations` first and resend every rule to keep"); `filters` should now carry the equivalent one. "Omit `enabledColumns` only for legacy …" gets partway there, but it frames omission as a legacy-compatibility choice rather than as a destructive default on rewrite.

### P2 — no documented way to express "no filter buttons at all"

`filterColumns:[]` has explicitly documented semantics ("keeps the range but omits `filterColumns` on readback"), so a bot will reasonably try `enabledColumns:[]` by analogy. The new text says it "must be non-empty" but not what the server does with an empty array, and the only coverage is the generic `422 sheet_cell_invalid` gate far down the op description. Worth one clause: either state that `enabledColumns:[]` is rejected, or state that removing the filter entirely (`null`) is the way to get zero controls.

### P2 (nit) — grammar in the new spec sentences

`docs.json:538`: "it must be non-empty, contain only columns inside ref, and **selects** exactly which columns show filter buttons" — `selects` breaks the parallel with `be`/`contain`. Same at `:479` ("must be a non-empty list … and selects …"). Note that `internal/registry/loader_test.go:732` asserts this exact substring, so the fix touches the test too.

### P2 (nit) — assertion style couples tests to prose line-wrapping

`skills/skills_test.go:59` and `loader_test.go:729-734,779-784` assert long exact prose substrings ("G:M backing range", "enables only G and M", "ref G:M plus enabledColumns:[6,12] enables only G and M, not H-L"). Reflowing a paragraph, or normalizing the `H–L` en-dash in `sheet.md:85` against the ASCII `H-L` in `docs.json`, breaks the build with no contract change. This is a pre-existing pattern in the file rather than something this PR introduced — flagged only so it does not keep growing.

## 3. Overall verdict

**REQUEST_CHANGES** — the spec gate passes, but the P1 leaves the skill page self-contradicting on the exact point (absolute vs. relative column indexing) it is trying to teach. The fix is a one-phrase edit; the P2s are optional wording improvements.

## 4. Suggested changes

1. `sheet.md:76`: `column B` → `column G` to match the new `colId: 6` example. **(blocking)**
2. `sheet.md:32`: change `enabledColumns?:[colId]` to `enabledColumns?:[<absolute 0-based column>]`.
3. Add one sentence covering `filterColumns` entries whose column is not in `enabledColumns`.
4. Add the read-then-resend warning to the `filters` surface, mirroring `dataValidations`.
5. State what `enabledColumns:[]` does (or point at `null` for "no controls").

## 5. Points for manual human verification

This PR was routed as security-sensitive, so calling these out explicitly:

- **No security-relevant code in the diff.** Nothing touches auth, credentials, middleware, or transport. The routing looks like a keyword hit on the pre-existing op description (`base-version`/`If-Match`/`protected_range` wording). No secrets, tokens, or endpoints are added.
- **`enabledColumns` is a UI affordance, not access control — please make that explicit.** "`enabledColumns:[6,12]` enables only G and M, not H–L" can be read by a bot (or a human integrator) as a data-visibility or protection guarantee for H–L. It is not: those columns still return their values from `docs sheet get` and remain writable through `docs sheet edit`. Given this text ships to autonomous agents, one clause ("this controls filter-button visibility only; it does not restrict reads or writes — use protected ranges for that") is worth adding so the contract cannot be mistaken for a privacy boundary.
- **Deployment ordering needs a human decision.** The PR body says the paired docs-backend MR must merge first. The CLI request body is deliberately an open object (`cmd/service/docs_test.go:1150-1152`) and forwards unknown keys verbatim, so this PR is documentation-only in effect — which also means merging it before the backend is deployed will have bots faithfully follow the docs and take `422 sheet_cell_invalid`. Please confirm the backend is live before this reaches a released CLI / published skill bundle.
- **Unverifiable from this repository.** The behavioral claims now asserted as contract — the server deduplicates and sorts `enabledColumns`, rejects columns outside `ref`, treats omission as all-columns-enabled, and preserves the field across historical version restore — can only be confirmed against the backend implementation. Worth one manual check against the paired MR, because once these sentences ship in the embedded skill, agents will treat them as guarantees.

Jerry-Xin (dismissed):
[Correction: the APPROVE vote attached to this review was emitted in error by a mechanical verdict gate and has been dismissed. The analysis below concludes REQUEST_CHANGES and stands unchanged; it is superseded by a REQUEST_CHANGES review with the same findings on this head.]

**Verdict: REQUEST_CHANGES** (reviewed at head 6083e623aa8f7365759858e288f732182c0a4067)

Full documentation-accuracy census on this change: every new `enabledColumns` contract statement was byte-verified against the authoritative backend and frontend sources, and the guard tests were exercised with mutation probes. The new contract text is accurate. However, the moved example leaves one pre-existing prose sentence self-contradictory, which is blocking for a docs-accuracy PR. A concurrent review by yujiawei on this same head raised it first; I independently byte-confirmed it and adjudicated each of the remaining claims below.

**Blocking**

1. `skills/octo-docs/sheet.md:76` — the prose still reads "The example below keeps rows whose column **B** raw value is `待处理`", but this PR moved the example directly below from `colId:1` / ref columns 0..1 (A:B) to `colId:6` / ref columns 6..12 (G:M), i.e. column **G**. Verified against merge-base: on main the sentence was accurate (the old example filtered colId 1 = column B), so the contradiction is newly introduced by this PR. It sits two lines above the sentence that exists specifically to prevent index confusion ("`colId` is the absolute 0-based worksheet column, not an offset from `ref.startColumn`"), on the page shipped to bots as the embedded skill — a wrong column mapping here carries the same practical weight as a wrong constant in code. One-phrase fix: "column B" → "column G". (yujiawei P1 — confirmed byte-level.)

**Verified accurate (no change needed)**

All new contract claims match the authoritative server implementations (backend MR !128 `validateSheetFilter` + contract tests; frontend MR !79 filter binding):
- "non-empty when present" — backend throws `SheetSnapshotInvalidError` on empty/non-array; contract test pins rejection of `[]`. ✓
- "contains only columns inside ref" — backend rejects entries outside `[ref.startColumn, ref.endColumn]`; test pins rejection of `13` against a ref ending at column 12. ✓
- "server deduplicates and sorts" — backend normalizes via `Set` + numeric sort at ingest; test pins `[12,6,12]` → `[6,12]`. ✓
- "omit only for legacy … every column in ref enabled" — frontend falls back via `stored.enabledColumns ?? columnsInFilterRange(stored.ref)`. ✓
- "selects exactly which columns show filter buttons" — frontend gates button rendering with `keepEnabledFilterButtons(objects, enabled)`. ✓
- G=6, M=12, H–L=7–11 math is consistent across all three docs.json descriptions, sheet.md prose/example, and the transport-test fixture (ref columns 6..12, colId 6, enabledColumns [6,12]). ✓
- Passthrough probe: mutating the transport test input to unsorted `[12,6,12]` with a verbatim wire expectation still passes — the CLI does no local dedup/sort, so attributing normalization to the server is accurate. ✓
- Mutation probes on shipped artifacts: mangling `enabledColumns?:[` in docs.json turns the registry schema test RED; deleting "G:M backing range" from sheet.md turns the embedded-skill test RED; restores turn both GREEN. The new assertions genuinely pin the doc strings. ✓
- Rendered-Markdown probe (CommonMark): the changed paragraph renders with balanced code spans and the intended visible text; both changed fenced blocks are intact; byte-level scans show zero backslash-backtick sequences in all five changed files; docs.json uses ASCII "not H-L" exactly as loader_test asserts. ✓
- `go test ./...` and `go vet ./...` green at head, matching CI build & test. ✓
- Consumer census: repo-wide grep for `filterColumns` / `sheetFilters` / `enabledColumns` — every shape-documenting site is updated by this PR; remaining mentions (sheet/get op description, versions.state `sheetFilters`, common.md) are shape-agnostic and non-contradictory. ✓

**Non-blocking (adjudicated advisories from the concurrent review + my own)**

1. Endorsed: `sheet.md:32` cheat-sheet `enabledColumns?:[colId]` — adjacent to `filterColumns?:[{colId,filters}]` it reads like an array of objects; matching the spec's `enabledColumns?:[<absolute 0-based column>]` would remove the ambiguity.
2. Endorsed: the interaction of `filterColumns` criteria on a column absent from `enabledColumns` is undefined in the docs. Byte-verified: the backend validates `colId` only against `ref`, not against `enabledColumns`, so the combination is accepted; the resulting behavior (criteria active behind a column with no visible button) is exactly the kind of trap worth one documented sentence.
3. Endorsed: read-modify-write silent widening. `filters` is a replace-style snapshot; resending it without `enabledColumns` re-enables every column in ref (backend omission + frontend fallback verified). The `dataValidations` surface already carries the equivalent "read first and resend every rule to keep" warning; `filters` now deserves the same.
4. Endorsed with evidence: `enabledColumns:[]` — backend throws "must be a non-empty array", surfacing via the generic 422 `sheet_cell_invalid` gate. The docs state "must be non-empty" but never say the empty array is rejected; one clause closes it (or points at `null` for removing the filter entirely).
5. Endorsed: grammar — "must be non-empty, contain only columns inside ref, and **selects** exactly which columns show filter buttons" breaks parallelism in both docs.json filter descriptions; note loader_test pins the exact substring "selects exactly which columns show filter buttons", so the fix must touch the test too.
6. Partially endorsed (style nit): the long exact prose-substring assertions do couple tests to line wrapping (a reflow splitting "G:M backing range" or "enables only G and M" would break the build) — a pre-existing pattern. The en-dash half of that concern is over-called, though: skills_test asserts no H–L/H-L substring from sheet.md, so normalizing that dash alone would not break the build.
7. Refinement of the merge-order outcome in the concurrent review: an out-of-order merge does **not** produce 422s. Byte-verified at the backend MR's target branch (`test`): `validateSheetFilter` selectively copies only `ref`/`filterColumns` into the stored object and silently drops unknown fields — unlike `validateSheetCell` and `validateSheetFreeze`, which explicitly reject unexpected fields. So if this PR shipped before backend MR !128, agents sending `enabledColumns` would silently get legacy all-columns behavior — benign, but divergent from the documented contract. The declared merge order (backend !128 first, then frontend !79 and this PR) remains a hard requirement; !128 was open but marked can_be_merged at review time.
8. Endorsed: the docs should state that `enabledColumns` controls filter-button visibility only, not data access — H–L columns remain readable via `docs sheet get` and writable via `docs sheet edit` (verified: nothing in the backend or frontend change gates reads or writes on this field; `protected_range` is the actual access boundary). Agents could otherwise misread "enables only G and M" as a visibility guarantee.
9. Process: the `check-sprint` gate is red (no board issue linked); the PR body flags this as a maintainer action.
10. Optional follow-up: the `docs.versions.state` `sheetFilters` description could mention `enabledColumns` once backend restore-preservation ships; it documents no shape details today, so there is no contradiction either way.

Bottom line: one blocking one-phrase fix (sheet.md:76 "column B" → "column G"); everything else is optional polish. With that fixed, the change is sound and should merge only after backend MR !128 per the declared order.

Jerry-Xin (changes requested):
**Verdict: REQUEST_CHANGES** (reviewed at head 6083e623aa8f7365759858e288f732182c0a4067; supersedes a now-dismissed review from this account whose APPROVE vote was emitted in error by a mechanical gate — the analysis is unchanged)

Full documentation-accuracy census on this change: every new `enabledColumns` contract statement was byte-verified against the authoritative backend and frontend sources, and the guard tests were exercised with mutation probes. The new contract text is accurate. However, the moved example leaves one pre-existing prose sentence self-contradictory, which is blocking for a docs-accuracy PR. A concurrent review by yujiawei on this same head raised it first; I independently byte-confirmed it and adjudicated each of the remaining claims below.

**🔴 Blocking (P1)**

1. 🔴 `skills/octo-docs/sheet.md:76` — the prose still reads "The example below keeps rows whose column **B** raw value is `待处理`", but this PR moved the example directly below from `colId:1` / ref columns 0..1 (A:B) to `colId:6` / ref columns 6..12 (G:M), i.e. column **G**. Verified against merge-base: on main the sentence was accurate (the old example filtered colId 1 = column B), so the contradiction is newly introduced by this PR. It sits two lines above the sentence that exists specifically to prevent index confusion ("`colId` is the absolute 0-based worksheet column, not an offset from `ref.startColumn`"), on the page shipped to bots as the embedded skill — a wrong column mapping here carries the same practical weight as a wrong constant in code. One-phrase fix: "column B" → "column G". (yujiawei P1 — confirmed byte-level.)

**Verified accurate (no change needed)**

All new contract claims match the authoritative server implementations (backend MR !128 `validateSheetFilter` + contract tests; frontend MR !79 filter binding):
- "non-empty when present" — backend throws `SheetSnapshotInvalidError` on empty/non-array; contract test pins rejection of `[]`. ✓
- "contains only columns inside ref" — backend rejects entries outside `[ref.startColumn, ref.endColumn]`; test pins rejection of `13` against a ref ending at column 12. ✓
- "server deduplicates and sorts" — backend normalizes via `Set` + numeric sort at ingest; test pins `[12,6,12]` → `[6,12]`. ✓
- "omit only for legacy … every column in ref enabled" — frontend falls back via `stored.enabledColumns ?? columnsInFilterRange(stored.ref)`. ✓
- "selects exactly which columns show filter buttons" — frontend gates button rendering with `keepEnabledFilterButtons(objects, enabled)`. ✓
- G=6, M=12, H–L=7–11 math is consistent across all three docs.json descriptions, sheet.md prose/example, and the transport-test fixture (ref columns 6..12, colId 6, enabledColumns [6,12]). ✓
- Passthrough probe: mutating the transport test input to unsorted `[12,6,12]` with a verbatim wire expectation still passes — the CLI does no local dedup/sort, so attributing normalization to the server is accurate. ✓
- Mutation probes on shipped artifacts: mangling `enabledColumns?:[` in docs.json turns the registry schema test RED; deleting "G:M backing range" from sheet.md turns the embedded-skill test RED; restores turn both GREEN. The new assertions genuinely pin the doc strings. ✓
- Rendered-Markdown probe (CommonMark): the changed paragraph renders with balanced code spans and the intended visible text; both changed fenced blocks are intact; byte-level scans show zero backslash-backtick sequences in all five changed files; docs.json uses ASCII "not H-L" exactly as loader_test asserts. ✓
- `go test ./...` and `go vet ./...` green at head, matching CI build & test. ✓
- Consumer census: repo-wide grep for `filterColumns` / `sheetFilters` / `enabledColumns` — every shape-documenting site is updated by this PR; remaining mentions (sheet/get op description, versions.state `sheetFilters`, common.md) are shape-agnostic and non-contradictory. ✓

**Non-blocking (adjudicated advisories from the concurrent review + my own)**

1. Endorsed: `sheet.md:32` cheat-sheet `enabledColumns?:[colId]` — adjacent to `filterColumns?:[{colId,filters}]` it reads like an array of objects; matching the spec's `enabledColumns?:[<absolute 0-based column>]` would remove the ambiguity.
2. Endorsed: the interaction of `filterColumns` criteria on a column absent from `enabledColumns` is undefined in the docs. Byte-verified: the backend validates `colId` only against `ref`, not against `enabledColumns`, so the combination is accepted; the resulting behavior (criteria active behind a column with no visible button) is exactly the kind of trap worth one documented sentence.
3. Endorsed: read-modify-write silent widening. `filters` is a replace-style snapshot; resending it without `enabledColumns` re-enables every column in ref (backend omission + frontend fallback verified). The `dataValidations` surface already carries the equivalent "read first and resend every rule to keep" warning; `filters` now deserves the same.
4. Endorsed with evidence: `enabledColumns:[]` — backend throws "must be a non-empty array", surfacing via the generic 422 `sheet_cell_invalid` gate. The docs state "must be non-empty" but never say the empty array is rejected; one clause closes it (or points at `null` for removing the filter entirely).
5. Endorsed: grammar — "must be non-empty, contain only columns inside ref, and **selects** exactly which columns show filter buttons" breaks parallelism in both docs.json filter descriptions; note loader_test pins the exact substring "selects exactly which columns show filter buttons", so the fix must touch the test too.
6. Partially endorsed (style nit): the long exact prose-substring assertions do couple tests to line wrapping (a reflow splitting "G:M backing range" or "enables only G and M" would break the build) — a pre-existing pattern. The en-dash half of that concern is over-called, though: skills_test asserts no H–L/H-L substring from sheet.md, so normalizing that dash alone would not break the build.
7. Refinement of the merge-order outcome in the concurrent review: an out-of-order merge does **not** produce 422s. Byte-verified at the backend MR's target branch (`test`): `validateSheetFilter` selectively copies only `ref`/`filterColumns` into the stored object and silently drops unknown fields — unlike `validateSheetCell` and `validateSheetFreeze`, which explicitly reject unexpected fields. So if this PR shipped before backend MR !128, agents sending `enabledColumns` would silently get legacy all-columns behavior — benign, but divergent from the documented contract. The declared merge order (backend !128 first, then frontend !79 and this PR) remains a hard requirement; !128 was open but marked can_be_merged at review time.
8. Endorsed: the docs should state that `enabledColumns` controls filter-button visibility only, not data access — H–L columns remain readable via `docs sheet get` and writable via `docs sheet edit` (verified: nothing in the backend or frontend change gates reads or writes on this field; `protected_range` is the actual access boundary). Agents could otherwise misread "enables only G and M" as a visibility guarantee.
9. Process: the `check-sprint` gate is red (no board issue linked); the PR body flags this as a maintainer action.
10. Optional follow-up: the `docs.versions.state` `sheetFilters` description could mention `enabledColumns` once backend restore-preservation ships; it documents no shape details today, so there is no contradiction either way.

Bottom line: one blocking one-phrase fix (sheet.md:76 "column B" → "column G"); everything else is optional polish. With that fixed, the change is sound and should merge only after backend MR !128 per the declared order.

luoshu0211:
Superseded by #164. The replacement fixes the column B/G contradiction, clarifies sparse-filter read-modify-write and authorization semantics, documents empty-list behavior, and pairs with backend validation that rejects hidden criteria.

PR #164, opened 2026-09-07
fix(sheet): document sparse filter columns

luoshu0211:
## Summary

Expose the sparse sheet-filter column contract to bots and the built-in Octo Docs skill.

This replaces #163 with a fresh review surface and addresses all review feedback from that PR.

## Linked Spec

Depends on backend MR !128 and should merge after it:
https://codex.mlamp.cn/dmwork/octo-docs-backend/-/merge_requests/128

## Review fixes

- correct the example from column B to column G so it matches `colId: 6`
- make the quick-reference type explicitly use absolute 0-based column numbers
- document that every criteria column must also be enabled; backend !128 now rejects mismatches with `422 sheet_cell_invalid`
- document replace-style read-modify-write behavior so omitting `enabledColumns` cannot silently widen visible controls
- document that `enabledColumns:[]` is rejected and `filters.<logicalId>: null` removes all controls
- clarify that filter-button visibility is not read/write authorization

## How verified

- `jq empty internal/registry/specs/docs.json`
- `go test ./internal/registry ./skills ./cmd/service -count=1`
- `go test ./...`
- `go vet ./...`

## COMPREHENSION

Univer stores one continuous filter range per worksheet. `enabledColumns` records the exact absolute columns that expose filter controls inside that range. For example, a G:M range with `enabledColumns:[6,12]` enables only G and M. Omitting the field is reserved for legacy snapshots where every column in the range is enabled.

This is a UI filter-control contract, not an authorization boundary. Protected ranges govern write restrictions.

The repository's `check-sprint` workflow still requires a real linked issue in the current sprint; this PR does not invent one.


yujiawei (approved):
# Code Review — PR #164 (octo-cli)

Reviewed at head SHA `8ff353a1e519df600495c02a2a9480ef1156dd54` against `main` (merge-base `7dc5aa6`). Diff is 5 files / +43 −14: three OpenAPI description strings, the embedded `octo-docs` skill page, and the tests that guard them. No runtime Go code paths are touched.

Verification performed at this SHA:

- `go test ./internal/registry ./skills ./cmd/service -count=1` — pass; `go test ./...`, `go vet ./...` — clean; `jq empty internal/registry/specs/docs.json` — valid.
- **Every new contract statement was checked against the paired backend change at its current head, not taken on faith.** All of them match the validator as written there, including the one that changed direction since the previous revision (see §1).
- Mutation probes to confirm the new guard tests are load-bearing: mangling `G:M backing range` in `sheet.md` turns `./skills` RED; mangling `every filterColumns colId must also appear in enabledColumns` in `docs.json` turns `./internal/registry` RED; both restore GREEN.
- Passthrough probe: rewriting the transport fixture to unsorted `[12,6,12]` with a verbatim wire expectation still passes, so the CLI performs no local dedup/sort and attributing normalization to the server is accurate.

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. The change is a revision of a closed predecessor PR and its body enumerates six fixes; all six landed, and each of the earlier advisory items is now covered on every surface it applies to:
  - example column corrected to G to match `colId: 6` — `skills/octo-docs/sheet.md:77`
  - quick-reference notation made explicit — `skills/octo-docs/sheet.md:32` (`enabledColumns?:[<absolute 0-based column>]`)
  - criteria-column/enabled-column interaction now specified and enforced — `internal/registry/specs/docs.json:479`, `:538`, `skills/octo-docs/sheet.md:82-84`
  - replace-style read-then-resend warning added — `internal/registry/specs/docs.json:479`, `:538`, `skills/octo-docs/sheet.md:88-90`
  - `enabledColumns:[]` rejection and the `null` escape hatch documented — `internal/registry/specs/docs.json:538`, `skills/octo-docs/sheet.md:91-92`
  - filter-button visibility explicitly separated from read/write access — all three descriptions plus `skills/octo-docs/sheet.md:87-88`
  - the `selects` → `select` parallelism fix landed with its matching test update — `internal/registry/loader_test.go:732`
- **Out-of-scope additions:** none. No new flags, fields, commands, or runtime behavior. The transport fixture move from an `A:B` range to `G:M` (`cmd/service/docs_test.go:1172`) is the minimum needed to exercise the sparse case and is in scope.
- **Deviations:** none. One statement reverses the previous revision's understanding — the earlier review recorded that the backend validated `colId` only against `ref`, so a criteria column absent from `enabledColumns` was *accepted*; this PR documents it as rejected with `422 sheet_cell_invalid`. I confirmed against the paired backend change at its current head that the validator now performs exactly this subset check and its contract test pins the rejection message. **The documentation is correct as written.** This is the resolution of the earlier finding, not a deviation.

## 2. Code quality

**Quality: Approved**

No P0 or P1. The contract text is accurate on all four surfaces and the numbering (G = 6, M = 12, H–L = 7–11) is consistent across `docs.json:451`, `:479`, `:538`, the `sheet.md` prose and example, and the transport fixture. Findings below are all advisory.

### P2 — `internal/registry/specs/docs.json:451`: subset rule is unscoped on the read surface

The read description states the subset requirement unconditionally:

> `enabledColumns` is non-empty when present, contains only columns inside ref, … ; **every filterColumns colId also appears in enabledColumns**. … Omitted `enabledColumns` identifies a legacy snapshot where every column in ref is enabled.

The `when present` qualifier attaches to `is non-empty`; the subset clause then reads as a global invariant, which is vacuous for the legacy snapshot the very next sentence describes. Both write surfaces get this right — `:479` and `:538` open with `When enabledColumns is present it must be …` and carry the subset clause inside that scope. Suggest mirroring that scoping at `:451` so all three read the same way.

### P2 — `skills/skills_test.go:59`: the example block's indices are unpinned

The skill guard asserts prose (`column G raw value`, `G:M backing range`, `enables only G and M`) but never the literal numbers in the fenced example — no assertion covers `"startColumn":6`, `"colId":6`, or `"enabledColumns":[6,12]`. That leaves the example free to drift back to `colId: 1` while the prose still says column G and the test still passes. Since this PR exists specifically to repair that class of prose/example drift, pinning `"enabledColumns":[6,12]` (and ideally `"colId":6`) would close the loop.

### P2 — `internal/registry/loader_test.go:900-916`: the operation-level contract is unguarded

The `enabledColumns` assertions cover `edit.RequestBody.Properties["filters"]` (`:729-738`) and `get.ResponseSchema.Properties["sheetFilters"]` (`:783-790`), but nothing covers `sheetPath.Patch.Description`, where this PR *also* duplicated the full contract (`docs.json:479`). The existing shared loop at `:900-916` already asserts a substring list against both operation descriptions and is the natural home for one or two of these strings. As it stands, the operation-level copy could be weakened or contradicted without turning any test red.

### P2 (nit) — notation drift between the operation description and the property schemas

`docs.json:479` writes the shape as `enabledColumns?:[absoluteColumnIds]` and the example as `ref G:M **with** enabledColumns:[6,12]`, while `:451` and `:538` use `enabledColumns?:[<absolute 0-based column>]` and `ref G:M **plus** …`. Same contract, three spellings. Harmless to a human, but these strings are what schema-driven consumers and agents read; aligning them costs nothing. (Only the `plus` variant is test-pinned, so `:479` is free to drift.)

### P2 (nit) — `skills/octo-docs/sheet.md:86` uses an en-dash where the spec uses ASCII

`not H–L` (U+2013) in the skill page vs `not H-L` (U+002D) in all three `docs.json` descriptions. No build risk — nothing asserts the `H-L` substring out of `sheet.md` — so this is consistency only.

### P2 (nit) — the prose-substring assertion pattern grew

This revision adds roughly a dozen long exact-prose assertions across `loader_test.go` and `skills_test.go` (e.g. `ref G:M plus enabledColumns:[6,12] enables only G and M, not H-L`). A paragraph reflow that splits any of them breaks the build with no contract change. This is a pre-existing pattern rather than something the PR introduces, and it does buy real protection (the mutation probes above confirm the assertions bite) — flagged only because the surface keeps expanding.

### P2 (nit) — `cmd/service/docs_test.go:1172`: fixture cannot distinguish client from server normalization

The fixture sends already-sorted `[6,12]` and `:1246` asserts `[6,12]` on the wire, so it cannot fail if the CLI ever started normalizing locally. I verified by probe that unsorted `[12,6,12]` currently round-trips verbatim, which is what makes the documented "the server deduplicates and sorts" attribution correct — using the unsorted value as the fixture input would pin that property instead of leaving it to a manual check.

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. The blocking issue from the predecessor PR is fixed, every advisory item of substance was addressed, and each newly asserted behavior matches the paired backend change at its current head. The remaining findings are wording alignment and test-strength improvements; none of them should hold up the merge.

## 4. Suggested changes (all optional)

1. `docs.json:451` — scope the subset clause to `when enabledColumns is present`, matching `:479`/`:538`.
2. `skills_test.go:59` — assert `"enabledColumns":[6,12]` (and `"colId":6`) so the example's indices are pinned, not just the prose.
3. `loader_test.go:900-916` — add one `enabledColumns` assertion for `sheetPath.Patch.Description`.
4. `docs.json:479` — align `[absoluteColumnIds]` → `[<absolute 0-based column>]` and `G:M with` → `G:M plus`.
5. `sheet.md:86` — normalize `H–L` to ASCII `H-L`.
6. `docs_test.go:1172` — use `[12,6,12]` as the fixture input to pin "the CLI does not normalize".

## 5. Points for manual human verification

- **Merge ordering is the one live gate, and it is not yet satisfied.** The paired backend change is still open and unmerged, targeting a non-default branch. Merging this PR first is *not* an error path: the backend's filter validator builds its stored object by selective field assignment rather than passthrough, so a pre-change deployment silently drops `enabledColumns` instead of returning `422`. The consequence is a divergence window in which agents follow the new documentation, get no error, and see every column in `ref` keep its filter button — precisely the silent widening the new text warns about. Please confirm the backend is deployed before this reaches a released CLI or a published skill bundle.
- **`check-sprint` is red.** No linked board issue for the current sprint; the PR body flags this as a maintainer action. Every other check (build & test, npm test, osv-scan, ancestor check, labels) is green at this head.
- **No security-relevant code in the diff.** Nothing touches auth, credentials, middleware, or transport. `enabledColumns` is a UI affordance, and the diff now says so explicitly on all four surfaces — columns excluded from it remain readable via `docs sheet get` and writable via `docs sheet edit`; protected ranges are the actual access boundary. That clarification is the right call given this text ships to autonomous agents.
- **Restore preservation is claimed by the backend change's title but asserted nowhere in this diff.** `docs.json:1025` (`docs.versions.state` → `sheetFilters`) describes the map without a value shape, so there is no contradiction either way. If restore-preservation is a guarantee you want agents to rely on, it needs its own sentence in a follow-up; leaving it undocumented is the safe default for now.

Jerry-Xin (approved):
## Review — PR #164 @ 8ff353a1

**Verdict: APPROVE**

Re-submission of #163, treated as a completely fresh review surface. Every load-bearing claim independently byte-verified at this head; the predecessor's blocker is fixed, all six declared review fixes landed, and every contract statement matches the paired backend and frontend MRs at their current heads.

### 1. Predecessor blocker — fixed (byte evidence)

The prose/example contradiction in the `skills/octo-docs/sheet.md` filter section is resolved:

- Repo-wide greps for `column B`, `"colId":1`, and `startColumn:0,"endColumn":1` return **zero** hits across all `.md`/`.go`/`.json` files.
- The prose now reads "keeps rows whose column G raw value is `待处理` within G:M" (:76), and the fenced example uses `ref` `startColumn:6`/`endColumn:12` with `colId:6` and `enabledColumns:[6,12]` — G=6, M=12, H–L=7–11 consistent across all four surfaces (docs.json GET/operation/request descriptions, sheet.md prose + example, the docs_test.go transport fixture, and the test pins).
- The drift class is now guarded by shipped tests: mangling `column G raw value` → `column B raw value` in sheet.md turns `./skills` RED (mutation probe; restore GREEN, tree clean).

### 2. Contract claims vs the paired MRs (verified at their current heads)

Backend MR !128 (head 26e49d4 — the new sparse-filter commit):

- Subset rule `every filterColumns colId must also appear in enabledColumns` — implemented in `validateSheetFilter` and pinned by a contract test expecting the exact rejection message.
- `enabledColumns:[]` rejection, non-integer and outside-ref rejection — test-pinned (`[]`/`[6.5]`/`[13]`).
- "The server deduplicates and sorts" — test-pinned (`[12,6,12]` → `[6,12]`), and literally true **on readback**: the GET serialization path re-runs the same validator over the stored snapshot.
- `422 sheet_cell_invalid` — the exact error mapping for validator throws in the edit service.
- `null` removes the filter and all controls — null deletes the stored map entry; the frontend removal path drops range, criteria, and buttons together.
- "replace-style … resend the complete filter object" — the commit path replaces the whole per-logicalId entry, so omitting `enabledColumns` on a resend drops the stored field.

Frontend module MR !79 (head ea20f50):

- Legacy fallback `stored.enabledColumns ?? columnsInFilterRange(stored.ref)` present on both remote-apply paths — "Omitting enabledColumns invokes legacy all-columns-in-ref behavior" is accurate.
- Only enabled columns keep filter buttons (`keepEnabledFilterButtons`); the toggle logic maintains `ref` = [min(enabled), max(enabled)], so the documented G:M + `[6,12]` sparse example satisfies the frontend invariant exactly.
- Visibility-not-authorization framing is correct: writes are gated by protected ranges (403), never by `enabledColumns`.

The CLI itself is a pure passthrough for this payload (no client-side filter handling exists in non-test Go code), so attributing dedup/sort to the server is accurate.

### 3. Verification battery at this head

- `go test ./...` — all packages pass; `go vet ./...` — clean; `docs.json` parses as valid JSON.
- Mutation probes on shipped artifacts: (1) sheet.md G→B → `./skills` RED; (2) weakening the subset sentence in docs.json → `./internal/registry` RED; (3) changing the pinned example `[6,12]`→`[6,13]` → RED; all restore GREEN.
- Every pinned substring independently re-verified byte-exact: all 25 skill-page pins present; every new loader_test assertion matches both the request-property and the readback descriptions.
- Unchanged-adjacent-reference census (the gap that missed the predecessor's blocker): enumerated every unchanged sentence in touched files and sibling docs that references a changed example/field/value — the null-removal example comment, the sorting paragraph's `sheetFilters` references, the `versions.state` description (no value shape, no contradiction), sibling skill pages (no filter references) — all still hold; zero stale references repo-wide.
- Rendered-Markdown probe: the changed paragraph renders cleanly through a CommonMark parser — every code span intact, `<logicalId>` correctly escaped inside its code span, no backslash-backtick (0x5c60) sequences anywhere in the file.
- The en-dash occurrences in sheet.md (`not H–L` :86; pre-existing `1–128` :165) are benign: no test pins the `H-L` substring out of the skill page, and docs.json uses ASCII hyphens exactly where its pins apply.

### 4. Prior-round advisory items — disposition

All five advisory items from the predecessor round are addressed, and the new wording is byte-accurate:

1. Cheat-sheet ambiguity → the quick reference now reads `enabledColumns?:[<absolute 0-based column>]`.
2. filterColumns/enabledColumns combination previously undefined → now specified on all surfaces **and** enforced by the backend validator (this contract direction changed between the two revisions; verified against the current MR head, not the earlier finding).
3. Silent widening on read-modify-write → replace-style + "silently widens the visible controls" warning added, matching the actual fallback behavior.
4. `enabledColumns:[]` → rejection plus the `{"filters":{"<logicalId>":null}}` escape hatch documented.
5. Grammar parallelism → "must be non-empty, contain only …, and select exactly …" fixed with its matching pin.

### 5. Peer review adjudication

yujiawei approved this head ahead of this review. I independently re-verified all six of his advisory findings and confirm each is VALID at the P2/nit level: the unscoped subset clause on the read description (:451), the unpinned example indices in the skill guard test, the unguarded operation-level contract copy (`sheetPath.Patch.Description` has no enabledColumns assertion — confirmed against the shared assertion loop), the with/plus and placeholder-notation drift, en-dash vs ASCII, and the already-sorted fixture not pinning the no-local-normalization property. None blocks merge; I endorse his suggested changes as optional follow-ups. Item 2 deserves emphasis: pinning the literal `"enabledColumns":[6,12]`/`"colId":6` from the fenced example is precisely the drift class that produced the predecessor's blocker, and current pins cover only the prose.

### 6. Merge gates (process, not code)

- The paired backend MR is still open and unmerged, targeting a non-default branch; the frontend MR is likewise open. The PR body's declared ordering (backend first) is a live gate: the **current** backend validator on the target branch builds the stored filter by selective field assignment and silently drops `enabledColumns` (byte-verified at that head) — shipping this documentation before the backend change would give agents no 422 and the exact silent widening the new text warns about. Merge in the declared order.
- `check-sprint` is red (no linked sprint issue); the PR body discloses this as a maintainer action. Every other check is green at this head.

Reviewed at head 8ff353a1e519df600495c02a2a9480ef1156dd54 against main (merge-base 7dc5aa6).

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #164 (octo-cli)

## Summary

This PR documents the sparse sheet-filter column contract (`enabledColumns`) on every surface that reaches a bot: the embedded skill reference (`skills/octo-docs/sheet.md`), the `docs.sheet.get` response schema plus the `docs.sheet.edit` operation and request descriptions in `internal/registry/specs/docs.json`, and the substring tests that pin that prose. It also moves the worked example from an A:B range with `colId: 1` to a G:M range with `colId: 6` and `enabledColumns: [6,12]`, and adds a transport assertion for the new field. The contract is internally consistent and the example arithmetic checks out; both items below are about what the doc tells an agent to do next, not about the shape it documents.

## Verification

Static analysis only at head `8ff353a1`; build and tests not executed in this environment.

- ✅ **`enabledColumns` really reaches the server** — `--data` is parsed into a `map[string]any` (`cmd/service/run.go:569`), the pre-send gates are only required fields, enums, variants and the scene-element index check (`cmd/service/validate.go:24`), and unknown-field rejection fires only where a schema declares `additionalProperties: false` (`cmd/service/run.go:1024`). The `filters` node declares `type: object` with no `properties`, so validation does not recurse and the field is forwarded verbatim; `cmd/service/docs_test.go:1172` and `:1246` now lock that.
- ✅ **Readback cannot drop it** — no non-test Go file mentions `filterColumns` or `sheetFilters` (zero `rg` hits), so no typed DTO exists in either direction, and `octo-cli schema` prints descriptions untruncated (`cmd/schema.go`).
- ✅ **Three-surface parity** — the get schema (`internal/registry/specs/docs.json:451`), the edit property (`:538`), the operation description (`:479`) and the skill reference state the same rules: absolute 0-based columns, non-empty when present, inside `ref`, every `filterColumns.colId` included, mismatch or `enabledColumns:[]` → `422 sheet_cell_invalid`, dedup + sort on readback, omitted → legacy all-columns-in-`ref`. "Filter-button visibility only, not read or write access" appears on all four, which is correct given hidden rows are still returned by `docs sheet get`.
- ✅ **Example arithmetic** — `startColumn: 6` / `endColumn: 12` is G:M, `colId: 6` is G and a member of `[6,12]`, the excluded set is exactly H–L, and the header row (`startRow: 0`) stays visible. Compared file-by-file against the closed predecessor, round 2 only adds rules (`colId` membership, the `422`, replace-style, `enabledColumns:[]` rejection, visibility ≠ authorization) and drops none — including fixing the old "column B" prose over a `colId: 6` example.
- ✅ **Every new assertion string exists verbatim** — all 12 `filters` and 10 `sheetFilters` substrings in `internal/registry/loader_test.go:728` and `:782`, and all 25 needles in `skills/skills_test.go:59`, match the shipped text as literal substrings.
- ✅ **No stale mirrors or fixtures** — README's sheet examples carry no filter body, `skills/octo-docs/common.md:103` only lists `sheetFilters` as an empty map, `docs.versions.state`'s `sheetFilters` (`internal/registry/specs/docs.json:1025`) is field-less, and there are no `testdata`/golden snapshots. The spec parses, the operation-level-rule guard (`internal/registry/loader_test.go:711`) still holds, and nothing caps description length.

## Findings

No P0/P1 issues; two non-blocking P2 items and one nit.

### P2 — "Use protected ranges" names a control this CLI does not expose (`skills/octo-docs/sheet.md:87`)

The new paragraph sends the reader elsewhere for real access control, and the spec twins repeat it (`internal/registry/specs/docs.json:538` and `:479`: "use protected ranges for access control"). No embedded spec declares a protected-range operation — searching every `operationId` in `internal/registry/specs/*.json` for `protect` returns nothing — and the concept's only other appearance is the `403 protected_range` write gate (`skills/octo-docs/sheet.md:507`). An agent asked to restrict writes on the filtered columns will hunt through `octo-cli schema --list`, find no such surface, and either burn the turns or report a restriction it never applied. This file is otherwise careful about that case: `skills/octo-docs/sheet.md:132` states plainly that drawings are not returned and cell comments are managed outside the sheet-edit surface. Qualify the pointer in all three places, e.g. "Use protected ranges for that — they are managed in Octo Docs and are not exposed by `octo-cli`."

### P2 — New read-modify-write rule omits the deployed-version caveat its sibling carries (`internal/registry/specs/docs.json:538`)

The same operation description already carries a deployment-skew idiom for the neighbouring resource: "read `sheetDataValidations` first and resend every rule to keep, **but do not write if the deployed read surface omits any validation family on that sheet**" (`internal/registry/specs/docs.json:479`). The filter sentence added on this line — and its twins at `:479` and `skills/octo-docs/sheet.md:88` — keeps the read-first half and drops the capability-check half. These strings ship inside released binaries (`internal/registry/loader.go:18`, `skills/skills.go:12`) while the field itself lands in the backend change the PR description says must merge first. Against a backend that predates it, `enabledColumns` is an unknown field, and this description documents that the validator tolerates and strips unexpected input ("The server removes an `operator` from otherwise-valid `list` / `listMultiple` rules", `skills/octo-docs/sheet.md:506`) — so the expected outcome is a silent no-op, not a `422`. The agent reports buttons narrowed to G and M while every column in `ref` still shows one, and "Omitted `enabledColumns` identifies a legacy snapshot" (`internal/registry/specs/docs.json:451`) leaves it no way to distinguish a legacy sheet from an un-upgraded backend. Mirror the `dataValidations` idiom in the same three places, e.g. "if the deployed read surface never returns `enabledColumns`, the backend predates sparse filter columns — do not report filter-button visibility as applied."

### Nit — Prose wrap and dash drift (`skills/octo-docs/sheet.md:77`)

`:77` runs 88 characters where the file otherwise wraps near 80, and the shared example reads "not H–L" (en dash) in the skill but "not H-L" in `internal/registry/specs/docs.json:451` and `:538` ("ref G:M **with** …" at `:479` vs "**plus**" at `:451`/`:538`). Nothing breaks — `internal/registry/loader_test.go:733` pins the spec's hyphen form — so rewrap `:77` and align the skill's dash to it.

## Human-verify

1. The asserted server behaviors — `422 sheet_cell_invalid` on a `colId` ∉ `enabledColumns` mismatch or `enabledColumns:[]`, dedup + sort on readback, omitted-means-legacy — live in the backend change named in the PR description, outside this checkout, so they could not be confirmed here. **Not a merge blocker** for a docs PR that states the dependency; worth confirming the code is `sheet_cell_invalid` and not `invalid_body`, since `skills/octo-docs/sheet.md:504` enumerates both.
2. Release ordering: if a CLI build carrying these docs can be pointed at a backend without sparse-filter support, the second P2 becomes user-visible. **Not** a blocker if the stated deploy order is enforced.

- Paired paths: adding controls (`enabledColumns`) and removing them (`{"filters":{"<logicalId>":null}}`) are documented on both the write and read side (`skills/octo-docs/sheet.md:91`, `:105`; `internal/registry/specs/docs.json:538` now says "null removes the filter and all controls"); neither surface was updated alone.
- Authorization boundary: no capability is widened, the text is explicit that button visibility is not read/write access, and the CLI's pre-send gates are untouched.
- Build vs runtime: no generated artifacts or golden fixtures, no `additionalProperties: false` on the touched node, no description truncation — the spec/skill text plus the transport test are the whole runtime effect.
- Doc self-consistency: no leftover "column B" / `colId: 1` example, no surviving merge-style claim for filters, and the paged-read rule (`sheetFilters` on the first page only) keeps the read-modify-write instruction satisfiable for large sheets.

## Verdict: APPROVED

Nothing here changes what the CLI sends, accepts, or renders: the field was already transportable, it is now documented consistently on all four surfaces, and the transport plus substring tests lock it. The two P2 items are guidance-precision gaps an agent could trip over — a recommended control with no CLI surface, and a missing deployed-version caveat the sibling `dataValidations` contract already has — and both are small follow-ups rather than blockers. The nit is cosmetic.

PR #169, opened 2026-09-14
feat(docs): add sheet row and column structural commands

luoshu0211:
## Summary
- expose docs sheet rows insert and delete
- expose docs sheet columns insert and delete
- require If-Match via --base-version and validate zero-based anchors and counts locally
- document formula refusal, absolute stored indexes, legacy default logical id, and rollout ordering

## Linked Spec
Paired backend: https://codex.mlamp.cn/dmwork/octo-docs-backend/-/merge_requests/136

## How verified
- go test ./...
- go vet ./...
- go build ./...
- registry reports 42 docs operations and 332 embedded operations
- request tests verify all four POST routes, headers, JSON bodies, zero anchors, and local bounds rejection

## COMPREHENSION
These commands are thin generated clients over guarded backend structural operations. Inserts use afterRow or afterColumn; deletes use startRow or startColumn. All coordinates are absolute zero-based stored indexes. If-Match prevents a transport retry from applying the same structural edit twice. The CLI must ship only after backend MR !136 is merged, deployed, and the new routes are proven live.


yujiawei (dismissed):
# Code Review — PR #169 (octo-cli)

Reviewed at head SHA `dc8e1ec9d1d2f3837fedff87ac86a22aaa3d576e`. Verified locally: `go build ./...`, `go vet ./...`, and `go test ./cmd/service/ ./internal/registry/ ./skills/` all pass.

## Summary

This PR adds the `docs sheet rows insert` command purely through the embedded OpenAPI registry (`internal/registry/specs/docs.json`) plus tests and documentation. There is no hand-written command logic — the leaf `docs sheet rows insert` is auto-derived from the `docs.sheet.rows.insert` operationId by the existing command-tree builder in `cmd/service/service.go` (`attachOperation`), so `rows` becomes an intermediate group and `insert` the leaf. This is the same mechanism already used across the codebase and needs no bespoke wiring.

## Spec compliance

The diff does exactly what the PR describes, and nothing beyond it:

- **New operation** `docs.sheet.rows.insert` → `POST /v1/bot/docs/{docId}/sheet/rows/insert`, registered and asserted in `cmd/service/docs_test.go` (`TestDocs_RegistryShape`).
- **Required flags** `--base-version` (sent as the `If-Match` header, mirroring `docs.sheet.replace`), `--logical-id`, `--after-row`, `--count`, all marked required with correct `x-octo-flag` mappings.
- **Local bounds**: `afterRow` `0..9999`, `count` `1..10000`, enforced client-side via `x-octo-strict-request-schema: true` so invalid input fails before any request is sent (covered by `TestDocsSheetRowsInsert_ValidatesRequiredFlagsAndBoundsLocally`).
- **Docs** are updated consistently: CHANGELOG, README example, `CLAUDE.md` op count (328→329), `skills/octo-docs/SKILL.md` capability list, and `skills/octo-docs/sheet.md`, all cross-checked by `skills/skills_test.go` and `internal/registry/loader_test.go` (docs count 38→39).

No scope creep, no missing pieces, no divergence from the stated approach.

## Correctness

- The happy-path test (`TestDocsSheetRowsInsert_SendsFlagsAsGuardedStructuralRequest`) pins method, path, `If-Match` header, and the exact JSON body (`{logicalId, afterRow, count}`), confirming the transport is a version-guarded structural write rather than a whole-sheet rewrite.
- The formula-safety refusal (`409 sheet_formula_unsupported`), staleness (`412`), and other failure modes are documented as server-owned behavior — appropriate, since the CLI is a thin registry-driven client and must not emulate structural insertion locally. `sheet.md` explicitly warns against emulating insertion with a rewrite, which is the right guardrail.
- The rollout dependency on `octo-docs-backend` MR !136 is clearly documented in both CHANGELOG and `sheet.md`, including the correct guidance that a 404 in a mixed-version environment means "backend not ready," not "fall back to rewrite."

## Non-blocking observations (suggestions only)

- `count` has a local max of `10000` while the true invariant is "resulting `rowCount` ≤ 10000." That final cap is necessarily server-enforced (the CLI doesn't know the current row count), so the generous local bound is fine — just noting the local schema is a coarse sanity gate, not the real limit. No change needed.
- The PR body notes no linked repository issue exists yet. That is a process/sprint-check item, not a code concern, and does not affect this review.

## Verdict

No P0/P1 issues. The change is minimal, internally consistent, well-tested, and follows the established registry + optimistic-concurrency conventions. Approving.

luoshu0211:
@Jerry-Xin @mochashanyao 麻烦帮忙审查一下，计划今天上线。当前 head `dc8e1ec9d1d2f3837fedff87ac86a22aaa3d576e` 的 Go CI 已通过；还需要 Maintainer 将本 PR 加入 Octo Board、设置为当前 `Sprint W37`，然后在 Actions 重跑 `check-sprint`（仓库内没有可关联的现成 Issue）。后端依赖需先合并并部署。谢谢。

Jerry-Xin (dismissed):
**APPROVE** — verified at head `dc8e1ec9`, including the mid-review amend that added the changelog entry and embedded-skill surface updates.

The change adds `docs sheet rows insert` as a registry-generated command wrapping `POST /v1/bot/docs/{docId}/sheet/rows/insert` with a required `If-Match` base-version guard, strict local schema bounds, and consistent documentation across README, CLAUDE.md, CHANGELOG, and the embedded octo-docs skill.

✅ Verified at head

- `go build ./...`, `go vet ./...`, and `go test ./... -count=1` all pass (12/12 packages, including the amended skills tests).
- Generated CLI help renders the new `rows` group and `insert` leaf correctly with all four required flags; four-segment operationIds have precedent (`mail.message.attachment.download`).
- The registry entry matches the sibling sheet write operations exactly (path `docId`, required `If-Match` → `--base-version`, `x-octo-risk: write`, strict request schema). The new tests pin the exact method/path/`If-Match`/body and prove invalid flags never reach the server (missing required flags, `--after-row -1`, `--count 0`, `--count 10001`).
- Cross-surface count consistency: CLAUDE.md "329 embedded operations" equals the sum of the per-domain loader counts (docs 38→39).
- Cross-repo contract parity: the route, body fields (`logicalId`/`afterRow`/`count`), `If-Match` semantics, and every documented error code (`base_version_stale`, `sheet_formula_unsupported`, `protected_range`, `sheet_not_found`, `sheet_row_out_of_bounds`, `sheet_row_limit_exceeded`) match the paired backend MR (`octo-docs-backend` !136, currently open — the changelog's minimum-rollout-dependency gate documents this correctly).
- The mid-review amend was reviewed as a byte-level delta: additions only (CHANGELOG entry with rollout dependency, SKILL.md seeding mention, sheet.md rollout/404 guidance, skills_test pins); the full battery was re-run green at the amended head.

💬 Non-blocking

- 🟡 `count` allows a local maximum of 10000, but `afterRow` must name an existing row (so at least one row exists) and the documented result cap is 10000 rows — a `count=10000` request can therefore never succeed and always returns `422 sheet_row_limit_exceeded` from the server. Tightening the schema maximum to 9999, and adding explicit `--after-row 10000` / `--count 10000` rejection cases beside the existing boundary tests, would make the "invalid flags fail locally" guarantee complete. No correctness impact: the server fails closed with the documented error.
- 🔵 `docs/octo-cli-search-command-tree.md` still shows the docs sheet line as `get · edit · replace`; adding `rows insert` would keep it in sync. The file's other staleness (missing ppt subtree, outdated docs count) predates this PR and is not attributed to it.
- 🔵 `skills/octo-docs/SKILL.md` frontmatter description and the spreadsheet routing-table capability list do not yet mention structural row insertion (the quick-start seeding comment now covers it and the skills test pins it). The previous sibling feature updated both lines; cosmetic parity only.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #169 (octo-cli)

## Summary

This PR exposes the backend's structural row-insertion endpoint as `octo-cli docs sheet rows insert`. Because the command tree is generated from the embedded OpenAPI registry, the whole production change is one operation object at `internal/registry/specs/docs.json:991`: a `POST /v1/bot/docs/{docId}/sheet/rows/insert` with a required `If-Match` header surfaced as `--base-version`, three required body fields promoted to `--logical-id` / `--after-row` / `--count`, and `x-octo-strict-request-schema: true` so the declared numeric bounds are enforced locally. The rest is documentation and tests.

The wiring is correct end to end and the new tests drive the real registry rather than a stub. My concerns are all in the agent-facing contract text, which in this repo is a shipped executable surface: the capability is not routed where agents look for it, and two words in the new prose ("visible", "404") collide with meanings the surrounding documentation already established.

## Verification

Static analysis only at head `dc8e1ec9d1d2f3837fedff87ac86a22aaa3d576e`; build and tests not executed in this environment.

- ✅ **Reachable as `docs sheet rows insert`** — `attachOperation` (`cmd/service/service.go:69`) drops the leading `docs`, reuses the `sheet` group, creates `rows` as a new group, and builds the `insert` leaf with `cobra.ExactArgs(1)` since no path flag is declared.
- ✅ **`--base-version` becomes `If-Match`** — `registerHeaderFlags` (`cmd/service/flags.go:168`) honours the rename and marks it required; `buildHeaders` (`cmd/service/run.go:542`) sends it only when set, and `rejectEmptyRequiredValue` refuses an empty value.
- ✅ **Bounds really are enforced locally** — `enforcePublicAPIConstraints` (`cmd/service/run.go:774`) is on via `StrictRequestSchema` (`internal/registry/loader.go:544`), `minimum`/`maximum` are parsed (`internal/registry/loader.go:908`), and `validateNumber` (`cmd/service/run.go:868`) range-checks exact decimals. The PR's "fails locally without sending a request" claim is accurate, not aspirational.
- ✅ **No silent zero-defaulting** — an omitted `--after-row`/`--count` stays absent from the merged body (`cmd/service/run.go:737` gates on `Changed()`) and is then refused as missing, rather than sent as `0`.
- ✅ **Registry invariants hold** — docs 38→39 (`internal/registry/loader_test.go:67`) matches the 39-entry map at `cmd/service/docs_test.go:85` and CLAUDE.md's 328→329; every exhaustive guard (pagination+transform, enum on header/path, secret-with-enum, required-parameter census) auto-covers the new op; `TestSkills_EveryDocumentedFlagExists` resolves the new command and all four documented flags against the real tree, with no engine or root-persistent name collision; `slices` is fine on `go 1.24.0` and the local `contains` helper keeps four callers so `unused` cannot fire.

## Findings

No P0/P1 issues. Four P2 items and one nit; all four P2s are in the agent-facing contract rather than the request wiring.

### P2 — Capability is not advertised where agents discover it, and the new guard passes for the wrong reason (`skills/octo-docs/SKILL.md:62`)

`skills/skills_test.go:49` adds `"row insertion"` to a list whose failure message reads "SKILL.md must advertise spreadsheet capability %q **and route it to sheet.md**". The assertion is a whole-file `strings.Contains`, and the only occurrence in SKILL.md is here at `:62`, inside the `docs create` seeding comment. It is absent from the frontmatter `description:` at `skills/octo-docs/SKILL.md:4` and from the routing-table row at `skills/octo-docs/SKILL.md:24` — the two surfaces an agent runtime uses to select the skill and then pick a reference file. The guard is now false-green for this capability and would stay green if the routing entry were deleted later.

Both preceding PRs in this family updated those two places: `de26e22` added "find & replace" to `:4` and `:24` and `docs.sheet.replace` to the introspection example at `:123`; `b3c9c43` added "create and edit PPT presentations" to `:4`. So this is new drift, not pre-existing. Listing row insertion as a way to *seed* a new doc is also misleading — the same comment says a new doc has NO body, while this command needs a `baseVersion` from `docs sheet get` and an `--after-row` naming an existing row.

```bash
# :4   … spreadsheets including find & replace, row insertion, cells, layout, …
# :24  | Read/edit a **spreadsheet** …: find & replace, row insertion, cells, … | **`sheet.md`** |
# :123 # + docs.sheet.edit / docs.sheet.replace / docs.sheet.rows.insert / docs.scene.edit / …
```

Mirror `de26e22`: add the capability to `:4` and `:24`, add the operation id to `:123`, and drop it from the `:62` seeding comment.

### P2 — "existing visible row" contradicts this repo's own meaning of "visible" (`internal/registry/specs/docs.json:996`)

The new description says "afterRow must name an existing **visible** row", mirrored at `skills/octo-docs/sheet.md:62`. In this documentation set "visible" already means *not hidden by an active shared filter*, explicitly contrasted with stored coordinates: `skills/octo-docs/sheet.md:209` reads "Rows hidden by an active shared filter still appear in `docs sheet get` … To sort only **visible** rows … build a sparse permutation that leaves hidden row **coordinates** unchanged". Everywhere else the row axis is an absolute stored index (`internal/registry/specs/docs.json:835`), and the new paragraph itself means absolute — "`--after-row 9` inserts immediately after row index 9, shifting the old row 10 … down" (`skills/octo-docs/sheet.md:48`).

The contract is also incomplete on its own terms: it states a "visible row" precondition but documents no rejection code for it, listing only `403 protected_range`, `404 sheet_not_found`, `422 sheet_row_out_of_bounds`, `422 sheet_row_limit_exceeded` and `409 sheet_formula_unsupported`. The consumer is a cold-start agent, which on a filtered sheet can pass a visible ordinal instead of the absolute index; the write then returns 200 having relocated rows, merges, hyperlinks, validation ranges, drawing anchors and protection ranges at the wrong coordinate — silent wrong data with no error to catch.

Replace "existing visible row" in both places with "an existing row index within the worksheet's current `rowCount` — an absolute 0-based stored index, not an ordinal among filter-visible rows", and document the rejection code if the backend enforces one.

### P2 — `--after-row 0` is untested, and it is exactly the value a zero-defaulted int flag would lose (`cmd/service/docs_test.go:1331`)

`--after-row` is registered as `IntVar(…, 0, …)`, so its unset value and its legitimate minimum are both `0`. The only things that make `--after-row 0` reach the wire are the `Changed()` gate at `cmd/service/run.go:737` and `validateRequiredProperties` treating a present non-nil `0` as satisfied at `cmd/service/run.go:1043`. Nothing pins that: the happy-path test at `cmd/service/docs_test.go:1300` uses `--after-row 9`, and this bounds test only asserts that `-1`, `0` and `10001` are *rejected*. Since `minimum: 0` makes row 0 a common input (insert above the header row), a regression in either gate turns "insert at the top" into `is missing required field(s): afterRow` with the suite still green.

Add one subtest on the same production path asserting `--after-row 0` produces `{"logicalId":"default","afterRow":0,"count":N}` and a 200.

### P2 — Rollout guidance conflates a route-level 404 with `404 sheet_not_found` (`skills/octo-docs/sheet.md:71`)

"A 404 in a mixed-version environment means the structural route is not live yet" (mirrored at `CHANGELOG.md:14`) tells the caller to read *any* 404 as a deployment problem, but the same operation documents `404 sheet_not_found` as a business error (`internal/registry/specs/docs.json:996`) — which is what a wrong `--logical-id` returns on a fully deployed backend. That mistake is newly easy here: `--logical-id` is required on this command though optional on `docs.sheet.replace`, and `docs sheet get` returns an empty `sheetList` for a legacy single-sheet document (`internal/registry/specs/docs.json:851`), so the caller cannot discover the id and must take `default` from prose. The cases are distinguishable — a business error arrives as an envelope carrying `error.code`, a missing route does not.

Qualify both places: a 404 *without* `error.code: sheet_not_found` means the route is not live; `404 sheet_not_found` means the `--logical-id` names no worksheet. The new section should also state that `default` is what to pass when `sheetList` comes back empty.

### Nit — Second idiom for the same check in one file (`internal/registry/loader_test.go:1043`)

This adds `slices.Contains` (import at `internal/registry/loader_test.go:8`) while the file's own `contains` helper at `internal/registry/loader_test.go:1275` still has four callers at `:559`, `:1231`, `:1234` and `:1237`. Not a build or lint problem, just two spellings three screens apart. Either use `contains` here or migrate all four call sites and delete the helper.

## Human-verify

1. The paired backend change (`octo-docs-backend` MR !136) is not in this repository, so the wire contract is unverifiable from the diff: request keys `logicalId`/`afterRow`/`count`, response keys `insertedRows`/`rowCount`/`bytes`/`newDocVersionSeq`, and the five error codes. The response names are corroborated indirectly — `bytes`, `baseVersion` and `newDocVersionSeq` all appear on sibling docs write ops — but `insertedRows` appears nowhere else. **Not a blocker for this diff**; worth recording the paired handler the way `cmd/service/enum_vocab_test.go:167` does for `docs.sheet.replace`.
2. Whether the backend resolves `afterRow` against absolute stored indices or filter-visible ordinals, and whether it rejects a hidden `afterRow`. **Not a blocker for the CLI wiring**, but the answer decides whether the "visible row" wording above is a phrasing fix or a behavioural gap needing a documented error code.

- **Retry and idempotence.** The client retries transport errors and 429/502/503/504 regardless of method (`internal/client/client.go:985`, `:1079`) and this op declares no `x-octo-retry: never`. That is safe rather than an oversight: a replay carries the same `If-Match`, so an applied first attempt makes the retry fail `412 base_version_stale` instead of inserting twice, and exhaustion surfaces `RESULT_UNKNOWN` with "do not retry automatically". It also matches the domain — `docs.json` declares `x-octo-retry` zero times, including for `docs.sheet.edit` and `docs.sheet.replace`.
- **Explicit null.** `--data '{"afterRow":null,…}'` is refused: a present-but-nil required field is reported missing (`cmd/service/run.go:1043`).
- **Identity routing and response shaping.** No `x-octo-allowed-token-kinds` / `x-octo-mount-by-token-kind`, identical to the three sibling sheet ops, so there is no asymmetry to close; authorization stays server-side and `x-octo-space-header: false` keeps the space unspoofable. No `x-octo-response-fields` / `x-octo-lossless-id-fields` / `x-octo-strict-response-schema` either, matching the siblings, so the envelope passes through and no response validation can reject a live payload.
- **Strict-mode blast radius and bounds.** Opting in adds only `validateNumber`/`validateString`/`validateObjectConstraints` for this body; `minProperties` is 0 and `additionalProperties` is undeclared, so `--data` with extra keys is still forwarded and `logicalId` (no length bounds) refuses nothing new. The bounds match the family: `afterRow` 0..9999 against "Rows are 0-based and may be 0..9999" (`internal/registry/specs/docs.json:835`) and `count` 1..10000 against the `rowCount` 1..10000 cap (`:851`), with an over-cap combination (`--after-row 9999 --count 10000`) correctly left to the documented `422 sheet_row_limit_exceeded`.
- **Rollout gating.** A prose-only rollout dependency with the command shipped enabled matches the accepted convention for `docs sheet replace` (`CHANGELOG.md:17`), and `x-octo-disabled` is service-scoped only (`internal/registry/loader.go:188`), so it is unavailable as a per-operation gate.
- **Pre-existing drift, not from this PR.** `docs/octo-cli-search-command-tree.md:14` still reads `docs (33)` and `:17` still reads `sheet get · edit · replace`. That file was last touched by `de26e22` and docs has been 38 ops since before this PR, so this change only extends existing staleness — noted, not counted against it. There is also no prior review round at this head, so no earlier blockers to re-verify.

## Verdict: APPROVED

The generated-command wiring is correct and I could not break it: the operation resolves to `docs sheet rows insert`, all four flags reach the wire in the right place with the right JSON types, the declared bounds are enforced before any request, an omitted or zero `--after-row` behaves correctly, and the mandatory `If-Match` guard makes the client's method-blind retry safe rather than a double-insert. Nothing here makes a working path unusable or produces wrong data on its own, so there is no blocker. The four P2s are worth fixing before this reaches agents and the first two are cheap: the capability is currently invisible in the two places an agent runtime actually reads, and "visible" and "404" each carry a second, already-established meaning elsewhere in the same documentation set that a cold-start agent will plausibly apply.

[Octo-Q] verdict: APPROVE — no P0/P1; four non-blocking P2 contract-text items (skill routing + false-green guard, "visible row" ambiguity, missing `--after-row 0` boundary test, 404 conflation) plus one nit.

yujiawei:
Follow-up to the approval: the wiring is correct and this stays **APPROVED** (no blocking issues). The items below are non-blocking P2 polish worth doing before this reaches agent consumers — they're all in the agent-facing contract text and tests, not the request path.

**1. Capability isn't advertised where the skill runtime looks (`skills/octo-docs/SKILL.md`).** "row insertion" only appears in the `docs create` seeding comment at line 62 — it's missing from the frontmatter `description:` (line 4) and the routing table row (line 24), which are the two surfaces used to select the skill and pick a reference file. The `skills_test.go` guard is a whole-file `strings.Contains`, so it's green for the wrong reason and would stay green if the routing entry were later deleted. Both preceding sheet PRs added their capability to lines 4 and 24. Listing it under *seeding a new doc* is also off, since that comment notes a new doc has no body while this command needs a `baseVersion` and an existing `--after-row`. Suggest: add the capability to lines 4 and 24 and drop it from the line-62 seeding comment.

**2. "existing visible row" collides with an established meaning (`internal/registry/specs/docs.json` and `skills/octo-docs/sheet.md:62`).** Elsewhere in this doc set "visible" specifically means *not hidden by an active shared filter* (see `sheet.md` sorting guidance), yet the insert paragraph itself means the absolute stored index (`--after-row 9` = row index 9). On a filtered sheet a cold-start agent could pass a filter-visible ordinal instead of the absolute index and get a 200 that relocates resources at the wrong coordinate — silent wrong data. Suggest wording like "an existing 0-based row index within the worksheet's current `rowCount` (an absolute stored index, not an ordinal among filter-visible rows)", and document the rejection code if the backend enforces the precondition.

**3. `--after-row 0` has no positive test (`cmd/service/docs_test.go`).** The flag registers with default `0`, and `minimum: 0` makes row 0 a legitimate (and common — insert above the header) input. The happy-path test uses `--after-row 9`; the bounds test only asserts `-1`/`count 0`/`count 10001` are rejected. A regression in the `Changed()` gate or required-field check would turn "insert at the top" into a spurious "missing afterRow" with the suite still green. Suggest one subtest asserting `--after-row 0` reaches the wire as `{"logicalId":"default","afterRow":0,"count":N}`.

**4. Rollout guidance conflates a route-level 404 with `404 sheet_not_found` (`skills/octo-docs/sheet.md:71`, `CHANGELOG.md`).** The docs tell callers to read any 404 as "backend not deployed", but the operation also documents `404 sheet_not_found` as a business error (wrong `--logical-id`), which is easy to hit since `--logical-id` is required here (unlike `docs sheet replace`) and a legacy single-sheet doc returns an empty `sheetList`. Suggest qualifying: a 404 *without* `error.code: sheet_not_found` means the route isn't live; `404 sheet_not_found` means the logical id names no worksheet — and note that `default` is the value to pass when `sheetList` comes back empty.

**Nit:** `internal/registry/loader_test.go` adds `slices.Contains` while the file's own `contains` helper (4 callers) does the same thing three screens away. Pick one spelling.

None of these block merge.

luoshu0211:
已补齐 docs sheet rows insert/delete 与 columns insert/delete，当前 head 458b744667caf70a04ac7499c8f33cce397da1b0；go test ./...、go vet ./...、go build ./... 均通过。当前 check-sprint 明确因 PR 未关联 issue/当前 Sprint 失败；请 Maintainer 将此 PR 加入 Octo Board 并设置 Sprint W37（或关联已有的本需求 issue 并设置当前 Sprint）后重跑。后端依赖：dmwork/octo-docs-backend!136，需先合并部署并验证新路由。

Jerry-Xin (approved):
**APPROVE** — re-review at head `458b7446`, fix-confirmation round over `dc8e1ec9`. All five peer suggestions verified fixed at the byte level, each mutation-killed where practical; the scope grew to four structural commands (rows insert/delete, columns insert/delete) and the new bytes were re-walked end to end with cross-repo contract verification.

## Peer-suggestion fixes — all verified

1. **Local `count` max 9999 (Jerry-Xin).** `docs.sheet.rows.insert` spec now caps `count` at 9999 instead of 10000, matching the fact that an insert anchor requires an existing row so the 10000-row cap can never validly consume the full range. The boundary test now rejects `--count 10000` locally. **Mutation-killed:** restoring `"maximum": 10000` in the spec turns that subtest red (`error = <nil>` — the over-cap request would be accepted); restore → green. The paired backend MR converges on the identical bound (`SHEET_MAX_ROWS - 1`, with the same "neither insertion nor deletion can validly use the full axis maximum" reasoning), so both sides of the wire now agree.
2. **Command-tree doc row (Jerry-Xin).** `docs/octo-cli-search-command-tree.md` gains the `rows insert · delete` / `columns insert · delete` lines and refreshes the docs count. See advisory A-1 below on the residual leaf gap.
3. **Skill discovery surfaces (Jerry-Xin + mochashanyao P2-1).** "structural row/column edits" now appears in both agent-discovery surfaces — the frontmatter `description:` and the routing-table row — plus the schema-introspection example lists `docs.sheet.rows.insert` / `docs.sheet.columns.delete`; the misleading seeding-comment mention is gone (correctly: the command cannot seed an empty doc). `skills_test.go` needle updated in lockstep. **Mutation-killed:** stripping the needle from SKILL.md turns `TestOctoDocsSkillEmbedded` red; restore → green. Residual note (pre-existing design, not new): the guard is a whole-file `strings.Contains`, so removing the phrase from only one of the two surfaces would stay green — same weakness class as every sibling needle in that list.
4. **"visible row" wording (mochashanyao P2-2).** Both the spec description and `sheet.md` now say absolute stored index, explicitly *not* an ordinal among filter-visible rows — the collision with this repo's established meaning of "visible" is gone, closing the silent-wrong-coordinate trap for cold-start agents on filtered sheets.
5. **`--after-row 0` untested (mochashanyao P2-3).** New `TestDocsSheetRowsInsert_PreservesZeroAnchor` drives the real production path and pins `afterRow: 0` reaching the wire body. **Mutation-killed:** neutering the `Changed()` merge gate for that flag reproduces exactly the predicted regression symptom (`missing required field(s): afterRow`) and turns the new test red; restore → green. This guard is real, not false-green.
6. **Rollout-404 conflation (mochashanyao P2-4) + `slices`/`contains` nit.** CHANGELOG and `sheet.md` now distinguish a route-level 404 (backend not deployed) from `404 sheet_not_found` (bad `--logical-id`), and document the legacy `default` fallback for an empty `sheetList`. `loader_test.go` drops the second idiom and uses the file's own `contains` helper.

## New scope: rows delete + columns insert/delete — fresh-eyes walk

- **Cross-repo contract parity (executable verification against the paired backend MR's diff):** all four routes exist (`POST /:docId/sheet/{rows,columns}/{insert,delete}`); request field names match exactly (`logicalId` + `afterRow`/`startRow`/`afterColumn`/`startColumn` + `count`); response keys match exactly (`insertedRows`/`deletedRows` + `rowCount`, `insertedColumns`/`deletedColumns` + `columnCount`); all six error codes the CLI documents (`sheet_row_out_of_bounds`, `sheet_row_limit_exceeded`, `sheet_column_out_of_bounds`, `sheet_column_limit_exceeded`, `sheet_formula_unsupported`, `sheet_not_found`) appear in the backend bytes; `base_version_stale`/`protected_range` ride the shared If-Match + writer-guard pipeline the handlers reuse. Backend count bounds are `SHEET_MAX_ROWS - 1` (9999) and `SHEET_MAX_COLUMNS - 1` (99) — identical to the CLI's local strict-schema bounds; the backend's own route test rejects `count: 10_000`, mirroring the CLI's local rejection.
- **Local bounds are the correct static subset:** anchors 0..9999 (rows) / 0..99 (columns) are absolute axis maxima; contextual limits ("at least one row/column must remain", resulting-size caps) depend on live `rowCount`/`columnCount` and are correctly delegated to the documented server 422s rather than faked locally.
- **New tests:** request-shape tests pin method, path, `If-Match: BV`, and exact JSON body for all three new commands; bounds tests pin local rejection (negative start, over-cap count, missing required field) with an explicit "must not reach the server" assertion. Registry-shape map and schema tests extended table-driven over all four operations.
- **Cross-surface census (whole-repo):** CHANGELOG entry generalized to all four commands with the rollout gate intact and sharpened; CLAUDE.md 329→332 equals the sum of the pinned per-domain counts (verified by arithmetic and by the green `TestAllDomainOperationCounts`, docs 39→42 matching the 42-entry registry-shape map); README shows all four examples; SKILL.md description/routing/introspection updated; `sheet.md` section generalized with correct caps (10,000 rows / 100 columns) and error semantics; `skills_test.go` needles cover the three new command strings. No stale "row insertion" / "count must be 1..10000" / "visible row" leftovers anywhere.
- **Schema introspection:** `docs.sheet.rows.delete` resolves through the real binary's registry.

## Verification at head

- `go build ./...`, `go vet ./...` clean; `go test ./... -count=1` green across all packages (including `cmd/service`, `internal/registry`, `skills`).
- Remote CI at this head: `build & test (1.24.x)` pass, `npm test` pass. `check-sprint` fails on board/sprint association (governance, acknowledged by the author, needs a Maintainer — not a code signal). `osv-scan` fails on dependency vulnerabilities while this PR touches no dependency manifest (no go.mod/go.sum bytes) — inherited, unrelated to this diff.
- Three mutation probes run and reverted (spec bound, flag-merge gate, skill needle); working tree confirmed clean afterwards, affected packages re-run green.

## Advisories (non-blocking)

- **A-1.** `docs/octo-cli-search-command-tree.md` now claims `docs (42)` but enumerates 37 leaves — the `ppt` row and `comments get · replies` remain absent. Those omissions predate this PR (inherited staleness of a design-snapshot doc), and every maintained authoritative surface (CLAUDE.md tree + pinned counts, README, SKILL.md, tests) is consistent, so this does not block. Since the count was refreshed to the true 42 in this round, a follow-up adding the five missing leaves would make the snapshot internally consistent again.
- **A-2.** The 9999/99 caps are pinned only on the rejection side (`--count 10000` / `--count 100` refused). No test asserts the accept side (`--count 9999` passes local validation), so an accidental future *lowering* of the spec maximum would stay green. One extra subtest per axis would close it.
- **A-3.** Rollout ordering remains a hard gate: the paired backend MR is still open, so these four routes 404 at the route level on any environment without it. The CHANGELOG/sheet.md guidance now states this precisely — release only after the backend MR merges and deploys.

Prior approvals from peers pinned `dc8e1ec9` and were dismissed by the force-push; re-review at this head is theirs to make.

luoshu0211:
@yujiawei @mochashanyao 当前 head `458b744667caf70a04ac7499c8f33cce397da1b0` 已补齐 rows insert/delete 与 columns insert/delete，Jerry-Xin 已基于当前 head APPROVE，build/npm/history 全绿。麻烦尽快补当前 head 复审；同时需要 Maintainer 将 PR 加入 Octo Board 并设为 Sprint W37。后端 !136 仍需先部署。

yujiawei (approved):
# Code Review — PR #169 (octo-cli)

Reviewed at head `458b744667caf70a04ac7499c8f33cce397da1b0` against base `main` (merge-base `b3c9c43`). Verified locally: `go build ./...`, `go vet ./...`, `gofmt`, and the full `go test ./...` suite all pass; the four new commands were smoke-tested via `--help` and `--dry-run`.

## Summary

This is a spec-driven change. The four new structural spreadsheet operations (`docs sheet rows insert|delete`, `docs sheet columns insert|delete`) are added purely as OpenAPI path entries in `internal/registry/specs/docs.json`; the generic command engine derives the cobra commands, flags, local schema validation, and the HTTP request from that spec. No hand-written command wiring was needed, which is the right pattern for this codebase.

## Scope check

The diff delivers exactly the four structural operations described in the PR body and nothing beyond them:
- Registry specs for the 4 routes, all `POST`, all `x-octo-risk: write`, all requiring `If-Match` via `--base-version` and `logicalId`/anchor/`count` in the body (`internal/registry/specs/docs.json:991+`).
- Registry-shape and count tests (`internal/registry/loader_test.go:66`, `:1024`), request/transport and local-bounds tests (`cmd/service/docs_test.go:1303+`).
- Docs kept in lockstep: `CHANGELOG.md`, `README.md:195`, `skills/octo-docs/SKILL.md`, `skills/octo-docs/sheet.md:45`, the search command tree, and the operation-count strings.

The registry now reports 42 docs operations / 332 embedded operations, matching the updated `CLAUDE.md` and `loader_test.go` assertions. No missing pieces and no undocumented additions.

## Findings

All findings below are non-blocking (P2 / advisory). None require changes before merge.

### 1. [P2] Structural mutations retain the default transport retry policy

The four new routes do not declare `x-octo-retry: "never"`, so they inherit the default retry-on-transient behavior (`internal/registry/specs/docs.json:991+`; retry maps to `DisableRetry`/`UnknownOutcomeOnNetworkFailure` via `x-octo-retry`, `cmd/service/run.go:84-85`). In principle a lost/transient-`5xx` response after the backend commits could be retried.

Why this is not a blocker:
- The mandatory `If-Match` (base-version) guard makes the operation fail *closed*: a retry reuses the same stale token, so the backend returns `412 base_version_stale` rather than applying the edit twice. There is no silent double-insert/double-delete.
- Retries only fire on `429/502/503/504` (`isRetryableStatus`, `internal/client/client.go:1302`), not on a delivered `200`.
- The pre-existing sibling writes `docs.sheet.replace` and `docs.sheet.edit` also do not set `x-octo-retry: "never"`, so this PR is consistent with the established convention, not a regression.

Suggestion (optional, follow-up): if you want ambiguous transport outcomes on these structural writes to surface as `RESULT_UNKNOWN` instead of a misleading `412`, mark all four ops (and ideally the existing `sheet.replace`/`sheet.edit`) with `x-octo-retry: "never"`. Doing it consistently across the sheet-mutation family is preferable to doing it for only the new four.

### 2. [P2] Local coordinate/count bounds are hard-coded to current server limits

`afterRow`/`startRow` and `count` are capped at `9999`, and column anchors/count at `99`, enforced client-side because the ops are `x-octo-strict-request-schema: true`. This is the correct current contract, but if the backend ever raises the 10,000-row / 100-column ceilings, older CLIs will reject valid operations until upgraded. Acceptable trade-off for fail-fast validation; noting it only as a forward-compat consideration.

### 3. [P2] No local test for the "at least one row/column must remain" rule

Existing tests cover negative indices, missing required flags, and count-over-limit. There is no test for deleting the entire sheet (leaving zero rows/columns). This is fine because that invariant is server-enforced (`422`, per the spec description) and cannot be validated from `count`/`start` bounds alone client-side — but a note in `sheet.md` already documents it, so behavior is well-specified.

## Coverage note

I could not assess from this diff alone: the `octo-docs-backend` MR !136 implementation (the `sheet_formula_unsupported`, `protected_range`, and atomic-relocation behavior all live server-side), nor live behavior in a mixed-version deployment. The PR body and `CHANGELOG.md`/`sheet.md` correctly gate rollout on that backend MR being merged and deployed first, and explicitly warn against emulating structural edits with a whole-sheet rewrite. That ordering must hold operationally; nothing in the CLI can enforce it.

## Verdict

Spec compliance: the diff implements exactly the stated scope with matching tests and documentation. Code quality: clean, idiomatic, spec-driven; the only findings are P2/advisory with no correctness, security, or data-loss risk (the `If-Match` guard closes the one retry concern). Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #169 (octo-cli)
Head `458b744667caf70a04ac7499c8f33cce397da1b0` · base `main` (merge-base `b3c9c43c`) · 10 files, +661/−47

## Summary

The PR adds four spec-driven write operations to the docs domain — `docs.sheet.rows.insert`, `docs.sheet.rows.delete`, `docs.sheet.columns.insert`, `docs.sheet.columns.delete` (`internal/registry/specs/docs.json:991`, `:1078`, `:1157`, `:1236`) — which the registry loader turns into the four-level commands `octo-cli docs sheet rows|columns insert|delete`. There is no hand-written Go production code: the whole surface is the OpenAPI entry plus the embedded skill doc (`skills/octo-docs/sheet.md:45`), README examples, CHANGELOG rollout note, and count updates. All four opt into `x-octo-strict-request-schema`, which switches on client-side `minimum`/`maximum` enforcement before any HTTP is sent.

The wiring is correct end to end and I verified it by building and running the binary, not just by reading the spec. The only real gap is documentation bookkeeping: the README domain inventory was not updated with the rest of the count surfaces.

## Verification

Local checks at head `458b744`, executed in this environment:
- `go build -o /tmp/cr169/octo-cli ./cmd/octo-cli` — clean
- `gofmt -l .` — clean; `go vet ./...` — clean
- `go test ./... -count=1` — pass (all 12 packages, including `cmd/service`, `internal/registry`, `skills`)

Runtime probes against the built binary (not static inference):
- `docs sheet --help` lists the new `rows` and `columns` groups beside `get`/`edit`/`replace`; `docs sheet rows --help` lists `insert`/`delete`.
- `docs sheet rows insert --help` shows `POST /v1/bot/docs/{docId}/sheet/rows/insert  [risk: write]` with `--after-row int`, `--count int`, `--logical-id`, `--base-version`, each marked `(required)`.
- `--dry-run` emits `{"afterRow":0,"count":1,"logicalId":"default"}` with `If-Match: BV==` — the zero anchor survives, confirming `applyBodyFlags`' `Changed()` gate (`cmd/service/run.go:735`) rather than an unconditional int-flag merge.
- Local gates fire with exit 2 and no HTTP: `--count 10000` → `count is outside its allowed numeric range`; `--after-column 100` → `afterColumn is outside its allowed numeric range`; missing `--logical-id` → `missing required field(s): logicalId`; missing `--base-version` → `required flag(s) "base-version" not set`.
- `--data '{"logicalId":"default","afterRow":9,"count":5}'` works, flags override `--data`, and a mistyped key (`logicalID`) is still caught by the required-field check.
- `octo-cli schema docs.sheet.rows.insert` resolves and exposes the bounds advertised in `skills/octo-docs/SKILL.md:123`.
- `octo-cli docs sheet rows upsert d1` exits 2 with `unknown subcommand`, so the new intermediate groups keep the fail-loud guard.

## Findings

No P0/P1 issues; two P2 items below, neither blocking.

### P2 — README domain inventory left stale by this count bump (`CLAUDE.md:44`)

Raising the embedded total to 332 is correct — it only sums with `docs` = 42, matching `internal/registry/loader_test.go:66` and `docs/octo-cli-search-command-tree.md:14`. But `README.md:40` still reports `| docs | 38 |`, and that value was accurate at the merge-base, so this PR introduces the drift rather than inheriting it. The two prior PRs that grew this domain updated that same row in the same commit (#167 32→33, #165 33→38), so the row is a maintained contract, not decoration. Nothing in CI catches a miss: no Go test reads `README.md`, which is why the suite is green. The row's Purpose cell also still reads "sheet cells (paged read and atomic replace)", so the four new structural commands are absent from the public capability inventory even though README's own examples at `README.md:197` demonstrate them.

```markdown
| `docs`    | 42  | … sheet cells (paged read, atomic replace, and structural row/column edits) … |
```

Fix: set `README.md:40` to 42 and mention structural row/column edits in the Purpose cell.

### P2 — Command-tree count no longer reconciles with its own listing (`docs/octo-cli-search-command-tree.md:14`)

`docs (42)` is now the true registry count, but the tree underneath enumerates only 37 leaves: the whole `ppt get · edit · export` group and `comments get · replies` are missing. Those lines are untouched by this PR (the omission dates to #165), so this is pre-existing and not a regression here; the only effect is that a reader cannot derive 42 from the tree. Optional follow-up: add a `ppt` row and the two missing `comments` verbs.

## 数据流回溯

Each consumed value traced from declaration to the production entry point:

- `operationId: docs.sheet.rows.insert` → `attachOperation` splits on `.` (`cmd/service/service.go:70`) and loops `for i := start; i < len(segs)-1` (`:85`), so a 4-segment id nests correctly; `findChild` reuses the existing `sheet` group instead of duplicating it. Verified in the rendered help, not inferred.
- `x-octo-flag` on body properties → `registerBodyFlags` (`cmd/service/flags.go:361`) → `promotableKind` yields `kindInt` for `type: integer` → `cmd.Flags().IntVar` → `applyBodyFlags` writes a Go `int` into the body (`cmd/service/run.go:741`). `validateNumber` accepts `int` (`:868`), so there is no string/number mismatch on the promoted path — the failure mode this trace exists to rule out.
- `minimum`/`maximum` → `schemaInfoFromNode` reads them as `float64` (`internal/registry/loader.go:908`), reachable only because `x-octo-strict-request-schema` sets `StrictRequestSchema` (`:544`) → `enforcePublicAPIConstraints` (`cmd/service/run.go:774`) → `validateByType` dispatches `integer` to `validateNumber` (`:845`). Without the strict opt-in the bounds would be help text only; confirmed live via the exit-2 probes above.
- `If-Match` `required: true` → `registerHeaderFlags` calls `MarkFlagRequired` (`cmd/service/flags.go:181`) → cobra rejects before body assembly. Same mechanism as the pre-existing `docs.sheet.edit`/`docs.sheet.replace`.
- Response schema → the new 200 bodies are flat `{docId,…,bytes,baseVersion,newDocVersionSeq}`, identical in shape to `docs.sheet.edit` and `docs.sheet.replace`; no `x-octo-response-unwrap` divergence, so envelope handling is unchanged.
- Bounds vs. real domain limits → rows are documented 0..9999 with a 10000-row cap and columns 0..99 with a 100-column cap (`skills/octo-docs/sheet.md:483`, `:500`). So `afterRow ≤ 9999` / `count ≤ 9999` and `afterColumn ≤ 99` / `count ≤ 99` are exactly the domain maxima, not narrower — this is the check that matters, because a gate narrower than the backend turns a working call into an unappealable local exit 2.
- Docs count claims → 42 ops in `docs.json`; the 14 per-domain counts in `internal/registry/loader_test.go:58` sum to exactly 332 with `matter`+`summary` withheld, i.e. 12 active domains. All three updated surfaces agree.

## 盲点 checklist (C1–C6)

- **C1 双路径 parity — clear.** insert↔delete checked as pairs for both axes: anchors (`afterRow`/`startRow`, `afterColumn`/`startColumn`), identical `count` bounds per axis, symmetric responses (`insertedRows`/`rowCount` vs `deletedRows`/`rowCount`), and error vocabularies that differ only where they should (`sheet_row_limit_exceeded` / `sheet_column_limit_exceeded` appear on insert only; delete cannot exceed a cap). README's examples are self-inverse (`--after-row 9 --count 5` then `--start-row 10 --count 5` removes exactly what was inserted), which independently confirms the anchor semantics are documented consistently.
- **C2 control-flow ordering / equivalent forms — clear.** `x-octo-strict-request-schema` newly applies to four ops, so I probed non-canonical input rather than trusting the flag: `--data` with `9.0` passes (`big.Rat.IsInt()`), a string `"9"` is correctly rejected, an unknown extra key is forwarded because `additionalProperties` is not declared, and flag/`--data` mixing merges as documented. `TestPptStrictSchemaOnlyAppliesToEditBody` (`internal/registry/loader_test.go:1296`) iterates a fixed id list that excludes the new ops, so the strict opt-in does not trip that guard.
- **C3 授权边界 ≠ 能力边界 — N/A with reason.** The CLI is a client; `x-octo-risk: write` is display-only (`cmd/service/service.go:230`) and grants nothing. Authorization stays server-side ("needs writer" in each summary). No endpoint, tool, or credential exposure is widened by this diff.
- **C4 授权生命周期 / 状态级联 — N/A.** No auth, membership, status, or container/member cascade logic is touched.
- **C5 build/note 通过 ≠ 运行期路径正确 — cleared by execution, not by note.** This is the exact trap for a spec-only PR: a green build would prove nothing about whether four-level operation ids actually nest. I built the binary and exercised help, `--dry-run` wire shape, header emission, local validation, `schema`, and the unknown-subcommand guard. `CHANGELOG.md:15` and `skills/octo-docs/sheet.md:93` carry the rollout caveat that these routes 404 until `octo-docs-backend` MR !136 ships; that is a real deployment dependency I cannot verify from this checkout, and the docs correctly tell operators not to paper over it with a whole-sheet rewrite.
- **C6 治理/策略/文档自洽性 — clear.** Cross-checked the new caps against the pre-existing ones in the same file (10000 rows / 100 columns at `skills/octo-docs/sheet.md:483`, `:500`) — no contradiction. Also checked the apparent conflict between CHANGELOG entries: `docs sheet replace` says a mixed-version 404 should fall back to `get` + `edit` (`CHANGELOG.md:22`), while the new entry forbids emulating a structural op with a whole-sheet rewrite (`:17`). These are deliberately different and each is justified — a value rewrite is safe, a structural rewrite cannot re-anchor merges/drawings/filters/comment anchors/protections — so it is not self-contradictory guidance. `skills/skills_test.go:60` pins all six new sheet.md strings and all six are present; not bumping `skills/octo-docs/SKILL.md` `version: 0.2.0` matches repo practice (#167, #165, #161, #152 all added capabilities without a bump).

## 跨轮 blocker 复检 (R6)

N/A — first review round for this PR; no prior blocker list exists on this issue or in the checkout.

- `TestAllDomainOperationCounts` (`internal/registry/loader_test.go:66`) updated to 42 and passing; `CLAUDE.md:44` 332 and 12 active domains both arithmetically correct.
- `TestEnum_EveryRequestSideVocabularyIsPinned` sweeps every enabled op; the new ops declare no enums, so no provenance entry is owed.
- No flag collisions: `--count`, `--logical-id`, `--after-row`, `--start-row`, `--after-column`, `--start-column` are absent from `reservedFlagNames`, and `--base-version` (header) correctly wins over any body mirror.
- No duplicate `operationId`s; the new ids are the only 4-segment ones in the docs spec and they nest without shadowing `get`/`edit`/`replace`.
- npm packaging (`npm/scripts/prepare-packages.test.js`) is unrelated to spec content and unaffected.

## Verdict: APPROVED

The four new operations are correctly declared, correctly wired through the loader into real four-level commands, correctly gated on `If-Match`, and correctly bounded — and I confirmed all of that by running the built binary rather than by reading the spec alone. Both findings are documentation bookkeeping with no runtime consequence: `README.md:40` still says 38 while every other count surface says 42, and the command-tree listing cannot be reconciled to its own (now correct) parenthesised count. Neither makes a working path unusable, produces incorrect runtime data, nor leaves the CLI worse than before, so under R1/R4 neither reaches P1. Fixing the README row before merge would be good hygiene; it is not a blocker.

[Octo-Q] verdict: APPROVE — 4 new spec-driven operations verified working end to end by building and probing the binary (command nesting, `If-Match`, `--dry-run` wire body, zero-anchor preservation, local bound gates exiting 2 with no HTTP); full `go test ./...`, `go vet`, and `gofmt` clean at head `458b744`. No P0/P1: bounds match the documented 10000-row/100-column caps exactly so the new strict gate is not narrower than the domain, insert↔delete parity is symmetric, and C1–C6 are clear or N/A. Two P2 documentation items only — stale `README.md:40` docs op count (38, should be 42) and a pre-existing unreconcilable command-tree listing.

yujiawei:
Confirming the doc-count follow-up (non-blocking, does not change the Approve):

`README.md:40` still reads `| `docs` | 38 |`, while every other count surface in this PR was bumped to 42 — `CLAUDE.md:44` (332 total), `docs/octo-cli-search-command-tree.md:14` (`docs (42)`), and `internal/registry/loader_test.go:66` (`"docs": 42`). I checked the history: #167 updated this row 32→33 and #165 updated it 33→38 in the same commit that grew the domain, so this row is a maintained inventory and the drift is introduced by this PR rather than inherited. No Go test reads `README.md`, which is why the suite stays green.

The Purpose cell on that row also still says "sheet cells (paged read and atomic replace)" and omits the new structural row/column commands, even though README's own examples further down demonstrate them.

Suggested one-line fix before merge (good hygiene, not a merge blocker):

```markdown
| `docs`    | 42  | … sheet cells (paged read, atomic replace, and structural row/column edits) … |
```

